PR #5, opened 2018-05-04
wlroots: fix build issues

SirCmpwn:
Summary of changes:

- wlroots requires C11, not C99
- Dependency change allows you to add wlroots as a subproject (useful if
  you're working with a local copy of wlroots)
- Several plugins needed to declare the wlroots dependencies to find the
  headers if installed in a nonstandard location


ammen99:
Thank you, this is definitely something I've missed.

PR #18, opened 2018-06-18
Add place plugin

soreau:
Add a basic window placement plugin with three placement modes.
This plugin was based off of the center-view plugin by Adrian
Vovk. It contains the same code as center-view for the center
placement mode. The valid mode config keys are cascade, center
and random.

### DIFF
--- a/plugins/single_plugins/place.cpp
+++ b/plugins/single_plugins/place.cpp
@@ -1,0 +1,106 @@
+#include <view.hpp>
+#include <core.hpp>
+#include <workspace-manager.hpp>
+#include <signal-definitions.hpp>
+
+class wayfire_place_window : public wayfire_plugin_t
+{
+	signal_callback_t created_cb;
+	signal_callback_t workarea_changed_cb;
+	wf_option placement_mode;
+	int cascade_x, cascade_y;
+
+	public:
+	void init(wayfire_config *config)
+	{
+		created_cb = [=] (signal_data *data)
+		{
+			auto view = get_signaled_view(data);
+
+			if (view->role != WF_VIEW_ROLE_TOPLEVEL ||
+				view->parent || view->fullscreen ||
+				view->maximized)
+				return;
+
+			auto workarea = output->workspace->get_workarea();
+			auto mode = placement_mode->as_string();
+
+			if (mode == "cascade")
+				cascade(view, workarea);
+			else if (mode == "random")
+				random(view, workarea);
+			else
+				center(view, workarea);
+		};
+
+		workarea_changed_cb = [=] (signal_data *data)
+		{
+			auto workarea = output->workspace->get_workarea();
+			cascade_x = workarea.x;
+			cascade_y = workarea.y;
+		};
+
+		auto section = config->get_section("place");
+		placement_mode = section->get_option("mode", "center");
+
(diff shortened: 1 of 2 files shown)


plugins/single_plugins/place.cpp
```diff
@@ -0,0 +1,106 @@
+#include <view.hpp>
+#include <core.hpp>
+#include <workspace-manager.hpp>
+#include <signal-definitions.hpp>
+
+using namespace std;
```
ammen99: don't `using namespace std;`

plugins/single_plugins/place.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	public:
+	void init(wayfire_config *config)
+	{
+		created_cb = [=] (signal_data *data)
+		{
+			auto view = get_signaled_view(data);
+
+			if (view->role != WF_VIEW_ROLE_TOPLEVEL ||
+				view->parent ||view->fullscreen ||
```
ammen99: space after the first  `||`

plugins/single_plugins/place.cpp
```diff
@@ -0,0 +1,106 @@
...
+				view->maximized)
+				return;
+
+			auto workarea = output->workspace->get_workarea();
+			auto section = config->get_section("place");
+			auto placement_mode = section->get_option("mode", "center");
```
ammen99: just a minor thing, you get get the section + option in `init()` , and save the placement_mode as `wf_option`, but it's fine now as well

plugins/single_plugins/place.cpp
```diff
@@ -0,0 +1,106 @@
...
+		if (cascade_x + window.width > workarea.x + workarea.width ||
+			cascade_y + window.height > workarea.y + workarea.height)
+		{
+			cascade_x = workarea.x;
+			cascade_y = workarea.y;
+		}
+
+		view->move(cascade_x, cascade_y);
+
+		cascade_x += 50;
```
ammen99: are we sure this should be done this way (hardcoded values)? Maybe add 2-3% of the resolution?

soreau: It could be done any way, I think the main idea for this primitive placement mode is so windows are not placed directly on top of each other, so the user can see them as they open.

plugins/single_plugins/place.cpp
soreau: I don't know what you mean by this.

plugins/single_plugins/place.cpp
soreau: Ok.

plugins/single_plugins/place.cpp
ammen99: but, on a 4k monitor and a low res one the plugin would have a different(IMO) behavior. How does compiz do it or you don't know?

plugins/single_plugins/place.cpp
ammen99: Just that you can store the result of get_option() as a member variable, but this isn't a big problem

ammen99:
LGTM, will merge it once I am able to get back to my computer and test it

ammen99:
Works as a charm, thanks!

PR #29, opened 2018-06-24
Introduce fisheye plugin

soreau:
This is a fisheye effect ported from compiz magnifier plugin fisheye fragment program.

### DIFF
--- a/plugins/single_plugins/fisheye.cpp
+++ b/plugins/single_plugins/fisheye.cpp
@@ -103,11 +103,14 @@ class wayfire_fisheye : public wayfire_plugin_t
                 render(fb, tex, target);
             };
 
-            toggle_cb = [=] (uint32_t key) {
-                    if (active) {
+            toggle_cb = [=] (uint32_t key)
+            {
+                    if (active)
+                    {
                         active = false;
                         duration.start(duration.progress(), 0);
-                    } else {
+                    } else
+                    {
                         active = true;
                         duration.start(duration.progress(), target_zoom);
 
--- a/plugins/single_plugins/fisheye.cpp
+++ b/plugins/single_plugins/fisheye.cpp
@@ -79,9 +79,9 @@ class wayfire_fisheye : public wayfire_plugin_t
     post_hook_t hook;
     key_callback toggle_cb;
     wf_duration duration;
-    float radius, target_zoom, current_zoom;
+    float target_zoom;
     bool active, hook_set;
-    wf_option radius_opt, zoom_opt;
+    wf_option radius, zoom_opt;
 
     GLuint program, posID, mouseID, resID, radiusID, zoomID;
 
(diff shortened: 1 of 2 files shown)


plugins/single_plugins/fisheye.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+        t1 = t1 * p1 + p2;
+
+        tc = texture2D(u_texture, t1.xy).rgb;
+
+        gl_FragColor = vec4(tc, 1.0);
+}
+)";
+
+class wayfire_invert_screen : public wayfire_plugin_t
```
ammen99: s/invert_screen/fisheye/ or something among these lines

plugins/single_plugins/fisheye.cpp
```diff
@@ -0,0 +1,201 @@
...
+            GL_CALL(glBindTexture(GL_TEXTURE_2D, 0));
+
+            if (!duration.running() && target_zoom == 0)
+            {
+                output->render->rem_post(&hook);
+                hook_set = false;
+            }
+
+            if (active)
+                duration.start(current_zoom, target_zoom);
```
ammen99: I can't get why you start at every frame ..

plugins/single_plugins/fisheye.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+)";
+
+class wayfire_invert_screen : public wayfire_plugin_t
+{
+
+    post_hook_t hook;
+    key_callback toggle_cb;
+    wf_duration duration;
+    float radius, target_zoom, current_zoom;
```
ammen99: why do you need current zoom? I think you can just use `duration.progress()` every time you need it, or make it a local variable for `render()`. Same for radius, just use `radius_opt->as_cached_double()` each time you need it, and in this case, you can rename it to just `radius`

plugins/single_plugins/fisheye.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+            GL_CALL(glDisable(GL_BLEND));
+            GL_CALL(glBindFramebuffer(GL_DRAW_FRAMEBUFFER, target));
+            GL_CALL(glDrawArrays (GL_TRIANGLE_FAN, 0, 4));
+            GL_CALL(glBindFramebuffer(GL_FRAMEBUFFER, 0));
+
+            GL_CALL(glDisableVertexAttribArray(posID));
+            GL_CALL(glBindTexture(GL_TEXTURE_2D, 0));
+
+            if (!duration.running() && target_zoom == 0)
```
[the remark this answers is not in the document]
soreau: It was being set in toggle_cb() if active is true here https://github.com/soreau/wayfire/blob/ed66f7d87826c437a7545b1ca91b229c92389f00/plugins/single_plugins/fisheye.cpp#L110

plugins/single_plugins/fisheye.cpp
soreau: You're basically right, I just made current_zoom local to render() and got rid of radius_opt.

plugins/single_plugins/fisheye.cpp
soreau: This is so changing target_zoom via wcm works.

plugins/single_plugins/fisheye.cpp
soreau: Ack.

plugins/single_plugins/fisheye.cpp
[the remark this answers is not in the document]
soreau: Actually that's current_zoom so you're right, it's never set to 0. I removed the condition.

ammen99:
LGTM (code- and aesthetics-wise), thanks!

PR #31, opened 2018-06-29
Transform mouse coordinates

soreau:
Apply output scale and rotation to mouse coordinates.


ammen99:
I trust you tested this ... LGTM

soreau:
Yes I tested compilation, scale and rotation.

PR #32, opened 2018-07-03
switcher: Split fast switcher into its own plugin

### LINKED ISSUE #16: Split switcher and fast-switch, and add touch support to switcher
Currently the two plugins live in the same file, but they have almost separate implementations, so it would make sense to separate them.

During this we can make the switcher act more like the fast-switch ability (so, once the user releases the modifier for activating, deactivate the plugin). This should be configurable behavior though, as it is useful for modifier activation bindings and also for touch activation.

soreau:
This only partially fixes #16 as it does not add touch support.


plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
+#include <output.hpp>
+#include <debug.hpp>
```
ammen99: it is highly probable I left this aside, but if you're not debugging, it's best to remove this

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <opengl.hpp>
```
ammen99: I think you can do without this as well ...

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <view.hpp>
+#include <workspace-manager.hpp>
+#include <linux/input-event-codes.h>
+
+/*
+ * This plugin rovides abilities to switch between views.
+ * It works similarly to the alt-esc binding in Windows or GNOME
+ */
+
+class fast_switcher : public wayfire_plugin_t
```
ammen99: add a `wayfire_` prefix

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+    key_callback init_binding;
+    wf_option activate_key;
+
+    signal_callback_t destroyed;
+
+    struct
+    {
+        bool active = false;
+
+        bool mod_released = false;
```
ammen99: I doubt you need those in fast switcher, also most of the bools below

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+        using namespace std::placeholders;
+        grab_interface->callbacks.keyboard.key = std::bind(std::mem_fn(&fast_switcher::handle_key),
+                this, _1, _2);
+
+        grab_interface->callbacks.keyboard.mod = std::bind(std::mem_fn(&fast_switcher::handle_mod),
+                this, _1, _2);
+    }
+
+    void stop_continuous_switch()
```
ammen99: no need to say `continuous_switch()` because this doesn't have two modes anyway

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+        state.in_continuous_switch = false;
+        if (state.in_switch)
+            switch_terminate();
+    }
+
+    void handle_mod(uint32_t mod, uint32_t st)
+    {
+        bool mod_released = (mod == activate_key->as_cached_key().mod && st == WLR_KEY_RELEASED);
+
+        if (mod_released && state.in_switch)
```
ammen99: you can remove the `state.in_switch` altogether because if we are in this function, there's an active grab, so we're switching

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    void handle_key(uint32_t key, uint32_t kstate)
+    {
+        if (kstate != WLR_KEY_PRESSED)
+            return;
+
+#define switch_on (state.in_switch && key == activate_key->as_cached_key().keyval)
+
+        if (!state.mod_released && (key == activate_key->as_cached_key().keyval || switch_on))
```
ammen99: dont' need this if and the body

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -0,0 +1,194 @@
...
+            state.in_continuous_switch = true;
+
+        if (switch_on && state.in_continuous_switch)
```
ammen99: remove the `in_continuous_switch`

plugins/single_plugins/meson.build
```diff
@@ -8,6 +8,7 @@ grid          = shared_module('grid',          'grid.cpp',          include_dire
 wrot          = shared_module('wrot',          'wrot.cpp',          include_directories: [wayfire_api_inc, wayfire_conf_inc], dependencies: [wlroots, pixman, wfconfig], install: true, install_dir: 'lib/wayfire/')
 expo          = shared_module('expo',          'expo.cpp',          include_directories: [wayfire_api_inc, wayfire_conf_inc], dependencies: [wlroots, pixman, wfconfig], install: true, install_dir: 'lib/wayfire/')
 switcher      = shared_module('switcher',      'switcher.cpp',      include_directories: [wayfire_api_inc, wayfire_conf_inc], dependencies: [wlroots, pixman, wfconfig], install: true, install_dir: 'lib/wayfire/')
+fast_switcher = shared_module('fast_switcher', 'fast-switcher.cpp',      include_directories: [wayfire_api_inc, wayfire_conf_inc], dependencies: [wlroots, pixman, wfconfig], install: true, install_dir: 'lib/wayfire/')
```
ammen99: maybe move this to the end, because it breaks the alignment of the last column

soreau:
I think everything has been addressed.

PR #69, opened 2018-08-26
Add alpha plugin

soreau:
This basic alpha plugin allows changes a views transparency
by hovering the mouse cursor using binding <mod> + scroll.
It is also intended to facilitate a way to adjust the initial alpha
of a view when it is first mapped.

### DIFF
--- a/plugins/single_plugins/alpha.cpp
+++ b/plugins/single_plugins/alpha.cpp
@@ -103,8 +103,7 @@ class wayfire_alpha : public wayfire_plugin_t
     {
         output->workspace->for_each_view([] (wayfire_view view)
         {
-            auto transformer = view->get_transformer("alpha").get();
-            if (transformer)
+            if (view->get_transformer("alpha"))
                 view->pop_transformer("alpha");
         }, WF_ALL_LAYERS);
 
--- a/plugins/single_plugins/alpha.cpp
+++ b/plugins/single_plugins/alpha.cpp
@@ -33,6 +33,8 @@
 class wayfire_alpha : public wayfire_plugin_t
 {
     axis_callback axis_cb;
+    wayfire_config_section *section;
+    wf_option modifier, min_value;
 
     public:
     void init(wayfire_config *config)
@@ -62,8 +64,8 @@ class wayfire_alpha : public wayfire_plugin_t
                 update_alpha(view, ev->delta);
         };
 
-        auto section = config->get_section("alpha");
-        auto modifier = section->get_option("modifier", "<alt>");
+        section = config->get_section("alpha");
+        modifier = section->get_option("modifier", "<alt>");
 
         output->add_axis(modifier, &axis_cb);
     }
@@ -77,6 +79,7 @@ class wayfire_alpha : public wayfire_plugin_t
             view->add_transformer(nonstd::make_unique<wf_2D_view> (view), "alpha");
 
         transformer = dynamic_cast<wf_2D_view*> (view->get_transformer("alpha").get());
+        min_value = section->get_option("min_value", "0.1");
         alpha = transformer->alpha;
 
         alpha -= delta * 0.003;
@@ -87,8 +90,8 @@ class wayfire_alpha : public wayfire_plugin_t
         if (alpha == 1.0)
             return view->pop_transformer("alpha");
 
-        if (alpha < 0)
-            alpha = 0;
+        if (alpha < min_value->as_double())
+            alpha = min_value->as_double();
 
         if (transformer->alpha != alpha)
         {
--- a/plugins/single_plugins/alpha.cpp
+++ b/plugins/single_plugins/alpha.cpp
@@ -59,7 +59,7 @@ class wayfire_alpha : public wayfire_plugin_t
                 return;
 
             if (ev->orientation == WLR_AXIS_ORIENTATION_VERTICAL)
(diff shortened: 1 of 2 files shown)


plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+        {
+            transformer->alpha = alpha;
+            view->damage();
+        }
+    }
+
+    void fini()
+    {
+        output->rem_axis(&axis_cb);
+    }
```
ammen99: We should also scan all views on the current output and remove any transformers left.

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        output->add_axis(modifier, &axis_cb);
+    }
+
+    void update_alpha_target(wayfire_view view, float delta)
+    {
+        auto transformer = dynamic_cast<wf_2D_view*> (view->get_transformer("alpha").get());
+        float alpha = transformer->alpha;
+
+        alpha -= delta * 0.003;
```
ammen99: maybe add an option for speed?

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        if (alpha > 1.0)
+            alpha = 1.0;
+
+        if (alpha == 1.0)
+            return view->pop_transformer("alpha");
+
+        if (alpha < 0)
```
ammen99: I'd also like an option for minimal value (so that I can't accidentally hide my windows)

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <output.hpp>
+#include <animation.hpp>
+#include "view-transform.hpp"
+#include "workspace-manager.hpp"
+#include <nonstd/make_unique.hpp>
+
+class wayfire_alpha : public wayfire_plugin_t
+{
+    axis_callback axis_cb;
+    wf_duration duration;
```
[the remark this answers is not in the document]
soreau: No it isn't. Pushed a fix to remove this and the animation header.

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+            output->deactivate_plugin(grab_interface);
+
+            auto focus = core->get_cursor_focus();
+
+            if (!focus)
+                return;
+
+            auto view = core->find_view(focus->get_main_surface());
+
```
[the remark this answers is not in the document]
soreau: Removed this newline.

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,112 @@
...
+            auto layer = output->workspace->get_view_layer(view);
+
+            if (layer == WF_LAYER_BACKGROUND)
+                return;
+
+            if (!view->get_transformer("alpha"))
```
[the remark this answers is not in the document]
soreau: Moved.

plugins/single_plugins/alpha.cpp
soreau: I pushed something that builds and works (alpha surfaces become opaque on plugin unload) but I'm not sure if it's correct. Should wrot do this too btw?

plugins/single_plugins/alpha.cpp
soreau: I'm not sure what you mean by this. How would speed value be interpreted?

plugins/single_plugins/alpha.cpp
soreau: Added min_value option.

soreau:
Regarding the brightness functionality, I think if it is added, the plugin name should be changed. For right now, I do not have an interest in this and would like to work on matching functionality to start windows with a certain alpha (like menus for example). Regarding focus-alpha, I think this belongs in a plugin that is specifically dedicated to focus-alpha functionality and we should not try to throw it all into the alpha plugin. For me, it would be ideal if this plugin remain purposed for simply changing alpha of surfaces with mod+scroll and setting the initial alpha values of windows.

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <plugin.hpp>
+#include <output.hpp>
+#include "view-transform.hpp"
+#include "workspace-manager.hpp"
+#include <nonstd/make_unique.hpp>
+
+class wayfire_alpha : public wayfire_plugin_t
+{
+    axis_callback axis_cb;
+    wayfire_config_section *section;
```
ammen99: don't store the section but the individual options

plugins/single_plugins/alpha.cpp
```diff
@@ -0,0 +1,122 @@
...
+            transformer->alpha = alpha;
+            view->damage();
+        }
+    }
+
+    void fini()
+    {
+        output->workspace->for_each_view([] (wayfire_view view)
+        {
+            auto transformer = dynamic_cast<wf_2D_view*> (view->get_transformer("alpha").get());
```
ammen99: it can be left like that if you want, but you don't need to cast it, so something like a simple `if (view->get_transformer("alpha"))` - unique_ptr can be implicitly converted to bool

PR #106, opened 2018-12-04
Blur v6

alex47:
Previous dual kawase blur implementation was not working correctly at all.
The down and upscaling was completely left out, shader "mode" was assigned in reverse.
Removed "if" from shader, it shouldn't be used.


plugins/blur/dual_kawase.cpp
```diff
@@ -0,0 +1,157 @@
...
+        OpenGL::render_begin();
+
+        programDown = OpenGL::create_program_from_source(dual_kawase_vertex_shader,
+            dual_kawase_fragment_shader_down);
+
+        posIDDown       = GL_CALL(glGetAttribLocation( programDown, "position"));
+        offsetIDDown    = GL_CALL(glGetUniformLocation(programDown, "offset"));
+        halfpixelIDDown = GL_CALL(glGetUniformLocation(programDown, "halfpixel"));
+
+	programUp = OpenGL::create_program_from_source(dual_kawase_vertex_shader,
```
ammen99: indent with spaces please

ammen99:
@alex47 Thanks for the interest in fixing up blur.

I'm sadly not familiar with how the algorithm works, lets wait for @alex47 soreau to take a look at it.

ammen99:
I just tested, this breaks blur. My guess is that we need to adjust calculation of blur radius, see for example the end of `box.cpp` (basically just override a function), but I can't be sure.

alex47:
did you try it with https://github.com/WayfireWM/wcm/pull/1 ?

ammen99:
I'm not using wcm at all. I just notice visual artifacts.

alex47:
what kind of artifacts?
could you post a screenshot?

ammen99:
https://imgur.com/a/cAhZW70

ammen99:
@alex47 can you say how those values were generated? They look way bigger than should be needed, but then again my knowledge of blur is limited.

alex47:
Those were generated manually by me. I increased them until I didn't see any artifact at the specific iteration. 
I tested these when I implemented this blur in KDE KWin, and we still use them, see: https://github.com/KDE/kwin/blob/master/effects/blur/blur.cpp#L193

Updating the content of the blurred window should no longer case artifacts.
Updating the content UNDER the blurred window causes artifacts: https://i.imgur.com/MgWeQ3D.jpg
I see every other blur has this problem too in Wayfire, so this is not specific for this method.

alex47:
>  `(1 << iterations) * offset` ?

The offset makes no difference in the calculation, only the iteration matters

ammen99:
> Updating the content of the blurred window should no longer cause artifacts.
> Updating the content UNDER the blurred window causes artifacts: https://i.imgur.com/MgWeQ3D.jpg
> I see every other blur has this problem too in Wayfire, so this is not specific for this method.

Isn't this how it is supposed to look like?

> The offset makes no difference in the calculation, only the iteration matters

Isn't offset the radius at which you sample pixels to calculate value for the current one (in the shader)?

ammen99:
I also just pushed a new commit to blur-v6, please tell me if the problems with content changing under a blurred window persist.

ammen99:
Some tests (didn't look very carefully though) show that `(1 << iterations) * offset` is enough as blur offset. For example, if `iterations = 4` `offset = 2` this gives `16 * 2 = 32` while your algo gives `150`, significantly more.

ammen99:
@alex47 do you have IRC? If so, please join the channel :) (IRC is better suited for discussions IMO)

soreau:
@alex47 Thank you for this. I have reviewed it and it looks pretty good. Just a few things:
1) blur-v6 has been force-pushed so please rebase to latest blur-v6
2) Do not s/kawase/dual_kawase/ everywhere, the kawase name is preferred internally for variables and config option names.
3) Please use this in calculate_blur_radius:
`pow(2, (wf_blur_base::iterations_opt->as_int() + 2) * offset_opt->as_double() * degrade_opt->as_int())`
4) Remove the #include "debug.hpp" from kawase.cpp

ammen99:
@alex47 We rebased you changes against blur and merged them (and blur has been merged into master): https://github.com/WayfireWM/wayfire/commit/61463746a54cbae79ce83e79e6894c1e220ef3db

Thanks for the help with dual-kawase and I hope you don't mind that we squashed everything together.

alex47:
nice
one last thing that should be added to this is to hide the iteration and offset values from the user
it's difficult for a user to pick the right values and avoid artifacts
here is an example that I wrote that can calculate sets of values on a custom scale (1-15 in this example):
https://github.com/KDE/kwin/blob/master/effects/blur/blur.cpp#L166

ammen99:
@alex47 We had such an algorithm at some point, but decided that we should let the user have full control over what happens. It might not be the easiest way to configure, but at least everybody can get what they want to.

PR #113, opened 2019-01-03
Toplevel list v2

ammen99:
With this we can implement docks, taskbars, etc.


soreau:
I looked through this and tested it with wf-dock. It works well. About the only question I have is regarding focusing of the next window. I would expect it to happen when the animation starts instead of after the animation completes. This is certainly trivial, the bulk LGTM.

ammen99:
The focus inconsistency will be fixed in another PR.

PR #114, opened 2019-01-03
cube: Fix cube when no skydome image is set

soreau:
When skydome was selected with no image set, the cube would not
activate at all. This was partially because the background methods
were clearing the depth buffer and skydome was only doing this
when there was a texture set. This makes it so the background is
cleared to green when skydome or cubemap are selected but the
texture for them are not loaded.

### DIFF
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -455,33 +455,34 @@ class wayfire_cube : public wayfire_plugin_t
     {
         update_workspace_streams();
 
-        if (program.id == (uint)-1)
+        if (program.id == (uint32_t)-1)
             load_program();
+
+        OpenGL::render_begin(dest);
+        GL_CALL(glClear(GL_DEPTH_BUFFER_BIT));
 
         reload_background();
         background->render_frame(dest, animation);
 
         auto vp = calculate_vp_matrix(dest);
-        OpenGL::render_begin(dest);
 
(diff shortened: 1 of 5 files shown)


plugins/cube/skydome.cpp
```diff
@@ -139,11 +139,14 @@ void wf_cube_background_skydome::render_frame(const wf_framebuffer& fb,
     fill_vertices();
     reload_texture();
 
-    if (tex == (uint32_t)-1)
+    if (tex == (uint)-1)
```
ammen99: I generally prefer to use the full `uint32_t`

plugins/cube/cube.cpp
```diff
@@ -479,6 +467,19 @@ class wayfire_cube : public wayfire_plugin_t
             0.0f, 0.0f
         };
 
+        update_workspace_streams();
```
ammen99: actually if you're gonna move things, the `vertexData` should be declared right before uploading it, also make it static please.

plugins/cube/skydome.cpp
soreau: I went ahead and changed all instances of `uint` to `uint32_t` in the cube plugin.

plugins/cube/cube.cpp
soreau: This should be better now.

ammen99:
Thanks!

PR #117, opened 2019-01-04
Alpha fixup

soreau:


plugins/single_plugins/alpha.cpp
```diff
@@ -99,6 +77,47 @@ class wayfire_alpha : public wayfire_plugin_t
...
+
+        if (ev->orientation == WLR_AXIS_ORIENTATION_VERTICAL)
+            update_alpha(view, ev->delta);
+    };
+
+    wf_option_callback min_value_changed = [=] ()
+    {
+        output->workspace->for_each_view([=] (wayfire_view view)
+        {
+            wf_2D_view *transformer;
```
ammen99: why not move that below (`wf_2D_view *transformer = ...`)

plugins/single_plugins/alpha.cpp
```diff
@@ -99,6 +77,47 @@ class wayfire_alpha : public wayfire_plugin_t
...
+
+            if (!view->get_transformer("alpha"))
+                return;
+
+            transformer = dynamic_cast<wf_2D_view*> (view->get_transformer("alpha").get());
+
+            if (transformer->alpha < min_value->as_double())
+                transformer->alpha = min_value->as_double();
+
+            view->damage();
```
ammen99: theoretically no need to damage if the alpha wasn't changed, could be moved inside the `if` on line 113

plugins/single_plugins/alpha.cpp
soreau: I was trying to keep declarations before code. Changed to the way you suggest.

plugins/single_plugins/alpha.cpp
soreau: True, fixed.

plugins/single_plugins/alpha.cpp
ammen99: as a general notice this isn't done in C++ at all AFAIK, and certainly isn't done in wayfire except perhaps at a limited number of places.

soreau: Noted.

ammen99:
Thanks!

PR #120, opened 2019-01-16
Example config file fixups

soreau:


wayfire.ini.default
```diff
@@ -44,36 +44,13 @@ disable_while_typing = 1
 cursor_size = 24
 cursor_theme = default
 
-xkb_layout = us,bg
+#xkb_layout = us,bg
```
ammen99: We might want to comment all the `xkb_*` options, too.

Issue #110, opened 2018-12-23
Send popup_done to dismiss xdg-popups


ammen99:
This is actually required by the xdg-shell protocol.

PR #121, opened 2019-01-16
FreeBSD support + catching up to wlroots updates + GLES2 typo fix

myfreeweb:
Hi, first of all, thanks for this project, this is EXACTLY what I've been looking for! I've started working on [a DE](https://github.com/myfreeweb/numbernine), and looks like Wayfire will be the best compositor for me. (And I'm so happy with wobbly windows!! :D)

This PR adds FreeBSD support and updates to latest wlroots master.


meson.build
```diff
@@ -54,14 +60,18 @@ if get_option('enable_graphics_debug')
   conf_data.set('WAYFIRE_GRAPHICS_DEBUG', true)
 endif
 
-if get_option('enable_gles32') and meson.get_compiler('cpp').has_header('GLES3/gl32.h')
+if get_option('enable_gles32') and meson.get_compiler('cpp').has_header('GLES3/gl32.h', args: '-I/usr/local/include')
```
soreau: It seems like this should be from `pkg-config --variable=includedir glesv2` with PKG_CONFIG_PATH correctly set to e.g. /usr/local/lib/pkgconfig

ammen99: Yes, I agree, even this might break on some non-standard distros like NixOS.

src/view/xdg-shell.cpp
```diff
@@ -338,7 +338,8 @@ std::string wayfire_xdg_view::get_title()
 
 void wayfire_xdg_view::close()
 {
-    wlr_xdg_surface_send_close(xdg_surface);
+    wlr_xdg_popup_destroy(xdg_surface);
```
ammen99: This function should just do a `wlr_xdg_toplevel_send_close` (views aren't popups in any case). For popups, we should add an override `wayfire_xdg_popup::close` which calls the `wlr_xdg_popup_destroy`, but that isn't necessary until #110  is resolved.

myfreeweb:
Yeah, sure.

PR #128, opened 2019-01-19
Add more input settings

myfreeweb:
settings

(that screenshot of n9-settings is from weston, on wayfire it doesn't show current devices yet, and there are some other changes)

### DIFF
--- a/src/core/seat/cursor.cpp
+++ b/src/core/seat/cursor.cpp
@@ -185,8 +185,11 @@ void input_manager::handle_pointer_axis(wlr_event_pointer_axis *ev)
         return;
     }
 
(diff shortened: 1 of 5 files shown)


src/core/seat/cursor.cpp
```diff
@@ -208,12 +211,15 @@ wf_cursor::wf_cursor()
     wl_signal_add(&cursor->events.motion_absolute, &motion_absolute);
     wl_signal_add(&cursor->events.axis, &axis);
 
-    init_xcursor();
-
     config_reloaded = [=] (signal_data*) {
         init_xcursor();
+        auto section = core->config->get_section("input");
```
ammen99: You need to get options just once, `wf_option` is actually an alias for a shared_ptr.

Issue #135, opened 2019-01-22
Build failure - GL_PATCHES was not declared in this scope


colemickens:
I'm not sure what I'm doing wrong, but my `wayfire` package definition stopped building recently:

```
[139/155] Compiling C++ object 'plugins/cube/643b030@@cube@sha/cube.cpp.o'.
FAILED: plugins/cube/643b030@@cube@sha/cube.cpp.o
/nix/store/isg8rxaxkijl9x3hr2gzsf8pqfnqxg3k-gcc-wrapper-7.4.0/bin/c++ -Iplugins/cube/643b030@@cube@sha -Iplugins/cube -I../plugins/cube -I../src/api -I. -I../ -I/nix/store/ppg31x6w5ss6rslacpbyph3sl58gblhq-wlroots-36ba56140ae325ae398229a689f20ded126608bc/include -I/nix/store/l0c5zzypk9n9qwvz3n7iai415q3shy5i-pixman-0.36.0/include/pixman-1 -I/nix/store/k2nwdw4wl1q3pazlz5lw70d7p6y2hg6b-wf-config-6d3426e216ac62ffa035035f9c1bea074e184018/include/wayfire -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -std=c++11 -O3 -DWLR_USE_UNSTABLE -Wno-unused-parameter -fPIC  -MD -MQ 'plugins/cube/643b030@@cube@sha/cube.cpp.o' -MF 'plugins/cube/643b030@@cube@sha/cube.cpp.o.d' -o 'plugins/cube/643b030@@cube@sha/cube.cpp.o' -c ../plugins/cube/cube.cpp
../plugins/cube/cube.cpp: In member function 'void wayfire_cube::load_program()':
../plugins/cube/cube.cpp:184:26: warning: unused variable 'tcs' [-Wunused-variable]
         GLuint vss, fss, tcs = -1, tes = -1, gss = -1;
                          ^~~
../plugins/cube/cube.cpp:184:36: warning: unused variable 'tes' [-Wunused-variable]
         GLuint vss, fss, tcs = -1, tes = -1, gss = -1;
                                    ^~~
../plugins/cube/cube.cpp:184:46: warning: unused variable 'gss' [-Wunused-variable]
         GLuint vss, fss, tcs = -1, tes = -1, gss = -1;
                                              ^~~
[...]

PR #137, opened 2019-01-23
cube: Build without gles32

soreau:


soreau:
I forgot to mention this should fix #135.

PR #129, opened 2019-01-20
Track wlroots changes

soreau:

PR #138, opened 2019-01-23
Fix 110 popup done

### LINKED ISSUE #110: Send popup_done to dismiss xdg-popups

ammen99:
Stacked on top of #129 
Fixes #110

### DIFF
--- a/src/core/wm.cpp
+++ b/src/core/wm.cpp
@@ -4,6 +4,10 @@
 #include "debug.hpp"
 #include "core.hpp"
 #include "workspace-manager.hpp"
+
+#include "../view/xdg-shell.hpp"
+#include "../view/xdg-shell-v6.hpp"
+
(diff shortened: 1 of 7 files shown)


soreau:
@ammen99 I was able to reproduce a crash you mentioned would be possible on IRC. [Here](https://www.hastebin.com/wogazuxico.sql) is the tail of the log containing the bt. This should be a blocker since it crashes the display server.

ammen99:
@soreau It should be fixed now, can you test again?

soreau:
@ammen99 I tested fix-110-popup-done and master branch, I couldn't reproduce the problem on either, though I'm not sure if I'm doing it right. I'm just dragging a window to the second output, clicking the close button and clicking on the panel or another window in the first output.

PR #142, opened 2019-01-27
cursor: do not set axis events if an axis binding has consumed it

### LINKED ISSUE #107: Don't pass button and axis events to clients for bindings
When using a button or axis binding on a view, the event is passed to the client as if no binding were matched. For example, if alt+drag happens when grabbing the close button of a decoration, the close button receives an event, activating the close sequence. In this case it causes wayfire to crash after the close animation ends. Perhaps the expected behavior would be to not send button or axis events to clients that were matched in a binding.

ammen99:
Fixes #107 

@soreau Can you test? I also can't reproduce the button issue, perhaps it is gone?

### DIFF
--- a/src/core/seat/cursor.cpp
+++ b/src/core/seat/cursor.cpp
@@ -8,8 +8,12 @@
 static void handle_pointer_button_cb(wl_listener*, void *data)
 {
     auto ev = static_cast<wlr_event_pointer_button*> (data);
-    core->input->handle_pointer_button(ev);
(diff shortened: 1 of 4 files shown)


soreau:
I tested and the axis issue is fixed. However the button issue remains. To reproduce: Alt+Click on wcm button. Expected: Since it's the move binding, it shouldn't click the button below the cursor. Actual: It does.

soreau:
Yes, I tested this and it fixes the axis and button bindings events being sent to clients.

PR #145, opened 2019-01-27
Add option to repeat key/button binding for command

ammen99:
Command now supports a new kind of binding, with the prefix `repeatable_`. This will make the command to be executed repeatedly after a certain delay. Works for key- and button activated commands.

Still TODO: enable the user to configure the repeat rate and delay. Do we reuse core options? Or do we add specific options for the command plugin?

### DIFF
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -106,12 +106,12 @@ class wayfire_cube : public wayfire_plugin_t
         };
 
         auto key_left = section->get_option("rotate_left", "<alt> <ctrl> KEY_LEFT");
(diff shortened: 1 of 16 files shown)


soreau:
It seems to me a key repeat rate option should be a core option. Setting different repeat speeds for different plugins seems archaic. In my opinion ,the key repeat should affect all plugins with keybindings, not just command.

ammen99:
@soreau Ok, updated to use core options.
@AdrianVovk I hope this is what you needed?

AdrianVovk:
@ammen99 What does using this look like? How would I configure it?

PR #147, opened 2019-01-28
Support new protocols

myfreeweb:
This goes after #129. This is separate from #129 as this adds new optional protocols.


ammen99:
Thanks!

PR #157, opened 2019-02-13
Improve modifier bindings

myfreeweb:
- remember which actual key activated a modifier binding (so that an [xcape](https://github.com/alols/xcape)-style plugin would be able to distinguish between left and right shift to do shifts-as-parens)
- add a configurable timeout for modifier bindings (nothing happens when a modifier is held for too long)
- use xkbcommon to apply the keymap to the modifier keys (so any xkb_options like swapping Ctrl and Caps Lock or the choice of Super would apply) (note: rootston uses xkbcommon for *everything* in bindings)

### DIFF
--- a/src/core/seat/input-manager.hpp
+++ b/src/core/seat/input-manager.hpp
@@ -93,9 +93,10 @@ class input_manager
         /* TODO: move this in a wf_keyboard struct,
          * This might not work with multiple keyboards */
         bool in_mod_binding = false;
+        uint32_t mod_binding_key = 0; /* e.g. whether L or R Shift caused a <shift> binding */
         std::chrono::steady_clock::time_point mod_binding_start;
         int count_other_inputs = 0;
-        std::vector<std::function<void()>> match_keys(uint32_t mods, uint32_t key);
+        std::vector<std::function<void()>> match_keys(uint32_t mods, uint32_t key, uint32_t mod_binding_key = 0);
 
         wayfire_view keyboard_focus;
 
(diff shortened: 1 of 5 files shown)


ammen99:
Hm, do we have to save the modifier for modifier binding? Can't we somehow get it from the wlroots pressed keys list? (Not sure if modifier is first removed from the list  or not though)

myfreeweb:
I can't find a pressed keys list in wlroots… I only see rootston keeping its own lists

src/core/seat/input-manager.hpp
```diff
@@ -92,8 +93,10 @@ class input_manager
         /* TODO: move this in a wf_keyboard struct,
          * This might not work with multiple keyboards */
         bool in_mod_binding = false;
+        uint32_t mod_binding_key = 0; /* e.g. whether L or R Shift caused a <shift> binding */
```
ammen99: Comment should say "The keycode which triggered the modifier binding". Also I think we no longer need `in_mod_binding`, if `mod_binding_key = 0` then this means we have no modifier binding.

src/core/seat/keyboard.cpp
```diff
@@ -171,8 +181,8 @@ std::vector<std::function<void()>> input_manager::match_keys(uint32_t mod_state,
             /* We must be careful because the callback might be erased,
              * so force copy the callback into the lambda */
             auto callback = binding->call.key;
-            callbacks.push_back([key, callback] () {
-                (*callback) (key);
+            callbacks.push_back([key, callback, mod_binding_key] () {
```
ammen99: you can just create a `uint32_t actual_key = .. ? ...` before both loops to avoid capturing both variables and having the check `key == 0` twice.

ammen99:
Just tested it, works pretty good. :+1: for the modifier binding timeout.

PR #163, opened 2019-02-16
Introduce `wf::safe_list_t`

ammen99:
We had a whole class of bugs which luckily haven't been hit yet. The difficulty when storing callbacks is that if one callback deletes another, we'd have still called the deleted one, resulting in a crash. The safe list type tries to provide the ability to erase elements from the list while iterating it.

Additionally animations have been reworked so that we avoid race conditions like map->unmap->map in a very short time.


ammen99:
There is a regression: Start Wayfire, go workspace down, open and close Expo. Screen starts to flicker.

PR #175, opened 2019-03-11
Implement foreign toplevel close handler

### LINKED ISSUE #174: Implement zwlr_foreign_toplevel_handle_v1_close()
The code is simply missing.

soreau:
Calling zwlr_foreign_toplevel_handle_v1_close() from a client using
the foreign toplevel protocol did not work because it was not
implemented. Now clients can request toplevels be closed and it
will actually work. Fixes #174.

### DIFF
--- a/src/view/view.cpp
+++ b/src/view/view.cpp
@@ -94,6 +94,14 @@ static void handle_toplevel_handle_v1_set_rectangle_request(wl_listener*, void *
     view->handle_minimize_hint(box);
 }
 
+static void handle_toplevel_handle_v1_close_request(wl_listener*, void *data)
+{
+    auto toplevel = static_cast<wlr_foreign_toplevel_handle_v1*> (data);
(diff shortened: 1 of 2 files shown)


src/view/view.cpp
```diff
@@ -1132,6 +1145,12 @@ void wayfire_view_t::minimize_request(bool state)
     }
 }
 
+void wayfire_view_t::close_request()
+{
+    destroy_toplevel();
```
ammen99: we don't need to destroy the toolevel here. The client might not close its window. In fact, the close() function is actually just a request to the client to close.

ammen99:
looking at the code, I'd say you can directly call `view->close`, so no nees to add a `close_request` function at all. aside from that lgtm

soreau:
Yes, that makes sense. Done.

ammen99:
Thanks!

Issue #97, opened 2018-10-30
Default config in /etc


ammen99:
It might be desirable to not require the user to install any config files and/or manually copy them from the repo. We can just install the default one to somewhere in `/etc` and fallback to it if no config in the home directory. Same for `wf-shell`

ammen99:
Fixed in #374

PR #191, opened 2019-03-17
Use system-wide configuration file if user does not have its own

wladmis:


src/main.cpp
```diff
@@ -155,11 +155,17 @@ int main(int argc, char *argv[])
...
+    const char *config_basename = "wayfire.ini";
 
     std::string config_dir = nonull(getenv("XDG_CONFIG_DIR"));
     if (!config_dir.compare("nil"))
         config_dir = std::string(nonull(getenv("HOME"))) + "/.config/";
-    config_file = config_dir + "wayfire.ini";
+    config_file = config_dir + config_basename;
+
+    /* use system-wide config file if local is not accessible */
+    /* FIXME: there can be a race */
```
soreau: Can you elaborate on this race potential here?

meson.build
```diff
@@ -90,6 +95,8 @@ subdir('src')
 subdir('plugins')
 
 install_subdir('shaders', install_dir: 'share/wayfire')
+install_data('wayfire.ini.default', rename: 'wayfire.ini',
+               install_dir: join_paths(sysconfdir, 'wayfire'))
```
soreau: I don't think we should install the default config file. This would require installation as root. Ideally, wayfire defaults should be enough to get things up and running without a config file. Perhaps wf-panel and wf-background could be ran by default if no config file is found, for example. Perhaps a choice background image could be used by default. The rest of the options already have hard-coded defaults.

The logic of looking for the config file in sysconfigdir as a fallback is fine, I just don't think the file should be installed by default. Perhaps another option could be added that installs the config file but is disabled by default.

src/main.cpp
wladmis: This can happen in case when an user simultaneously run Wayfire and erases the config file. The right thing to do is to fallback to system config file when Wayfire fails to open user config.

meson.build
wladmis: The default paths used for installation already require privileges. But I like idea with hard-coded defaults instead of default config installation but it could be undesirable for some use-cases.

ammen99: I like this idea, and see no problem with installing a default file, many projects do that. The only thing I want is that if the user specifies a nonstandard prefix, then the config file is installed to that prefix, thus requiring no root access.

ammen99:
Btw this would fix https://github.com/WayfireWM/wayfire/issues/97

src/main.cpp
ammen99: I think this is fine, if the config file is erased, we just stop reloading changes dynamically (at least in theory).

meson.build
wladmis: > The only thing I want is that if the user specifies a nonstandard prefix, then the config file is installed to that prefix, thus requiring no root access.
It already works that way. I build rpm package using that.

soreau: If the config file is to be installed, it probably should be processed by meson so the default paths hard coded to /usr in the config file are changed to whatever prefix wayfire is installed with, and perhaps rename wayfire.ini.default to wayfire.ini.in. I think the only two places are in [autostart] section for wf-panel and wf-background.

ammen99: @wladmis So then it is ok I think. But it would be cool if you could add what @soreau described.

wladmis: I would try $XDG_CONFIG_DIR, $HOME/.config, /etc/wayfire and then /usr/share/wayfire paths order to find the config.

ammen99: /usr/share/wayfire??

meson.build
```diff
@@ -7,12 +7,16 @@ project(
...
-        'c_std=c11',
+		'c_std=c11',
 		'warning_level=2',
 		'werror=false',
+		'sysconfdir=/etc',
 	],
 )
 
+prefixdir=get_option('prefix')
+sysconfdir=join_paths(prefixdir, get_option('sysconfdir'))
```
ammen99: hm, I am not sure about this one. This will be something like `/usr/etc` which is a pretty weird place to put config into. Maybe just `sysconfdir`?

soreau: You're saying check if the prefix is the default of /usr/local and if so, use /etc/wayfire/ for the config directory, otherwise, use $prefix/etc/wayfire/? This doesn't sound right. For starters, how do you determine if --prefix=/usr/local was passed or not?

ammen99:
Okay, after a bit of discussion in IRC we want the following behavior:

1. Install the default config file to `<sysconfdir>/wayfire/`. Notice this is NOT `<prefix>/<sysconfdir>`
2. Have an option to not install the config file at all

ammen99:
@wladmis Are you going to finish this PR? No pressure, I can pick your changes and finalize things if you have no time/desire.

wladmis:
> @wladmis Are you going to finish this PR? No pressure, I can pick your changes and finalize things if you have no time/desire.

Sorry guys, I'm not sure I'll do it any time soon :(

ammen99:
This is going to be implemented as part of wf-config.

PR #208, opened 2019-04-19
[WIP] Stabilize api

ammen99:
This first needs another wlroots release so that we can merge the `track-wlroots` branch.

The goal is to have well-defined API for plugins, which means that all internal implementations (private variables, non-API methods,etc.) must be hidden either via PIMPL or abstract base classes. This is also a great opportunity to document what most of the API functions do, change names for some functions, etc.

What classes are affected:

- [x] render-manager

- [x] wayfire_output

- [x] wayfire_core - should be made a singleton

- [ ] wayfire_view,  wayfire_surface and compositor views/surfaces

- [ ] opengl - hide non-api functions

Others like `workspace-manager`, `output-layout` are already in a good shape.

After this is done, the last step would be:

- [ ] Implement versioning of the API/ABI and enforce checks when loading plugins


soreau:
With this PR, gridding a wobbly window sets the position down and to the right 100,100 until the wobbly animation is finished, then it snaps back to the correct position.

ammen99:
Closing in favor of #243

PR #212, opened 2019-04-27
zoom: Don't sample from undefined area

### LINKED ISSUE #211: Fix zoom with multiple outputs
We should confine the zoomed region to the output on which zoom is active, otherwise we sample from undefined area.

soreau:
Fixes #211.

### DIFF
--- a/plugins/single_plugins/zoom.cpp
+++ b/plugins/single_plugins/zoom.cpp
@@ -68,10 +68,13 @@ class wayfire_zoom_screen : public wayfire_plugin_t
         {
             auto w = output->handle->width;
             auto h = output->handle->height;
-            GetTuple(x, y, output->get_cursor_position());
+            GetTuple(_x, _y, output->get_cursor_position());
+            double x, y;
+            wlr_box b = output->get_relative_geometry();
+            wlr_box_closest_point(&b, _x, _y, &x, &y);
 
             /* get rotation & scale */
-            wlr_box box = {x, y, 1, 1};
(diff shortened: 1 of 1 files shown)


plugins/single_plugins/zoom.cpp
```diff
@@ -39,6 +40,9 @@ class wayfire_zoom_screen : public wayfire_plugin_t
 
             duration = wf_duration(smoothing_duration);
             duration.start(1, 1); // so that the first value we get is correct
+
+            GetTuple(x, y, output->get_cursor_position());
+            last_cursor_position = wf_point{x, y};
```
ammen99: You should still use `wlr_box_closest_point` here, nothing guarantees that the cursor will be on the output (since an output is focused when you click on it)

soreau: I don't really understand what you mean. Do you want the zoom area to move even though the pointer is not on the output?

ammen99: Okay, maybe we need another check. What I mean is this: the scroll binding will be sent to the focused output, not the output that the mouse is on.

soreau: Yes, that is how it currently works. How should it work instead?

soreau: I updated to use `wlr_box_closest_point` and it still works the same way except the zoom area moves even when the cursor is not on the output. Is this what you wanted?

ammen99:
Thanks!

PR #214, opened 2019-05-01
api: add missing include <algorithm> for std::min/max

myfreeweb:
Looks like this problem might have been hidden by gnu libstdc++ including it somewhere, with llvm libc++ std::min/max are not found w/o `<algorithm>`


ammen99:
Thanks!

PR #215, opened 2019-05-01
[track-wlroots] output-layout, render-manager: track frame submission API changes

myfreeweb:
(swaywm/wlroots#1660)


ammen99:
Thanks!

PR #218, opened 2019-05-06
Fixup x11 backend

soreau:
This fixes a couple problems that were causing black window with x11 backend.


ammen99:
+1, thanks for investigating!

PR #231, opened 2019-06-03
Fix RTTI/dynamic_cast across plugins

myfreeweb:
RTTI is awesome https://stackoverflow.com/a/2352208

This fixes snap preview assertion failures and crashes.


ammen99:
Thanks!

PR #233, opened 2019-06-06
fast-switcher: fix SIGFPE (modulo by zero) when last window is destroyed

myfreeweb:
Modulo by zero exception (views.size is zero): https://gist.github.com/myfreeweb/b6365b31dc3b0b416243c6b291221e57

> how i got that is by fast-switching on an output where the only window was from a hanging client, which could've crashed or something

I *guess* deactivating makes sense but idk.


ammen99:
I already pushed the fix :)

PR #241, opened 2019-06-12
grid: Add way to restore with keybinding

### LINKED ISSUE #240: Grid: Add a restore keybinding
For reference: GNOME's keybindings

soreau wants an option to toggle when repeating the same button, so maybe if the option = "last" function like soreau intends.

Example:
```
output->add_activator(restore_opt, restore());
output->add_activator(opts[5], center());
[...]

center() {
   if (restore_opt->as_string() == "last" && view->maximized) {
      restore();
      return;
   }
   maximize();
}

[...]
```

soreau:
The default with no keybinding specified is to restore when the same keybinding
is used twice. Set 'restore' key in [grid] section to a keybinding to use a
dedicated keybinding instead. Fixes #240.

### DIFF
--- a/plugins/single_plugins/grid.cpp
+++ b/plugins/single_plugins/grid.cpp
@@ -222,6 +222,23 @@ class wayfire_grid : public wf::plugin_interface_t
 
     wf_option animation_duration, animation_type;
 
+    wf_option restore_opt;
+    std::string restore_opt_str;
+    const std::string restore_opt_default = "toggle";
+
+    activator_callback restore = [=] (wf_activator_source, uint32_t)
+    {
+        auto view = output->get_active_view();
+        handle_slot(view, 0, true);
+    };
+    wf_option_callback restore_opt_changed = [=] ()
+    {
+        output->rem_binding(&restore);
+        restore_opt_str = restore_opt->as_string();
+        if (restore_opt_str != restore_opt_default)
+            output->add_activator(restore_opt, &restore);
+    };
+
     nonstd::observer_ptr<wayfire_grid_view_cdata>
         ensure_grid_view(wayfire_view view)
     {
@@ -255,6 +272,11 @@ class wayfire_grid : public wf::plugin_interface_t
             output->add_activator(keys[i], &bindings[i]);
         }
 
+        restore_opt = section->get_option("restore", restore_opt_default);
+        restore_opt_str = restore_opt->as_string();
+        restore_opt_changed();
+        restore_opt->add_updated_handler(&restore_opt_changed);
+
         output->connect_signal("reserved-workarea", &on_workarea_changed);
         output->connect_signal("view-snap", &on_snap_signal);
         output->connect_signal("query-snap-geometry", &on_snap_query);
@@ -265,11 +287,19 @@ class wayfire_grid : public wf::plugin_interface_t
     void handle_slot(wayfire_view view, const wf_geometry workarea, int slot, bool comes_from_maximize_request)
     {
         wf_geometry target = get_slot_dimensions(slot, workarea);
-        if (view->maximized && view->get_wm_geometry() == target)
+        if (view->tiled_edges && view->get_wm_geometry() == target)
         {
-            return;
+            if (restore_opt_str == restore_opt_default)
+            {
+                slot = 0;
+                comes_from_maximize_request = true;
+            }
+            else
+            {
+                return;
+            }
         }
-        else if (!view->has_data<wf_grid_saved_view_geometry>() && slot)
+        if (!view->has_data<wf_grid_saved_view_geometry>() && slot)
         {
             view->get_data_safe<wf_grid_saved_view_geometry>()
                 ->geometry = view->get_wm_geometry();
@@ -277,6 +307,7 @@ class wayfire_grid : public wf::plugin_interface_t
         else if (view->has_data<wf_grid_saved_view_geometry>() && slot == 0)
         {
             target = view->get_data_safe<wf_grid_saved_view_geometry>()->geometry;
+
             /* If we have a request from another plugin, for ex. move, we want
              * to make sure the view stays under the pointer. However, when it was
              * a normal restore caused by a client request, we want to restore the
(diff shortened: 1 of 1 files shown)


ammen99:
please reopen against master.

soreau:
> please reopen against master.

Done :thumbsup:

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
...
+            snap_wobbly(view, target, false);
+        }
+
+        view->erase_data<wf_grid_saved_view_geometry>();
+    }
+
     void handle_slot(wayfire_view view, const wf_geometry workarea, int slot, bool comes_from_maximize_request)
     {
         wf_geometry target = get_slot_dimensions(slot, workarea);
-        if (view->maximized && view->get_wm_geometry() == target)
```
ammen99: We DO need to check `view->maximized` (actually `view->tiled_edges`) since we can have a window which already has "good" size but isn't maximized/tiled.

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
...
+        restore_opt->add_updated_handler(&restore_opt_changed);
+
         output->connect_signal("reserved-workarea", &on_workarea_changed);
         output->connect_signal("view-snap", &on_snap_signal);
         output->connect_signal("query-snap-geometry", &on_snap_query);
         output->connect_signal("view-maximized-request", &on_maximize_signal);
         output->connect_signal("view-fullscreen-request", &on_fullscreen_signal);
     }
 
+    void restore_view(wayfire_view view)
```
ammen99: Why do we need `restore_view()`? AFAICT `handle_slot(.., slot = 0)` does all we need.

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
             output->add_activator(keys[i], &bindings[i]);
         }
 
+        restore = [=] (wf_activator_source, uint32_t)
+        {
+            auto view = output->get_active_view();
+            restore_view(view);
+        };
+
+        restore_opt = section->get_option("restore", "last");
```
ammen99: not sure about "last", maybe "toggle" ?

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
...
+        restore_opt_str = restore_opt->as_string();
+        restore_opt_changed = [=] ()
```
ammen99: Please move the lambda outside of the function whenever possible (here, it is)

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
             output->add_activator(keys[i], &bindings[i]);
         }
 
+        restore = [=] (wf_activator_source, uint32_t)
```
ammen99: Lambda should be outside of the function.

plugins/single_plugins/grid.cpp
```diff
@@ -255,24 +261,69 @@ class wayfire_grid : public wf::plugin_interface_t
...
+        {
+            auto view = output->get_active_view();
+            restore_view(view);
+        };
+
+        restore_opt = section->get_option("restore", "last");
+        restore_opt_str = restore_opt->as_string();
+        restore_opt_changed = [=] ()
+        {
+            if (restore_opt_str != "last")
```
ammen99: IIRC you don't need to check, if the binding isn't found, `rem_binding` is a no-op.
And then you also don't need `restore_opt_str` at all.

plugins/single_plugins/grid.cpp
```diff
@@ -222,6 +223,11 @@ class wayfire_grid : public wf::plugin_interface_t
 
     wf_option animation_duration, animation_type;
 
+    activator_callback restore;
```
ammen99:
```suggestion
    activator_callback restore = [=] (wf_activator_source, uint32_t)
    {
    ...
    }
```

soreau:
@ammen99 I pushed a v2 of the patch that should address your concerns, let me know what you think.

plugins/single_plugins/grid.cpp
```diff
@@ -265,24 +288,29 @@ class wayfire_grid : public wf::plugin_interface_t
     void handle_slot(wayfire_view view, const wf_geometry workarea, int slot, bool comes_from_maximize_request)
     {
         wf_geometry target = get_slot_dimensions(slot, workarea);
-        if (view->maximized && view->get_wm_geometry() == target)
+        if (view->tiled_edges && view->get_wm_geometry() == target)
         {
-            return;
+            if (restore_opt_str == restore_opt_default)
```
ammen99: Add braces here. Preferably on the same line as the if/else since both statements are short enough.

plugins/single_plugins/grid.cpp
```diff
@@ -222,6 +222,24 @@ class wayfire_grid : public wf::plugin_interface_t
 
     wf_option animation_duration, animation_type;
 
+    wf_option restore_opt;
+    std::string restore_opt_str;
+    int last_slot = 0;
```
ammen99: wait, what? Grid can be executed on different windows in different order.

plugins/single_plugins/grid.cpp
```diff
@@ -265,24 +288,29 @@ class wayfire_grid : public wf::plugin_interface_t
...
         else if (view->has_data<wf_grid_saved_view_geometry>() && slot == 0)
         {
-            target = view->get_data_safe<wf_grid_saved_view_geometry>()->geometry;
             /* If we have a request from another plugin, for ex. move, we want
              * to make sure the view stays under the pointer. However, when it was
              * a normal restore caused by a client request, we want to restore the
              * view to where it was */
-            if (!comes_from_maximize_request)
+            wf_geometry last_target = get_slot_dimensions(last_slot, workarea);
+            if (!comes_from_maximize_request && (view->get_wm_geometry() != last_target))
```
ammen99: can you explain why we need the second condition?

soreau: I see it's wrong, will fix.

ammen99:
WFM, thanks for the patience!

PR #242, opened 2019-06-15
Option to disable Xwayland, missing includes, dead code etc.

myfreeweb:


src/core/output-layout.cpp
```diff
@@ -158,6 +158,7 @@ namespace wf
         for (auto& view : views)
             from->workspace->remove_view(view);
 
+        if (to) /* views would be empty if !to, but clang-analyzer detects null deref */
```
ammen99: okay, but then this needs braces since it is a multi-line if.

ammen99:
LGTM, just add braces around the `if` in `output-layout.cpp`.

myfreeweb:
done

PR #244, opened 2019-06-18
blur: use cached config functions

myfreeweb:
kawase caught my eye when DTracing malloc..

---

also https://github.com/WayfireWM/wf-config/pull/11


ammen99:
Makes sense, thanks!

PR #246, opened 2019-06-20
Remove protocols removed by wlroots

myfreeweb:
https://github.com/swaywm/wlroots/commit/3dec88e4555ee5fd95ffd69133623816cb0c25c4

(I don't think we need a track-wlroots branch yet, since this change is backwards compatible)


ammen99:
Thanks!

PR #249, opened 2019-06-22
vswipe: touchpad swipe support for switching workspaces

myfreeweb:
This is the proper nice trackpad workspace switching, like on macOS, not the cheap fake discrete gesture thing many people settle for with external programs like libinput-gestures. With this, you can "play around", move the viewport left-right-left-right, which is very relaxing, at least for me :)

- [x] "rubberband" "soft limit" for swipe distance (is there a proper word for this?), if you go too far (more than one workspace) you're smoothly slowed to a crawl, but you can change direction and that is not slow
- [x] ultra fast jerky motion can change direction instantly (you could be very far into the workspace to the right, and a very sudden swipe to the left would go to the workspace to the left)
- [x] works properly with multiple outputs, respects their independence
- [x] the slowdown should begin instantly when going out of bounds, i.e. going to the left of the leftmost workspace should not pretend there's an extra `-1`st workspace there
- [ ] enqueue multiple super-quick swipes instead of ignoring when animation is still going. or have the swipe take over from the animation? (that one sounds hard)
- [x] these events should not go through to apps — update/end events should go through the grab? (grab does not work for me on an unfocused output!)
- [x] if this should be a separate plugin, how would it interact with vswitch? grabbing would prevent another plugin from doing its thing?
- [x] reduce copy-paste?
- [x] [EXTERNAL] libinput currently delays swipe events because it thinks that 3-4 finger gestures could be pinches. which doesn't make sense. feels so much better when this is patched out. I'll discuss that in the libinput issue tracker.
- [x] [FUTURE] only moving xdg surfaces looks awkward (moving between empty workspaces looks like nothing is happening at all), layer-shell panel/background should also move. that would need mirroring, right?
- [x] [FUTURE] when we have the above, I'd love to see gaps (black bars) between the workspaces

### DIFF
--- a/plugins/single_plugins/vswipe.cpp
+++ b/plugins/single_plugins/vswipe.cpp
@@ -1,0 +1,262 @@
+#include <plugin.hpp>
+#include <output.hpp>
+#include <core.hpp>
+#include <debug.hpp>
+#include <view.hpp>
+#include <view-transform.hpp>
+#include <render-manager.hpp>
+#include <workspace-stream.hpp>
+#include <workspace-manager.hpp>
+#include <signal-definitions.hpp>
+#include <glm/glm.hpp>
+#include <glm/gtc/matrix_transform.hpp>
+#include <animation.hpp>
+
+#include <cmath>
+#include <utility>
+#include <animation.hpp>
+
+#include "vswipe-processing.hpp"
+
+class vswipe : public wf::plugin_interface_t
+{
+    private:
+        struct {
+            wf::workspace_stream_t left, curr, right;
+        } streams;
+
+        struct {
+            bool swiping = false;
+            double dx = 0.0;
+            double gap = 0.0;
+
+            double delta_prev = 0.0;
+            double delta_last = 0.0;
+
+            int vx = 0;
+            int vy = 0;
+            int vw = 0;
+            int vh = 0;
+        } state;
+
+        wf::render_hook_t renderer;
+
+        wf_duration duration;
+        wf_transition transition;
+
+        wf_option animation_duration;
+        wf_option background_color;
+        wf_option enable;
+        wf_option ignore_cancel;
+        wf_option fingers;
+        wf_option gap;
+        wf_option threshold;
+        wf_option delta_threshold;
+        wf_option speed_factor;
+        wf_option speed_cap;
+
+    public:
+
+    void init(wayfire_config *config)
+    {
+        grab_interface->name = "vswipe";
+        grab_interface->capabilities = wf::CAPABILITY_MANAGE_COMPOSITOR;
+        grab_interface->callbacks.cancel = [=] () { finalize_and_exit(); };
+
+        auto section = config->get_section("vswipe");
+
+        animation_duration = section->get_option("duration", "180");
+        duration = wf_duration(animation_duration);
+
+        enable = section->get_option("enable", "1");
+        ignore_cancel = section->get_option("ignore_cancel", "1");
+        fingers = section->get_option("fingers", "4");
+        gap = section->get_option("gap", "32");
+        threshold = section->get_option("threshold", "0.35");
+        delta_threshold = section->get_option("delta_threshold", "24");
+        speed_factor = section->get_option("speed_factor", "256");
+        speed_cap = section->get_option("speed_cap", "0.05");
+        wf::get_core().connect_signal("pointer-swipe-begin", &on_swipe_begin);
+        wf::get_core().connect_signal("pointer-swipe-update", &on_swipe_update);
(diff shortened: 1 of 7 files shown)


ammen99:
With your last points (gaps, moving all windows) it seems that you want another plugin completely :)

It is easily doable, instead of operating like vswitch does, the other plugin can draw workspace streams at specific positions.

myfreeweb:
oh right, that wasn't hard! :)

ammen99:
I noticed that you are only handling horizontal scrolling, is vertical planned too?

myfreeweb:
Yes, I guess I can do vertical too..

ammen99:
Would it be easy to add animations when making the workspaces smaller? Currently there is a bit of (IMO) rough transition between the normal workspace size and the slightly scaled down size when transitioning.

myfreeweb:
uh, the unzoom effect has been removed? I think you might be running an old version..

(I might add it back as an option)

Anyway, I don't think animations in between events make sense. The display is usually slower than the trackpad, so the only way to animate between events is to add horrible latency.

myfreeweb:
Rebased.

ammen99:
Can you do vertical swipe too?

myfreeweb:
I don't really have the time or motivation (don't personally need it, I rarely use more than two workspaces haha) for vertical right now, sorry.

// It would require axis lock (deciding whether it's a horizontal or vertical swipe in the first couple events), I'm not sure if just two events is good enough, that would require playing around tuning the algorithm to see what feels right

ammen99:
Ok, I'll try to review the implementation soon.

plugins/single_plugins/vswipe.cpp
```diff
@@ -0,0 +1,387 @@
...
+        wf::render_hook_t renderer;
+
+        wf_duration duration;
+        wf_transition transition;
+
+        wf_option animation_duration;
+        wf_option background_color;
+        wf_option enable_horizontal;
+        wf_option enable_vertical;
+        wf_option ignore_cancel;
```
ammen99: Can you explain when cancel happens?

plugins/single_plugins/vswipe.cpp
```diff
@@ -0,0 +1,387 @@
...
+    void fini()
+    {
+        if (state.swiping)
+            finalize_and_exit();
+
+        OpenGL::render_begin();
+        streams.prev.buffer.release();
+        streams.curr.buffer.release();
+        streams.next.buffer.release();
+        OpenGL::render_end();
```
ammen99: Missing `disconnect()` on the swipe gesture signals.

plugins/single_plugins/vswipe.cpp
```diff
@@ -0,0 +1,387 @@
...
+        output->deactivate_plugin(grab_interface);
+
+        if (streams.prev.running)
+            output->render->workspace_stream_stop(streams.prev);
+        output->render->workspace_stream_stop(streams.curr);
+        if (streams.next.running)
+            output->render->workspace_stream_stop(streams.next);
+
+        output->render->set_renderer(nullptr);
+        output->render->set_redraw_always(false);
```
ammen99: It seems that the `set_redraw_always(true)` won't be always called, right? So the the `set_redraw_always(false)` shouldn't be called always too, otherwise we may corrupt the counter.

plugins/single_plugins/vswipe.cpp
myfreeweb: grep for the usage of this :) `(!ev->cancelled || ignore_cancel->as_int())` — `cancelled` is a property of the `end` event. libinput has some weird heuristics for whether the gesture was "cancelled", in my experience it doesn't make sense for this plugin, it was usually true when I don't want it.

I guess we can just ignore it always and not have the setting.

plugins/single_plugins/vswipe.cpp
myfreeweb: there is a counter?! hmm okay

ammen99: yes, if you call set_redraw_always() it increases the counter ....

plugins/single_plugins/vswipe.cpp
ammen99: in my tests it also didn't seem necessary to change it. So maybe we should just ignore it always.

AdrianVovk:
I can switch back to horizontal workspaces and we can just use horizontal vswitch

ammen99:
@Adrianvovk, wym? I wanted vertical workspaces because I am going to use them.

AdrianVovk:
@ammen99 Ah okay. I thought I was the only one who needed them :)

ammen99:
Thanks a lot!

PR #252, opened 2019-07-04
input-manager: do not crash when the keyboard is gone

myfreeweb:
Found by using a remote keyboard via [netevent](https://github.com/Blub/netevent) and quitting that at the "right" moment after dragging a window (lucky!).


ammen99:
I think I already have a patch for this in `osk-tablet-support` can you try there?

myfreeweb:
oh okay, yeah, [found that](https://github.com/WayfireWM/wayfire/commit/b57c20ee48791d82f32c4e0041c50eff44535e2b#diff-c9c95a9859baf8704dd0a568b30dc81dR202). would be nice to cherry-pick that into master :)

ammen99:
yeah, sorry for wasting your effort

PR #261, opened 2019-07-10
Pointer constraints

### LINKED ISSUE #254: issue with Minecraft/Xwayland mouse
When I run Minecraft (not the launcher, the actual game), it shows an X as the cursor (it's running in Xwayland if that's important). And when I move my mouse to move the character's view around, the view goes all over the place. I think the issue is that Minecraft cannot capture the mouse or something. Also, this seems to only be an issue with Wayfire because Minecraft runs fine in swaywm.

ammen99:
Fixes #254 

As far as I know, this doesn't introduce any regressions. To anybody interested, check this, and report if it works (and also any regressions that have happened, although I don't expect any).


sreehax:
I can confirm that the pointer constraints and relative pointer protocols in Minecraft on Arch Linux

ammen99:
Thanks for testing @myfreeweb @sreehax

soreau:
For reference, this also fixes many FPS type games found in steam where the mouse seems to 'go crazy'.

PR #264, opened 2019-07-17
Drop unused dependencies

jbeich:
- libevdev isn't required since 268d4d26bacf
- freetype2 isn't required since cce90043176c

See [build log](https://github.com/WayfireWM/wayfire/files/3401597/wayfire-0.2.8.log.txt)


ammen99:
Indeed, libevdev is now only an indirect dependency (all calls are inside wf-config), and freetype2 isn't needed in any case. Thanks for spotting this!

PR #267, opened 2019-07-27
Drop root on startup like Sway

jbeich:
On FreeBSD running KMS+DRM session requires setuid bit but it'd needs to be dropped as applications may refuse to run. Taken from Sway.

```
$ env -u DISPLAY wayfire
2019-07-27 23:50:19 - [backend/session/direct-freebsd.c:163] Cannot open tty: Permission denied
2019-07-27 23:50:19 - [backend/session/session.c:96] Failed to load session backend
2019-07-27 23:50:19 - [backend/backend.c:321] Failed to start a DRM session
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -129,6 +129,25 @@ namespace wf
     }
 }
 
+static bool drop_permissions(void)
+{
+    if (getuid() != geteuid() || getgid() != getegid())
+    {
+	if (setuid(getuid()) != 0 || setgid(getgid()) != 0)
+	{
+	    log_error("Unable to drop root, refusing to start");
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -193,6 +212,12 @@ int main(int argc, char *argv[])
     core.egl = egl_for_renderer[core.renderer];
     assert(core.egl);
 
+    if (!drop_permissions()) {
```
jbeich: Oops, missed this when adjusting sway->wayfire braces style.

ammen99:
This doesn't seem to break anything for me.

Thanks!

PR #268, opened 2019-07-27
Unbreak with wlroots master

jbeich:
Regressed by https://github.com/swaywm/wlroots/pull/1763 but probably an improvement even with wlroots 0.6.0.


ammen99:
Looks good, probably missed it since this doesn't break compilation. Thanks!

PR #270, opened 2019-07-28
Add basic tablet support and support for on-screen keyboards via the wf-shell proto

### LINKED ISSUE #183: Tablet support
Supporting tablet devices is something missing from Wayfire. However, this can't be currently implemented since I do not have any such hardware, so we'll need somebody who has a tablet, wants to see support for it in Wayfire and is ready to test.

### LINKED ISSUE #203: Implement ability to always run commands
For ex. we want to be able to adjust brightness even when the screen is locked. This means we should add a new type of commands which don't attempt to active the grab interface.

### LINKED ISSUE #204: Proper multikeyboard support
We need to make sure that we handle multiple keyboards properly (including virtual ones). Another thing to consider is output removal. If a keyboard is removed, `wlr_seat_get_keyboard` will return null, and we need to check those cases.

ammen99:
Fixes #204 
Fixes #183 
Fixes #203 

Stacked on top of #269


Ongy:
Expanding a bit on https://github.com/swaywm/wlroots/issues/1778 since I don't think it really belongs there:

Tablet pad buttons are on nearly every tablet out there. Stylus' are extempt from that rule.
Strips / Rings are on the "slightly" above entry level tier. I'd say that's where hobbiest top out. 
Group modes are supported by some of those devices, but IMO they aren't a main feature.

There's only 1 device I know of, that has more than one group (one of those 2k+Eur cintiqs), but that's probably the easiest part if everything else works decently.
Cintiqs are somewhat easy to support, you "just" have to map the input device to the output. If you find a good way to do it (without user configuration), might be nice as plugin library, so other compositors can share the mechanism (I have access to wacom signing tablets, so I might be able to help here)

Do you have a tablet? I could send you mine if you can't implement this for a lack of hardware. I'm more less than more working on stuff either way these days.

ammen99:
@Ongy Thanks, I hope sending hardware back and forth won't be necessary. What I would really appreciate is a bit of explanation how those things work. I.e are tablet pad ring/buttons/whatever events sent to the surface which has the tool focus? Or do they have a separate focus? (Since you said multiple groups are rare, I'd rather leave that for now).

I can certainly implement simple forwarding of events to clients. If I did this, would you be able to test it?

Ongy:
Pretty much. The pad has an enter/leave event. And then you send the events of any elements on it to the surface that it last entered.
GNOME has a secondary focus set for tablets afaik (and also secondary cursor, not using main one).
The rootston implementation just uses the main cursor and sends enter/leave of tablet pads with keyboard. The finicky thing here is pointer enter/leave on the cursor side, but you have that with only tool as well.

ammen99:
@Ongy I know have a basic implementation for tablet pads. Can you test it? (I have no tablet pads, so there could even be crashes on startup or similar). You could stop by at #wayfire when you have time :)

ammen99:
Merging this. Tablet pad support will be tracked in another PR.

PR #273, opened 2019-08-07
Make tile great again

### LINKED ISSUE #45: Fix tile plugin
Having a tiling support would be quite nice. It should have the following form - each workspace has a tiling and floating layer (the Z order depends on the focused view). We should also support moving views to and from the tiling layer, and add the ability to tile by default.

ammen99:
This is a completely new tile plugin to replace the old one.

Stacks on top of #271 
Fixes #45 

TODO:

- [x] Add transformer so that views are always at the needed size (i.e scale if minimum size is greater than slot size)

- [ ] Test thoroughly, the only clients I tested with are weston-terminal and terminator

- [x] Make sure tile interacts well with vswitch and expo

- [x] Make sure tile interacts well with minimizing views


myfreeweb:
> scale if minimum size is greater than slot size

wow, that's… an interesting solution :) How does sway handle this case?

> great again

urgh not that annoying meme >_< (Is it even technically correct? Was there a good tiling situation in the past? Before wlroots?)

ammen99:
> > scale if minimum size is greater than slot size
> 
> wow, that's… an interesting solution :) How does sway handle this case?

I think sway just cuts it (so you don't see the whole window), but before it used to scale it (or was that orbital that scaled?)
> 
> > great again
> 
> urgh not that annoying problematic meme >_< (Is it even technically correct? Was there a good tiling situation in the past? Before wlroots?)

Yes, the tiling plugin was a thing in the libweston version of Wayfire, and it was working. The current tiling plugin is however going to be much better! So it should technically be `make-tile-greater-than-ever`

PR #276, opened 2019-08-17
Check wlr headers for xwayland and display enabled state

soreau:


emersion:
Better way to do it: https://github.com/swaywm/sway/blob/master/meson.build#L73

ammen99:
I don't understand, is this just a message when compiling? Also I like the one-liner which emersion linked :)

soreau:
@emersion This was a nice change for simplification, thanks. Is there a way to silence the message "Fetching value of define "WLR_HAS_XWAYLAND" with dependency wlroots: 1" in the configure output?

emersion:
>I don't understand, is this just a message when compiling?

If you try building wayfire with xwayland while wlroots has been built without it, you can error out.

>This was a nice change for simplification, thanks. Is there a way to silence the message "Fetching value of define "WLR_HAS_XWAYLAND" with dependency wlroots: 1" in the configure output?

No idea, sorry.

ammen99:
Thanks to both @emersion and @soreau

PR #283, opened 2019-08-26
Add missing cmath include

myfreeweb:
for libc++


ammen99:
Thanks, it's good to have somebody checking builds on BSD (we should have CI but that's another issue)

PR #287, opened 2019-09-01
Add workspace background color option

soreau:
Fixes wf-shell #26.

### DIFF
--- a/src/api/workspace-stream.hpp
+++ b/src/api/workspace-stream.hpp
@@ -18,7 +18,7 @@ struct workspace_stream_t
     float scale_y = 1.0;
 
     /* The background color of the stream, when there is no view above it */
-    wf_color background = {0.0f, 0.0f, 0.0f, 1.0f};
+    wf_color background = {0.0f, 0.0f, 0.0f, -1.0f};
 };
 
 /** Emitted whenever a workspace stream is being started or stopped */
--- a/src/api/workspace-stream.hpp
+++ b/src/api/workspace-stream.hpp
@@ -17,7 +17,14 @@ struct workspace_stream_t
     float scale_x = 1.0;
     float scale_y = 1.0;
 
-    /* The background color of the stream, when there is no view above it */
+    /* The background color of the stream, when there is no view above it.
+     * All streams start with -1.0 alpha to indicate that the color is
+     * invalid. In this case, we use the default color, which can
+     * optionally be set by the user. If a plugin changes the background,
+     * the color will be valid and it will be used instead. This way,
+     * plugins can choose the background color they want first and if
+     * it is not set (alpha = -1.0) it will fallback to the default
+     * user configurable color. */
     wf_color background = {0.0f, 0.0f, 0.0f, -1.0f};
 };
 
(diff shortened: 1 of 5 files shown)


soreau:
> A few issues:
> 
>     1. Why does this all originate in output-layout? Can't it be contained inside render-manager?
I figured all the code would be called once per output if was done in render_manager, eg. get_option and setting up the changed callback.

> 
>     2. This won't work for custom workspace streams, for ex. the ones that expo uses.
I see you are correct, though I'm not sure why.

> 
> 
> So what I suggest: Set the default workspace color (in workspace-stream.hpp) to an invalid color. When clearing the background, check whether the color is invalid, and then use the default color (so that plugins can still override background color if they wish so).

I don't really understand what you mean here.

ammen99:
> > A few issues:
> > ```
> > 1. Why does this all originate in output-layout? Can't it be contained inside render-manager?
> > ```
> 
> I figured all the code would be called once per output if was done in render_manager, eg. get_option and setting up the changed callback.
> 

It's okay to make this multiple times, changing a config option isn't a costly operation (and doesn't happen every frame)

> > ```
> > 2. This won't work for custom workspace streams, for ex. the ones that expo uses.
> > ```
> 
> I see you are correct, though I'm not sure why.
Because the plugin can create its own workspace stream. And you set the background for only default workspace streams.

> 
> > So what I suggest: Set the default workspace color (in workspace-stream.hpp) to an invalid color. When clearing the background, check whether the color is invalid, and then use the default color (so that plugins can still override background color if they wish so).
> 
> I don't really understand what you mean here.

Maybe the comment about having non-default workspaces would be helpful in understanding. 
Let me try to explain it in another way: plugins can create custom workspaces and can request custom background. We just need some way for plugins to signal when they want the default background color and when they want to override it. So I proposed, setting the `stream.background` by default to some invalid value, then have render-manager take the default color in this case when updating the workspace stream.

soreau:
I think I see what you mean now. I have it working, let me know if it looks more inline now.

src/api/workspace-stream.hpp
```diff
@@ -18,7 +18,7 @@ struct workspace_stream_t
     float scale_y = 1.0;
 
     /* The background color of the stream, when there is no view above it */
```
ammen99: Maybe explain what the negative alpha represents (in the comment)

src/output/render-manager.cpp
```diff
@@ -864,7 +888,12 @@ class wf::render_manager::impl
         }
 
         check_schedule_surfaces(repaint, stream);
-        clear_empty_areas(repaint, stream.background);
```
ammen99: braces around this

src/output/render-manager.cpp
```diff
@@ -864,7 +888,12 @@ class wf::render_manager::impl
...
+
+        if (stream.background.a < 0)
```
ammen99: Braces around if/else

src/output/render-manager.cpp
```diff
@@ -348,6 +353,18 @@ class wf::render_manager::impl
...
+
+        background_color_opt_changed = [=] ()
+        {
+            auto color = background_color_opt->as_color();
+            update_background_color(color);
+        };
+
+        auto section = wf::get_core().config->get_section("core");
+        background_color_opt = section->get_option("background_color", "0 0 0 1");
+        background_color_opt->add_updated_handler(&background_color_opt_changed);
```
ammen99: One more thing: when the render manager is destroyed, we need to disconnect the updated handler.

ammen99:
This PR LGTM. Thanks!

PR #289, opened 2019-09-07
Fix build when wlroots is built without x11-backend

soreau:


src/core/output-layout.cpp
```diff
@@ -240,7 +242,7 @@ namespace wf
             position_opt  = output_section->get_option("layout", default_value);
             scale_opt     = output_section->get_option("scale", "1");
             transform_opt = output_section->get_option("transform", "normal");
-
+#if WLR_HAS_X11_BACKEND
```
ammen99: This breaks resizing on wayland backend, if x11 backend is disabled.

soreau: @ammen99 I pushed a new version that should make more sense.

ammen99: I'd prefer to make it without an additional function. Maybe something along the lines of

`bool is_nested_compositor = is_wyaland..()` and then inside an ifdef `is_nested... |= is_xwl()`

soreau: You're right, that's cleaner.

ammen99:
Looks good now, thanks!

PR #295, opened 2019-09-29
Wobbly rework

### LINKED ISSUE #280: Rewrite wobbly logic
The current wobbly logic is getting complicated. We need to rethink the logic handling, preferably with an explicit state listing (snapped, grabbed, nothing + tiled, nothing) and handle the view resizing, etc. appropriately.

ammen99:
Fixes #280


ammen99:
Thanks @soreau for testing this.

PR #297, opened 2019-09-29
vswipe: add smoothing of scrolling

ammen99:
@myfreeweb I'd appreciate if you can test this and tell me if it makes the experience better or worse. On my setup, I think smoothening the transitions makes it much better, but it'd make sense to gather some more opinions :)

### DIFF
--- a/plugins/single_plugins/vswipe.cpp
+++ b/plugins/single_plugins/vswipe.cpp
@@ -39,8 +39,6 @@ class vswipe : public wf::plugin_interface_t
             swipe_direction_t direction;
 
             wf_pointf initial_deltas;
-
-            double delta_sum = 0.0;
             double gap = 0.0;
 
             double delta_prev = 0.0;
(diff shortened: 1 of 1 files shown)


myfreeweb:
Mostly worse I'd say. Not very bad, but sometimes the latency appears and it feels sort of jelly-like. I absolutely do not like it when I can feel a big movement (whole screen in this case) lagging behind my input.

ammen99:
Okay I'll have to find a way to turn this off then. It probably depends on the hardware - without this, movement feels somehow unstable.

ammen99:
@myfreeweb The new version should make smoothing entirely optional while the user is actually swiping (smoothing happens only after swipe ends). I hope I didn't break anything.

myfreeweb:
btw, here's a GTK patch for scroll interpolation https://gitlab.gnome.org/GNOME/gtk/merge_requests/1117 I haven't tried it yet, but might be worth looking at the interpolation algorithm there

PR #300, opened 2019-10-01
Add packaging status

tim77:
Please add packaging status. Wayfire with all required components [packaged for Fedora](https://src.fedoraproject.org/rpms/wayfire) (for Rawhide at this moment). Waiting for your new 0.3 version. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ meson build --prefix=/usr --buildtype=release
 ninja -C build && sudo ninja -C build install
 ```
 
+# Packaging status
+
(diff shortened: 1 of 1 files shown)


ammen99:
Would this also pull wf-shell as well? (`dnf install wayfire`) or just Wayfire itself?

tim77:
Yes, `wf-shell` and `wayfire-config-manager` as weak depencies.

soreau:
> Yes, `wf-shell` and `wayfire-config-manager` as weak depencies.

:+1:

ammen99:
I figure we can merge this even without having a release.
Thanks!

PR #303, opened 2019-10-04
Fix blur

### LINKED ISSUE #159: Blur results in garbage background in mirror views

ammen99:
Fixes #159


soreau:
This looks good to me, thanks for finding and fixing the problem! :+1:

PR #305, opened 2019-10-04
place: Fix window placement using cascade when place is loaded after wayfire starts

### LINKED ISSUE #304: Loading place plugin with cascade mode after wayfire is started causes incorrectly placed windows
Windows start being placed under the panel if the panel is at the top. This is because cascade_x/y are initialized to 0 on plugin init and rely on workspace area change signal to update. This only happens on wayfire start.

soreau:
This was happening because cascade_x/y are initialized to 0 on plugin init
and rely on workarea change signal to update. However, this only happens if
place is loaded when wayfire starts. Fix this by initializing cascade_x/y
with the workarea.x/y on init. Fixes #304.

### DIFF
--- a/plugins/single_plugins/place.cpp
+++ b/plugins/single_plugins/place.cpp
@@ -14,7 +14,9 @@ class wayfire_place_window : public wf::plugin_interface_t
     public:
     void init(wayfire_config *config)
     {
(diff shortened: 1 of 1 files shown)


ammen99:
This seems reasonable, thanks!

PR #308, opened 2019-10-05
xwayland: Use configure request mask to position surfaces

### LINKED ISSUE #307: Some xwayland windows get wrong position
xterm can end up under the panel and steam can show a single first-frame of one position, immediately followed by another frame with the client-intended position, causing a visual glitch. There might be problems related to this with other clients as well.

soreau:
The mask contains information about whether or not to use the
position and size fields. Since size is determined by commit,
we only concern ourselves with the position here. This requires
[wlroots #1843](https://github.com/swaywm/wlroots/pull/1843). This fixes bugs with positioning xwayland surfaces
with and without place plugin enabled. Fixes #307.

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -32,6 +32,9 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
     int last_server_width = 0;
     int last_server_height = 0;
 
+    /** The geometry requested by the client */
+    bool self_positioned = false;
+
     wf::signal_callback_t output_geometry_changed =
         [this] (wf::signal_data_t*)
     {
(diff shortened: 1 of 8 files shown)


src/view/xwayland.cpp
```diff
@@ -49,6 +52,13 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
         on_destroy.set_callback([&] (void*) { destroy(); });
         on_configure.set_callback([&] (void* data) {
             auto ev = static_cast<wlr_xwayland_surface_configure_event*> (data);
+            if ((ev->mask & XCB_CONFIG_WINDOW_X) && (ev->mask & XCB_CONFIG_WINDOW_Y))
+            {
+                self_positioned = true;
+            } else {
```
ammen99: Why do we need the else here? Wlroots will already set x and y to `xsurface->x/y`

src/view/xwayland.cpp
```diff
@@ -285,15 +305,17 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
         if (xw->fullscreen)
             fullscreen_request(get_output(), true);
 
-        auto real_output = get_output()->get_layout_geometry();
-        if (!tiled_edges && !fullscreen && !parent)
+        wf::wlr_view_t::map(surface);
+
+        /* Freely-floating X windows might have set a geometry hint, try to
+         * take this into account */
+        if (!tiled_edges && !fullscreen && !parent && !self_positioned)
```
ammen99: I just realize this, but we may not need this at all. Can you try removing the if and its contents?

My logic is the following: if we had a configure request before mapping, we have already applied it so the view is at the correct place.

src/view/xwayland.cpp
soreau: This is needed or else xterm is placed at 0,0 instead of 100,100 without place enabled.

src/view/xwayland.cpp
soreau: This is needed or else xterm is placed as if it doesn't have decorations, without place enabled.

src/view/xwayland.cpp
ammen99: Okay but this is still wrong because ev->x/y is supposed to be global coordinates, and geometry.x/y is geometry relative to the view output.

soreau: Should be fixed now.

src/view/xwayland.cpp
soreau: As discussed on IRC, it doesn't really matter if place is disabled.

PR #310, opened 2019-10-05
xwayland: Use mask flag to configure appropriate geometry elements

### LINKED ISSUE #309: Some xwayland windows start with wrong size
xterm can end up at 3x3 instead of the expected size, with decoration plugin enabled.

soreau:
Pass mask to wlr_xwayland_surface_configure() so as to not configure
a window twice with the same size. If this happens between commits,
the client might not end up with its intended size. This requires
[wlroots #1843](https://github.com/swaywm/wlroots/pull/1843). Fixes #309.

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -121,7 +121,7 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
         wf::wlr_view_t::close();
     }
 
-    void send_configure(int width, int height)
+    void send_configure(int width, int height, uint16_t mask)
     {
         if (!is_mapped())
             return;
@@ -147,19 +147,19 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
         }
 
         wlr_xwayland_surface_configure(xw,
-            configure_x, configure_y, width, height);
+            configure_x, configure_y, width, height, mask);
     }
 
-    void send_configure()
+    void send_configure(uint16_t mask)
     {
-        send_configure(last_server_width, last_server_height);
+        send_configure(last_server_width, last_server_height, mask);
     }
 
     void move(int x, int y) override
     {
         wf::wlr_view_t::move(x, y);
         if (!view_impl->in_continuous_move)
-            send_configure();
+            send_configure(XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y);
     }
 
     virtual void set_output(wf::output_t *wo) override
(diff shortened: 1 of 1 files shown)


soreau:
I see what you mean. Is there a case you have in mind that might exhibit a problem?

ammen99:
Yes, if a window is moved, the client needs to be notified because it needs to reposition its menus (VLC is a good test client, with the X11 backend)

soreau:
Ok thanks, I'll look into it.

src/view/xwayland.cpp
```diff
@@ -178,7 +178,7 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
                 &output_geometry_changed);
         }
         /* Update the real position */
-        send_configure();
+        send_configure(XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y | XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT);
```
ammen99: why do we need width and height here? It is again position-update-only

src/view/xwayland.cpp
```diff
@@ -344,7 +343,11 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
 
         last_server_width = w;
         last_server_height = h;
-        send_configure(w, h);
+
+        if (w != geometry.width && h != geometry.height)
+            send_configure(w, h, XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y |
+                XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT |
+                XCB_CONFIG_WINDOW_BORDER_WIDTH);
```
ammen99: why do we configure border width at all?

src/view/xwayland.cpp
```diff
@@ -356,7 +359,7 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
         {
             last_server_width = xw->size_hints->base_width;
             last_server_height = xw->size_hints->base_height;
-            send_configure();
+            send_configure(XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y);
```
ammen99: huh, i think here we want to change width and height as well?

soreau: Yes, it seems so.

src/view/xwayland.cpp
soreau: You're right, this shouldn't be needed.

src/view/xwayland.cpp
soreau: Without this, xterm ends up starting with 3x3 size.

ammen99:
As discussed in IRC, this isn't necessary at all.

Issue #284, opened 2019-08-30
Editing wayfire.ini causes wm to crash


Malmik:
As in description. After that I am beeing kicked to tty but I briefly see some red error messages appearing for a second. After executing "wayfire" command wm starts without problems with updated settings. Where are the logs stored?

ammen99:
Logs are printed to stdout.

Which settings are you changing?

luispabon:
I can confirm changing any setting causes wayfire to crash. Doesn't matter what. I'll try generate some debug.

soreau:
Please follow the instructions in [the wiki](https://github.com/WayfireWM/wayfire/wiki/Debugging-problems) to produce a meaningful log.

ammen99:
@luispabon I just merged pr #312 which fixes the original problem of @Malmik (we have debugged it a bit on IRC). Can you check whether the crash still happens after updating to latest master?

luispabon:
Yes, it doesn't crash for me no more.

ammen99:
Ok, great. Thanks for the test.

PR #312, opened 2019-10-06
Avoid crash when calling set_decoration with null frame

### LINKED ISSUE #311: Unloading decoration plugin crashes wayfire

soreau:
The decoration plugin calls set_decoration with a null frame
argument on unload which causes wayfire to crash. Handle this
case properly by resetting the frame and emitting decoration
changed signal. Fixes #311.

### DIFF
--- a/src/view/view.cpp
+++ b/src/view/view.cpp
@@ -482,6 +482,18 @@ bool wf::view_interface_t::should_be_decorated()
 
 void wf::view_interface_t::set_decoration(surface_interface_t *frame)
 {
+    if (!frame)
+    {
+        damage();
+        if (view_impl->decoration)
+            view_impl->decoration->unref();
+        view_impl->decoration = nullptr;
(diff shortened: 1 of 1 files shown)


ammen99:
Btw this should also fix #284

src/view/view.cpp
```diff
@@ -482,6 +482,18 @@ bool wf::view_interface_t::should_be_decorated()
 
 void wf::view_interface_t::set_decoration(surface_interface_t *frame)
 {
+    if (!frame)
+    {
+        damage();
+        if (view_impl->decoration)
+            view_impl->decoration->unref();
+        view_impl->decoration = frame;
```
ammen99: i'd like to see it set to `nullptr` explicitly (frame is null anyway)

soreau: Yes, makes sense.

ammen99:
Thanks!

PR #315, opened 2019-10-08
Don't build on breaking wlroots version.

travankor:
wlroots versions can still break on major version bumps. It would be helpful for users and packagers to know if they are compiling wayfire on an incompatible wlroots version.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,7 +29,7 @@ libinput       = dependency('libinput', version: '>=1.7.0')
 pixman         = dependency('pixman-1')
 threads        = dependency('threads')
 xkbcommon      = dependency('xkbcommon')
-wlroots        = dependency('wlroots', version: '>=0.8.0', fallback: ['wlroots', 'wlroots'])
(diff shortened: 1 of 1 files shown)


ammen99:
The reason I haven't done this already is that we don't know whether wlroots 0.9.0 will break Wayfire, because for ex. you can compile Wayfire 0.3 with wlroots 0.7 (although I don't recommend it because there are some bugs).

But maybe it makes more sense to just make minor releases (i.e 0.3.1) when a new wlroots release happens.

travankor:
>But maybe it makes more sense to just make minor releases (i.e 0.3.1) when a new wlroots release happens.

This is a lot better than having users second guess their distro's packages IMO. I hope additional release tags is not too much of a hassle. [Sway](https://github.com/swaywm/sway/commit/08b3eaf93074daf4e7fa0831a63c46bba55fbbd7#diff-969b60ad3d206fd45c208e266ccfed38) recently added a similar change.

I hear in the long-term wlroots 1.0.0 will be  backwards-compatible, but 1.0.0 is probably not happening anytime soon.

ammen99:
I am not sure wlroots 1.0 will happen at all :)

I agree with the strategy do make minor releases, thanks for this fix!

PR #316, opened 2019-10-10
fisheye: Initialize activate variable

soreau:
This was causing fisheye to not activate until the second time
the binding is pressed after loading the plugin.


ammen99:
Nice catch!

PR #317, opened 2019-10-10
fisheye: Initialize hook_set variable

soreau:
This was causing fisheye to not start at all in some cases.


ammen99:
Thanks, another good catch :)

PR #320, opened 2019-10-17
layer-shell: Track wlroots changes

soreau:
This fixes building against wlroots as of 17242619.


soreau:
> This isn't enough, because the view is never moved to the appropriate layer, right?

I'm not sure.

ammen99:
Fixes #334 #323

PR #328, opened 2019-11-04
Send bound events when input is inhibited

soreau:
When a client inhibits input, the compositor should send all
events, even those that match a binding.

### DIFF
--- a/plugins/blur/blur.cpp
+++ b/plugins/blur/blur.cpp
@@ -152,15 +152,20 @@ class wayfire_blur : public wf::plugin_interface_t
         /* Toggles the blur state of the view the user clicked on */
         button_toggle = [=] (uint32_t, int, int)
         {
+            if (!output->can_activate_plugin(grab_interface))
+                return false;
+
             auto view = wf::get_core().get_cursor_focus_view();
             if (!view)
-                return;
+                return false;
 
             if (view->get_transformer(transformer_name)) {
                 view->pop_transformer(transformer_name);
             } else {
                 add_transformer(view);
             }
+
+            return true;
         };
         output->add_button(section->get_option("toggle", "<super> <alt> BTN_LEFT"),
             &button_toggle);
--- a/plugins/single_plugins/expo.cpp
+++ b/plugins/single_plugins/expo.cpp
@@ -23,11 +23,15 @@ class wayfire_expo : public wf::plugin_interface_t
     activator_callback toggle_cb = [=] (wf_activator_source, uint32_t)
     {
         if (!state.active) {
-            activate();
+            return activate();
         } else {
-            if (!zoom_animation.running() || state.zoom_in)
+            if (!zoom_animation.running() || state.zoom_in) {
                 deactivate();
+                return true;
+            }
         }
+
+        return false;
     };
 
     wf_option action_button;
@@ -123,10 +127,10 @@ class wayfire_expo : public wf::plugin_interface_t
         output->connect_signal("view-disappeared", &view_removed);
     }
 
-    void activate()
+    bool activate()
     {
         if (!output->activate_plugin(grab_interface))
-            return;
+            return false;
 
         grab_interface->grab();
 
@@ -141,6 +145,8 @@ class wayfire_expo : public wf::plugin_interface_t
 
         output->render->set_renderer(renderer);
         output->render->set_redraw_always();
+
+        return true;
     }
 
     void deactivate()
--- a/plugins/single_plugins/fisheye.cpp
+++ b/plugins/single_plugins/fisheye.cpp
@@ -125,6 +125,9 @@ class wayfire_fisheye : public wf::plugin_interface_t
     public:
         void init(wayfire_config *config)
         {
+            grab_interface->name = "fisheye";
+            grab_interface->capabilities = 0;
+
             auto section = config->get_section("fisheye");
             auto toggle_key = section->get_option("toggle", "<super> KEY_F");
             radius = section->get_option("radius", "300");
@@ -139,22 +142,27 @@ class wayfire_fisheye : public wf::plugin_interface_t
             hook_set = active = false;
             toggle_cb = [=] (wf_activator_source, uint32_t)
             {
-                    if (active)
-                    {
-                        active = false;
-                        duration.start(duration.progress(), 0);
-                    } else
-                    {
-                        active = true;
-                        duration.start(duration.progress(), target_zoom);
+                if (!output->can_activate_plugin(grab_interface))
+                    return false;
 
-                        if (!hook_set)
-                        {
-                            hook_set = true;
-                            output->render->add_post(&hook);
-                            output->render->set_redraw_always();
-                        }
+                if (active)
+                {
+                    active = false;
+                    duration.start(duration.progress(), 0);
+                } else
+                {
+                    active = true;
+                    duration.start(duration.progress(), target_zoom);
+
+                    if (!hook_set)
+                    {
+                        hook_set = true;
+                        output->render->add_post(&hook);
+                        output->render->set_redraw_always();
                     }
+                }
+
+                return true;
             };
             output->add_activator(toggle_key, &toggle_cb);
 
--- a/plugins/single_plugins/grid.cpp
+++ b/plugins/single_plugins/grid.cpp
@@ -240,8 +240,11 @@ class wayfire_grid : public wf::plugin_interface_t
 
     activator_callback restore = [=] (wf_activator_source, uint32_t)
     {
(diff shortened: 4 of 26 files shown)


plugins/single_plugins/switcher.cpp
```diff
@@ -126,6 +126,9 @@ class WayfireSwitcher : public wf::plugin_interface_t
         };
 
         touch_activate = [=] (wf_touch_gesture*) {
+            if (!output->activate_plugin(grab_interface))
```
ammen99: This is wrong, the plugin is activated only in the `!active` case, check the code of `handle_switch_request`

plugins/single_plugins/oswitch.cpp
```diff
@@ -24,6 +24,9 @@ class wayfire_output_manager : public wf::plugin_interface_t
 
             switch_output = [=] (wf_activator_source, uint32_t)
             {
+                if (!output->activate_plugin(grab_interface))
```
ammen99: Not sure that I like this check, but I am also not sure how we can implement it (maybe we'll have to extend the `activate_plugin` function). Use-case I have in mind: we can have Expo open on the focused output, and it is perfectly fine to use a keybinding to switch to the other output.

plugins/single_plugins/invert.cpp
```diff
@@ -61,13 +61,19 @@ class wayfire_invert_screen : public wf::plugin_interface_t
...
+        grab_interface->capabilities = 0;
+
         hook = [=] (const wf_framebuffer_base& source,
             const wf_framebuffer_base& destination) {
             render(source, destination);
         };
 
 
         toggle_cb = [=] (wf_activator_source, uint32_t) {
+            if (!output->activate_plugin(grab_interface))
```
ammen99: we also need to deactivate the plugin somewhere.

plugins/single_plugins/idle.cpp
```diff
@@ -101,10 +101,18 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
     {
         singleton_plugin_t::init(config);
 
+        grab_interface->name = "idle";
+        grab_interface->capabilities = 0;
+
         auto binding = config->get_section("idle")
             ->get_option("toggle", "<super> <shift> KEY_I");
         toggle = [=] (wf_activator_source, uint32_t) {
+            if (!output->activate_plugin(grab_interface))
```
ammen99: where do we deactivate the plugin?

plugins/single_plugins/grid.cpp
```diff
@@ -240,8 +240,11 @@ class wayfire_grid : public wf::plugin_interface_t
 
     activator_callback restore = [=] (wf_activator_source, uint32_t)
     {
+        if (!output->activate_plugin(grab_interface))
```
ammen99: we need to deactivate the plugin after this.

plugins/single_plugins/fisheye.cpp
```diff
@@ -139,22 +142,27 @@ class wayfire_fisheye : public wf::plugin_interface_t
             hook_set = active = false;
             toggle_cb = [=] (wf_activator_source, uint32_t)
             {
-                    if (active)
-                    {
-                        active = false;
-                        duration.start(duration.progress(), 0);
-                    } else
+                if (!output->activate_plugin(grab_interface))
```
ammen99: we need to deactivate the plugin.

plugins/single_plugins/expo.cpp
```diff
@@ -22,12 +22,20 @@ class wayfire_expo : public wf::plugin_interface_t
 
     activator_callback toggle_cb = [=] (wf_activator_source, uint32_t)
     {
+        if (!output->activate_plugin(grab_interface))
```
ammen99: pretty sure `activate` does try to activate the plugin so just get the value from there.

plugins/blur/blur.cpp
```diff
@@ -152,15 +152,20 @@ class wayfire_blur : public wf::plugin_interface_t
         /* Toggles the blur state of the view the user clicked on */
         button_toggle = [=] (uint32_t, int, int)
         {
+            if (!output->activate_plugin(grab_interface))
```
ammen99: deactivate the plugin somewhere?

plugins/single_plugins/switcher.cpp
soreau: I pushed an update for this, please check if it's doing The Right Thing.

src/output/output.cpp
```diff
@@ -242,6 +242,28 @@ wayfire_view wf::output_impl_t::get_active_view() const
     return active_view;
 }
 
+bool wf::output_impl_t::can_activate_plugin(const plugin_grab_interface_uptr& owner)
```
ammen99: Since we have all the checks here, maybe we can just reuse this function in `activate_plugin`? (So we don't have duplicate logic)

plugins/single_plugins/switcher.cpp
```diff
@@ -128,12 +128,16 @@ class WayfireSwitcher : public wf::plugin_interface_t
         touch_activate = [=] (wf_touch_gesture*) {
             if (!active)
             {
+                if (!output->can_activate_plugin(grab_interface))
```
ammen99: I still don't get why we need this check instead of returning the return value of handle_switch_request() directly ...

plugins/single_plugins/expo.cpp
```diff
@@ -23,11 +23,17 @@ class wayfire_expo : public wf::plugin_interface_t
     activator_callback toggle_cb = [=] (wf_activator_source, uint32_t)
     {
         if (!state.active) {
-            activate();
+            if (!activate())
```
ammen99: You can just `return active();`

src/output/output.cpp
soreau: That mostly works. There however there's a check to load the plugin again sandwiched between the initial checks and the loop to check compatibility. If I put the loop in can_activate() then call it with the other initial checks, it becomes before the condition that might load the plugin again. Not sure what to do here exactly.

plugins/single_plugins/switcher.cpp
soreau: I see now.

plugins/single_plugins/expo.cpp
soreau: True.

soreau:
@ammen99 Updated and rebased on master as well.

src/output/output.cpp
ammen99: Oh, you should move the sandwiched check to the can_activate_plugin function too. We need to have the same behavior in both functions, otherwise it will just get confusing(also add the ignore_inhibit parameter to can_activate)

soreau: This sandwiched check potentially adds the plugin and returns early. We don't want to actually add any plugins to the active list in can_activate().

src/api/output.hpp
```diff
@@ -76,6 +76,14 @@ class output_t : public wf::object_base_t
      */
     wf_pointf get_cursor_position() const;
 
+    /**
+     * Checks if a plugin can activate. This may not succeed if a plugin
+     * with the same abilities is already active or if in idle state.
```
ammen99: Not `in idle state` but if input is inhibited via input-inhibit.

soreau: Fixed.

src/output/output.cpp
soreau: Ok, I put all the checks in can_activate except ignore_inhibit and put the sandwiched check again in activate just to output the right message.

src/output/output.cpp
```diff
@@ -269,8 +264,21 @@ bool wf::output_impl_t::activate_plugin(const plugin_grab_interface_uptr& owner,
             return false;
     }
 
+    return true;
+}
+
+bool wf::output_impl_t::activate_plugin(const plugin_grab_interface_uptr& owner,
+    bool ignore_inhibit)
+{
+    if (!can_activate_plugin(owner) || !ignore_inhibit)
```
ammen99: Oh hm I didn't see this until now. `ignore_inhibit` should also be a parameter of `can_activate_plugin` (as I said, we do not want any differences between the two functions in terms of when one succeeds or fails.

soreau: It seems to me that it is pointless to pass an argument that just makes the function fail. Why would you not handle this in the plugin(s) and just avoid calling (can_)activate in the first place?

soreau: Fixed.

ammen99:
Tested it, seems to run fine. Thanks!

PR #332, opened 2019-11-10
Implement screensaver timeout with rotating cube

soreau:


ammen99:
Why do you  need to create a second wlr_idle instance?

soreau:
> Why do you need to create a second wlr_idle instance?

One is for the screensaver and the other for dpms.

ammen99:
Can't you just have multiple timers on the same `wlr_idle`?

soreau:
> Can't you just have multiple timers on the same `wlr_idle`?

Not that I am aware of.

soreau:
@ammen99 I dropped the second wlr_idle.

soreau:
However it only works for the focused output. I'm not clear exactly why this is, but if I click on the other output while the cube is settling, then both settle together.

soreau:
That is, inhibit works for all outputs. Cube rotation screensaver from idle only works on the focused output.

soreau:
[This](https://github.com/WayfireWM/wayfire/blob/master/src/output/output.cpp#L251) check needs to be removed for all outputs to have working cube screensaver at the same time. I think this should be harmless to remove, @ammen99 what do you think?

soreau:
I am able to get into a state where I cannot exit the screensaver. This only happens after editing the screensaver timeout value in wcm with keyboard. The screensaver times out at the previous value set and no input action exits the screensaver. It seems like a bug in the wlroots idle implementation. However I cannot seem to trigger this problem just by editing the config file, I must use wcm and enter a value without using the up/down buttons. Here is the [idle.xml](http://ix.io/21rk).

soreau:
> I am able to get into a state where I cannot exit the screensaver. This only happens after editing the screensaver timeout value in wcm with keyboard. The screensaver times out at the previous value set and no input action exits the screensaver. It seems like a bug in the wlroots idle implementation. However I cannot seem to trigger this problem just by editing the config file, I must use wcm and enter a value without using the up/down buttons. Here is the [idle.xml](http://ix.io/21rk).

I found out this was because the timeout was being reset while it was in idle state. If a timeout is destroyed while in idle state, the resume event never happens for it and it becomes stuck in idle. This seems like a wlroots bug to me but I've done as much as I can to prevent this on our side. The way I was reproducing this was wcm would write the config file timeout value after typing it manually in the spinner and wcm losing focus. I found out this was because wcm applies options on focus lost which happens when screensaver kicks in. This condition can still happen if a user types in the spinner without pressing enter and waiting for screensaver to kick in. This probably will only happen on short timeouts for testing since wcm might not have lost focus but idle now takes this into account and does not allow the screensaver to get in a stuck state.

soreau:
I've rebased on master and fixed an issue when the activate cube binding is used while screensaver is transitioning to end.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+    wf_option dpms_timeout, screensaver_timeout;
+    wf_option cube_zoom_speed, cube_rotate_speed, cube_zoom_end;
     wf_option_callback dpms_timeout_updated = [=] () {
-        create_timeout(dpms_timeout->as_int());
+        create_dpms_timeout(dpms_timeout->as_int());
+    };
+    wf_option_callback screensaver_timeout_updated = [=] () {
+        create_screensaver_timeout(screensaver_timeout->as_int());
+    };
+    wf_option_callback cube_zoom_speed_updated = [=] () {
```
ammen99: You don't need this, if you create a wf_duration from a wf_option, it will automatically update when the option updates.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+        });
+        on_resume_screensaver.connect(&timeout_screensaver->events.resume);
+    }
+
+    void screensaver_frame()
+    {
+        cube_control_signal data;
+        bool carried_out = true;
+
+        if (screensaver_stopping)
```
ammen99: Style: missing braces

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+            rotation = duration.progress(rot_transition);
+        else
+            rotation += cube_rotate_speed->as_double();
+        if (rotation >= M_PI * 2)
```
ammen99: We should at least `#include <cmath>`

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+            rotation = 0.0;
```
ammen99: Hm hm, are you sure this doesn't result in some glitches? Like, if the rotate speed is too big, you might get `rotation = 2*PI + 0.1` and then you jump back to `0` but we wanted `0.1`. Would be better to `rotation -= 2 * M_PI` if I understand the idea correctly.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+        data.zoom = duration.progress(zoom_transition);
+        data.last_frame = false;
+
+        if (screensaver_stopping && !duration.running())
+        {
+            cube_control_signal data;
+            data.angle = 0.0;
+            data.zoom = ZOOM_BASE;
+            data.last_frame = true;
+            for (auto& output : wf::get_core().output_layout->get_outputs()) {
```
ammen99: Style: `{` on new line

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+        else
+            rotation += cube_rotate_speed->as_double();
+        if (rotation >= M_PI * 2)
+            rotation = 0.0;
+
+        data.angle = rotation;
+        data.zoom = duration.progress(zoom_transition);
+        data.last_frame = false;
+
+        if (screensaver_stopping && !duration.running())
```
ammen99: I would like to see this `if` block moved to the beginning of the function.
Reason: the calculations above don't seem to be necessary for anything in this if-block.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+                    output->render->add_inhibit(false);
+                    output->render->damage_whole();
+                }
+            }
+            outputs_inhibited = false;
+            screensaver_stopping = false;
+            return;
+        }
+
+        for (auto& output : wf::get_core().output_layout->get_outputs()) {
```
ammen99: Style `{` on a new line, as well as on the lines below

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+                    output->render->add_inhibit(true);
+                    output->render->damage_whole();
+                }
+                carried_out = false;
+            }
+        }
+        if (!carried_out)
+            outputs_inhibited = true;
+
+        if (screensaver_stopping)
```
ammen99: Missing braces on a multiline if

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+    }
+
+    void start_screensaver()
+    {
+        cube_control_signal data;
+        data.angle = 0.0;
+        data.zoom = ZOOM_BASE;
+        data.last_frame = false;
+        bool carried_out = true;
+        for (auto& output : wf::get_core().output_layout->get_outputs()) {
```
ammen99: style {

src/output/output.cpp
```diff
@@ -248,9 +248,6 @@ bool wf::output_impl_t::can_activate_plugin(const plugin_grab_interface_uptr& ow
     if (!owner)
         return false;
 
-    if (wf::get_core().get_active_output() != this)
```
ammen99: I would like to get this into a separate commit.

ammen99: Oh nvm it is a separate commit.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+        screensaver_enabled = true;
+        screensaver_stopping = false;
+        rotation = 0.0;
+        zoom_transition = {ZOOM_BASE, cube_zoom_end->as_double()};
+        duration.start();
+    }
+
+    void stop_screensaver()
+    {
+        if (outputs_inhibited) {
```
ammen99: style {

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
 {
+    wf::effect_hook_t render_hook;
+    double rotation = 0.0;
+    wf_duration duration;
+    wf_transition rot_transition, zoom_transition;
+    bool screensaver_stopping = false;
     bool idle_enabled = true;
-    wlr_idle_timeout *timeout = NULL;
-    wf::wl_listener_wrapper on_idle, on_resume;
+    bool screensaver_enabled = false;
```
ammen99: We are too reliant on boolean flags, where actually what we want to represent is just the current state. Use an `enum`, with states like the following: `SCREENSAVER_DISABLED, SCREENSAVER_RUNNING, SCREENSAVER_STOPPING`

This way, we have fewer bools we need to keep track of.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,250 @@ extern "C"
...
+            }
+            outputs_inhibited = false;
+            screensaver_stopping = false;
+            return;
+        }
+
+        for (auto& output : wf::get_core().output_layout->get_outputs()) {
+            output->emit_signal("cube-control", &data);
+            if (!data.carried_out) {
+                if (!outputs_inhibited) {
```
ammen99: Okay so we have a problem here. Cube may fail on some outputs, but not on the others. Which means some output will get inhibited, others not.

Then upon resume, you will try to add_inhibit(false) to all outputs (because outputs_inhibitied is `true`) but it is not true for all outputs.

What I'd suggest: first iterate over all outputs, and try to activate cube. If you can't activate it for some output, just iterate over all outputs again, disable cube on outputs where it was activated, then just turn off all outputs.

soreau:
@ammen99 I believe I have addressed all of your concerns, please have a look. I also took the liberty of squashing the patches into just two.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,297 @@ extern "C"
...
+
+        render_hook = [=] () { screensaver_frame(); };
 
         dpms_timeout->add_updated_handler(&dpms_timeout_updated);
         dpms_timeout_updated();
+        screensaver_timeout->add_updated_handler(&screensaver_timeout_updated);
+        screensaver_timeout_updated();
+        duration = wf_duration(cube_zoom_speed);
+        
+        for (auto& output : wf::get_core().output_layout->get_outputs())
```
ammen99: no need for this. Operator [] will return the default value for bool (i.e false) if the element is not in the map.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,297 @@ extern "C"
...
+                    output->render->rem_effect(&render_hook);
+                    hook_set[output] = false;
+                }
+                if (state == SCREENSAVER_DISABLED)
+                {
+                    output->render->add_inhibit(false);
+                    output->render->damage_whole();
+                }
+            }
+            state = SCREENSAVER_DISABLED;
```
ammen99: Guess we should clear the map at this point.

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,297 @@ extern "C"
...
 {
+    wf::effect_hook_t render_hook;
+    double rotation = 0.0;
+    wf_duration duration;
+    wf_transition rot_transition, zoom_transition;
     bool idle_enabled = true;
-    wlr_idle_timeout *timeout = NULL;
-    wf::wl_listener_wrapper on_idle, on_resume;
+    screensaver_state state = SCREENSAVER_DISABLED;
+    std::map<wf::output_t*, bool> hook_set;
```
ammen99: Can we have a better name for this? Maybe output_inhibited?

plugins/single_plugins/idle.cpp
```diff
@@ -4,69 +4,297 @@ extern "C"
...
+        }
+        state = SCREENSAVER_DISABLED;
+    }
+
+    void screensaver_frame()
+    {
+        cube_control_signal data;
+        bool all_outputs_active = true;
+
+        if (state == SCREENSAVER_STOPPING && !duration.running())
```
ammen99: Now that I see it, I'd like to have the body of the if in a separate function.

plugins/single_plugins/idle.cpp
soreau: Ok.

plugins/single_plugins/idle.cpp
soreau: What do you mean by clear the map? If you mean set all `hook_set` elements to false, this already done in the loop above this line.

plugins/single_plugins/idle.cpp
soreau: It can be renamed but output_inhibited is a bad name because this is actually true when the output is not inhibited and the cube screensaver is running.

plugins/single_plugins/idle.cpp
ammen99: Oh hm I meant map.clear() so that we don't keep pointers which may become dangling...

plugins/single_plugins/idle.cpp
ammen99: I see. So what about screensaver_enabled?

soreau: I went with screensaver_hook_set since that's what it means.

plugins/single_plugins/idle.cpp
soreau: Ok.

plugins/single_plugins/idle.cpp
soreau: Done.

soreau:
@ammen99 All of the changes requested should be addressed.

ammen99:
Seems to work. Thanks for the hard work!

Issue #15, opened 2018-06-16
Dynamic shell configuration


ammen99:
The built-in shell needs a proper event loop and dynamic configuration. On top of that, it'll be nice to have a better architecture for widgets.

ammen99:
The built-in shell has been demoted to just a demo. Use https://github.com/WayfireWM/wf-shell instead. It has dynamic config, better outlook and more.

PR #337, opened 2019-11-18
Add wlroots as subproject

soreau:
Requires wf-config #15.


soreau:
At first I built wlroots as a static library because I didn't want to overwrite wlroots in the same prefix when installing wayfire. This worked but I had to 1) Patch wlroots to not install headers when built with default_library=static 2) Pass -Ddefault_library=static to wayfire. As it turns out, this option is global to the project and cannot be different for subprojects. There is a patch to fix this though (meson PR #4157). As it stands currently, the version of wlroots included with this PR is unpatched but is set at b81bb2ef30, which is the commit directly before the major x11 backend breakage found in a3c3b928a3.

PR #342, opened 2019-11-30
Avoid damage on vswipe (credit: ammen99)

szbergeron:
Integrates changes posted in chat by @ammen99  that avoid damaging buffers on workspace switch (reducing stutter while using vswipe)


ammen99:
Thanks :)

PR #356, opened 2019-12-27
Wf config update

ammen99:
Update for the changes in https://github.com/WayfireWM/wf-config/pull/16

This is a major breaking change. Simple guideline for updating external plugins:

1. All wayfire headers have been moved in a `wayfire/` subdirectory. So for ex. `#include <core.hpp>` -> `#include <wayfire/core.hpp>`

2. Plugins' `init()` no longer receives the configuration as an argument, simply remove it.
3. Loading options generally happens through `wf::option_wrapper_t`. Look at any existing plugin for information how to use it.
4. If you need access to the raw configuration manager, use `wf::get_core().config`
5. Create and install a XML containing configuration options and their types. The correct installation path is provided in the .pc file.

EDIT: further breaking changes:

6. Many functions and types were moved in the `wf` namespace. So you need to change `wf_point` -> `wf::point_t`, `wf_geometry` -> `wf::geometry_t`, etc. There has been no functional change, but this is an effort to have a consistency across the database.

I hope this will be the last major breakage I do :)


soreau:
>     2. Plugins' `init()` no longer receives the configuration as an argument, simply remove it.

Also need to add override after init() and fini().

>     5. Create and install a XML containing configuration options and their types. The correct installation path is provided in the .pc file.

I don't see this PR. I'm guessing we'll want something like [this](http://ix.io/25Eo).

ammen99:
> Also need to add override after init() and fini().

This is not strictly necessary, but I consider it mandatory for in-house plugins.

PR #358, opened 2019-12-27
build: Add icondir variable to wayfire.pc

soreau:
This also introduces icondir variable to wayfire.pc file.


ammen99:
I'd specify only icondir in the .pc file and then put icons in WCM (because icons are not needed by Wayfire at all).

soreau:
> I'd specify only icondir in the .pc file and then put icons in WCM (because icons are not needed by Wayfire at all).

That makes sense, done.

PR #361, opened 2019-12-31
screensaver: Add ease deformation to cube control

soreau:
This allows a way for the idle plugin to set cube ease value
so deformation is included in the duration. Without this, the
deformation is enabled and disabled without a transition,
which isn't very pretty. This also fixes glitches every second
when the cube is rotating and seconds are enabled in the panel
clock.


ammen99:
Thanks!

PR #362, opened 2019-12-31
Track wlroots

soreau:


soreau:
> Why do we compare to pending state? Shouldn't the comparison be done with the current state?

I was trying to follow [this sway patch](https://github.com/swaywm/sway/commit/ef2332f7a648c208bb0c601f670823162f584664). Wayfire still starts without changing to pending but changing modes via the config file is still broken either way.

PR #371, opened 2020-01-03
Fix broken background_color option

### LINKED ISSUE #370: core/background_color option doesn't work
From irc:
> adrianvovk: I have `background_color = 0.0 0.0 1.0 1.0` but the background is still black and not blue
> soreau: where are you setting this?
> soreau: core section?
> adrianvovk: Yep
> soreau: yea it broke somehow
> soreau: file an issue
> adrianvovk: Alright will do

soreau:
This fixes #370.

### DIFF
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -337,8 +337,6 @@ class wf::render_manager::impl
     std::unique_ptr<postprocessing_manager_t> postprocessing;
 
     wf::option_wrapper_t<wf::color_t> background_color_opt;
-    /* The default color which is user configurable */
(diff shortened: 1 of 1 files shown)


ammen99:
You should rather remove the default_background and get the value fr the option directly.

soreau:
> You should rather remove the default_background and get the value fr the option directly.

NAK, see https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/workspace-stream.hpp#L20 and https://github.com/WayfireWM/wayfire/blob/master/src/output/render-manager.cpp#L883

ammen99:
Nice

PR #372, opened 2020-01-03
core: Add plugins option

soreau:
Reference https://github.com/WayfireWM/wf-config/issues/19


ammen99:
I thought you wanted to hide this option in WCM :)

PR #374, opened 2020-01-03
Pipe sysconfdir into build_config() for optional defaults

soreau:


meson.build
```diff
@@ -48,6 +48,7 @@ conf_data.set('INSTALL_PREFIX', get_option('prefix'))
 conf_data.set('PLUGIN_PATH', join_paths(get_option('prefix'), get_option('libdir'), 'wayfire'))
 metadata_dir_suffix = 'share/wayfire/metadata'
 conf_data.set('PLUGIN_XML_DIR', join_paths(get_option('prefix'), metadata_dir_suffix))
+conf_data.set('SYSCONFDIR', '/' + get_option('sysconfdir'))
```
ammen99: use join_paths

soreau: > use join_paths

I don't think this is what it looks like. SYSCONFDIR is a value being set and `/` is a prefix for `etc`, which is the value of `get_option('sysconfdir')`. So there's not more than one path to join here.

ammen99:
Thanks :)

PR #375, opened 2020-01-04
Submodules

ammen99:
Supersedes #337

Note to package maintainers: ~~With this pull request, wf-config will be built together with Wayfire by default. You can use `-Duse_system_wlroots` and `-Duse_system_wfconfig` to configure the behavior.~~

wf-config and wlroots are now built by default **only** if they are not already installed, or if explicitly configured.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "subprojects/wf-config"]
 	path = subprojects/wf-config
 	url = https://github.com/ammen99/wf-config
+[submodule "subprojects/wlroots"]
+	path = subprojects/wlroots
+	url = https://github.com/swaywm/wlroots
--- a/README.md
+++ b/README.md
@@ -16,6 +16,13 @@ meson build --prefix=/usr --buildtype=release
 ninja -C build && sudo ninja -C build install
 ```
 
+If you want to build `wf-config` as a submodule of Wayfire, add `-Duse_system_wfconfig=disabled` to the second line.
+
+If you want to build `wlroots` as a submodule of Wayfire, add `-Duse_system_wlroots=disabled` to the second line.
+
+Note that **any** of `wf-config` and `wlroots` will be built as submodules if they are not already available and if you don't explicitly turn these submodules off.
+
+
 # Packaging status
 
 - [Fedora](https://apps.fedoraproject.org/packages/wayfire) (31+): `sudo dnf install wayfire`
--- a/meson.build
+++ b/meson.build
@@ -112,11 +112,11 @@ summary = [
 	'',
     'system wfconfig: @0@'.format(use_system_wfconfig),
     ' system wlroots: @0@'.format(use_system_wlroots),
-	'   debug output: @0@'.format(conf_data.get('WAYFIRE_DEBUG_ENABLED')),
-	'        imageio: @0@'.format(conf_data.get('BUILD_WITH_IMAGEIO')),
-	'         gles32: @0@'.format(conf_data.get('USE_GLES32')),
-	'   graphics dbg: @0@'.format(conf_data.get('WAYFIRE_GRAPHICS_DEBUG')),
-	'----------------',
-	''
+    '   debug output: @0@'.format(conf_data.get('WAYFIRE_DEBUG_ENABLED')),
+    '        imageio: @0@'.format(conf_data.get('BUILD_WITH_IMAGEIO')),
+    '         gles32: @0@'.format(conf_data.get('USE_GLES32')),
+    '   graphics dbg: @0@'.format(conf_data.get('WAYFIRE_GRAPHICS_DEBUG')),
+    '----------------',
+    ''
 ]
 message('\n'.join(summary))
--- a/meson.build
+++ b/meson.build
@@ -4,7 +4,7 @@ project(
 	'cpp',
 	version: '0.3.1',
 	license: 'MIT',
-	meson_version: '>=0.43.0',
+	meson_version: '>=0.47.0',
 	default_options: [
 		'cpp_std=c++17',
 		'c_std=c11',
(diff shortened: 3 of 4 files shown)


meson.build
```diff
@@ -55,10 +66,6 @@ conf_data.set('SYSCONFDIR', join_paths('/', get_option('sysconfdir')))
 add_project_arguments(['-DWLR_USE_UNSTABLE'], language: ['cpp', 'c'])
 add_project_link_arguments(['-rdynamic', '-Wl,-E'], language: 'cpp')
 
-ccp = meson.get_compiler('cpp')
-wlroots_has_xwayland = ccp.get_define('WLR_HAS_XWAYLAND', prefix: '#include <wlr/config.h>', dependencies: wlroots) == '1'
-wlroots_has_x11_backend = ccp.get_define('WLR_HAS_X11_BACKEND', prefix: '#include <wlr/config.h>', dependencies: wlroots) == '1'
-
```
soreau: It would be nice to see these in the configure output if get_option('use_system_wlroots'). Can we output these with the rest of the variables conditionally dependent on get_option('use_system_wlroots')?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "subprojects/wf-config"]
 	path = subprojects/wf-config
 	url = https://github.com/ammen99/wf-config
+[submodule "subprojects/wlroots"]
+	path = subprojects/wlroots
+	url = https://github.com/swaywm/wlroots
```
soreau: Do we want to use the upstream url or one under our control? If we use a url that is within our control, we can update wlroots and fix up wayfire at our leisure and this should make it less likely for users to experience breakage.

ammen99: I don't think we will have many patches for wlroots, and it generally makes sense to keep as close to upstream as possible. Of course, we can adjust the submodule to any commit we want, so I see no problems to change the submodule to our own fork if the need arises.

meson.build
ammen99: fixed

meson.build
```diff
@@ -99,18 +106,31 @@ subdir('plugins')
 
 install_subdir('shaders', install_dir: 'share/wayfire')
 
-summary = [
+wlroots_xwayland_info = []
+
+if use_system_wlroots
+  ccp = meson.get_compiler('cpp')
```
soreau: Maybe the variable should be named cpp instead of ccp? I think this was my mistake but probably should fix it while it's on the table.

.gitmodules
soreau: I would like to see https://github.com/swaywm/sway/issues/1865#issuecomment-450659415 for one.

meson.build
ammen99: done

meson.build
```diff
@@ -99,18 +106,31 @@ subdir('plugins')
 
 install_subdir('shaders', install_dir: 'share/wayfire')
 
-summary = [
+wlroots_xwayland_info = []
```
soreau: A nitpick, maybe rename this to wlroots_x_info since it includes xwayland and x11 backend info.

.gitmodules
ammen99: alright, I'll fork wlroots

ammen99: you can send it as a PR to WayfireWM/wlroots

README.md
```diff
@@ -16,6 +16,11 @@ meson build --prefix=/usr --buildtype=release
 ninja -C build && sudo ninja -C build install
 ```
 
+If you want to use a system-wide installation of `wf-config`, just add `-Duse_system_wfconfig=enabled` to the second line.
+
+If you want to build wlroots automatically or you want to use the one provided by Wayfire, add `-Duse_system_wlroots=disabled` to the second line. This will also overwrite any previous installations in the destination that you have. 
+
```
soreau: I had to read these comments a few times to understand them, I think they could stand to be clarified a bit more. But the real problem I see is, as far as I understand it, that package maintainers can't really use -Duse_system_wlroots=disabled because then wayfire would conflict with the wlroots package because they would both provide the same files. And if package maintainers want to package wf-config, then if wayfire is built with -Duse_system_wfconfig=disabled, it would conflict with the wf-config package the same way. So either we should leave a note to maintainers saying that they should use -Duse_system_w*=enabled to package and 'disabled' is for users that are building manually or provide some solution to rectify the situation.

ddevault:
Can you clarify why it seems necessary to fork wlroots?

ammen99:
@ddevault The idea is to gather patches which have been rejected upstream (again, there aren't many of those, and I'll regularly synchronize the fork with upstream). There won't be massive changes for sure.

ddevault:
Can you enumerate those patches?

ammen99:
> Can you enumerate those patches?

Currently there is just a single patch, the last commit in the repo.

ddevault:
Two things:

1. As @emersion pointed out, you would probably be wiser to maintain your own xwm code than an entire wlroots fork.
2. Maybe we can reconsider this upstream as a workaround for a specific client problem, by allowing this only if focus is changing from an X window with class=steam to an X window with class=steam.

ammen99:
> 2\. Maybe we can reconsider this upstream as a workaround for a specific client problem, by allowing this only if focus is changing from an X window with class=steam to an X window with class=steam.

I was under the impression that no such hack (it is a hack, I agree) will be accepted under any circumstances. If you are willing to accept such a patch, then I was mistaken, and I'd gladly go back to using a specific wlroots commit for the submodule.

ddevault:
I might not have been there for all of the discussion about this, but I think that hack would be a reasonable add. I can't speak for anyone else who might have comments.

emersion:
A client can fake its class pretty easily. Maybe using the PID would be good enough. Its a shame we can't have a per-client identifier (can we?).

ddevault:
Yeah, let's let focus move in X if the target and destination are the same pid. Seems fine.

ammen99:
> Yeah, let's let focus move in X if the target and destination are the same pid. Seems fine.

I like this solution, because it can also help in other cases (I assume there could be other clients doing stupid things)

emersion:
Hmm, a client can probably fake its PID too by setting the property. Oh well, this is X11 anyway.

README.md
ammen99: Wait what? I agree the second line is confusing, I'll edit this, but I dont' understand what the problem with the rest is.

Current default behavior: Build wf-config as a submodule. If you are a packager, you "want to use system wf-config", so just use the option.

Also default: do not build the wlroots submodule. This is reasonable since I expect a lot of people already have wlroots installed/packaged.

PR #377, opened 2020-01-04
Fix crash on closing wireshark

masm11:
Sorry for no crash dump.

1. open gnome-terminal.
2. open wireshark from the terminal.
3. start capture.
4. wait for a packet to be displayed.
5. close by `Ctrl+Q`.
6. click `Stop and Quit without Saving`.
7. click the gnome-terminal window.
8. wayfire crashes.

I fixed this problem.

In `src/core/wm.cpp`:

Here, view is checked whether it is mapped. Otherwise return.
```
    if (!view || !view->is_mapped() || !view->get_keyboard_focus_surface()
        || !output->activate_plugin(grab_interface))
    {
        return;
    }
```

Here, get the root of the view.
```
    /* Raise the base view. Modal views will be raised to the top by
     * wayfire_handle_focus_parent */
    while (view->parent)
        view = view->parent.get();
```

Here, the view is focused.
```
    set_last_focus(view->self());
```

However, root view may not be mapped even if child view is mapped.

Here, in `set_last_focus()`, `last_focus` hold the view.
```
    last_focus = view;
```

`last_focus` is cleared when the view is disappeared, not when it is destroyed.

Totally, if the view is created, the child view gets focused (i.e. `last_focus` becomes the view), and the view is destroyed, then last_focus holds the destroyed view, and may be passed to `send_done()`.

So the view needs check again whether it is mapped.

(I don't know why the child view get focused even if the view is not mapped.
But I confirmed that occurs by inserting `printf()`.)

### DIFF
--- a/src/core/wm.cpp
+++ b/src/core/wm.cpp
@@ -115,6 +115,9 @@ void wayfire_focus::check_focus_surface(wf::surface_interface_t* focus)
     while (view->parent)
         view = view->parent.get();
 
+    if (!view->is_mapped())    /* check again */
+        return;
+
     view->get_output()->focus_view(view->self(), true);
     set_last_focus(view->self());
 }


ammen99:
Thanks for debugging this. Have you tried #363 ? It changes a lot in the way in which dialogues are handled, and I think it may resolve this issue as well. Otherwise, please rebase your changes on top of #363

masm11:
@ammen99 
I tried #363. It seems to resolve this issue, so I'll close it.
Thanks.

However, the dialog of wireshark is not shown until I click the icon in the wf-shell dock 3 times.

ammen99:
I reproduced the problem with wireshark. The issue stems from how the dialogues are handled in Qt/Wireshark. It creates an unmapped window (the WM leader) and then sets the dialogue as a child of that window. Since the leader is not mapped, we don't show the child view unless something strange happens (like minimizing a dialogue)

Can you try this patch: http://ix.io/26yL ? It fixes the problem with wireshark for me.

masm11:
@ammen99 I confirmed it is fixed. Thanks.

ammen99:
> @ammen99 I confirmed it is fixed. Thanks.

Nice, I pushed it to child-views, which should be merged as soon as #375 is merged.

PR #381, opened 2020-01-16
Respect --prefix when looking for wayfire/defaults.ini

jbeich:
By default on FreeBSD packages are installed under `/usr/local` and configuration is stored under `/usr/local/etc`.
This matches GNU autoconf default where `sysconfdir` under `prefix` by default.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -48,7 +48,7 @@ conf_data.set('INSTALL_PREFIX', get_option('prefix'))
 conf_data.set('PLUGIN_PATH', join_paths(get_option('prefix'), get_option('libdir'), 'wayfire'))
 metadata_dir_suffix = 'share/wayfire/metadata'
 conf_data.set('PLUGIN_XML_DIR', join_paths(get_option('prefix'), metadata_dir_suffix))
-conf_data.set('SYSCONFDIR', join_paths('/', get_option('sysconfdir')))
+conf_data.set('SYSCONFDIR', join_paths(get_option('prefix'), get_option('sysconfdir')))
 
 # needed to dlopen() plugins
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -48,7 +48,7 @@ conf_data.set('INSTALL_PREFIX', get_option('prefix'))
 conf_data.set('PLUGIN_PATH', join_paths(get_option('prefix'), get_option('libdir'), 'wayfire'))
 metadata_dir_suffix = 'share/wayfire/metadata'
 conf_data.set('PLUGIN_XML_DIR', join_paths(get_option('prefix'), metadata_dir_suffix))
-conf_data.set('SYSCONFDIR', join_paths('/', get_option('sysconfdir')))
+conf_data.set('SYSCONFDIR', join_paths(get_option('prefix'), get_option('sysconfdir')))
```
soreau: I think the intent is that --prefix can be set independently of --sysconfdir. For example, with your proposal, you wouldn't be able to set --prefix to /usr and SYSCONFDIR to /etc. With the current state, it shouldn't be a problem if you wanted to set sysconfdir relative to prefix, because prefix is known at configure time (e.g. --prefix=/usr --sysconfdir=/usr/etc). That said, is there a specific case where this might not work?

jbeich: `--prefix=/usr` defaults to `/etc` and `--sysconfdir=/usr/etc` works fine. I suspect `join_paths()` is magical. ;)

soreau:
I have tested this as well and it works as described. join_paths will discard prefix if sysconfdir is an absolute path.

ammen99:
Thanks!

PR #384, opened 2020-01-19
Add virtual pointer protocol

soreau:
This makes things like wayvnc pointer input work.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -4,6 +4,7 @@ extern "C"
 #include <wlr/types/wlr_data_control_v1.h>
 #include <wlr/types/wlr_data_device.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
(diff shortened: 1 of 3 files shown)


src/core/core.cpp
```diff
@@ -159,6 +160,14 @@ void wf::compositor_core_impl_t::init()
     });
     vkbd_created.connect(&protocols.vkbd_manager->events.new_virtual_keyboard);
 
+    protocols.vptr_manager = wlr_virtual_pointer_manager_v1_create(display);
+    vptr_created.set_callback([&] (void *data) {
+        auto event = (wlr_virtual_pointer_v1_new_pointer_event*) data;
+	auto ptr = event->new_pointer;
```
ammen99: indentation seems wrong here, maybe a mix of tabs and spaces? (should be 8 spaces here)

PR #389, opened 2020-01-23
Fix crash introduced by transformer optimizations merge

soreau:
This can be triggered when moving a view to another output
with keycolor enabled or trying to move a compositor view,
such as in magnifier.

### DIFF
--- a/src/view/surface.cpp
+++ b/src/view/surface.cpp
@@ -164,8 +164,11 @@ void wf::surface_interface_t::impl::scale_opaque_region(
      * different scales, we just shrink by 1 to compensate for the ceil/floor
      * discrepancy */
     int ceil_factor = 0;
-    if (output->handle->scale != (float)wsurface->current.scale)
+    if ((wsurface && output->handle->scale != (float)wsurface->current.scale) ||
+        (!wsurface && output->handle->scale != std::round(output->handle->scale)))
+    {
         ceil_factor = 1;
+    }
 
     region.expand_edges(-shrink - ceil_factor);
 }


src/view/surface.cpp
```diff
@@ -164,7 +164,7 @@ void wf::surface_interface_t::impl::scale_opaque_region(
      * different scales, we just shrink by 1 to compensate for the ceil/floor
      * discrepancy */
     int ceil_factor = 0;
-    if (output->handle->scale != (float)wsurface->current.scale)
+    if (wsurface && output->handle->scale != (float)wsurface->current.scale)
```
ammen99: In the `!wsurface` case we should check whether the output's scale is an integer. If not, we need to expand as well.
So maybe add an `else if (!wsurface && output->handle->scale != (float)(int)output->handle->scale)`

soreau: Did you mean to double cast output->handle->scale? What would be in the body of this conditional?

ammen99: the same as in the `if` case, which probably means it should be an `||` not else

src/view/surface.cpp
```diff
@@ -164,7 +164,8 @@ void wf::surface_interface_t::impl::scale_opaque_region(
      * different scales, we just shrink by 1 to compensate for the ceil/floor
      * discrepancy */
     int ceil_factor = 0;
-    if (output->handle->scale != (float)wsurface->current.scale)
+    if ((wsurface && output->handle->scale != (float)wsurface->current.scale) ||
+        (!wsurface && output->handle->scale != std::round(output->handle->scale)))
         ceil_factor = 1;
```
ammen99: style: since this is a multiline if, it needs parenthesis.
Tbh I'm getting more and more inclined towards saying we should always use parenthesis so that the rule is simple

soreau: > style: since this is a multiline if, it needs parenthesis.
> Tbh I'm getting more and more inclined towards saying we should always use parenthesis so that the rule is simple

My personal rule is to only use braces when necessary. This basically means use braces unless it's a conditional with only one LOC in the body. It seems strange to me that braces would be required depending on how many lines the conditional happens to be. I think it would make more sense to require braces always since you seem to be quite pedantic about it.

ammen99:
Good catch.

PR #390, opened 2020-01-23
Set core xwayland_display string

soreau:
This was previously an unused variable. It allows external
plugins to get the display string, since xwayland_get_display()
is not exported.


soreau:
Optionally, we could make this an int instead of string.

ammen99:
I'd like to see this as a function (I know it is a variable currently, but let's do it properly since we're already at it)

soreau:
Is that to say you would like to get rid of xwayland_display string altogether and replace it with a function that internally calls xwayland_get_display()?

ammen99:
LGTM. Thanks!

PR #395, opened 2020-01-26
build: compatible with wlroots-0.10.0

### LINKED ISSUE #393: wlroots updated to 0.10.0, breaks build
So, wlroots was recently updated to 0.10.0, which broke the version requirement on the meson.build file. Considering it must have been there for a reason, even if the compilation step with 0.10.0 worked without issues (after quickly patching meson.build), I'm going to test it out tomorrow and see how it does. Are there any specific reasons for the version requirement?

travankor:
closes #393


ammen99:
You should also bump the wlroots submodule version.

travankor:
oh neat, I didn't realize wlroots can be built as a submodule.

ammen99:
Thanks!

PR #400, opened 2020-02-08
View fullscreen

soreau:
This PR emits a view-fullscreen signal for the output and adds state to the view fullscreen signal. In addition, it renames autohide-panels signal to fullscreen-layer-focused, disable_dpms_on_fullscreen to disable_on_fullscreen and makes idle plugin listen for the fullscreen-layer-focused signal to disable idle completely if disable_on_fullscreen is enabled.

### DIFF
--- a/plugins/single_plugins/idle.cpp
+++ b/plugins/single_plugins/idle.cpp
@@ -9,10 +9,12 @@ extern "C"
 #include "wayfire/core.hpp"
 #include "wayfire/output-layout.hpp"
 #include "wayfire/workspace-manager.hpp"
+#include "wayfire/signal-definitions.hpp"
 #include "../cube/cube-control-signal.hpp"
 
 #include <cmath>
 #include <wayfire/util/duration.hpp>
+#include <wayfire/util/log.hpp>
 
 #define ZOOM_BASE 1.0
 
@@ -50,6 +52,7 @@ class wayfire_idle
     std::map<wf::output_t*, bool> screensaver_hook_set;
     bool outputs_inhibited = false;
     bool idle_enabled = true;
+    int idle_inhibit_ref = 0;
     uint32_t last_time;
 
     wlr_idle_timeout *timeout_screensaver = NULL;
@@ -343,6 +346,33 @@ class wayfire_idle
         wf::get_core().output_layout->apply_configuration(config);
     }
 
+    void idle_enable()
+    {
+        idle_inhibit_ref--;
+
+        if (idle_inhibit_ref < 0)
+        {
+            LOGE("idle_inhibit_ref < 0: ", idle_inhibit_ref);
+        }
+
+        if (idle_inhibit_ref == 0 &&
+            !idle_enabled)
+        {
+            toggle_idle();
+        }
+    }
+
+    void idle_inhibit()
+    {
+        idle_inhibit_ref++;
+
+        if (idle_inhibit_ref == 1 &&
+            idle_enabled)
+        {
+            toggle_idle();
+        }
+    }
+
     void toggle_idle()
     {
         idle_enabled ^= 1;
@@ -361,6 +391,13 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
         return true;
     };
 
+    wf::signal_connection_t fullscreen_state_changed{[this] (wf::signal_data_t *data)
+    {
+        bool state = data ? true : false;
+
+        state ? get_instance().idle_inhibit() : get_instance().idle_enable();
+    }};
+
     void init() override
     {
         singleton_plugin_t::init();
@@ -370,6 +407,12 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
         output->add_activator(
             wf::option_wrapper_t<wf::activatorbinding_t>{"idle/toggle"},
             &toggle);
+        output->connect_signal("fullscreen-layer-focused", &fullscreen_state_changed);
+
+        for (auto& view : output->workspace->get_views_in_layer(wf::LAYER_FULLSCREEN))
+        {
+            get_instance().idle_inhibit();
+        }
     }
 
     void fini() override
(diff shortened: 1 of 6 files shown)


src/view/view.cpp
```diff
@@ -403,7 +403,13 @@ void wf::view_interface_t::set_fullscreen(bool full)
...
         get_output()->workspace->add_view(self(), wf::LAYER_WORKSPACE);
 
-    this->emit_signal("fullscreen", nullptr);
+    view_fullscreen_signal data;
+    data.view = self();
+    data.state = full;
+    data.desired_size = get_output()->get_relative_geometry();
+
+    get_output()->emit_signal("view-fullscreen", &data);
+    this->emit_signal("fullscreen", reinterpret_cast<signal_data_t*> (full));
```
ammen99: Maybe we can just emit the same data as for the `view-fullscreen` signal and change any previous "subscribers"

plugins/single_plugins/idle.cpp
```diff
@@ -361,15 +350,39 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
...
+                wlr_idle_set_enabled(wf::get_core().protocols.idle, NULL, true);
+            }
+        };
+
+        disable_on_fullscreen.set_callback(disable_on_fullscreen_changed);
+
         output->add_activator(
             wf::option_wrapper_t<wf::activatorbinding_t>{"idle/toggle"},
             &toggle);
+        output->connect_signal("view-fullscreen", &fullscreen_state_changed);
```
ammen99: Use the `autohide-panels` signal here.
This is pretty unreliable with multiple fullscreen views, fullscreen view getting unfocused even if it remains focused, etc.

plugins/single_plugins/idle.cpp
```diff
@@ -361,15 +350,38 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
         return true;
     };
 
+    wf::signal_connection_t fullscreen_state_changed{[this] (wf::signal_data_t *data)
+    {
+        bool state = data ? true : false;
```
ammen99: This will not work with multiple outputs and multiple fullscreen views, right? You can have fullscreen on both outputs, then if one of them is unfullscreened ...

Maybe you can change idle_enabled to an int counter to keep track of this.

soreau: I made a counter for fullscreen but kept the toggle function independent so it would always work.

plugins/single_plugins/idle.cpp
```diff
@@ -329,6 +330,24 @@ class wayfire_idle
...
+        {
+            idle_ref++;
+
+            if (idle_ref == 1 && idle_enabled)
+                toggle_idle();
+        }
+        else
+        {
+            idle_ref--;
+
```
ammen99: would probably be also nice to print an error in case idle_ref gets below zero since that'd indicate a bug in the plugin.

plugins/single_plugins/idle.cpp
```diff
@@ -58,6 +58,7 @@ class wayfire_idle
 
   public:
     bool idle_enabled = true;
```
ammen99: Why use idle_enabled when we have idle_ref ? Also, please choose something better than `idle_ref`, it isn't clear whether `idle_ref` is a count for things that want to enable or to disable idling.

soreau: Since there is a toggle_idle function bound to a keybinding, to retain this functionality, it needs idle_enabled.

soreau:
Should idle ref count for each fullscreen view when it's first loaded? I think if it's enabled while there are views already fullscreen, things could get into a bad state.

plugins/single_plugins/idle.cpp
```diff
@@ -343,6 +344,38 @@ class wayfire_idle
...
+    void idle_enable()
+    {
+        idle_inhibit_ref--;
+
+        if (idle_inhibit_ref < 0)
+        {
+            LOGE("idle_inhibit_ref < 0: ", idle_inhibit_ref);
+        }
+
+        if (idle_inhibit_ref == 0 &&
```
ammen99: new line here seems weird? the line is short anyway

plugins/single_plugins/idle.cpp
```diff
@@ -343,6 +344,38 @@ class wayfire_idle
...
+    void idle_inhibit()
+    {
+        idle_inhibit_ref++;
+
+        if (!disable_on_fullscreen)
+        {
+            return;
+        }
+
+        if (idle_inhibit_ref == 1 &&
```
ammen99: same as above

src/view/view.cpp
```diff
@@ -403,7 +403,13 @@ void wf::view_interface_t::set_fullscreen(bool full)
     if (!fullscreen && get_output() && needs_lowering)
         get_output()->workspace->add_view(self(), wf::LAYER_WORKSPACE);
 
-    this->emit_signal("fullscreen", nullptr);
+    view_fullscreen_signal data;
```
ammen99: I would like to document this signal in `signal-definitions.hpp` (the goal is to eventually have all signals documented, along with the appropriate types).

Also, I think we could leave out desired size, since it is not always the full geometry. Let's leave it "undefined" or "0,0 0x0"

src/api/wayfire/signal-definitions.hpp
```diff
@@ -67,6 +67,13 @@ struct view_tiled_signal : public _view_signal
     wf::geometry_t desired_size;
 };
 
+/* sent when the view fullscreen state changes.
```
ammen99:
```suggestion
/**
 * The view-fullscreen-request and view-fullscreen signals are emitted on the view's output when the view's fullscreen state changes.
 * view-fullscreen-request is emitted when the view needs to be fullscreened, but has not been fullscreened yet.
 * view-fullscreen is emitted whenever the view's fullscreen state actually changes.
```

ammen99:
Thanks for the patience :)

PR #413, opened 2020-02-14
vswipe: Merge vswitch capabilities

soreau:
Merging vswitch capabilities into vswipe means the shell is animated
when switching using keybindings.

### DIFF
--- a/plugins/single_plugins/vswipe.cpp
+++ b/plugins/single_plugins/vswipe.cpp
@@ -11,6 +11,7 @@
 #include <glm/glm.hpp>
 #include <glm/gtc/matrix_transform.hpp>
 #include <wayfire/util/duration.hpp>
+#include "view-change-viewport-signal.hpp"
 
 #include <cmath>
 #include <utility>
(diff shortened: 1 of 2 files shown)


plugins/single_plugins/vswipe.cpp
```diff
@@ -76,6 +80,118 @@ class vswipe : public wf::plugin_interface_t
...
+        struct wlr_event_pointer_swipe_update swipe_update_ev = {
+                .device = nullptr,
+                .time_msec = 0,
+                .fingers = 0,
+                .dx = x * -100,
+                .dy = y * -100,
+        };
+        swipe_update.event = &swipe_update_ev;
+
+        on_swipe_begin(&swipe_begin);
```
ammen99: Would be nice if you could extract the functionality of `swipe_begin`, etc. to a separate function which is called both here and at the other place.

Simulating events is usually a hack which in this case can be avoided quite easily.

ammen99:
I just realized this is going to be more work than I initially thought.

The vswipe plugin currently renders only 2 workspaces, but this will probably need to change. There are two reasons for that:

1. It would be nice to be able to switch almost diagonally with 2 fast keybindings pressed.

2. The default workspace switch plugin (we want vswipe to become this one) needs to handle arbitrary workspace changes, because when view is activated from the panel's window-list, we need to somehow switch to its workspace. We have a request which does exactly this, but vswipe needs to be able to handle it.

If you don't feel like it, I could probably implement it soon.

soreau:
I had considered both these points and I think you'd better handle it.

tschuchortdev:
I just have to say this is seriously awesome. I have been looking for a DE that supports workspace gestures like this for a long time. I have high hopes for this project!

PR #414, opened 2020-02-17
Warnings and stuff

myfreeweb:
make clang less annoyed + add [editorconfig](https://editorconfig.org/)

### DIFF
--- a/plugins/wobbly/wobbly.cpp
+++ b/plugins/wobbly/wobbly.cpp
@@ -504,8 +504,6 @@ class wf_wobbly : public wf::view_transformer_t
             wf::OUTPUT_EFFECT_PRE);
     };
 
-    const wf::plugin_grab_interface_uptr& iface;
-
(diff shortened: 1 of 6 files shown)


plugins/wobbly/wobbly.cpp
```diff
@@ -532,8 +530,7 @@ class wf_wobbly : public wf::view_transformer_t
     }
 
   public:
-    wf_wobbly(wayfire_view view, const wf::plugin_grab_interface_uptr& _iface)
-        : iface(_iface)
+    wf_wobbly(wayfire_view view, const wf::plugin_grab_interface_uptr&)
```
ammen99: since the interface is no longer needed, maybe we can just remove the parameter altogether?

myfreeweb: lol sorry, when I see a constructor in a plugin type thing, I immediately assume it has to conform to some interface from the plugin system :D

PR #419, opened 2020-02-22
Revamp option defaults

alexherbo2:


metadata/zoom.xml
```diff
@@ -7,12 +7,12 @@
 		<option name="modifier" type="key">
 			<_short>Zoom in the desktop by scrolling with modifier</_short>
 			<_long>Scrolls with the specified modifier to zoom in and out.</_long>
-			<default>&lt;super&gt;</default>
+			<default>&lt;super&gt; &lt;ctrl&gt;</default>
```
soreau: Since this is an accessibility feature, it would make sense for the default to have a single modifier, so it's easier to find. My personal preference is the super key, as it stands.

metadata/vswitch.xml
```diff
@@ -7,47 +7,47 @@
 		<option name="binding_left" type="activator">
 			<_short>Left</_short>
 			<_long>Switches to workspace left with the specified activator.</_long>
-			<default>&lt;super&gt; KEY_LEFT</default>
```
soreau: I don't see the need to change the KEY_[DIRECTION] defaults for vswitch. The arrow keys are what a user is more likely to expect as a default.

metadata/simple-tile.xml
```diff
@@ -18,42 +18,42 @@
 		<option name="button_move" type="button">
 			<_short>Button move</_short>
 			<_long>When the specified button is held down, you can drag and drop tiled windows to reorder them and to change the tiling direction.</_long>
-			<default>&lt;super&gt; BTN_LEFT</default>
```
soreau: I don't see a reason to set the simple-tile bindings to none.

metadata/grid.xml
```diff
@@ -30,52 +30,52 @@
 		<option name="slot_bl" type="activator">
 			<_short>Bottom left</_short>
 			<_long>Positions the window in the bottom left corner of the screen with the specified activator.  The window takes 1/4 of the screen.</_long>
-			<default>&lt;alt&gt; &lt;ctrl&gt; KEY_KP1</default>
+			<default>&lt;super&gt; &lt;shift&gt; KEY_HOME | &lt;super&gt; KEY_KP1</default>
```
soreau: It seems conflicting that there's a set of keys with two modifiers and a set with one, to do the same action. I would rather drop the former set in favor of the latter alone, because the key positioning doesn't make much sense on a traditional keyboard layout.

metadata/fisheye.xml
```diff
@@ -7,12 +7,12 @@
 		<option name="toggle" type="activator">
 			<_short>Toggle</_short>
 			<_long>Toggles fisheye with the specified key.</_long>
-			<default>&lt;super&gt; KEY_F</default>
+			<default>&lt;super&gt; &lt;ctrl&gt; BTN_LEFT</default>
```
soreau: I would rather see this combined with super+F than replace it.

metadata/fast-switcher.xml
```diff
@@ -7,7 +7,7 @@
 		<option name="activate" type="key">
 			<_short>Activate</_short>
 			<_long>Activates fast switcher with the specified key.</_long>
-			<default>&lt;alt&gt; KEY_TAB</default>
+			<default>&lt;alt&gt; KEY_ESC</default>
```
soreau: Alt+Tab is a very common binding for switching windows and should be included.

metadata/expo.xml
```diff
@@ -7,12 +7,12 @@
 		<option name="toggle" type="activator">
 			<_short>Toggle</_short>
 			<_long>Shows an overview of all workspaces with the specified activator.  Pressing again exits.</_long>
-			<default>&lt;super&gt; KEY_E</default>
+			<default>&lt;super&gt;</default>
```
soreau: I don't think this is a good choice for a default but maybe @ammen99 can chime in here.

metadata/decoration.xml
```diff
@@ -28,12 +28,12 @@
 		<option name="active_color" type="color">
 			<_short>Color when window is active</_short>
 			<_long>Sets the color when the window is active.</_long>
-			<default>#222222aa</default>
+			<default>0.6 0.6 0.6 1.0</default>
```
soreau: Is there a reason to change these or are these values your preference?

metadata/cube.xml
```diff
@@ -11,17 +11,17 @@
 			<option name="activate" type="button">
 				<_short>Activate</_short>
 				<_long>Activates the cube with the specified button.</_long>
-				<default>&lt;alt&gt; &lt;ctrl&gt; BTN_LEFT</default>
+				<default>&lt;super&gt; &lt;shift&gt; BTN_LEFT</default>
```
soreau: NAK. Ctrl+Alt+Drag should be the default.

metadata/cube.xml
```diff
@@ -11,17 +11,17 @@
...
 			</option>
 			<option name="rotate_left" type="activator">
 				<_short>Rotate left</_short>
 				<_long>Rotates left with the specified activator.</_long>
-				<default>&lt;alt&gt; &lt;ctrl&gt; KEY_LEFT</default>
+				<default>&lt;super&gt; &lt;ctrl&gt; KEY_H</default>
```
soreau: You should either keep the left/right bindings as they are or just disable them.

metadata/core.xml
```diff
@@ -7,34 +7,12 @@
...
-				wrot
-				zoom
-			</default>
+			<default>alpha animate autostart command decoration expo fast-switcher fisheye grid idle invert move oswitch place resize switcher vswitch window-rules wobbly wrot zoom</default>
 		</option>
 		<option name="close_top_view" type="activator">
 			<_short>Close view</_short>
 			<_long>Closes the currently focused window with the specified key.</_long>
-			<default>&lt;super&gt; KEY_Q | &lt;alt&gt; KEY_F4</default>
+			<default>&lt;super&gt; KEY_W | &lt;super&gt; BTN_MIDDLE | &lt;alt&gt; KEY_F4</default>
```
soreau: super+BTN_MIDDLE might happen by mistake when trying to do another binding. This is not likely to happen with alt+f4. super+KEY_W seems strange to me, KEY_Q makes more sense here.

metadata/alpha.xml
```diff
@@ -7,7 +7,7 @@
 		<option name="modifier" type="key">
 			<_short>Modifier</_short>
 			<_long>When the modifier key is held down, you can scroll down and up to adjust the opacity of the window.</_long>
-			<default>&lt;alt&gt;</default>
+			<default>&lt;super&gt;</default>
```
soreau: This should be retained so zoom can remain super+scroll.

metadata/zoom.xml
alexherbo2: I chose it to match applications with <kbd>Control</kbd> + scroll.

metadata/simple-tile.xml
alexherbo2: `simple-tile` is not enabled by default and these bindings will likely conflict with the user config.

metadata/grid.xml
alexherbo2: The grid functionality is unreachable without a keypad.

metadata/fisheye.xml
alexherbo2: It is already used for the grid.
Rationale for a button is it affects the cursor.
I rather prefer the fancy effects be on the <kbd>Control</kbd> in general.

metadata/decoration.xml
alexherbo2: I did not know the equivalent in this format.
I took the occasion to experiment variants.

metadata/cube.xml
alexherbo2: It matches well with the config as a whole.

https://github.com/WayfireWM/wayfire/wiki/Tutorial#default-key-bindings

metadata/cube.xml
alexherbo2: It matches well with the config as a whole.  <kbd>Super</kbd> being the default modifier.

https://github.com/WayfireWM/wayfire/wiki/Tutorial#default-key-bindings

metadata/alpha.xml
alexherbo2: I rather prefer a single modifier.  <kbd>Super</kbd> being the default.

metadata/vswitch.xml
alexherbo2: It is used for the grid.
`HJKL` is used for navigating workspaces and the arrow keys and keypad for the grid.

https://github.com/WayfireWM/wayfire/wiki/Tutorial#default-key-bindings

metadata/simple-tile.xml
ammen99: The plugins are designed in such a way that this is fine. I intentionally made both bindings the same: simple-tile will be used only for tiled windows, and the move binding will be used only for floating windows.

metadata/grid.xml
ammen99: Here I agree with @alexherbo2, these make sense for a 14' laptop :)

However I think an even better solution would be: Retain the keypad bindings for big keyboards.
For small keyboards, add only left/right/center slots with <super> + ARROW, and make <super> + DOWN restore the window (the rest are only reachable via a keypad). These actions match the behavior of other DEs like GNOME.

metadata/fisheye.xml
ammen99: maybe ctrl-super-F? I don't think this should be a button binding by default.

metadata/fast-switcher.xml
ammen99: I think alt-tab should go to switcher actually, and here alt-esc as proposed.

metadata/expo.xml
ammen99: I use `<super>` for expo personally so +1 from me :P

metadata/alpha.xml
ammen99: I think zoom is generally more important than alpha, so we can use super+scroll for zoom, and here something else.

ammen99: Maybe put super+alt for alpha

ammen99:
I took some of your changes here, and updated them with our suggestions in #440. Hence I'm closing this PR.

Anyway, thank you for bringing this up, and starting the effort to improve Wayfire's default configuration.

PR #422, opened 2020-02-26
add some changes to wayfire.ini

librewish:
some changes that i propose

### DIFF
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -224,9 +224,9 @@ command_logout = wlogout
 # https://wayland.emersion.fr/grim/
 # https://wayland.emersion.fr/slurp/
 binding_screenshot = KEY_SYSRQ | KEY_PRINT
-command_screenshot = grim $HOME/Pictures/screenshot-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'screenshot saved in ~/Pictures'
+command_screenshot = grim $HOME/Pictures/screenshot-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'saved in ~/Pictures'
 binding_screenshot_interactive = <shift> KEY_SYSRQ | <shift> KEY_PRINT
-command_screenshot_interactive = slurp | grim -g - ~/Pictures/slurped-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'screenshot saved in ~/Pictures'
+command_screenshot_interactive = slurp | grim -g - ~/Pictures/slurped-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'saved in ~/Pictures'
 
 # Volume controls using alsa
 # https://alsa-project.org
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -224,9 +224,9 @@ command_logout = wlogout
 # https://wayland.emersion.fr/grim/
 # https://wayland.emersion.fr/slurp/
 binding_screenshot = KEY_SYSRQ | KEY_PRINT
-command_screenshot = grim $(date '+%F_%T').webp
+command_screenshot = grim $HOME/Pictures/screenshot-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'screenshot saved in ~/Pictures'
 binding_screenshot_interactive = <shift> KEY_SYSRQ | <shift> KEY_PRINT
-command_screenshot_interactive = slurp | grim -g - $(date '+%F_%T').webp
+command_screenshot_interactive = slurp | grim -g - ~/Pictures/slurped-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'screenshot saved in ~/Pictures'
 
 # Volume controls using alsa
 # https://alsa-project.org
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -200,6 +200,11 @@ command_terminal = alacritty | gnome-terminal
 binding_launcher = <super> <shift> KEY_ENTER
 command_launcher = wofi
 
+# sgtk-grid
+# https://github.com/nwg-piotr/sgtk-menu
+binding_sgtk-grid = <super> KEY_SPACE
+command_sgtk-grid = sgtk-grid
+
 # clipman - A simple clipboard manager for Wayland 
 # https://github.com/yory8/clipman 
 binding_clipman = <super> KEY_H
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -15,7 +15,9 @@
 # [input]
 # xkb_layout = us,fr
 # xkb_variant = dvorak,bepo
-#
+# cursor_theme = xcursor-breeze
+# tap_to_click = true
+
 # See Input options for a complete reference.
 # https://github.com/WayfireWM/wayfire/wiki/Configuration#input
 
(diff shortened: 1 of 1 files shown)


wayfire.ini
```diff
@@ -174,12 +223,12 @@ command_logout = wlogout
 # Screenshots
 # https://wayland.emersion.fr/grim/
 # https://wayland.emersion.fr/slurp/
-binding_screenshot = KEY_PRINT
-command_screenshot = grim $(date '+%F_%T').webp
-binding_screenshot_interactive = <shift> KEY_PRINT
-command_screenshot_interactive = slurp | grim -g - $(date '+%F_%T').webp
+binding_screenshot = KEY_SYSRQ | KEY_PRINT
```
ammen99: +1

wayfire.ini
```diff
@@ -174,12 +223,12 @@ command_logout = wlogout
...
+command_screenshot = grim $HOME/Pictures/screenshot-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'saved in ~/Pictures'
+binding_screenshot_interactive = <shift> KEY_SYSRQ | <shift> KEY_PRINT
+command_screenshot_interactive = slurp | grim -g - ~/Pictures/slurped-$(date "+%Y-%m-%d-%H:%M:%S").png && notify-send -i "camera" 'screenshot' 'saved in ~/Pictures'
 
-# Volume controls
+# Volume controls using alsa
```
ammen99: +1

wayfire.ini
```diff
@@ -151,8 +190,8 @@ idle = swayidle before-sleep swaylock
 
 # Start a terminal
 # https://github.com/alacritty/alacritty
-binding_terminal = <super> KEY_ENTER
-command_terminal = alacritty
+binding_terminal = <super> KEY_ENTER | <ctrl> <alt> KEY_T
+command_terminal = alacritty | gnome-terminal
```
ammen99: I don't think this is a good idea since if the user has both installed, both will be started.

wayfire.ini
```diff
@@ -122,6 +128,9 @@ outputs = kanshi
 # Notifications
 # https://wayland.emersion.fr/mako/
 notifications = mako
+# if using elogind
```
ammen99: +1

wayfire.ini
```diff
@@ -73,11 +76,14 @@ plugins = \
...
 # Close focused window.
-close_top_view = <super> KEY_W | <super> BTN_MIDDLE | <alt> KEY_F4
+close_top_view = <super> KEY_W | <super> BTN_MIDDLE | <alt> KEY_FN_F4 | <alt> KEY_F4
 
 # Workspaces arranged into a grid: 3 × 3.
 vwidth = 3
 vheight = 3
+# Enable Xwayland
+# also install xorg-xhost to run xwaland apps with elevated privilage
+xwayland = true
```
ammen99: +1, you can link to the Tips & Tricks wiki page instead of explaining xorg-xhost, since we already have information for xwayland apps and sudo there.

wayfire.ini
```diff
@@ -73,11 +76,14 @@ plugins = \
 # https://github.com/WayfireWM/wayfire/wiki/Configuration#blur
 
 # Close focused window.
-close_top_view = <super> KEY_W | <super> BTN_MIDDLE | <alt> KEY_F4
+close_top_view = <super> KEY_W | <super> BTN_MIDDLE | <alt> KEY_FN_F4 | <alt> KEY_F4
```
ammen99: Does this depend on the keyboard model or something?

wayfire.ini
```diff
@@ -15,7 +15,9 @@
 # [input]
 # xkb_layout = us,fr
 # xkb_variant = dvorak,bepo
-#
+# cursor_theme = xcursor-breeze
```
ammen99: +1

librewish:
> since a lot of the proposed changes are tighly tied to GNOME

okay removed anything tightly integrated 
its only tracker 
and none of above is tightly integrated now

wayfire.ini
librewish: some keyboards directly use fn keys without needing to use fn + f1 etc

wayfire.ini
```diff
@@ -223,7 +311,7 @@ prev_view = <super> <shift> KEY_TAB
 
 # Simple active window switcher.
 [fast-switcher]
-activate = <alt> KEY_ESC
+activate = <alt> KEY_TAB
```
ammen99: This one stays as KEY_ESC, alt+tab is provided by switcher.

wayfire.ini
```diff
@@ -188,12 +244,44 @@ command_volume_down = amixer set Master 5%-
...
-repeatable_binding_light_up = KEY_BRIGHTNESSUP
-command_light_up = sudo light -A 5
-repeatable_binding_light_down = KEY_BRIGHTNESSDOWN
-command_light_down = sudo light -U 5
+# repeatable_binding_light_up = KEY_BRIGHTNESSUP
+# command_light_up = sudo light -A 5
+# repeatable_binding_light_down = KEY_BRIGHTNESSDOWN
+# command_light_down = sudo light -U 5
+
+# Screen brightness using brightnessctl
```
ammen99: Why do you think we should use `brightnessctl` instead of `light` by default?

wayfire.ini
```diff
@@ -151,15 +197,25 @@ idle = swayidle before-sleep swaylock
...
 
 # Start your launcher
 # https://hg.sr.ht/~scoopta/wofi
 # Note: Add mode=run or mode=drun to ~/.config/wofi/config.
 # You can also specify the mode with --show option.
 binding_launcher = <super> <shift> KEY_ENTER
-command_launcher = wofi
+command_launcher = wofi --show=run
+
+# sgtk-grid
```
ammen99: I think we spoke about this, you could create a section in the wiki for this, I do not think it belongs to the default config.

wayfire.ini
```diff
@@ -151,15 +197,25 @@ idle = swayidle before-sleep swaylock
...
+# https://github.com/nwg-piotr/sgtk-menu
+binding_sgtk-grid = <super> KEY_SPACE
+command_sgtk-grid = sgtk-grid
+
+# clipman - A simple clipboard manager for Wayland
```
ammen99: Same thing here, add it to the wiki instead.

wayfire.ini
```diff
@@ -134,6 +144,42 @@ redshift = redshift -m wayland
 # https://github.com/swaywm/swaylock
 idle = swayidle before-sleep swaylock
 
+# clipman - A simple clipboard manager for Wayland
```
ammen99: See below, this belongs to the wiki.

wayfire.ini
```diff
@@ -134,6 +144,42 @@ redshift = redshift -m wayland
...
+
+# trcker daemon autostart if you need gnome apps to work properly as they should
+# https://wiki.gnome.org/Projects/Tracker
+# tracker = tracker daemon -s
+
+# libinput-gestures - Actions gestures on your touchpad using libinput 
+# https://github.com/bulletmark/libinput-gestures
+# Note:- to use without root
+# groupadd -r input && usermod -a -G input $USER
+libinput-gesture = libinput-gestures-setup start
```
ammen99: for the wiki I guess.

wayfire.ini
```diff
@@ -134,6 +144,42 @@ redshift = redshift -m wayland
...
+
+# ydotool - xdotool replacement for wayland
+# https://github.com/ReimuNotMoe/ydotool
+# Note:- to use without root
+# groupadd -r input && usermod -a -G input $USER
+# groupadd -r plugdev && usermod -a -G plugdev $USER
+ydotool = ydotoold
```
ammen99: Also belongs to the wiki.

PR #425, opened 2020-03-02
Low latency stuff: render delay, presentation time protocol

myfreeweb:
[presentation-time](https://github.com/wayland-project/wayland-protocols/blob/master/stable/presentation-time/presentation-time.xml) can be used by e.g. `weston-presentation-shm` from Weston, which is suggested for [testing wlroots adaptive sync support](https://github.com/swaywm/sway/pull/5063) among other things.

I think `render_views` is the right (and only) place for this? Sway has another call in an optimized fullscreen scanout path, but we don't have that yet, right?

(Seems like we don't have a [repaint delay](https://ppaalanen.blogspot.com/2015/02/weston-repaint-scheduling.html) yet btw.. I'll look into that)

### DIFF
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -20,6 +20,7 @@ extern "C"
 #include <wlr/render/wlr_renderer.h>
 #undef static
 #include <wlr/types/wlr_output_damage.h>
+#include <wlr/types/wlr_presentation_time.h>
 #include <wlr/util/region.h>
 }
 
@@ -892,17 +893,39 @@ class wf::render_manager::impl
 
         for (auto& ds : wf::reverse(repaint.to_render))
         {
+            wlr_surface *sampled_surf = nullptr;
+
             if (ds->view)
             {
                 repaint.fb.geometry.x = ds->pos.x;
                 repaint.fb.geometry.y = ds->pos.y;
                 ds->view->render_transformed(repaint.fb, ds->damage);
+                sampled_surf = ds->view->get_wlr_surface();
+                for (auto& child : ds->view->enumerate_surfaces({0, 0}))
+                {
+                    if (child.surface->get_wlr_surface() != nullptr)
+                    {
+                        wlr_presentation_surface_sampled_on_output(
+                            wf::get_core_impl().protocols.presentation,
+                            child.surface->get_wlr_surface(),
+                            output->handle);
+                    }
+                }
             }
             else
             {
                 repaint.fb.geometry = fb_geometry;
                 ds->surface->simple_render(repaint.fb,
                     ds->pos.x, ds->pos.y, ds->damage);
+                sampled_surf = ds->surface->get_wlr_surface();
+            }
+
+            if (sampled_surf != nullptr)
+            {
+                wlr_presentation_surface_sampled_on_output(
+                    wf::get_core_impl().protocols.presentation,
+                    sampled_surf,
+                    output->handle);
             }
         }
     }
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -368,11 +368,19 @@ class wf::render_manager::impl
             if (total <= 0 || max_render_time <= 0 || this->renderer)
                 total = 0;
 
-            output->handle->frame_pending = true;
-            repaint_timer.set_timeout(total, [=] () {
-                output->handle->frame_pending = false;
+            // We cannot really wait less than 1ms, render right away in that case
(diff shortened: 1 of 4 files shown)


ammen99:
fwiw I have tried to add repaint delay here: https://github.com/WayfireWM/wayfire/tree/delay-experiments

I it has been some time since then so I am not sure whether this thing I did was complete or not ... Might be useful if you start working on it :)

src/output/render-manager.cpp
```diff
@@ -849,18 +850,26 @@ class wf::render_manager::impl
 
         for (auto& ds : wf::reverse(repaint.to_render))
         {
+            wlr_surface *sampled_surf = nullptr;
             if (ds->view)
             {
                 repaint.fb.geometry.x = ds->pos.x;
                 repaint.fb.geometry.y = ds->pos.y;
                 ds->view->render_transformed(repaint.fb, ds->damage);
+                sampled_surf = ds->view->get_wlr_surface();
```
ammen99: In this case, all subsurfaces are rendered as well (`view->enumerate_surfaces()` iirc)

myfreeweb:
> I have tried to add repaint delay here

Nice. I've been thinking about almost this code, but I didn't know about the extra `on_present` thing.

myfreeweb:
hm, currently for me Firefox `widget.wayland_vsync.enabled` likes to not refresh when there's no input now, so I'll need a bit more time to test things

myfreeweb:
oops, I wasn't reading sway code right: sending the frame callback early actually requires doing more fancy stuff with the extra per-surface delay they're doing or something. Removed that change, now everything (well, Firefox vsync) works great

src/output/render-manager.cpp
```diff
@@ -377,6 +377,8 @@ class wf::render_manager::impl
             {
                 output->handle->frame_pending = true;
                 repaint_timer.set_timeout(total, [=] () {
+                    if (output->handle == nullptr)
```
ammen99: Is this necessary (i.e have you experienced a crash because of this)?

When the output is destroyed, its render_manager is also destroyed, leading to automatic disconnect of this timer IIRC.

myfreeweb: Haven't experienced a crash, just saw Sway doing this. And they do `wl_event_source_remove(output->repaint_timer)` in `output_destroy`..

ammen99: I think you can drop this check, output->handle is never supposed to get NULL. If that happens, then a much larger bug is there and Wayfire will crash in other places.

src/output/render-manager.cpp
```diff
@@ -849,18 +897,32 @@ class wf::render_manager::impl
...
         for (auto& ds : wf::reverse(repaint.to_render))
         {
+            wlr_surface *sampled_surf = nullptr;
             if (ds->view)
             {
                 repaint.fb.geometry.x = ds->pos.x;
                 repaint.fb.geometry.y = ds->pos.y;
                 ds->view->render_transformed(repaint.fb, ds->damage);
+                sampled_surf = ds->view->get_wlr_surface();
+                for (auto& child : ds->view->enumerate_surfaces({0, 0}))
```
ammen99: braces {} for the `if` and `else` (in general, I have been allowing skipping braces only for short single-line ifs, and I think I'll start required braces even then)

src/output/render-manager.cpp
```diff
@@ -849,18 +897,32 @@ class wf::render_manager::impl
...
+                            output->handle);
             }
             else
             {
                 repaint.fb.geometry = fb_geometry;
                 ds->surface->simple_render(repaint.fb,
                     ds->pos.x, ds->pos.y, ds->damage);
+                sampled_surf = ds->surface->get_wlr_surface();
             }
+            if (sampled_surf != nullptr)
```
ammen99: same as above, braces {}

ammen99:
This PR breaks the wayland backend, and it is the fault of my early commits :) Can you add http://ix.io/2elN to this PR?

Last time I tested I also noticed some new lags when playing youtube videos in chrome, but I'll have to investigate further whether this is caused by this PR, or maybe it was just my perception which was wrong.

myfreeweb:
> This PR breaks the wayland backend, and it is the fault of my early commits :)

yeah, just noticed that recently, did not even think this PR could be the cause, thought something weird was happening with my graphics stack

myfreeweb:
sure.

ammen99:
Thanks!

PR #426, opened 2020-03-03
plugins: add "maximize" mode to the place plugin

dvn0:
This mode will maximize all newly opened windows.

Not sure if the function can be simplified further. Looking for feedback. :)

Thanks!

### DIFF
--- a/plugins/single_plugins/place.cpp
+++ b/plugins/single_plugins/place.cpp
@@ -36,6 +36,8 @@ class wayfire_place_window : public wf::plugin_interface_t
             std::string mode = placement_mode;
             if (mode == "cascade")
                 cascade(view, workarea);
+            else if (mode == "maximize")
+                maximize(view, workarea);
             else if (mode == "random")
                 random(view, workarea);
             else
@@ -102,6 +104,11 @@ class wayfire_place_window : public wf::plugin_interface_t
         window.x = workarea.x + (workarea.width / 2) - (window.width / 2);
         window.y = workarea.y + (workarea.height / 2) - (window.height / 2);
         view->move(window.x, window.y);
+    }
+
+    void maximize(wayfire_view &view, wf::geometry_t workarea)
+    {
+        view->tile_request(wf::TILED_EDGES_ALL);
     }
 
     void fini() override


plugins/single_plugins/place.cpp
```diff
@@ -36,6 +36,8 @@ class wayfire_place_window : public wf::plugin_interface_t
             std::string mode = placement_mode;
             if (mode == "cascade")
                 cascade(view, workarea);
+            else if (mode == "max")
```
soreau: Since 'max' doesn't really make sense for a placement type, maybe this should be 'maximize' to make it clear.

plugins/single_plugins/place.cpp
```diff
@@ -104,6 +106,15 @@ class wayfire_place_window : public wf::plugin_interface_t
         view->move(window.x, window.y);
     }
 
+    void max(wayfire_view &view, wf::geometry_t workarea)
+    {
+        wf::geometry_t window = view->get_wm_geometry();
+        window.x = workarea.x + (workarea.width);
+        window.y = workarea.y + (workarea.height);
```
soreau: Can you explain why you're adding workarea.width/height to x/y?

erikvanhamme:
Window rules will support a rule type without condition, but care should be taken if maximizing every view. The panel and doc are also considered views which we don't want maximized. I don't know if there is an option to detect that the panel and docs are 'builtin' views and should be excluded from the window-rules application. This is an architectural decision.

As it stands now, the rule would have to include a condition element that specifically excludes the views that should not be maximized. It can be done, but it might not be an elegant rule if there are more and more exceptions.

plugins/single_plugins/place.cpp
dvn0: I agree. I think "maximum" also makes sense, no? I changed it to this in the following commit.

plugins/single_plugins/place.cpp
dvn0: It was misguided... was not thinking it through. I've removed this in the latest commit.

dvn0:
NOTE: If this PR gets approved, then I would like to squash commits before it's merged.

ammen99:
ping, just needs update to `maximize` instead of `maximum`

plugins/single_plugins/place.cpp
dvn0: changed to maximize

PR #427, opened 2020-03-07
Blur fixup

soreau:


ammen99:
Is this based on master or not? (Wonder why the CI is not running)

soreau:
Yes it's based on master.

ammen99:
Oh I see why it doesn't run, the changes are only to the XML files and these have been explicitly marked as ignored ...

Anyway, thanks :)

Issue #432, opened 2020-03-11
Compilation error in src/view/surface.cpp


ericonr:
I tested this with both clang and gcc, both generated errors. This is the error output from gcc:

```
[72/160] Compiling C++ object 'src/25a6634@@wayfire@exe/view_surface.cpp.o'.
FAILED: src/25a6634@@wayfire@exe/view_surface.cpp.o
c++ -Isrc/25a6634@@wayfire@exe -Isrc -I../src -I. -I../ -I../src/api -I/usr/lib/libffi-3.2.1/include -I/usr/include/libdrm -I/usr/include/pixman-1 -I/usr/include/libevdev-1.0/ -I/usr/include/libxml2 -I/usr/include/libpng16 -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -std=c++17 -g -DWLR_USE_UNSTABLE -Wno-unused-parameter -DHAS_ADDR2LINE=1 -MD -MQ 'src/25a6634@@wayfire@exe/view_surface.cpp.o' -MF 'src/25a6634@@wayfire@exe/view_surface.cpp.o.d' -o 'src/25a6634@@wayfire@exe/view_surface.cpp.o' -c ../src/view/surface.cpp
../src/view/surface.cpp: In member function ‘virtual wlr_buffer* wf::wlr_surface_base_t::get_buffer()’:
../src/view/surface.cpp:325:25: error: cannot convert ‘wlr_client_buffer*’ to ‘wlr_buffer*’ in return
  325 |         return surface->buffer;
      |                ~~~~~~~~~^~~~~~
      |                         |
      |                         wlr_client_buffer*
```

Misc:

* GCC: c++ (Arch Linux 9.2.1+20200130-2) 9.2.1 20200130
* git commit: `1b87bff15`

ericonr:
So, this seems to be due to a wlroots update. They changed the structure of the wlr_surface struct. I've found a fix, I'm going to try it out now.

ammen99:
Fixed in #433

PR #433, opened 2020-03-11
view: fix wlr_surface usage due to upstream change

ericonr:
The wlroots changed the organization of the wlr_surface struct, which
led to breakage in the surface.cpp code. This commit fixes the
issue in wlr_surface_base_t::get_buffer(), which was the only place
affected (at least in regards to compile-time errors).

Fixes issue #432 

I have tested it on my machine, and wf-background and wf-panel both launched and worked perfectly.

EDIT: btw, this fix is only necessary in wlroots git, for now. I'd guess wlroots will make it necessary either in some 0.10.x revision or the 0.11.0 version. I don't know how that influences what you merge to master. And the CI builds seem to have failed because their version of wlroots isn't as recent.

### DIFF
--- a/src/view/surface.cpp
+++ b/src/view/surface.cpp
@@ -322,7 +322,7 @@ void wf::wlr_surface_base_t::unmap()
 wlr_buffer* wf::wlr_surface_base_t::get_buffer()
 {
     if (surface && wlr_surface_has_buffer(surface))
-        return surface->buffer;
+        return &surface->buffer->base;
 
     return nullptr;
 }


ericonr:
The fix is mentioned here: https://github.com/swaywm/wlroots/pull/2043

Should be pretty safe.

ammen99:
You should also update the wlroots subproject to the appropriate version so that the CI can successfully build.

Anyway I am in the process of preparing the next release (which should be aligned with wlroots 0.10.1) so this won't be merged immediately.

ericonr:
>You should also update the wlroots subproject to the appropriate version so that the CI can successfully build.

How would I do that? `meson.build` already lists wlroots as `"<0.11.0"`. That should include the latest version.

ammen99:
Sorry, I was not clear. I meant the git submodule at `subprojects/wlroots`

soreau:
You would run `git pull` in the wlroots subproject directory, then run `git commit -a` in the wayfire root directory.

Small request, but can you possibly not the breaking wlroots commit id in the commit message somewhere? The first 8 characters should work. Thanks for beating us to fixing this!

ericonr:
Ok, done with both! Thanks for the help!

soreau:
LGTM, we should merge this once 0.4 is released.

soreau:
Well I hadn't noticed you put both the wlroots update and the fix in the same commit, you might want to separate these.

ericonr:
I had thought about separating them, but ended up deciding to do it together because one commit would depend on the other to work. I can split them no problem, though.

soreau:
The main idea is to not have a single line change buried together with other changes, and with a commit message that doesn't explain what's happening exactly. Since these will go together under a single merge commit, it will be obvious what's happening when looking at the history or bisecting, and it's not likely anyone will pull one commit without the other.

ericonr:
That makes sense! Thanks a lot for the patience!

PR #434, opened 2020-03-11
CI: fix the names used for the CI jobs

ericonr:
The previous naming scheme didn't accurately describe the jobs being run.

Would this impact any of the CI currently running in the project? It's just that the naming scheme bothered me. If you don't think it's necessary to fix, no problem, I can close the PR.


ammen99:
well I see no problem with changing it.

ammen99:
Thanks!

PR #438, opened 2020-03-18
Optimizations for cube and expo

soreau:


ammen99:
Tested, and it works. I had some doubts because I hadn't noticed that wlroots does set `needs_frame` on the output whenever we call `schedule_frame` ... But it should work quite well now.

Actually, this means that we probably don't have to use `redraw_always` in most places where it is used ...

ammen99:
Thanks btw :)

soreau:
> Tested, and it works. I had some doubts because I hadn't noticed that wlroots does set `needs_frame` on the output whenever we call `schedule_frame` ... But it should work quite well now.

That's good. This actually makes showrepaint behave correctly in the case where moving the mouse would cause white screen sometimes, because the render function was being called but not actually rendering.

> 
> Actually, this means that we probably don't have to use `redraw_always` in most places where it is used ...

Yes indeed. I also found some things in zoom that could be improved in this regard.

PR #443, opened 2020-03-20
Trigger config reload when config file is created

soreau:
This requires adding a watch on the directory containing the config file.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -32,13 +32,15 @@ extern "C"
 wf_runtime_config runtime_config;
 
 #define INOT_BUF_SIZE (1024 * sizeof(inotify_event))
-char buf[INOT_BUF_SIZE];
+static char buf[INOT_BUF_SIZE];
 
-static std::string config_file;
+static std::string config_dir, config_file;
+
 static void reload_config(int fd)
 {
     wf::config::load_configuration_options_from_file(
         wf::get_core().config, config_file);
+    inotify_add_watch(fd, config_dir.c_str(), IN_CREATE);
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -31,23 +31,53 @@ extern "C"
...
 }
 
 static int handle_config_updated(int fd, uint32_t mask, void *data)
 {
-    LOGD("Reloading configuration file");
+    int length, i = 0;
 
-    /* read, but don't use */
-    read(fd, buf, INOT_BUF_SIZE);
+    length = read(fd, buf, INOT_BUF_SIZE);
```
ammen99: This is unfortunately still not enough. `read()` can be interrupted after reading just a single byte, in which case we'll try to read an invalid event below, and potentially corrupt all subsequent events since we'll read out-of-sync.

I researched a bit to confirm my suspicion, and it seems that it is indeed the case. Here is 
 "the solution" to reading a fixed number of bytes: https://unix.stackexchange.com/a/16487

I guess we could adapt it so that we read until we get a number divisible by `sizeof(inotify_event)`, or maybe even just read events one by one. I'll leave the choice to you :P

soreau: It would be nice if there was a way to trigger this case for testing and debugging. For now I pushed something that I think could work in this case, and of course works for the normal case.

ammen99:
Nice, thanks for noticing this.

We can merge once the CI is finished.

src/main.cpp
soreau: As discussed on IRC, this extra code is only to avoid reloading the config when new files are created in ~/.config/ and it's not really a big deal if we do.

PR #451, opened 2020-03-28
update README.md

noname-3:


noname-3:
but in the dependencies on the wiki page "Home", the wf-shell is not marked as an optional dependency. and does wayfire start without wf-shell?

ammen99:
The wiki should then also be updated, I missed it. The thing is this: wf-shell is recommended to get a functional desktop, however Wayfire can function without wf-shell (won't be very usable, I agree)

ammen99:
Looks good now, thanks!

PR #458, opened 2020-04-03
blur: Implement some optimizations

soreau:
- Don't blur if opaque region occludes the bounding box of the view
- Scissor out the opaque region when we are blurring
- Use damage region for scissor to avoid unneeded blurring


szbergeron:
Seems to pass the "unrealistic expectations from an iGPU" test :)

2 4k@60 + 1 1440@144, smooth as butter with a couple blurred windows open and a bunch of opaque ones. No noticeable perf diff from blur disabled.

Settings: kawase with degrade=4, iter=4, offset=1.0

Issue #353, opened 2019-12-24
Allow greater flexibility for workspace layers.


ammen99:
Currently we have just a fixed set of layers, however these may not be enough. We can provide the current layers by default, with a given Z index, and then allow plugins to position one or more layers between those fixed Z order layers.

Another thing to consider is that we handle different layers differently. It would be desirable to figure out a list of these different uses and be able to "tag" each layer with its uses, so that custom layers can fit in with the rest of the plugins.

PR #461, opened 2020-04-05
view: add always-on-top state

myfreeweb:
This adds a state that keeps the view on `LAYER_TOP` and a hotkey to toggle that.

Eventually this is also going to be used for a protocol that would allow apps to control this (e.g. Firefox picture-in-picture).. fun fact, gtk-shell seems to actually have a request for this, but no one wants to use gtk-shell.

Not tested:
- [ ] multiple outputs
- [ ] interaction with minimize (???)
- [ ] interaction with fullscreen

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -14,6 +14,11 @@
 			<_long>Closes the currently focused window with the specified key.</_long>
 			<default>&lt;super&gt; KEY_Q | &lt;alt&gt; KEY_F4</default>
 		</option>
+		<option name="toggle_always_on_top" type="activator">
+			<_short>Toggle always-on-top</_short>
+			<_long>Toggles the always-on-top state of the currently focused window with the specified key.</_long>
+			<default>&lt;super&gt; &lt;shift&gt; KEY_F</default>
+		</option>
 		<!-- Horizontal/Vertical workspaces -->
 		<option name="vwidth" type="int">
 			<_short>Horizontal virtual size</_short>
--- a/src/api/wayfire/view.hpp
+++ b/src/api/wayfire/view.hpp
@@ -218,14 +218,19 @@ class view_interface_t : public surface_interface_t, public wf::object_base_t
      * set_activated() or focus_request() */
     bool activated = false;
     /** Whether the view is in minimized state, usually you want to use either
-     * set_minizied() or minimize_request() */
+     * set_minimized() or minimize_request() */
     bool minimized = false;
+    /** Whether the view is in always-on-top state, usually you want to use either
+     * set_on_top() or on_top_request() */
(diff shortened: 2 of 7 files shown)


ammen99:
A few things to consider:

1. `LAYER_TOP` is used with layer-shell surfaces as well. We probably want a new layer.
2. Is always-on-top also sticky (I would think this is not the case but I haven't used these features in other compositors so I don't really understand them)? Otherwise, you need to add the new layer in the `WM_LAYERS` or whatever it was called.
3. Do we really need a function `set_on_top` for the view? This does not change any state inside the view. I think the plugin should do that functionality manually.
4. Multiple outputs? You probably want to adjust `move_view_to_output` in core.
5. Interaction with minimize - now this is going to be tricky, I agree.
6. Interaction with fullscreen - I could imagine several possible behaviors, but I think fullscreen is above always-on-top?

myfreeweb:
> LAYER_TOP is used with layer-shell surfaces as well

Yeah. I couldn't think of any problems with sharing the layer though?

> Is always-on-top also sticky

As in, appears on all workspaces like layer-shell panels? hmmmm that's a hard question. I like that they are. Picture-in-picture videos are intended to be always visible no matter what you're working on, even with switching workspaces. I guess ideally the stickiness could be controlled separately? but I'm too lazy to implement *that*

> Multiple outputs? You probably want to adjust move_view_to_output in core

I did, just haven't tested that it actually works.

> Do we really need a function set_on_top for the view?

I mean.. currently there's only a hotkey, but someday the same functionality would be available through a protocol. I don't think leaking details like the particular layer we use into hotkey/protocol implementations is good?

> Interaction with fullscreen - I could imagine several possible behaviors, but I think fullscreen is above always-on-top?

Yes. I mean more like.. I haven't tested that going fullscreen and back to non-fullscreen would restore to the top layer, etc.

ammen99:
> Yeah. I couldn't think of any problems with sharing the layer though?

One potential problem is that you can drag the window on top of the panel. That alone isn't bad.

However, by sharing the layer, you exclude the view from fast-switcher, switcher, etc.

When an output is unplugged, core goes through the WM_LAYERS and moves them to the other output. Other views (including always-on-top in this case) are closed.

All in all, I think it is easier to use another layer. However, once #353 is solved (I hope to do that one soon), it should be possible to add new (sub)layers, which should fix this.

Also, parallel to #353 I'll add a dedicated sticky option and stop treating layer-shell surfaces specially (they'll be simply sticky).

So maybe it is best to wait until then?

myfreeweb:
Sure, no rush to get this merged, sublayers sound cool

ammen99:
This was implemented in #503

PR #463, opened 2020-04-05
Add input-method/text-input support

myfreeweb:
This is a port of https://github.com/swaywm/sway/pull/4740

- seems to work the same as sway with squeekboard and the wlroots input-method example
- looks like no crashes anymore after I added all the `disconnect`s on listeners..
- [ ] cleanup blank constructors
- style: `input->im_relay->setup_listeners()` called after the protocols are added is meh..
- too many `std::unique_ptr`?

### DIFF
--- a/src/api/wayfire/signal-definitions.hpp
+++ b/src/api/wayfire/signal-definitions.hpp
@@ -112,6 +112,17 @@ struct surface_map_state_changed_signal : public wf::signal_data_t
  * argument: unused
  */
 
+/**
+ * name: keyboard-focus-changed
+ * on: core
+ * when: Keyboard focus is changed (may change to nullptr).
+ */
+struct keyboard_focus_changed_signal : public wf::signal_data_t
+{
+    wayfire_view view;
+    wlr_surface *surface;
+};
+
 /* ----------------------------------------------------------------------------/
  * Output signals
  * -------------------------------------------------------------------------- */
--- a/src/core/seat/input-method-relay.cpp
+++ b/src/core/seat/input-method-relay.cpp
@@ -1,0 +1,309 @@
+#include <wayfire/util/log.hpp>
+#include "input-method-relay.hpp"
+#include "../core-impl.hpp"
+#include "../../output/output-impl.hpp"
+
+#include <algorithm>
+
+wf::input_method_relay::input_method_relay()
+{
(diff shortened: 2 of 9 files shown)


src/core/seat/input-method-relay.cpp
```diff
@@ -0,0 +1,293 @@
...
+    {
+        auto wlr_text_input = static_cast<wlr_text_input_v3*> (data);
+        assert(input == wlr_text_input);
+
+        if (input->current_enabled)
+        {
+            relay->disable_text_input(input);
+        }
+        set_pending_focused_surface(nullptr);
+        on_text_input_enable.disconnect();
```
ammen99, line 252: Are these the disconnects which you needed to not get a crash?

Seems to me that the object will be immediately deleted so they should be auto-disconnected.

myfreeweb:
Did the signal thing and the namespace, resolved style via uncrustify. I don't remember anything about disconnects anymore and what any of this really means.. :D

did encounter a crash related to uninitialized pointers though, added `= nullptr` init.

src/api/wayfire/signal-definitions.hpp
```diff
@@ -320,6 +320,16 @@ using view_unmapped_signal = _view_signal;
  */
 using view_set_output_signal = _output_signal;
 
+/**
+ * name: keyboard-focus-changed
+ * on: view
```
ammen99, line 325: this signal doesn't seem to be emitted on the view, but on core.

src/core/seat/input-method-relay.cpp
```diff
@@ -0,0 +1,309 @@
...
+        LOGI("Disabling text input, but input method is gone");
+
+        return;
+    }
+
+    wlr_input_method_v2_send_deactivate(input_method);
+    send_im_state(input);
+}
+
+wf::text_input*wf::input_method_relay::find_focusable_text_input()
```
ammen99, line 135: uncrustify doesn't have bugs, does it :P

ammen99: (you can't do much about it, just saying)

src/api/wayfire/signal-definitions.hpp
myfreeweb: oh, I completely misunderstood what "on" means

ammen99:
Thanks!

PR #469, opened 2020-04-10
idle: Overhaul screensaver

soreau:
This makes screensaver per-output instead of single-instance. It fixes
several problems which aren't readily noticeable. For one, it makes it
so each instance is driven at the respective output refresh rate instead
of a single output driving all instances. This makes for a smoother
experience. Secondly, if an output was unplugged and another remained
active, the cube would get stuck grabbed until button press and then
would only restore back to untransformed when grabbing cube again. The
last bug this fixes is when an output was reconnected, the screensaver
was not running on it. This also hides mouse cursor when screensaver is
active.

### DIFF
--- a/plugins/single_plugins/idle.cpp
+++ b/plugins/single_plugins/idle.cpp
@@ -16,13 +16,13 @@ extern "C"
 #include <wayfire/util/duration.hpp>
 #include <wayfire/util/log.hpp>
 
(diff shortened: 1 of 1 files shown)


ammen99:
Seems to work well enough, and I don't think there will be compilation errors, so merging even with failing builds (the CI doesn't seem to be having a good day).

PR #472, opened 2020-04-11
matcher: fix build without xwayland

myfreeweb:


ammen99:
This matcher is going to be removed as soon as we finish #454 , although LGTM.

PR #474, opened 2020-04-13
xwayland: Clamp geometry to workspace area instead of output

soreau:
This was causing xwayland views to jump workspaces when they configured
themselves. Instead of clamping to the output geometry, we clamp to
the desktop area, which is made up of all the workspaces.

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -115,15 +115,20 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
          * coordinates */
         if (get_output())
         {
+            auto current_workspace = get_output()->workspace->get_current_workspace();
+            auto wsize = get_output()->workspace->get_workspace_grid_size();
             auto og = get_output()->get_layout_geometry();
             configure_geometry.x -= og.x;
             configure_geometry.y -= og.y;
 
-            /* It is possible the client requests to position itself on another
-             * output. However, we want to make sure that the view stays on its
-             * output, since this is how Wayland works */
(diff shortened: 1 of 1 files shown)


src/view/xwayland.cpp
```diff
@@ -90,6 +91,8 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
         on_configure.connect(&xw->events.request_configure);
         on_set_title.connect(&xw->events.set_title);
         on_set_app_id.connect(&xw->events.set_class);
+
+        origin_workspace = get_output()->workspace->get_current_workspace();
```
soreau, line 95: Should I check if get_output() isn't null or use core get_active_output?

src/view/xwayland.cpp
```diff
@@ -115,15 +115,17 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
...
             auto og = get_output()->get_layout_geometry();
             configure_geometry.x -= og.x;
             configure_geometry.y -= og.y;
-
-            /* It is possible the client requests to position itself on another
-             * output. However, we want to make sure that the view stays on its
-             * output, since this is how Wayland works */
-            configure_geometry = wf::clamp(configure_geometry,
-                get_output()->get_relative_geometry());
+            wlr_box wsg{
```
ammen99, line 123: Can we have a comment explaining why we do this (we do it so that the client doesn't go off the screen)?

ammen99:
Thanks :)

Issue #475, opened 2020-04-14
Closing some windows (pavucontrol, heaptrack) causes Wayfire to unload all plugins


myfreeweb:
I have no idea how the hell this is possible, but it is kinda funny. This started happening today.
Clicking the close button on the windows of pavucontrol (gtk) and heaptrack (qt) started crashing. The crash was caused by `switcher` plugin unload (fix coming). But it's still unloading plugins when reacting to `detach-view`. ??? o_0

Nothing interesting in the log. Doesn't seem like it could be caused by my GSettings plugin setting core/plugins (it logs when it reloads config), I have actually used it to reload everything after this happened (and it did log of course).

```
[…]
DD 14-04-20 20:12:56.997 - [types/wlr_surface.c:615] New wlr_surface 0x6180000f5880 (res 0x60c000340fc0)
DD 14-04-20 20:12:56.997 - [types/wlr_surface.c:615] New wlr_surface 0x6180000f5c80 (res 0x60c000341200)
DD 14-04-20 20:12:57.649 - [types/wlr_surface.c:615] New wlr_surface 0x6180000f6080 (res 0x60c000342340)
DD 14-04-20 20:12:57.649 - [types/xdg_shell/wlr_xdg_surface.c:459] new xdg_surface 0x6120001c6cc0 (res 0x60c000342400)
II 14-04-20 20:12:57.649 - [src/view/xdg-shell.cpp:209] new xdg_shell_stable surface: Volume Control app-id: <unknown>
DD 14-04-20 20:12:57.650 - [types/wlr_server_decoration.c:107] new server_decoration 0x60b000147d30 (res 0x60c0003430c0)
DD 14-04-20 20:12:58.166 - [src/core/seat/pointer.cpp:108] change cursor focus 0x61c0000ca1c0 -> 0x61c0000d01c0
DD 14-04-20 20:12:58.180 - [src/core/seat/pointer.cpp:108] change cursor focus 0x61c0000d01c0 -> 0x61a0000b89c0
EE 14-04-20 20:12:58.182 - [backend/drm/atomic.c:38] Atomic test failed: Invalid argument
DD 14-04-20 20:12:58.316 - [src/core/seat/pointer.cpp:108] change cursor focus 0x61a0000b89c0 -> 0x61c0000d01c0
DD 14-04-20 20:12:58.364 - [src/core/seat/pointer.cpp:108] change cursor focus 0x61c0000d01c0 -> 0x61a0000b89c0
[...]

PR #476, opened 2020-04-14
switcher: implement fini()

myfreeweb:
Found via some bizzare unloading thing (issue #475) + AddressSanitizer + RTLD_NODELETE (to see which plugin is to blame in the ASan trace — speaking of which, maybe a good idea to NODELETE.. always? in debug builds?).

### DIFF
--- a/plugins/single_plugins/switcher.cpp
+++ b/plugins/single_plugins/switcher.cpp
@@ -783,6 +783,8 @@ class WayfireSwitcher : public wf::plugin_interface_t
 
     void fini() override
     {
+        if (output->is_plugin_active(grab_interface->name))
+            deinit_switcher();
         output->rem_binding(&next_view_binding);
         output->rem_binding(&prev_view_binding);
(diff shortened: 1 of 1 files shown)


plugins/single_plugins/switcher.cpp
```diff
@@ -781,7 +781,13 @@ class WayfireSwitcher : public wf::plugin_interface_t
         views.push_back(std::move(sv));
     }
 
-    // TODO:!!! fini
+    void fini() override
+    {
```
ammen99, line 785:
```suggestion
    {
        if (output->is_plugin_active(grab_interface->name))
            deinit_switcher();
```

ammen99: This would be if the plugin is disabled while it is active. I can't imagine how this can happen during regular usage, the user will probably have to try to intentionally break it, but still ..

ammen99:
Thanks, I think we have postponed this long enough ;)

PR #477, opened 2020-04-15
decoration: plug cairo surface memory leak (fixes #473)

### LINKED ISSUE #473: memory leak
memory leak is exclusive to wayfire - sway is not affected.
Wayfire 0.4.0

1. Open a terminal
2. killall wf-panel
3. wf-panel
4. repeat 2 & 3 a couple of times

The memory usage of wayfire will increase each time by 5-10Mb. (have reached 600Mb)
Note that this affects only created surfaces that use layer shell somehow.

myfreeweb:


ammen99:
Nice find, thanks :)

PR #482, opened 2020-04-23
Add water plugin

soreau:


ammen99:
ftr, this is going to be part of plugins-extra.

PR #490, opened 2020-04-28
Adjust command line parameters

### LINKED ISSUE #61: Add a --help option to wayfire

damianatorrpm:
Adds -v --version.
Adds -h --help.
Adjust --damage-debug /  from -d to -D
Adjust --verbose -v to --debug -d
Also adds a new line for 'unrecognized parameter' / typo

Fixes #61

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -286,7 +286,7 @@ int main(int argc, char *argv[])
     signal(SIGABRT, signal_handler);
 #endif
 
-    LOGI("Starting wayfire", WAYFIRE_VERSION);
+    LOGI("Starting wayfire version ", WAYFIRE_VERSION);
     /* First create display and initialize safe-list's event loop, so that
      * wf objects (which depend on safe-list) can work */
     auto display = wl_display_create();
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -56,6 +56,24 @@ static int handle_config_updated(int fd, uint32_t mask, void *data)
     return 0;
 }
 
+static void print_version()
+{
+     std::cout << WAYFIRE_VERSION << std::endl;
+     exit(0);
+}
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -219,31 +237,39 @@ int main(int argc, char *argv[])
...
                 break;
-            case 'v':
+            case 'h':
+                print_help();
+                break;
+            case 'd':
                 log_level = wf::log::LOG_LEVEL_DEBUG;
                 break;
+            case 'v':
+                print_version();
```
ammen99, line 269: Would be nice to also print the version in the regular log, line 289.

ammen99:
Just a note, this is a breaking change, `-v` now prints version and `-d` prints debug information.

src/main.cpp
```diff
@@ -260,7 +286,7 @@ int main(int argc, char *argv[])
     signal(SIGABRT, signal_handler);
 #endif
 
-    LOGI("Starting wayfire");
+    LOGI("Starting wayfire", WAYFIRE_VERSION);
```
soreau, line 289: You should add a space after 'wayfire', at least, like you did [here](https://github.com/WayfireWM/wayfire/pull/490/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR66). I would suggest something like `LOGI("Starting wayfire version ", WAYFIRE_VERSION);`

damianatorrpm:
I think the test failed due to some unrelated network issue,  do I need to restart it?

soreau:
> I think the test failed due to some unrelated network issue, do I need to restart it?

I wouldn't worry too much about it. Unfortunately, this is common.

ammen99:
I tested this manually, since the builds have gone wrong ... Thanks!

PR #491, opened 2020-04-28
Erase views immediately

soreau:
We no longer need to wait for idle to erase views from the list.
This fixes a crash when a plugin with a color_rect_view_t is unloaded
when calling close() in fini().


damianatorrpm:
It seems to me that this is also in the ownership-fixes branch/patch

ammen99:
This patch is indeed wished for but it would introduce quite a lot of other problems. I incorporated this into ownership-fixes branch which appears to work so I will close this PR and merge the other branch.

Issue #493, opened 2020-04-29
Search path for plugins


alyssais:
Currently, all plugins have to be installed into Wayfire’s prefix, or be referenced with absolute paths in the configuration file. There are a few scenarios in which this is not ideal:

- In an FHS-based distribution, Wayfire might have been installed by the distribution into /usr, but plugins built and installed by the user should go into /usr/local.
- In a purely functional distribution, like Nixpkgs or Guix, a package for a Wayfire plugin will not be allowed to install files into Wayfire’s unique prefix, e.g. /nix/store/[hash]-wayfire-0.4.0. This case has an even worse user experience with the current behaviour, because using absolute paths in the configuration file will require the configuration file to be edited or regenerated every time the hash of the plugin, or any of its dependencies changes, because it will be at a new prefix.

So, there should be a search path in the environment that Wayfire should use to discover plugins. I think there should probably be `WAYFIRE_PLUGIN_PATH`, which would point to the directories where plugin libraries are installed, and `WAYFIRE_PLUGIN_XML_PATH`, which contains the directories containing the XML plugin metadata.

wucke13:
I just finished writing a PR for a nixpkgs wayfire package, and this was the only bummer for me.

alyssais:
That’s exactly what I’ve been working on!

ammen99:
Fixed in #497

PR #497, opened 2020-05-02
Support plugin installation outside of Wayfire’s prefix

alyssais:
Part 2 of #493. Implemented as proposed there. Also changes autostart to start wf-panel and wf-background from PATH rather than Wayfire’s prefix.

### DIFF
--- a/src/output/plugin-loader.cpp
+++ b/src/output/plugin-loader.cpp
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include <set>
 #include <memory>
+#include <filesystem>
 #include <dlfcn.h>
 
 #include "plugin-loader.hpp"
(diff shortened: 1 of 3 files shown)


src/output/plugin-loader.cpp
```diff
@@ -147,16 +148,38 @@ void plugin_manager::reload_dynamic_plugins()
...
     while(stream >> plugin_name)
     {
         if (plugin_name.size())
         {
             if (plugin_name.at(0) == '/')
+            {
                 next_plugins.push_back(plugin_name);
+                continue;
+            }
             else
```
ammen99, line 171: I imagine you don't want to have this `else` block anymore ...

ammen99: ping ?

alyssais: Sorry for being slow.

> I imagine you don't want to have this `else` block anymore ...

Probably not here. We should either drop it, or add the library name as a fallback after the loop through plugin_prefixes. The latter would have the effect of following back to the standard library search paths. What do you think?

ammen99: I think simply removing is fine. Searching the `/usr/lib` or whatever doesn't seem a good idea, and the code doesn't really do it - if you look closely, it searches in the default plugin path which is checked below anyway.

ammen99:
Thanks a lot!

PR #503, opened 2020-05-07
Sublayers

### LINKED ISSUE #353: Allow greater flexibility for workspace layers.

### LINKED ISSUE #480: Glitched, cloned Firefox window
This bug may occur at very specific configuration, so bear with me.

My setup:
 - triple-head output
 - - all outputs are side-by-side, arranged
 - - second (center) output is rotated by 270°
```
[DP-1]
layout = 0,0
mode = 3840x2160@60Hz
scale = 1.000000
transform = normal

[DP-2]
layout = -1200,256
mode = 1920x1200@30Hz
scale = 1.000000
transform = 270

[HDMI-A-1]
layout = -3120,950
mode = 1920x1080@20Hz
scale = 1.000000
transform = normal
```
 - Firefox (wayland native) with more than one window open
 - - all Firefox windows tiled with simple-tiling -plugin
 - - the last active Firefox window at DP-1
 - Alacritty (wayland native) on DP-2
 - - maximized, but not tiled
 - Wayfire plugins enabled:
 - - alpha
 - - animate
 - - autostart
 - - command
 - - expo
 - - fast-switcher
 - - move
 - - place
 - - resize
 - - switcher
 - - zoom
 - - wobbly
 - - decoration
 - - simple-tile
 - - fisheye
 - - idle 
 - - cube

== Steps to reproduce ==
 - click on a link in the terminal (Alacritty) window on DP-2
 - - the link _needs_ to point to a site (http. https and ftp not tested) which will ask user credentials in a gtk pop-up window. So the usual html login page that's made with html isn't the kind op pop-up window. An ftp site which requires user credentials might also work. Firefox should pop-up similar gtk window there too.
 - insert your credentials
 - log in
 - a clone of the Firefox window that was last active becomes visible on DP-2

The "clone" window is somewhat glitched, for example if there is a video playing on the page in the link it will stutter between few frames on the clone window, but will work fine on the original window.
Also usually when trying to move the clone window it disappears, but can be found via using expose plugin on DP-2.

I'd assume this can happen with some other pop-ups too. So the title may to really reflect the problem here. Those who have the permission can then change the title as they see fit. :)

**EDIT**: Made a [recording of the glitch](https://youtu.be/mMQpl85iF0U), but there it behaves somewhat differently.

### LINKED ISSUE #481: simple-tiling squeezing/compressing window contents (and decorations too)
It's rather easy to have window content to compress (visually) when moving tiled windows with simple-tiling plugin.

Here's a [video showing it happen](https://www.youtube.com/watch?v=WGb9KFc284s). It does not require much to reproduce. I've reproduced this with all kinds of windows. Xwayland and wayland native too.
The video drags (retiles) two terminals.
The one with black background is st, Xwayland.
The one with semi-transparent background is alacritty, wayland native.

ammen99:
Supersedes #461 

Fixes #353 
Fixes #480 
Fixes #481

@myfreeweb Would you like to test this?


damianatorrpm:
@soreau @ammen99 
Love the feature to toggle always on top.

One regression though, with the sublayer patch enabled when changing the output of a view
(you need multiple monitors to test this) it will eventually crash wayfire.

> Without the sublayer patch using `view->set_output(output);` will sent it to the output requested.
> One still needs to click on the wf-panel window-list on the window a few times, till it's restored.

> With the sublayer patch `view->set_output(output);` will sent it to the output requested. 
(the window-list suggest so)
> Clicking on the window in the window-list reproducible crashes wayfire.

With and without the patch the `set_output `changes the view not being visible (even it was on the previous output). The crash with the patch mentions something with layer bring_to_front in the stacktrace.


PS: I currently switched from nested wayfire to running wayfire as main
and some things are not yet set-up keyrings etc. and I still run into some stuff.

ammen99:
@damianatorrpm Wayfire shouldn't crash, thanks for the report, I'll see how to fix it.

Anyway, just setting the view's output does not automatically add it to a layer - you are responsible for that. You can also use things like core::move_view_to_output() which will do both things for you.

damianatorrpm:
core::move_view_to_output() works without problem with sublayers patch

ammen99:
> One regression though, with the sublayer patch enabled when changing the output of a view
> (you need multiple monitors to test this) it will eventually crash wayfire.
> 
> >

Should be fixed now.

ammen99:
Thanks to all who tested this.

PR #506, opened 2020-05-10
Use transformed geometry when computing visibility if view is transformed

soreau:
This removes the wm_only flag from get_views_on_workspace which was an artifact
of when wayfire used libweston and a shadow discrepancy called for this hack.

### DIFF
--- a/plugins/single_plugins/fast-switcher.cpp
+++ b/plugins/single_plugins/fast-switcher.cpp
@@ -60,7 +60,7 @@ class wayfire_fast_switcher : public wf::plugin_interface_t
     {
         current_view_index = 0;
         views = output->workspace->get_views_on_workspace(
-            output->workspace->get_current_workspace(), wf::WM_LAYERS, true);
+            output->workspace->get_current_workspace(), wf::WM_LAYERS);
     }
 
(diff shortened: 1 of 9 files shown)


ammen99:
As discussed on IRC, there are also functions in sublayers which were following these changes ...

ammen99:
> As discussed on IRC, there are also new functions in sublayers which had this flag, so you'll probably want to make sure the flag is removed there as well.

I also suspect that the many conflicts come from the removal of FULLSCREEN_LAYER? idk

ammen99:
Haven't found out the culprit yet, but this seems to break switcher.

soreau:
> I haven't searched for the problem, but this seems to break switcher.

The problem seems to be [here](https://github.com/WayfireWM/wayfire/blob/master/src/view/view.cpp#L786). This transforms a box with sane coordinates to (-2147483648,-2147483648 0x0) when switcher fails to switch.

soreau:
[This](http://ix.io/2lPr) fixes switcher.

ammen99:
correct fix for switcher: http://ix.io/2s4m

ammen99:
Thanks @soreau

PR #528, opened 2020-05-24
Updated window-rules plugin using the rule engine in wf-utils.

erikvanhamme:
Adds the view action interface.

Not sure about the curly braces style for the single line lambdas in window-rules.cpp. I treated them as methods rather then single line ifs.


ammen99:
Hmm, I recently had an idea which IMO would be cool - make it possible for window-rules to be extended by other plugins. I have imagined something like attaching custom data to an output to store a list of callbacks, where other plugins could append their own handlers for rules, and window-rules would then be responsible for calling those.

So for example, instead of emitting the snap-signal from window-rules, you'd have grid add a handler for that action to the shared state, then window-rules will iterate over all callbacks, they verify whether the command is correct and if yes, handle it.

Note that I do not consider this feature mandatory for this PR to be merged - if you do not have time/motivation for this then I'll implement it later (and I'll then review this PR as it is).

plugins/common/lambda-rules-registration.cpp
```diff
@@ -0,0 +1,70 @@
+#include "lambda-rules-registration.hpp"
```
ammen99, line 1: You could put the contents of this file in the header, making this header-only, that's the approach taken by other similar classes.

plugins/common/lambda-rules-registration.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    auto instance = get_core().get_data<lambda_rules_registrations_t>();
+    if (instance == nullptr)
+    {
+        get_core().store_data(std::unique_ptr<lambda_rules_registrations_t>(new lambda_rules_registrations_t()));
+
+        instance = get_core().get_data<lambda_rules_registrations_t>();
+
+        if (instance == nullptr) {
+            std::cerr << "Window lambda rules: Lazy-init of lambda registrations failed." << std::endl;
```
ammen99, line 27: Do not use `std::cerr`, but rather use our logging infrastructure (i.e `LOGE`, `LOGW`, etc).

plugins/common/lambda-rules-registration.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "wayfire/nonstd/observer_ptr.h"
+#include "wayfire/parser/lambda_rule_parser.hpp"
+#include "wayfire/rule/lambda_rule.hpp"
+
+namespace wf
+{
+
+nonstd::observer_ptr<lambda_rules_registrations_t> lambda_rules_registrations_t::getInstance()
+{
+    auto instance = get_core().get_data<lambda_rules_registrations_t>();
```
ammen99, line 19: A much easier way to achieve this is to use `get_data_safe` (which does exactly what you are doing here).

plugins/common/lambda-rules-registration.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace wf
+{
+
+struct lambda_rule_registration_t;
+
+using map_type = std::map<std::string, std::shared_ptr<lambda_rule_registration_t>>;
+
+/**
+ * @brief The lambda_rule_registration_t struct represents registration information for a single lambda rule.
```
ammen99, line 21: I would like to also see how the rule text is used.

plugins/common/lambda-rules-registration.hpp
```diff
@@ -0,0 +1,131 @@
...
+    wf::lambda_t else_lambda;
+
+    /**
+     * @brief access_interface Access interface to be used when evaluating the rule.
+     *
+     * @note If this is left blank (nullptr), the standard view_access_interface_t instance will be used.
+     */
+    std::shared_ptr<wf::access_interface_t> access_interface;
+
+    /**
```
ammen99, line 62: If this is not supposed to be set by plugins, I can imagine we can declare it as private and add the other class as a friend?

plugins/common/lambda-rules-registration.hpp
```diff
@@ -0,0 +1,131 @@
...
+ */
+class lambda_rules_registrations_t : public custom_data_t
+{
+  public:
+    /**
+     * @brief getInstance Static accessor for the singleton.
+     *
+     * @return Observer pointer to the singleton instance, fetched from wf::core.
+     */
+    static nonstd::observer_ptr<lambda_rules_registrations_t> getInstance();
```
ammen99, line 88: Naming: wayfire uses `snake_case` for everything

ammen99: also same for functions below

plugins/common/lambda-rules-registration.hpp
```diff
@@ -0,0 +1,131 @@
...
+     * @brief registerLambdaRule Registers a lambda rule with its associated key.
+     *
+     * This method will return error result if the key is not unique or the registration struct is incomplete.
+     *
+     * @param[in] key Unique key for the registration.
+     * @param[in] registration The registration structure.
+     *
+     * @return <code>True</code> in case of error, <code>false</code> if ok.
+     */
+    bool registerLambdaRule(std::string key, std::shared_ptr<lambda_rule_registration_t> registration);
```
ammen99, line 100: Trying to clarify things for myself: why do we use a `shared_ptr` here? Who is the owner of the memory? Isn't this always the plugin using this interface? (so it could be observer_ptr, if I understand correctly).

plugins/common/view-action-interface.cpp
```diff
@@ -0,0 +1,308 @@
+#include "wayfire/output.hpp"
```
ammen99, line 1: I have a feeling that this file belongs exclusively to window-rules, doesn't it? So it shouldn't be in common.

plugins/common/view-action-interface.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+void view_action_interface_t::_move(int x, int y)
+{
+    // Clamp x and y to sane values. Do not allow to move outside of the output.
+    auto output = _view->get_output();
+    if (output != nullptr)
+    {
+        auto dimensions = output->get_screen_size();
+
+        x = std::clamp(x, 0, (dimensions.width - 40));
```
ammen99, line 286: Any rationale for this limit?

plugins/single_plugins/window-rules.cpp
```diff
@@ -1,307 +1,161 @@
...
-        {
-            auto conv = static_cast<view_fullscreen_signal*> (data);
-            assert(conv);
-
-            if (!conv->state || conv->carried_out)
-                return;
+            error = registration->rule_instance->apply(signal, _access_interface);
+        }
+        else {
+            error = registration->rule_instance->apply(signal, *registration->access_interface);
```
ammen99, line 150: Shouldn't you set the access interface view here too ?

plugins/common/lambda-rules-registration.hpp
erikvanhamme: I used shared_ptr here because the registration object will be changed (compiled rule added). I will read up on observer_ptr to see if I can use that instead. This whole smart pointer stuff is not something I use a lot (Qt and embedded dev) and I am bound to fuck up here and there.

plugins/common/view-action-interface.cpp
erikvanhamme: I put it in common because other plugins that use the rule engine may use it. Also, window-rules was a single plugin and I did not want to add it there. I will move window-rules to a folder as you suggested, but we should still weigh potential reuse of the view access interface versus the grouping of all files. I am fine with either decision, but I did have a good reason to put that in common.

plugins/common/lambda-rules-registration.hpp
ammen99: Hmm, in that case, maybe you want to pass a unique_ptr ? This way the plugin calling this transfers the object to window-rules and does not own it anymore.

plugins/common/view-action-interface.cpp
ammen99: Alright, but view-access-interface is in core, right? It is reusable by all plugins already.

plugins/common/view-action-interface.cpp
erikvanhamme: I wanted to avoid moving the window to a position offscreen where it would be impossible to grab the title bar and manually move it back. To avoid "losing" the window due to a bad rule. I used the 40/30 pixels again as minimum size in resize method below.

ammen99: Hmm, I see. However, moving to negative coordinates or something else also has its use: for example, if I move to negative coordinates, I will be moving my view to the workspace on the left. Also, maybe I want to put part of the view offscreen.

All in all, here is what I think can fulfill both usecases: just reject moving the view iff it is not visible on any workspace. For example using this function: https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/workspace-manager.hpp#L115

plugins/single_plugins/window-rules.cpp
erikvanhamme: The access interface type referred to here may use another source of information then the view. If operation on the view data is desired, it is better that the plugin making the registration leaves this at nullptr. I only added the option of setting a non-standard access interface to allow plugins to use different data source type. If a custom access interface is used, Window-rules does not know anything about it. We can discuss changing this feature because we never really set hard requirements for it before.

plugins/common/view-action-interface.cpp
erikvanhamme: I did not know about the workspaces thing. I will change the code to check if the move would go to another workspace and reject it if not.

Are you good with the minimum size limit I put on, or is there a trick that I missed there as well?

ammen99: No, there is no other trick (which is why I didn't comment on the resize code :)

plugins/single_plugins/window-rules.cpp
ammen99: Hmm, I see. However, currently all signals window-rules connect to are about views, so the event is only meaningful if there is a view bound to it? Or maybe I am misunderstanding.

ammen99: On the other hand, you are right, we might want to add events like output-added in the future ...

ammen99:
Gentle ping ;)

If you are not going to have possibility to fix this up in the next few weeks I might be able to do the remaining changes, it is not too much work, and it would be a waste to let this PR rot.

plugins/single_plugins/window-rules.cpp
```diff
@@ -1,307 +1,161 @@
...
-        result.signal = event;
-        result.func = [exec] (wayfire_view view)
-        {
-            if (exec.verify(view, exec.verification_string))
-                exec.action(view);
-        };
+void wayfire_window_rules_t::fini()
+{
+    output->disconnect_signal("map-view", &_created);
+    output->disconnect_signal("view-maximized", &_maximized);
```
ammen99, line 109: There is only `view-tiled` now

plugins/common/lambda-rules-registration.cpp
erikvanhamme: I looked into this, and I found that get_data_safe needs the constructor of the object to be public. This violates the intent of the registrations a bit since I wanted it to be a singleton.

I'll leave this be for now unless you tell me to change it.

plugins/common/lambda-rules-registration.cpp
erikvanhamme: Done

plugins/common/lambda-rules-registration.cpp
erikvanhamme: Done

plugins/common/lambda-rules-registration.hpp
erikvanhamme: Done

plugins/common/lambda-rules-registration.hpp
erikvanhamme: Done

plugins/common/lambda-rules-registration.hpp
erikvanhamme: Done

plugins/common/view-action-interface.cpp
erikvanhamme: Done

plugins/common/lambda-rules-registration.hpp
ammen99: I guess a shared_ptr is fine enough.

ammen99 (commented):
A few more things I noticed this time.

A question: what is the purpose of the separated action interface? I thought its purpose is so that other plugins can also add their own action interfaces (not whole rules), but this doesn't seem the case.

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+        LOGE("Window-rules: Error while registering test rule.");
+    }
+
+    // Build rule list.
+    auto section = wf::get_core().config.get_section("window-rules");
+    for (auto opt : section->get_registered_options())
+    {
+        _lexer.reset(opt->get_value_str());
+        auto rule = wf::rule_parser_t().parse(_lexer);
+        if (rule != nullptr)
```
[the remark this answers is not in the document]
erikvanhamme: The rule parser will spit out a detailed error, including a pointer to the symbol in the rule string that is wrong.

Unfortunately it will be spit out on std::cerr because the logging facilities are not available in wf-utils.

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+        // Run the lambda rule.
+        error = registration->rule_instance->apply(signal, _access_interface);
+
+        // Unload wrappers.
+        registration->rule_instance->setIfLambda(nullptr);
+        registration->rule_instance->setElseLambda(nullptr);
+
+        if (error)
+        {
+            LOGE("Window-rules: Error while executing rule on ", signal, " signal.");
```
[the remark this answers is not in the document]
erikvanhamme: Error can be set true in the following conditions:
* The lambda returns true to signal an error during the execution of whatever code is in the lambda.
* The access interface returns an error because it is looking for a property that does not exist (in the case a rule mentions a non-existing variable, but in a semantically correct way).

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+        if (registration->access_interface != nullptr)
+        {
+            access_iface = *registration->access_interface;
+        }
+
+        // TODO: Remove debug/test code.
+        LOGW("Apply rule to view: ", view->get_app_id(), ": ");
+
+        // Load if lambda wrapper.
+        if (registration->if_lambda != nullptr)
```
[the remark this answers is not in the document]
erikvanhamme: I'll keep this one and remove the inner one. I fucked this up with my sleepyhead... :)

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    nonstd::observer_ptr<wf::lambda_rules_registrations_t> _lambda_registrations;
+};
+
+void wayfire_window_rules_t::init()
+{
+    // Get the lambda rules registrations.
+    _lambda_registrations = wf::lambda_rules_registrations_t::get_instance();
+
+    // TODO: Remove debug/test code.
```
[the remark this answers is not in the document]
erikvanhamme: > A question: what is the purpose of the separated action interface? I thought its purpose is so that other plugins can also add their own action interfaces (not whole rules), but this doesn't seem the case.

It was split to get the ability to use the conditions separate from the action rules. You already do this in the matcher I think. The condition only needs the access interface where the rules (which act on the result of the condition check) need both interfaces.

I basically split this to build in more flexibility in the architecture with eye on re-usability of the rule/condition stuff. I would advise against merging these again.

plugins/window-rules/lambda-rules-registration.hpp
```diff
@@ -0,0 +1,217 @@
...
+                lambda_rules_registrations_t()));
+
+            instance = get_core().get_data<lambda_rules_registrations_t>();
+
+            if (instance == nullptr)
+            {
+                LOGE("Window lambda rules: Lazy-init of lambda registrations failed.");
+            } else
+            {
+                LOGI(
```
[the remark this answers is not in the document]
erikvanhamme: You mean the LOGI on success right? Because the init failing is an error that would cause the lambda rules not to work.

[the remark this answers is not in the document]
ammen99: Yes, I mean LOGI -> LOGD, the error is good as it is.

plugins/window-rules/window-rules.cpp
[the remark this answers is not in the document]
ammen99: Alright, then it may make sense to also log the whole rule text, if possible.

plugins/window-rules/window-rules.cpp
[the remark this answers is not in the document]
ammen99: Ok, at this stage it is important the user gets any error. Eventually, the logging should be moved to wf-utils.

plugins/window-rules/window-rules.cpp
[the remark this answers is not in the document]
ammen99: Ok.

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+            });
+        }
+
+        // Load else lambda wrapper.
+        if (registration->else_lambda)
+        {
+            registration->rule_instance->setElseLambda(
+                [registration, signal, view] () -> bool
+            {
+                if (registration->else_lambda != nullptr)
```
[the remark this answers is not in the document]
erikvanhamme: Done

plugins/window-rules/window-rules.cpp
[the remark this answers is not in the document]
erikvanhamme: Done

plugins/window-rules/lambda-rules-registration.hpp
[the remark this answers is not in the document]
erikvanhamme: Done

plugins/window-rules/window-rules.cpp
```diff
@@ -0,0 +1,234 @@
...
+        _access_interface.set_view(view);
+        wf::access_interface_t & access_iface = _access_interface;
+
+        // If a custom access interface is set in the regoistration, use this one.
+        if (registration->access_interface != nullptr)
+        {
+            access_iface = *registration->access_interface;
+        }
+
+        // TODO: Remove debug/test code.
```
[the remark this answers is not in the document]
erikvanhamme: Done

plugins/window-rules/window-rules.cpp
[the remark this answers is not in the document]
erikvanhamme: Test code removed

ammen99:
Just tested, works like a charm. Thank you for your work!

ammen99 (approved):

PR #531, opened 2020-05-24
cube: cubemap: Reset state on error

### LINKED ISSUE #530: [cube] cubemap crashes if no file / invalid file set
I have never used cubemap and TBH don't know what it is.
Although when no file is specified or a txt file is specified wayfire crashes.

https://stackoverflow.com/questions/12774207/fastest-way-to-check-if-a-file-exist-using-standard-c-c11-c
I wanted to implement the fastest method to check if the file exists but a txt file specified would crash wf as well and I don't know how to check if file == cubemap

soreau:
Call OpenGL::render_end() where needed and unbind cubemap texture.
Without this, rotate cube crashes when no cubemap image is set.
Fixes #530.

### DIFF
--- a/plugins/cube/cubemap.cpp
+++ b/plugins/cube/cubemap.cpp
@@ -80,6 +80,7 @@ void wf_cube_background_cubemap::render_frame(const wf::framebuffer_t& fb,
     {
         GL_CALL(glClearColor(TEX_ERROR_FLAG_COLOR));
         GL_CALL(glClear(GL_COLOR_BUFFER_BIT));
+        OpenGL::render_end();
         return;
     }
     program.use(wf::TEXTURE_TYPE_RGBA);


plugins/cube/cubemap.cpp
```diff
@@ -51,7 +51,7 @@ void wf_cube_background_cubemap::reload_texture()
 
             GL_CALL(glDeleteTextures(1, &tex));
             tex = -1;
-            break;
+            goto render_end;
```
ammen99, line 54: I definitely don't see why this thing here is necessary?

plugins/cube/cubemap.cpp
```diff
@@ -80,6 +81,7 @@ void wf_cube_background_cubemap::render_frame(const wf::framebuffer_t& fb,
     {
         GL_CALL(glClearColor(TEX_ERROR_FLAG_COLOR));
         GL_CALL(glClear(GL_COLOR_BUFFER_BIT));
+        OpenGL::render_end();
```
ammen99, line 84: +1

plugins/cube/cubemap.cpp
soreau: Ah, I read it as `return` and not `break`. Indeed, it is not needed.

PR #532, opened 2020-05-24
Add get_view_workspaces

damianatorrpm:


src/output/workspace-impl.cpp
```diff
@@ -398,6 +398,50 @@ class output_viewport_manager_t
...
+        if (threshold != 1.0)
+         {
+             bounding_box = view->get_bounding_box();
+             output_relative_geometry = view->get_output()->get_relative_geometry();
+             intersection = wf::geometry_intersection(bounding_box, output_relative_geometry);
+             area = 1.0 * intersection.width * intersection.height;
+             area /= 1.0 * bounding_box.width * bounding_box.height;
+         }
+
+        for (int row = output_workspaces.width; row-- > 0;)
```
ammen99, line 427: 1. Please don't use `--` and `>` in the same expression, I can understand what's going on but I don't think it adds anything but makes the code a little more complex.
2. Workspaces are generally numbered from 0 to vwidth-1, it is strange that this function returns them in {1 .. vwidth}

src/output/workspace-impl.cpp
```diff
@@ -398,6 +398,50 @@ class output_viewport_manager_t
...
+        wf::geometry_t output_relative_geometry;
+        wf::geometry_t intersection;
+        double area;
+        if (threshold != 1.0)
+         {
+             bounding_box = view->get_bounding_box();
+             output_relative_geometry = view->get_output()->get_relative_geometry();
+             intersection = wf::geometry_intersection(bounding_box, output_relative_geometry);
+             area = 1.0 * intersection.width * intersection.height;
+             area /= 1.0 * bounding_box.width * bounding_box.height;
```
ammen99, line 424: I thought the idea is to calculate visible area for **each** workspace? In this way you'll be doing it for just the current workspace, which is also completely wrong - the view is possibly completely unvisible on current workspace, but it is still visible on others.

src/api/wayfire/workspace-manager.hpp
```diff
@@ -111,6 +111,14 @@ enum sublayer_mode_t
 class workspace_manager
 {
   public:
+    /**
+     * @param threshold Threshold of the view to be counted
```
ammen99, line 115:
```suggestion
     * Calculate a list of workspaces the view is visible on.
     * @param threshold How much of the view's area needs to overlap a workspace to be counted as visible on it.
```

src/api/wayfire/workspace-manager.hpp
```diff
@@ -111,6 +111,14 @@ enum sublayer_mode_t
 class workspace_manager
 {
   public:
+    /**
+     * @param threshold Threshold of the view to be counted 
+     *        on that workspace. 1.0 for 100% visible, 0.1 for 10%
```
ammen99, line 116:
```suggestion
     *    1.0 for 100% visible, 0.1 for 10%.
```

ammen99:
A few (IMO) improvements to the docstring, but as for the impl, I think you forgot to put the area calculation in the loops ?

src/output/workspace-impl.cpp
```diff
@@ -398,6 +398,48 @@ class output_viewport_manager_t
...
+
+        for (int row = 0; row < output_workspaces.width; row++)
+        {
+           for (int column = 0; column < output_workspaces.heigh; column++)
+           {
+              wf::point_t workspace = {row, column};
+              if (output->workspace->view_visible_on(view, workspace))
+              {
+                 bounding_box = view->get_bounding_box();
+                 output_relative_geometry = view->get_output()->get_relative_geometry();
```
ammen99, line 427: The box for a workspace has x/y coordinates equal to vx*output_width / vy*output_height, where vx/vy are the coordinates of a workspace RELATIVE to the current workspace.

FWIW there is render_manager::get_ws_box() which should actually be moved to workspace-manager because that's where it belongs ...

damianatorrpm:
> LGTM, indentation was a little bit off and I could see some potential simplifications, so I added a commit to address them (would've probably taken longer to explain them rather than write them myself lol)

Faster for me too. Thanks a lot!

PR #533, opened 2020-05-25
Move resize workspace changed

damianatorrpm:
Adds view-viewport-changed signal to move and resize plugin as workspaces the view is on may have changed


damianatorrpm:
argh nvm I have to fix the other PR first and than rebase/recreate fork

PR #536, opened 2020-05-27
Add signals to move/resize that views workspace maybe changed

damianatorrpm:


plugins/tile/tile-plugin.cpp
```diff
@@ -315,7 +315,8 @@ class tile_plugin_t : public wf::plugin_interface_t
     signal_callback_t on_view_change_viewport = [=] (signal_data_t *data)
     {
         auto ev = (view_change_viewport_signal*) (data);
-        change_view_workspace(ev->view, ev->to);
+        if (ev->old_viewport_invalid)
```
ammen99, line 318: indentation is off

plugins/single_plugins/resize.cpp
```diff
@@ -249,6 +250,11 @@ class wayfire_resize : public wf::plugin_interface_t
                 view->set_moving(false);
             view->set_resizing(false);
             end_wobbly(view);
+
+            view_change_viewport_signal workspace_may_changed;
+            workspace_may_changed.view = this->view;
+            workspace_may_changed.old_viewport_invalid = false;
```
ammen99, line 256: what about setting current viewport? We know the viewport that we're moving to (current one)

plugins/single_plugins/move.cpp
```diff
@@ -266,6 +267,11 @@ class wayfire_move : public wf::plugin_interface_t
                 update_slot(0);
             }
 
+            view_change_viewport_signal workspace_may_changed;
+            workspace_may_changed.view = this->view;
+            workspace_may_changed.old_viewport_invalid = false;
+            output->emit_signal("view-change-viewport", &workspace_may_changed);
```
ammen99, line 273: same as below (setting target workspace)

plugins/common/wayfire/plugins/common/view-change-viewport-signal.hpp
```diff
@@ -8,4 +8,11 @@
 struct view_change_viewport_signal : public _view_signal
 {
     wf::point_t from, to;
+
+    /**
+     * from and to maybe null
+     * if old_viewport_invalid is false
+     * as it should trigger a full re-read
```
ammen99, line 15:
```suggestion
     * Indicates whether the old viewport is known.
     * If false, then the `from` field should be ignored.
```

ammen99: I want to change the wording because whether full re-read should be triggered or not is something the plugin connecting to this signal receives.

plugins/single_plugins/resize.cpp
damianatorrpm: That was the reason I omitted these, it made only sense to me in a scenario where all workspaces need to be recalculated and setting the current view-port probably adds a byte somewhere, that's why the tile -plugin was adjusted to not react if old_viewport_invalid=true

PR #537, opened 2020-05-27
Add gtk_shell v3 and add function implementations

damianatorrpm:


damianatorrpm:
One thing on coding style though https://github.com/WayfireWM/wayfire/pull/537/commits/e39974b0a4af77b83240e775741cc416066c93aa

IMHO this is much easier to read
```
std::string wf_gtk_shell_get_custom_app_id(wf_gtk_shell *shell,
                                                                       wl_resource *surface);
```

than this
```
std::string wf_gtk_shell_get_custom_app_id(wf_gtk_shell *shell,
    wl_resource *surface);
```

and not only that it makes more sense with wayfire code-base, why?
For example the if's are aligned as well so the 2 constructs align:
```
   if (||view && view != nullptr)
       ||view->erase_data("gtk-shell-modal");
```

ammen99:
> [](https://github.com/WayfireWM/wayfire/commit/e39974b0a4af77b83240e775741cc416066c93aa)
> 
> IMHO this is much easier to read
> 
> ```
> std::string wf_gtk_shell_get_custom_app_id(wf_gtk_shell *shell,
>                                            wl_resource *surface);
> ```

As with all style preferences, this is subjective. I find the other format much more readable.




> For example the if's are aligned as well so the 2 constructs align:
> 
> ```
>    // and must gtk-connected people actually design 
>   // the function so that the first argument is always the longest
>    if (||view && view != nullptr)
>        ||view->erase_data("gtk-shell-modal");
> ```

The ifs are aligned with 4 spaces, which is coincidentally alignment with the first line.



> As a 'bonus', it's how it is used throughout whatever gnome people touch.
> I see that wlroots sway does it with the 4 as well, but than they also prefer `{ `on the same line.

The brace on the other line may change for ifs and fors, since I wanted to try using automatic formatting with clang-format and it can't really do what  I want it to yet (I'd like same-line for short statements, brake-to-newline for multi-statement if). But anyway this hasn't been decided yet, please do not change the formatting.

And btw gnome people can do whatever they want :)

src/api/wayfire/signal-definitions.hpp
```diff
@@ -67,6 +67,10 @@ struct view_tiled_signal : public _view_signal
     wf::geometry_t desired_size;
 };
 
+struct view_self_request_focus_signal : public _view_signal
```
ammen99, line 70:
```suggestion
/**
 * view-self-request-focus signal is emitted on the view's output whenever the client indicates the
 * view should become active.
 */
struct view_self_request_focus_signal : public _view_signal
```

ammen99:
> But anyway this hasn't been decided yet, please do not change the formatting.

Actually, please change formatting at places which are not indented as they should - I noticed you did that for the callbacks list, which is of course perfectly fine.

src/output/gtk-shell.cpp
```diff
@@ -2,75 +2,227 @@
...
+    .request_focus = handle_gtk_surface_request_focus,
 };
 
-static void handle_gtk_shell_get_gtk_surface(wl_client *client, wl_resource *resource,
-    uint32_t id, wl_resource *surface)
+/**
+ * Passes the gtk_surface object to the client.
+ */
+static void handle_gtk_shell_get_gtk_surface(wl_client *client,
+    wl_resource *resource,
```
ammen99, line 128: I usually try to binpack arguments on the same line, as it was before ...

damianatorrpm: done.

src/output/gtk-shell.cpp
```diff
@@ -2,77 +2,205 @@
...
+ * Modal dialogs may be handled differently than non-modal dialogs.
+ * It is a hint that this should be attached to the parent surface.
+ * In gnome this does not affect input-focus.
+ * This function sets the modal hint.
+ */
+static void handle_gtk_surface_set_modal(wl_client *client, wl_resource *resource)
+{
+    auto surface = static_cast<wl_resource*> (wl_resource_get_user_data(resource));
+    wayfire_view view = wf::wl_surface_to_wayfire_view(surface);
+    if (view && view != nullptr)
```
ammen99, line 44:
```suggestion
    if (view)
```
Seems like you're doing the same check twice?

src/output/gtk-shell.cpp
```diff
@@ -2,77 +2,205 @@
...
+ * Modal dialogs may be handled differently than non-modal dialogs.
+ * It is a hint that this should be attached to the parent surface.
+ * In gnome this does not affect input-focus.
+ * This function removes the modal hint.
+ */
+static void handle_gtk_surface_unset_modal(wl_client *client, wl_resource *resource)
+{
+    auto surface = static_cast<wl_resource*> (wl_resource_get_user_data(resource));
+    wayfire_view view = wf::wl_surface_to_wayfire_view(surface);
+    if (view && view != nullptr)
```
ammen99, line 59:
```suggestion
    if (view)
```
Same as above

src/output/gtk-shell.cpp
```diff
@@ -2,77 +2,205 @@
...
+ * gnome-control-center, gnome-clocks, dconf-editor are single instance and if
+ * they are already running and launched again, this will request that they get focused.
+ * This function is superseeded by handle_gtk_surface_request_focus a newer equivalelent
+ * used by gtk-applications now. This function is for compatibility reasons.
+ */
+static void handle_gtk_surface_present(wl_client *client, wl_resource *resource, uint32_t time)
+{
+    auto surface = static_cast<wl_resource*> (wl_resource_get_user_data(resource));
+    wayfire_view view = wf::wl_surface_to_wayfire_view(surface);
+    if (view && view != nullptr)
```
ammen99, line 75:
```suggestion
    if (view)
```

src/output/gtk-shell.cpp
```diff
@@ -2,77 +2,205 @@
...
+/**
+ * The surface requests focus, for example single instance applications like
+ * gnome-control-center, gnome-clocks, dconf-editor are single instance and if
+ * they are already running and launched again, this will request that they get focused.
+ */
+static void handle_gtk_surface_request_focus(struct wl_client *client, struct wl_resource *resource, const char* startup_id)
+{
+    auto surface = static_cast<wl_resource*> (wl_resource_get_user_data(resource));
+    wayfire_view view = wf::wl_surface_to_wayfire_view(surface);
+    if (view && view != nullptr)
```
ammen99, line 93:
```suggestion
    if (view)
```

src/output/gtk-shell.cpp
```diff
@@ -2,77 +2,205 @@
...
+        view_self_request_focus_signal data;
+        data.view = view;
+        view->get_output()->emit_signal("view-self-request-focus", &data);
+    }
+}
+
+
+/**
+ * Destroys the gtk_surface object.
+ * TODO: May this be required for some kind of leak?
```
ammen99, line 104:
```suggestion
```

There shouldn't be any leaks, because we are not allocating any objects when the surface is created aside from the wayland protocol object, and this is its destructor.

damianatorrpm: Oh, I didn't realize that this is the destructor.

damianatorrpm:
okay, anything else?

ammen99:
Indeed, this seems good for merging.

PR #539, opened 2020-05-29
README: sync with wiki

ericonr:
- Dependencies
- wf-shell
- Formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ It aims to create a customizable, extendable and lightweight environment without
 
 ## Dependencies
 
+### Wayfire Dependencies
+
+These are the dependencies needed for building Wayfire.
+
 - [Cairo](https://cairographics.org)
 - [FreeType](https://freetype.org)
 - [GLM](https://glm.g-truc.net)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,27 @@ It aims to create a customizable, extendable and lightweight environment without
 - [wf-config](https://github.com/WayfireWM/wf-config)
 - [wlroots](https://github.com/swaywm/wlroots)
 
+### Login Provider (optional)
+
+- [systemd](https://systemd.io/) **or**
+- [elogind](https://github.com/elogind/elogind)
+
+### XWayland Support (optional)
```
ammen99, line 53: Some of these are hard to get except if you don't compile from source. For example, I do not have a package `xcb-errors`, but neither is this really required. Can you list only required packages here? (for login providers, I'd say it is okay).

README.md
```diff
@@ -45,6 +45,27 @@ It aims to create a customizable, extendable and lightweight environment without
 - [wf-config](https://github.com/WayfireWM/wf-config)
 - [wlroots](https://github.com/swaywm/wlroots)
 
+### Login Provider (optional)
```
ammen99, line 48: I also think that it might be worth to say login provider and xwayland dependencies are optional, but highly recommended (after all, except if you don't really know what you're doing, you probably want to have these)

soreau: > I also think that it might be worth to say login provider and xwayland dependencies are optional, but highly recommended (after all, except if you don't really know what you're doing, you probably want to have these)

Maybe we can use (optional, recommended) 'tag'.

ericonr: Sounds good! I will make the changes here till there's a definite agreement, then I can update the wiki page,

ericonr:
Sorry for the wait, updated.

ammen99:
On a second glance, we should somehow say that Xwayland and login provider deps need to be installed before wlroots is built (because these are actually deps for wlroots).

PR #543, opened 2020-05-31
ci: Don't ignore any files

soreau:
This makes it so ci will run even if the only files modified are in
metadata/ or proto/.


ammen99:
LGTM

PR #552, opened 2020-06-08
Fix almost all game related issues

damianatorrpm:
Requires https://github.com/swaywm/wlroots/pull/2264

* Adds minimizing support for chrome / steam csd
  and games that want to minimize on focus loss
* Adds view-self-focus-request for xwayland like gtk-shell has
* Adds view-hints-changed-signal (for urgency hint, possibly others)
* Adds view-system-bell signal (currently used in gtk-shell only)
* Fixes fullscreen games, the move behavior of fullscreen views,
   is to snap back to fullscreen after move. I would demo how 
  smooth and nice this works with multiple outputs but wf-recorder
  can only record one.
* Disallow resizing & snap off for fullscreen as this breaks 95 % of steam games,
   since they expect the size to be the resolution it is set to.

I have to say that in some occasions the switcher plugin crashes wayfire,
but I am unable to find the cause. The crash never happens if you 
reload the wobbly plugin at the beginning of your session,
maybe the load order is important here...

The fast switcher work perfectly fine.
It seems to be a 'floating point exception in switcher' and I found switcher
to also have other kind of strange behavior as setting views alpha to 0.0
when minimized and sometimes after restore it is not restored to 1.0,
also the other issue I demoed with the switcher (can't switch).

I have disabled this plugin and hope for a soon present plugin :100:

### DIFF
--- a/plugins/single_plugins/move-snap-helper.hpp
+++ b/plugins/single_plugins/move-snap-helper.hpp
@@ -106,18 +106,15 @@ class move_snap_helper_t : public wf::custom_data_t
   protected:
     virtual bool should_enable_snap_off() const
     {
-        return enable_snap_off &&
-            (view->tiled_edges || view->fullscreen);
+        return enable_snap_off && (view->tiled_edges || view->fullscreen);
     }
 
     /** Move the view out of its slot */
     virtual void snap_off()
     {
         view_in_slot = false;
-        if (view->fullscreen)
-            view->fullscreen_request(view->get_output(), false);
 
-        if (view->tiled_edges)
+        if (view->tiled_edges && !view->fullscreen)
             view->tile_request(0);
     }
 
--- a/plugins/single_plugins/move.cpp
+++ b/plugins/single_plugins/move.cpp
@@ -210,10 +210,10 @@ class wayfire_move : public wf::plugin_interface_t
                 return false;
             }
 
-            view->store_data(std::make_unique<wf::move_snap_helper_t> (
-                    view, get_input_coords()));
-
             output->focus_view(view, true);
+            view->store_data(std::make_unique<wf::move_snap_helper_t> (
+                view, get_input_coords()));
+
             if (enable_snap)
                 slot.slot_id = 0;
 
@@ -256,7 +256,21 @@ class wayfire_move : public wf::plugin_interface_t
             }
 
             /* Snap the view */
-            if (enable_snap && slot.slot_id != 0)
+            /*
+             * A view that is fullscreened may break if resized i.e
+             * tiled_vertical or tiled_horizontal, games usually have
+             * their own option to be fullscreen.
+             * If the compositor resizes them and thus removes
+             * for xwayland _NET_WM_STATE fullscreen (there
+             * are not much wayland games around to test what wayland games
+             * do when they have fullscreen in their settings and
+             * the compositor puts them in another state) they break.
+             *
+             * This way fullscreened view can be moved from one
+             * output to another and be restored to fullscreen
+             * on a different output (hopefully with the same resolotion)
+            */
+            if (enable_snap && slot.slot_id != 0 && !this->view->fullscreen)
             {
                 snap_signal data;
                 data.view = view;
@@ -272,6 +286,14 @@ class wayfire_move : public wf::plugin_interface_t
             workspace_may_changed.to = output->workspace->get_current_workspace();
             workspace_may_changed.old_viewport_invalid = false;
             output->emit_signal("view-change-viewport", &workspace_may_changed);
+
+            /*
+             * Restore the fullscreen state so the window is
+             * not in a weird position afterwards, fullscreen
+             * are by spec the geometry of the output (or span outputs)
+             */
+            if (this->view->fullscreen)
+                view->fullscreen_request(output, true);
 
             this->view = nullptr;
         }
@@ -537,7 +559,7 @@ class wayfire_move : public wf::plugin_interface_t
             /* View might get destroyed when updating multi-output */
             if (view)
             {
-                if (enable_snap && !MOVE_HELPER->is_view_fixed())
+                if (enable_snap && !MOVE_HELPER->is_view_fixed() && !this->view->fullscreen)
                     update_slot(calc_slot(input.x, input.y));
             } else
             {
--- a/src/api/wayfire/signal-definitions.hpp
+++ b/src/api/wayfire/signal-definitions.hpp
@@ -76,6 +76,23 @@ struct view_self_request_focus_signal : public _view_signal
 };
 
 /**
+ * view-hints-changed signal is emitted on the view's output whenever the client indicates the
+ * views hints have changed (example urgency hint).
+ */
+struct view_hints_changed_signal : public _view_signal
+{
+    bool demands_attention = false;
+};
+
+/**
+ * view-system-bell signal is emitted on the view's output whenever the client indicates the
+ * view wants to invoke the system bell if such is available.
+ */
+struct view_system_bell_signal : public _view_signal
+{
+};
+
+/**
  * The view-fullscreen-request and view-fullscreen signals are emitted on the view's output when the view's fullscreen state changes.
  * view-fullscreen-request is emitted when the view needs to be fullscreened, but has not been fullscreened yet.
  * view-fullscreen is emitted whenever the view's fullscreen state actually changes.
--- a/src/output/gtk-shell.cpp
+++ b/src/output/gtk-shell.cpp
@@ -86,7 +86,7 @@ static void handle_gtk_surface_present(wl_client *client, wl_resource *resource,
  * gnome-control-center, gnome-clocks, dconf-editor are single instance and if
  * they are already running and launched again, this will request that they get focused.
  */
-static void handle_gtk_surface_request_focus(struct wl_client *client, struct wl_resource *resource, const char* startup_id)
+static void handle_gtk_surface_request_focus(wl_client *client, wl_resource *resource, const char* startup_id)
 {
     auto surface = static_cast<wl_resource*> (wl_resource_get_user_data(resource));
     wayfire_view view = wf::wl_surface_to_wayfire_view(surface);
@@ -147,7 +147,18 @@ static void handle_gtk_shell_set_startup_id(wl_client *client, wl_resource *reso
  *  A view could use this to invoke the system bell, be it aural, visual or none at all.
  *  Not implemented.
  */
-static void handle_gtk_shell_system_bell(wl_client *client, wl_resource *resource, wl_resource *surface) {}
+static void handle_gtk_shell_system_bell(wl_client *client, wl_resource *resource, wl_resource *surface)
+{
+    auto wl_surface = static_cast<wl_resource*> (wl_resource_get_user_data(surface));
+    wayfire_view view = wf::wl_surface_to_wayfire_view(wl_surface);
+
+    if (view)
+    {
+        view_system_bell_signal data;
+        data.view = view;
+        view->get_output()->emit_signal("view-system-bell", &data);
+    }
+}
 
 
 /**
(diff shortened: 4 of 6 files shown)


damianatorrpm:
Confirming: the order is important - wobbly plugin must be before switcher plugin otherwise a rare issue might crash wayfire

src/api/wayfire/signal-definitions.hpp
```diff
@@ -75,6 +75,22 @@ struct view_self_request_focus_signal : public _view_signal
 {
 };
 
+/**
+ * view-hints-changed signal is emitted on the view's output whenever the client indicates the
+ * views hints have changed (example urgency hint).
```
ammen99, line 80: What is the difference between hints-changed and system-bell, except that they come from different sources?

src/view/view-impl.cpp
```diff
@@ -227,7 +227,7 @@ wlr_surface *wf::wlr_view_t::get_keyboard_focus_surface()
 
 bool wf::wlr_view_t::should_be_decorated()
 {
-    return role == wf::VIEW_ROLE_TOPLEVEL && !has_client_decoration;
+    return role == wf::VIEW_ROLE_TOPLEVEL && !has_client_decoration && !fullscreen;
```
ammen99, line 230: the decoration plugin is responsible for this, dont' put it here. If you see some bug related to this, let's look into how to fix it in the decoration plugin.

src/view/xwayland.cpp
```diff
@@ -305,9 +306,24 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
...
+                this->get_output()->emit_signal("view-self-request-focus", &data);
+            }
+        });
+        on_request_minimize.set_callback([&] (void*) {
+            minimize_request(xw->minimized);
+        });
         on_request_maximize.set_callback([&] (void*) {
-            tile_request((xw->maximized_horz && xw->maximized_vert) ?
-                wf::TILED_EDGES_ALL : 0);
+            if (xw->maximized_horz && xw->maximized_vert)
```
ammen99, line 322: wait why the change? It doesn't do anything ...

ammen99: I mean it is the same as before

src/view/xwayland.cpp
```diff
@@ -327,11 +343,29 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
             update_decorated();
         });
 
+        on_set_hints.set_callback([&] (void*) {
+            /*
+             * By storing urgency data instead of a signal
+             * it can be manually "acknowledged" by removing the data
+            */
+            if (xw->hints_urgency && !this->has_data("view-demands-attention"))
+                this->store_data(std::unique_ptr<wf::custom_data_t>(), "view-demands-attention");
```
ammen99, line 352: why is this useful? I think that emitting the signal is enough. Plugins can store this data themselves if they want/need to.

src/view/xwayland.cpp
```diff
@@ -400,7 +438,10 @@ class wayfire_xwayland_view : public wayfire_xwayland_view_base
         }
 
         if (xw->fullscreen)
-            fullscreen_request(get_output(), true);
+        {
+                this->view_impl->last_windowed_geometry = get_output_geometry();
```
ammen99, line 442: 1) output geometry is the wrong thing to take, you should pick wm_geometry
2) what bug(s) does this fix? This should already be handled in fullscreen_request

plugins/single_plugins/move-snap-helper.hpp
```diff
@@ -106,16 +106,13 @@ class move_snap_helper_t : public wf::custom_data_t
   protected:
     virtual bool should_enable_snap_off() const
     {
-        return enable_snap_off &&
```
ammen99, line 109: I don't see how this change is useful. If snap-off is enabled, this means the view stays in place until a certain threshold is achieved - looks like exactly what we want for fullscreen?

src/api/wayfire/signal-definitions.hpp
damianatorrpm: well, if you have gnome terminal and test it on gnome, you will see that if your cursor is already most-left at the terminal and press left you will hear woof if that is your alert sound.
hints changed is really like IM message received

src/view/xwayland.cpp
damianatorrpm: You looked at the newer commits which almost all say test,
because of the switcher plugin :)

There is one other issue I have been looking into:
https://github.com/ValveSoftware/steam-for-linux/issues/3792

this change was basically a result though that these are different properties and possibly at some point this should be changed further to actually make partial maximize recognized (i.e horz / vert)

src/view/xwayland.cpp
damianatorrpm: kinda unsure since hints_change may represent !xw->hints_urgency at some point, than what? view-no-longer-demands-attention?

src/view/xwayland.cpp
damianatorrpm: none really as mentioned above a desperate try to see why swticher crashes.

plugins/single_plugins/move-snap-helper.hpp
damianatorrpm: the important part here is that it removes the condition || fullscreen.
`If snap-off is enabled, this means the view stays in place until a certain threshold is achieved - looks like exactly what we want for fullscreen?`
Not sure what you mean but a definitive no.
With the patch you can freely move fullscreen views between outputs etc.
But snap-off resizes and un-fullscreens a fullscreen view because it thinks that right, 
which is not, hence snap-off is disabled for fullscreen.
It's just how applications work, hence the long comment, 
when you take games they decide (via option) if fullscreen or not, not the compositor.

src/view/view-impl.cpp
damianatorrpm: fair enough, will remove and we'll fix whatever happens

plugins/single_plugins/move.cpp
```diff
@@ -273,6 +292,17 @@ class wayfire_move : public wf::plugin_interface_t
             workspace_may_changed.old_viewport_invalid = false;
             output->emit_signal("view-change-viewport", &workspace_may_changed);
 
+            /*
+             * Restore the fullscreen state so the window is
+             * not in a weird position afterwards, fullscreen
+             * are by spec the geometry of the output (or span outputs)
+             */
+            if (this->view->fullscreen && this->view->activated
```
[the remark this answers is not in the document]
damianatorrpm: these are actually too many checks this->view->fullscreen was enough but,
remember the minimize bug? clicking on the surface that is being minimized?
well this is a result of finding out why dragging a different view, during the minimize action of a game that has "minimize on focus loss" actually is restored if a different view is dragged during it's minimize animation.

src/view/xwayland.cpp
ammen99: Just look at the code before and now, you have not changed ANYTHING.

ammen99: Well, except that before it was using ternary operator and now it uses if/else ..

src/view/xwayland.cpp
ammen99: put a flag in the signal?

src/view/xwayland.cpp
ammen99: alright, then, since this doesn't change anything, revert the change, most probably the bug is in switcher itself.

src/view/xwayland.cpp
damianatorrpm: you're right, my usual thinking is more toward the goal, whether my mind is currently located in plugin or not. The (theoretical, sure let's not get ahead of ourselves) issue I have is this though,
this requires all plugins that use this to use the same semantics for the data i.e the data name,
which hasn't been 'defined' anywhere. I understand this may not seem real at this point but the virtual barrier this detail puts, well I can't unsee it...

damianatorrpm: Nevertheless of course I will change it too a flag in the signal

plugins/single_plugins/move-snap-helper.hpp
ammen99: Snap-off is not responsible for unfullscreening and resizing the view! Snap-off is the thing which holds the view in place, so that you don't accidentally move it. 

However, I can see what you mean. Wouldn't it be better to just not unfullscreen the view in the `snap_off()` function? This way the snap-off-threshold is still respected (as it should) but the view is not unfullscreened.

damianatorrpm: I can see what you mean. Practically it's that you can't start a move operation by the decoration,
for fullscreen views since the have no decoration (maybe with some rare exceptions though) 
and one has to use the shortcut. Nevertheless I absolutely agree that it should respect the snap-off-threshold.

src/view/xwayland.cpp
ammen99: But each plugin that cares about this data will receive the signal, right? So all of them will know that the view has or doesn't have the urgent hint.

src/view/xwayland.cpp
damianatorrpm: If you implement at one point vert/horz maximization in this function it will have to change from the ternary operator proably. I wanted to implement this but wlroots has no signal for vert/horz tile changes yet. I can revert this.

plugins/single_plugins/move.cpp
```diff
@@ -210,11 +210,30 @@ class wayfire_move : public wf::plugin_interface_t
...
+             * It is possible to make the below clause an option,
+             * but not having that option enabled will break any
+             * game that is fullscreen or set's their own size,
+             * as tiling will resize them.
+             *
+             * This way fullscreened view can be moved from one
+             * output to another and be restored to fullscreen
+             * on a different output (hopefully with the same resolotion)
+             */
+            if (enable_snap && !view->fullscreen)
```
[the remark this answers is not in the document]
damianatorrpm: Will see now :)

src/view/xwayland.cpp
ammen99: I'd keep it as it was before, until the moment we implement separate maximization.

src/api/wayfire/signal-definitions.hpp
ammen99: Maybe a stupid question, but is this different from self-focus-request? My understanding was that the two are very similar.

damianatorrpm: Yes it's different :) 
The hint is not necessarily a request for focus.
It can be interpreted by a notify-send, a blinking taskbar icon, a sound,
whatever.

damianatorrpm: Also I kept it as view-hints-changed not view-urgency or something as there maybe other kind of hints

src/view/xwayland.cpp
damianatorrpm: Done

src/output/gtk-shell.cpp
```diff
@@ -147,7 +147,18 @@ static void handle_gtk_shell_set_startup_id(wl_client *client, wl_resource *reso
  *  A view could use this to invoke the system bell, be it aural, visual or none at all.
  *  Not implemented.
  */
-static void handle_gtk_shell_system_bell(wl_client *client, wl_resource *resource, wl_resource *surface) {}
+static void handle_gtk_shell_system_bell(wl_client *client, wl_resource *resource, wl_resource *surface)
+{
+    auto wl_surface = static_cast<wl_resource*> (wl_resource_get_user_data(surface));
```
kode54, line 152: This entire function needs a null pointer guard on the `surface` parameter, and should somehow handle a case for firing a bell signal to the shell itself, or something else. I fixed a crash simply adding a null guard to my local copy, but that doesn't send a bell signal anywhere in that case.

ammen99: seeing this, maybe it makes more sense to just emit the signal on core?

ammen99:
You said you'll send separate PRs for these changes? Maybe we can close this one.

Issue #500, opened 2020-05-03
Artifac when using Drag and drop


bluebyt65:
Whenever using Drag and drop artifact on the screen appeared.
- Drag and drop file from nautilus to Totem, bookmarks etc..
-Also Drag and and drop file from Archive manager to nautilus is not working.

 Video link
https://youtu.be/EPEDOCYQvJw

PR #555, opened 2020-06-09
Add scale plugin

### LINKED ISSUE #449: Present windows
`wayfire` currently has no way to show in one view all windows present on a virtual desktop. The fastest one can get to another window is to successively alt-Tab through stuff. This is hugely unergonomic. 

One way to deal with this is with a panel / dock. The panel in `wf-shell` does not do this. I have not seen / been able to make the dock work. 

But we could (I think should) have a way to do this is like a 'present windows' (as in Plasma) / 'expose' (as in Mac) / or best of all, 'activities overview' (as in GNOME, which would also integrate with the launcher, like dmenu or whatever). 

There already exists something like this with `expo` for virtual desktops. Could we have a similar plugin for windows?

soreau:
Fixes #449.


damianatorrpm:
Two things I noticed:
Like you mentioned on irc the position is wrong if a view is opened while the plugin is active
Un-minimizing a view while plugin is active causes a crash.

damianatorrpm:
Probably the right place to ask:
What I want to achieve is to be able to present() view based on some criteria (e.g all matching an app_id or by view-id vector), or present() everything from all outputs (what gnome-shell does).

@ammen99 said that the first one may not be a good idea for core,
but that this can be achieved using a signal like
view-present-signal (which has a data member, a list/vector of views to present)

My question here to @ammen99 @soreau is whether or not if I PR enhancements for all of 
the above this will be accepted by you two into scale plugin, since this is a plugin not core,
or if I should do it elsewhere?

ammen99:
@damianatorrpm I do not think this belongs to the core plugins. Even if we implemented such functionality, it cannot be tested except with your plugin.

What would be nice is to make the logic reusable. For example, plugin-wall: it has the rendering logic about rendering multiple workspaces, and then you have vswitch, vswipe and expo which are different controllers on top of this. So maybe @soreau can come up with some class which implements the logic of scale, then you(@damianatorrpm) could create a plugin based on this helper class.

damianatorrpm:
While I don't understand what the helper class should do,
it would be the easiest approach for me to copy the scale plugin 
and make it do what I want. This would have the side effect that I am unlikely
to contribute to scale plugin. You say only my plugin could test this, 
what about  wm-rules? Maybe a user want to present stuff.
(not that I am willing to implement this in other plugins, though)
:)

soreau:
> Probably the right place to ask:
> What I want to achieve is to be able to present() view based on some criteria (e.g all matching an app_id or by view-id vector), or present() everything from all outputs (what gnome-shell does).
> 
> @ammen99 said that the first one may not be a good idea for core,
> but that this can be achieved using a signal like
> view-present-signal (which has a data member, a list/vector of views to present)
> 
> My question here to @ammen99 @soreau is whether or not if I PR enhancements for all of
> the above this will be accepted by you two into scale plugin, since this is a plugin not core,
> or if I should do it elsewhere?

I think it would be feasible to add a signal that scale handles, which anything can call, with a view list (pointer? passing null could assume all workspace layer views) to be scaled. This wouldn't be much different than how idle plugin calls into cube for the rotating cube screensaver. See [this](https://github.com/WayfireWM/wayfire/blob/master/plugins/cube/cube-control-signal.hpp), [this](https://github.com/WayfireWM/wayfire/blob/master/plugins/single_plugins/idle.cpp#L297) and [this](https://github.com/WayfireWM/wayfire/blob/master/plugins/cube/cube.cpp#L202) for reference.

damianatorrpm:
I still got it to crash:
```
Jun 15 11:43:49 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:49.970 - [src/main.cpp:225] Fatal error: Segmentation fault
Jun 15 11:43:49 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:49.995 - #1  signal_handler(int) ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.013 - #2  __restore_rt ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.023 - #3  wayfire_scale::transform_views() ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.035 - #4  wayfire_scale::toggle_cb::{lambda(wf::activator_source_t, unsigned int)#1}::operator()(wf, unsigned int) const [clone .constprop.0] ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.049 - #5  std::_Function_handler<bool (), input_manager::match_keys(unsigned int, unsigned int, unsigned int)::{lambda()#2}>::_M_invoke(std::_Any_data const&) ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.060 - #6  input_manager::handle_keyboard_key(unsigned int, unsigned int) ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.072 - #7  std::_Function_handler<void (void*), wf_keyboard::setup_listeners()::{lambda(void*)#1}>::_M_invoke(std::_Any_data const&, void*&&) ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.083 - #8  handle_wrapped_listener(wl_listener*, void*) ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.093 - #9  wlr_signal_emit_safe ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.104 - #10 wlr_keyboard_notify_key ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.127 - #11 handle_libinput_readable ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.136 - #12 wl_event_loop_dispatch ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.145 - #13 wl_display_run ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.156 - #14 main ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.169 - #15 __libc_start_main ??:?
Jun 15 11:43:50 localhost.localdomain wayfire[29762]: EE 15-06-20 11:43:50.180 - #16 _start ??:?
```

what I do is open terminal, open wcm, sleep 5 && gnome-clocks; trigger scale, select terminal, ctrl-c, trigger scale again. If it didn't crash I'd try gnome-clocks & and than sleep 5 && killall gnome-clocks while scale is active

I think it should cancel scale and revert to previous active view if for example wf-panel is clicked.

soreau:
> I still got it to crash:
> 
> what I do is open terminal, open wcm, sleep 5 && gnome-clocks; trigger scale, select terminal, ctrl-c, trigger scale again. If it didn't crash I'd try gnome-clocks & and than sleep 5 && killall gnome-clocks while scale is active

I was able to reproduce this and pushed a fix for it.

So far as the other features you mentioned, they could be added, but the biggest priority for now is that it doesn't cause any crashes. It would help if you were able to describe these features in more detail. For one, I think that canceling scale when clicking the panel or desktop background should only happen if interact is not enabled. When you speak of gnome, that doesn't help me because I'm not familiar with gnome at all.

Thanks for testing.

soreau:
@ammen99 @dkondor Can you possibly test latest changes and see if things work well enough for you? I think I've ironed out most of the corner cases by now and I'm casually optimistic about an API where another plugin can scale a certain group of views.

SnooSnoo-ui:
Excuse my intrusion but I have tested it too. It was very smooth and worked very well! So I had to find a way to break it. While the scale plugin view was open I called things like  `for i in {1..20}; do (sleep 0.1 && kcalc &); done`, open mpv, open flatpaks, `sleep 2; killall ...` in a background script. I made Wayfire crash a few times.

I'm new to this and I don't like posting logs publicly, let me know if you need more http://paste.debian.net/plainh/a18cace9

A small detail, when you select a window your mouse input works imediatelly and also keyboard input for the app itself, but wayfire keys, or maybe modifiers, like alt tab only works when the animation is finished. Not saying it's a bug but something to think about, it was with interact=false.

I hope it helps!

soreau:
@SnooSnoo-ui It's no intrusion, thanks for testing. I can't really understand what might be happening from your backtrace but maybe @ammen99 can spot something.

soreau:
If you have a reliable way to reproduce a crash, please post it.

dkondor:
Hi,

@soreau thank you for the updates, I just tested the newest version and it works fine for me.

I've tried reproducing the problem by @SnooSnoo-ui, but I don't have the KDE apps and I did not see any crash with any other apps I tried (e.g. opening and closing 20 gnome-calculators works without problems for me). My guess based on the stacktrace is a view might be removed / destroyed without being removed from the scale_data map, but I have no idea how that could happen. @SnooSnoo-ui , if you can reproduce this crash reliably (or at least relatively easily even if it is random), it would be great if you could provide more info how it happens.

@soreau now I realized what I would really like is to have two separate key combinations to start the scale plugin either only for the current workspace or for all workspaces. I understand this is somewhat different approach to having a bool setting for this; I'm mainly thinking it would be great if there was a way to enable / disable keybindings, since there are already a lot of keybindings from all the plugins and a use might not need all of them. This way, a user might have only one of the bindings for scale set or both. I feel this would require a lot of changes to the code.

SnooSnoo-ui:
I can crash by running a youtube video in the io.github.celluloid_player.Celluloid flatpak from flathub. Activate Scale and selected the video. When I quit the video Wayfire crashes.

https://paste.debian.net/plainh/801d9498

The other crash I suspect is just a Qt issue that crashes KDE in Wayland too so I ignored it but if you want to try.. Open Dolphin and Konsole. Start a file drag operation in Dolphin, don't release it, open the Scale plugin, drop the file to Konsole. The dragging also shows artifacts when you move it around.

soreau:
> I can crash by running a youtube video in the io.github.celluloid_player.Celluloid flatpak from flathub. Activate Scale and selected the video. When I quit the video Wayfire crashes.

Do you have interact option enabled? How are you quitting the video exactly? (Esc?)

dkondor:
Hi @SnooSnoo-ui 

I tried and cannot reproduce this so far (with celluloid installed with flatpak on Ubuntu 18.04). I have three questions regarding this:
 - Are you running a debug or release build of wayfire? (you can check with 'meson configure', it should give the 'buildtype' variable)
 - Are you running a separate session or running wayfire inside another (X or Wayland) session?
 - Which distribution / release are you using?

Thanks!

dkondor:
> I'm mainly thinking it would be great if there was a way to enable / disable keybindings

I just realized there is a way just there is not a GUI for it, i.e. in the ini file or in wcm, I can manually set any of the keybindings to 'disabled' :)

SnooSnoo-ui:
>Do you have interact option enabled? How are you quitting the video exactly? (Esc?)

No. Oh it's only when you use the app's quit keybind that it crashes. Q is the default for Celluloid.

>Are you running a debug or release build of wayfire? (you can check with 'meson configure', it should give the 'buildtype' variable)

`--buildtype=release -D use_system_wfconfig=disabled -D use_system_wlroots=disabled` but it always says debug.
I changed the default_options value to buildtype=release, confirmed with meson configure that is now `release` with a clean build, and it still crashes.

>Are you running a separate session or running wayfire inside another (X or Wayland) session?
TTY

Since my logs were useless maybe you can ignore and see if anyone more experienced gets the same crash. It could just be something wrong with my setup.

@soreau I'm curious, do you also get visual artifacts when dragging files on top of Scale?

soreau:
> Since my logs were useless maybe you can ignore and see if anyone more experienced gets the same crash. It could just be something wrong with my setup.

I don't think your logs are useless. It seems that wayfire is actually crashing in safe_list when it shouldn't be, so this might be a wayfire core bug.

> @soreau I'm curious, do you also get visual artifacts when dragging files on top of Scale?

I forgot to mention, this is a known, unrelated issue. See i.e. #500.

soreau:
> A small detail, when you select a window your mouse input works imediatelly and also keyboard input for the app itself, but wayfire keys, or maybe modifiers, like alt tab only works when the animation is finished. Not saying it's a bug but something to think about, it was with interact=false.

Alt+tab switcher should work during end animation now.

soreau:
I seem to have found a reliable way to crash it. Enable animate and scale, open two terminals and in one, run `sleep 2; pkill $terminal` and then activate scale. When the close animation begins for the terminals, deactivate scale with keybinding and wayfire will soon after crash. I was able to fix it here..

Does [this patch](http://ix.io/2qZh) fix the crashes related to closing views?

SnooSnoo-ui:
>Does this patch fix the crashes related to closing views?

Yes! Before the patch it was crashing reproducibly. Now even inside a for loop and with multiple apps it's not crashing anymore.

plugins/single_plugins/scale.cpp
```diff
@@ -0,0 +1,856 @@
...
+        auto view = get_signaled_view(data);
+
+        if (output->workspace->get_view_layer(view) != wf::LAYER_WORKSPACE)
+        {
+            return;
+        }
+
+        if (view->get_transformer(transformer_name))
+        {
+            goto end;
```
ammen99, line 608: -1 for goto

dkondor:
Hi,
I've tested the latest version with the additional commits, and it works well, thanks @soreau for all the work you are putting in this!
I've created several pull requests with small changes in functionality and style -- let me know if those look reasonable :)

Also, I've noticed a strange bug or weirdness: scale does not work for modal dialogs, e.g. the confirmation dialog from gedit when trying to close it with unsaved changes -- see the attached screenshot.
Screenshot from 2020-07-09 01-29-24

SnooSnoo-ui:
I can't test #578 right now (at work) but wouldn't it be better if windows had the same container size even if transparent? When you have 5 open windows the bottom two are larger.

>duration = 1
>inactive_alpha = 0.1

Should duration affect alpha too? _inactive_alpha_duration_?

If anyone needs anything tested you can ping me. For any feature or problem, not only this issue.

dkondor:
I did some investigation with the problem with the unsaved changes popup from gedit. I found that it's technically not a popup, at least according to the following messages:
II 09-07-20 03:12:09.825 - [src/view/xdg-shell.cpp:209] new xdg_shell_stable surface: gedit app-id: gedit
II 09-07-20 03:12:12.904 - [src/view/xdg-shell.cpp:209] new xdg_shell_stable surface:  app-id: gedit
the first one is the normal gedit window, the second one is the popup.

Despite this, it is a view that does not exist in any of the layers, i.e.
output->workspace->get_views_in_layer(wf::ALL_LAYERS);
will not return it (this is why it is ignored by the scale plugin), and calling
output->workspace->get_view_layer(view)
on it returns zero.

On the other hand, calling
wf::get_core().get_all_views()
returns it, and this way, even the scale transform can be applied to it.

I've made a hack-ish solution here:
https://github.com/dkondor/wayfire/tree/scale_gedit_popup
that works at least when gedit is on the same workspace or the all_workspaces option is enabled. This still has problems: (1) the view_focused() signal gets into an endless call loop resulting in stack overflow, so I had to remove functionality from it; (2) the popup window seems to be present in all workspaces in an "invisible" way, but clicking on it can result in a crash (it turns invisible even on its own workspace if I switch away -- this can be seen in the workspace switcher as well); (3) selecting the popup will not trigger a workspace switch even if it is needed (if all_workspaces == true)

I think a better solution would be to "hide" these views in some way. I'll see if I can come up with a solution for this option instead. I wonder if any other types of views might end up in this state.

Or alternatively, the scale plugin could take over the whole rendering and then only render the views it knows about (similarly to the switcher plugin); but this would make it more complex?

I find the whole behavior a bit strange and wondering if it is intended, but quirky behavior or there might be a problem in handling these views in wayfire? Is there maybe something I'm missing here? I was trying to look at the view and workspace APIs, but I'm not sure I understand everything, especially when it comes to the layers.

soreau:
> Hi, I've tested the latest version with the additional commits, and it works well, thanks @soreau for all the work you are putting in this! I've created several pull requests with small changes in functionality and style -- let me know if those look reasonable :)

I looked over your commits and the ideas seem ok but there are several style and whitespace issues. I will review them on the respective PR's later.

> Also, I've noticed a strange bug or weirdness: scale does not work for modal dialogs, e.g. the confirmation dialog from gedit when trying to close it with unsaved changes -- see the attached screenshot.

I was discussing this with @ammen99 on IRC and I think what we need to do is in view_attached, first check if view->parent is in the scale_data map and if so, call layout_slots. There, we should also transform any view children to be centered with the parent always.

soreau:
Something like [this](http://ix.io/2req) could work for dialogs. Apparently a view can also change dialog state from/to normal (not having a parent) which would mean handling view layer attach/detach signals but this patch doesn't do that and I don't know of any clients that do this yet.

damianatorrpm:
I noticed https://github.com/WayfireWM/wayfire/pull/555#issuecomment-655657140 as well. 
For example the properties window of nautilus (right click file => properties).

Also I think that modal dialogs should not be shown separately from the main window
(the move plugin has the option 'Disallow independently moving dialogues')

It is not possible that the main window gets focues, hence the modal dialog should always be rendered on top.

dkondor:
Hi,
@soreau thanks for the quick response and patch; I wasn't aware how the parent - child relations add this complications. In compiz, such child windows are displayed independently, which I always thought was a bit weird. I'm trying the patch you sent and I like the way it works; so far, all dialogs I tried are displayed properly with their parent.

Some of the things I noticed further:
 - clicking on a dialog does not work yet, since the child views are not in any layer; [here](https://www.dropbox.com/s/kgwkr0549wbbmrb/scale_child_select.diff?dl=0) is a patch that does that (on top of your last patch), but maybe in a very simplistic way (by only excluding based on role); a better option could be checking the parent; also, there could be a crash when selecting a child view in the focus event handler, so that needs to handle this case separately
 - with my patch, workspace switching does not work if selecting a child view; I assume this need to be handled based on the parent, but I'm unsure if there is a legitimate case when the child will be on a different workspace? It seems to me that child views are strictly constrained to the workspace of the parent.
 - when switching workspaces child views disappear temporarily (e.g. if switching to a view that is on a different workspace by clicking on the parent view)
 - sometimes a child view becomes (relatively) too large; I think it may be nicer if the scale of child views was required to be always less than that of parent views
 - menus (and I assume other xdg-popups) still show up without being scaled; I think that if scale is run in non-interactive mode, it could be OK (and the simplest solution) to close popups (or at least menus); on the other hand, when in interactive mode, the user might expect menus to work (and possibly to be scaled according to their parent); I'm attaching a screenshot of this behavior

image

soreau:
Actually, I had only tested with dconf-editor. [This patch](http://ix.io/2req) should fix gedit save dialog and I hope others.

@dkondor @damianatorrpm If you want to test and voice any issues, otherwise I'll push in a couple days.

dkondor:
It seems to work well now, with the exception of switching workspaces if the dialog is clicked. Selecting the parent in select_view() fixes that for me:

256a257
>         while (view->parent) view = view->parent;

soreau:
> It seems to work well now, with the exception of switching workspaces if the dialog is clicked. Selecting the parent in select_view() fixes that for me:
> 
> 256a257
> 
> > ```
> >     while (view->parent) view = view->parent;
> > ```

Nice find, this actually helped me fix another bug that I was chasing. [Here's the latest iteration](http://ix.io/2req).

dkondor:
Just tested this, I can confirm it works well with clicking on the dialog.

soreau:
Wayfire crashed with [this](https://hastebin.com/ewilinuziv.cpp) when I middle-click-to-close a scaled view.

kode54:
Middle-click-to-close? Is that a non-default binding for something?

soreau:
> Middle-click-to-close? Is that a non-default binding for something?

Yes, see https://github.com/WayfireWM/wayfire/pull/555/commits/0abf22439864b3cc65dd3888e45bdf5d0ef140de

kode54:
Oh, I didn't realize you did more than rebase it.

soreau:
> Oh, I didn't realize you did more than rebase it.

The commit was there before the recent rebase, see #580.

ammen99:
Does #673 help?

soreau:
> Does #673 help?

Actually, I haven't found a way to reproduce it reliably yet.

ammen99:
> > Does #673 help?
> 
> Actually, I haven't found a way to reproduce it reliably yet.

The said PR was merged, I think it should help with that particular crash.

In general, I'll be reviewing this next week. Do you really want to keep all 47 commits separated? (I'd prefer fewer but if it is difficult to squash because relevant changes are across non-sequential commits it's fine to keep it as is).

soreau:
> > > Does #673 help?
> > 
> > 
> > Actually, I haven't found a way to reproduce it reliably yet.
> 
> The said PR was merged, I think it should help with that particular crash.

I think it did.

> In general, I'll be reviewing this next week. Do you really want to keep all 47 commits separated? (I'd prefer fewer but if it is difficult to squash because relevant changes are across non-sequential commits it's fine to keep it as is).

I don't have objections to squashing.

plugins/single_plugins/scale.cpp
```diff
@@ -0,0 +1,1403 @@
...
+        layout_slots(get_views());
+
+        if (interact)
+        {
+            connect_button_signal();
+        }
+
+        output->connect_signal("view-layer-attached", &view_attached);
+        output->connect_signal("view-attached", &view_attached);
+        view_detached.disconnect();
```
ammen99, line 1298: disconnect here?

plugins/single_plugins/scale.cpp
```diff
@@ -0,0 +1,1403 @@
...
+        output->connect_signal("view-minimized", &view_minimized);
+        output->connect_signal("view-unmapped", &view_unmapped);
+        output->connect_signal("view-focused", &view_focused);
+
+        view_geometry_changed.disconnect();
+        for (auto& e : scale_data)
+        {
+            auto view = e.first;
+            view->connect_signal("geometry-changed", &view_geometry_changed);
+            if ((view == initial_focus_view) || (view->parent == initial_focus_view))
```
ammen99, line 1310: seeing this again here, maybe it is worth adding helper functions `get_topmost_view()` because there may be nested dialogues ...

plugins/single_plugins/vswitch.cpp
```diff
@@ -0,0 +1,348 @@
+#include <wayfire/plugin.hpp>
```
ammen99, line 1: This file should go eventually before merging :)

plugins/single_plugins/meson.build
```diff
@@ -13,4 +13,4 @@ foreach plugin : plugins
       dependencies: all_deps,
       install: true,
       install_dir: conf_data.get('PLUGIN_PATH'))
-endforeach
+endforeach
```
ammen99, line 16: I prefer to have a trailing newline

plugins/single_plugins/scale.cpp
```diff
@@ -0,0 +1,1277 @@
...
+        if (col >= current_row_cols)
+        {
+            col = 0;
+        }
+
+        view = find_view_in_grid(row, col);
+        if (view && (current_focus_view != view))
+        {
+            // view_focused handler will update the view state
+            output->focus_view(view, true);
```
ammen99, line 630: Here, do we really want to change the stacking order? (which is what the second parameter indicates) I'd think we want to not change the stacking order except when exiting (ENTER or ESC)

soreau: Make a PR?

ammen99 (approved):

ammen99:
Thanks to all of you who have tested and especially to @soreau @dkondor!

PR #558, opened 2020-06-13
Scale

dkondor:
Hi,

I've recently found wayfire and I think it is a great project! (As a long-time compiz user, I'm happy to see it as a potential replacement for wayland)

One of the things I found lacking is a scale / present windows functionality, so I thought I give it a try to implement this.

I just realized I'm duplicating effort here:
https://github.com/WayfireWM/wayfire/pull/555

Nevertheless, things you might find useful from here:
 - show windows from all workspaces and switch to the workspace of the window selected
 - use the arrow keys to select windows (enter to switch, esc to cancel)


ammen99:
Thanks for the PR! It seems interesting, I'll look into both PRs after next week. We determined that there are indeed a ton of different ways to change the scale plugin so the idea is to create sort of scale plugin "base" class which can then be extended to provide various capabilities (because obviously there are a ton of options, etc.). So it is good having more variants, this way I can create a more feature-complete and flexible base, mixing different ideas from the PRs :)

soreau:
@dkondor Thanks for your work on this! As @ammen99 mentioned, the more ideas, the better. Also, code contributions are quite welcome. It sounds like the scale plugin will be pretty robust, once it's complete.

soreau:
@dkondor FYI, #560 was merged, which provides an easier way to get a view from global coordinates.

soreau:
@dkondor I've updated #555 with the features you introduced here. If you try it, let me know how it works for you.

dkondor:
Hi, @soreau thanks!
So far the switcher works fine, but in some cases, it causes a crash. I'll investigate a bit more. Also, I believe using the workspace switcher animation should work (I saw the TODO note for that), I'll create a pull request for that separately.

soreau:
> Hi, @soreau thanks!
> So far the switcher works fine, but in some cases, it causes a crash. I'll investigate a bit more.

What do you mean by 'the switcher'? Do you mean scale in general, or something else? I have updated #555 with a crash fix related to using scale after closing a view while scale was active. If you are still able to reproduce, please post the backtrace from wayfire log there.

> Also, I believe using the workspace switcher animation should work (I saw the TODO note for that), I'll create a pull request for that separately.

Awesome, thanks!

dkondor:
Sorry, by switcher, I meant "scale switcher" :)

I've added a small PR for workspace switch animation.

Also, the crash is fixed in the updated version.

PR #559, opened 2020-06-13
small fixes for vswitch

dkondor:
Two small issues I noticed with the vswitch plugin while working on https://github.com/dkondor/wayfire/tree/scale
 - need wf::CAPABILITY_CUSTOM_RENDERER
 - set-workspace-request signal should only set carried_out == true if workspace switch was successful

Let me know if I should create an issue for these or a minimum example of the issue I had.

### DIFF
--- a/plugins/single_plugins/vswitch.cpp
+++ b/plugins/single_plugins/vswitch.cpp
@@ -49,7 +49,10 @@ class vswitch : public wf::plugin_interface_t
     void init()
     {
         grab_interface->name = "vswitch";
-        grab_interface->capabilities = wf::CAPABILITY_MANAGE_DESKTOP;
+        /* note: workspace_wall_t sets a custom renderer, so we
+         * need the capability for that */
+        grab_interface->capabilities = wf::CAPABILITY_MANAGE_DESKTOP |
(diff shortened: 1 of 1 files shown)


ammen99:
And btw thanks :)

dkondor:
Hi,

my first attempt at scale had a custom renderer (https://github.com/dkondor/wayfire/blob/6c74e083dcd81d60b41b313f94eb2c6ef12899bb/plugins/single_plugins/scale.cpp -- adapted from the switcher plugin).
Calling workspace->request_workspace() resulted in messed-up result since the vswtich plugin replaced the renderer (and some of the logic for scale was in the renderer, so it never had the chance to deactivate).

Adding wf::CAPABILITY_CUSTOM_RENDERER then resulted in workspace->request_workspace() silently failing, since vswitch set carried_out == true, even if it could not do the switch (since now the capabilities conflicted with the running scale plugin).

In the end, I realized that scale does not need a custom renderer, it is enough to set transformers on the views, so this is not relevant in this case.

BTW, based on this, it seems that output->render->set_renderer() does not check if the plugin calling it has wf::CAPABILITY_CUSTOM_RENDERER -- it might be good to do this check if technically possible?

PR #560, opened 2020-06-13
core: Add get_view_at function to API

soreau:
Create and expose a new core function get_view_at, which takes a point and
returns a view, or null if none is found. This is useful for plugins like
scale where it makes sense to be able to pick views while holding a grab.

### DIFF
--- a/src/api/wayfire/core.hpp
+++ b/src/api/wayfire/core.hpp
@@ -155,10 +155,17 @@ class compositor_core_t : public wf::object_base_t
      */
     virtual wf::surface_interface_t *get_touch_focus() = 0;
 
+    /**
+     * @return The surface at point, or null if none.
+     */
+    virtual wf::surface_interface_t *get_surface_at(wf::pointf_t point) = 0;
+
     /** @return The view whose surface is cursor focus */
     wayfire_view get_cursor_focus_view();
     /** @return The view whose surface is touch focus */
     wayfire_view get_touch_focus_view();
+    /** @return The view whose surface is at point */
+    wayfire_view get_view_at(wf::pointf_t point);
 
     /**
      * @return A list of all currently attached input devices.
--- a/src/api/wayfire/core.hpp
+++ b/src/api/wayfire/core.hpp
@@ -164,7 +164,8 @@ class compositor_core_t : public wf::object_base_t
     wayfire_view get_cursor_focus_view();
     /** @return The view whose surface is touch focus */
     wayfire_view get_touch_focus_view();
-    /** @return The view whose surface is under the give global coordinates,
+    /**
+     * @return The view whose surface is under the given global coordinates,
      *  or null if none */
     wayfire_view get_view_at(wf::pointf_t point);
 
--- a/src/api/wayfire/core.hpp
+++ b/src/api/wayfire/core.hpp
@@ -156,7 +156,7 @@ class compositor_core_t : public wf::object_base_t
     virtual wf::surface_interface_t *get_touch_focus() = 0;
 
     /**
-     * @return The surface at point, or null if none.
+     * @return The surface under the given global coordinates, or null if none.
      */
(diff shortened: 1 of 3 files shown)


src/core/core.cpp
```diff
@@ -327,6 +327,20 @@ wayfire_view wf::compositor_core_t::get_cursor_focus_view()
...
+wf::surface_interface_t *wf::compositor_core_impl_t::get_surface_at(wf::pointf_t point)
+{
+    wf::pointf_t local = {0.0, 0.0};
+    return input->input_surface_at(point, local);
+}
+
+wayfire_view wf::compositor_core_t::get_view_at(wf::pointf_t point)
+{
+    auto surface = get_surface_at(point);
+    auto view = dynamic_cast<wf::view_interface_t*> (surface);
```
ammen99, line 339: this is wrong. You should try to cast `surface->get_main_surface()`, otherwise, this will fail if you are over a subsurface.

src/api/wayfire/core.hpp
```diff
@@ -155,10 +155,17 @@ class compositor_core_t : public wf::object_base_t
      */
     virtual wf::surface_interface_t *get_touch_focus() = 0;
 
+    /**
+     * @return The surface at point, or null if none.
```
ammen99, line 159: Both functions are missing comments about the coordinate systems of their parameters (output-local, output-layout ..)

src/api/wayfire/core.hpp
```diff
@@ -155,10 +155,18 @@ class compositor_core_t : public wf::object_base_t
...
+    /**
+     * @return The surface under the given global coordinates, or null if none.
+     */
+    virtual wf::surface_interface_t *get_surface_at(wf::pointf_t point) = 0;
+
     /** @return The view whose surface is cursor focus */
     wayfire_view get_cursor_focus_view();
     /** @return The view whose surface is touch focus */
     wayfire_view get_touch_focus_view();
+    /** @return The view whose surface is under the give global coordinates,
```
ammen99, line 167:
```suggestion
    /**
     * @return The view whose surface is under the given global coordinates,
```

src/core/core.cpp
```diff
@@ -327,6 +327,20 @@ wayfire_view wf::compositor_core_t::get_cursor_focus_view()
...
+wf::surface_interface_t *wf::compositor_core_impl_t::get_surface_at(wf::pointf_t point)
+{
+    wf::pointf_t local = {0.0, 0.0};
+    return input->input_surface_at(point, local);
+}
+
+wayfire_view wf::compositor_core_t::get_view_at(wf::pointf_t point)
+{
+    auto surface = get_surface_at(point);
+    auto view = dynamic_cast<wf::view_interface_t*> (surface->get_main_surface());
```
ammen99, line 339: but what if surface itself is null? xD

soreau: Hm, good point..

src/core/core.cpp
```diff
@@ -327,6 +327,26 @@ wayfire_view wf::compositor_core_t::get_cursor_focus_view()
...
+wf::surface_interface_t *wf::compositor_core_impl_t::get_surface_at(wf::pointf_t point)
+{
+    wf::pointf_t local = {0.0, 0.0};
+    return input->input_surface_at(point, local);
+}
+
+wayfire_view wf::compositor_core_t::get_view_at(wf::pointf_t point)
+{
+    auto surface = get_surface_at(point);
+
```
ammen99, line 339:
```suggestion
```

src/core/core.cpp
```diff
@@ -327,6 +327,26 @@ wayfire_view wf::compositor_core_t::get_cursor_focus_view()
...
+    if (!surface)
+    {
+        return nullptr;
+    }
+
+    auto view = dynamic_cast<wf::view_interface_t*> (surface->get_main_surface());
+
```
ammen99, line 346:
```suggestion
```

ammen99:
Thanks!

PR #562, opened 2020-06-16
use request_workspace() to change workspaces

dkondor:
this will use vswitch plugin if enabled

note: this requires to change grab_interface->capabilities to only include wf::CAPABILITY_GRAB_INPUT, otherwise they would clash with vswitch; as far as I understand this is OK since only this is actually used

There is some strangeness in during the workspace switch animation: the windows seem to "jump" a bit and the inactive ones stay transparent and revert later (instead of fading); I'll try to look into this more

the other crash I mentioned yesterday is fixed since then (it was caused by remove_transformers() and scale_data.clear() being reversed when deactivating the plugin

### DIFF
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -73,7 +73,7 @@ class wayfire_scale : public wf::plugin_interface_t
     void init() override
     {
         grab_interface->name = "scale";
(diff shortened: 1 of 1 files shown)


soreau:
Thanks for this. I noticed the views were animating with the wrong position so I added a patch on top of this that fixes that amongst other various issues.

dkondor:
Thank, I confirm it now works great!

PR #564, opened 2020-06-17
No need for grid member variable anymore

dkondor:
I originally used it to keep track of views in the grid. Since find_view_in_grid() always goes through the list of views returned by get_views(), there is no need to keep this, we only need to save the number of rows, columns, and the number of views in the last row.

Alternatively, we could store the view pointers in the grid member for a very slight optimization of lookups. But I think removing it is more in line with the rest of the code.


soreau:
Thanks!

Issue #551, opened 2020-06-07
[Question] How to remove window decorations?


kanyck:
Fiddling with wayfire for some time and quite impressed. I've been a compiz user for over a decade and watching sadly how it was gradually dying... So I'm happy that wayfire shows up. 
However.
I want to remove title bar from native wayland windows (at least from Alacritty's, maybe from all). Because wayfire allows to move/resize windows without using title bar, it looks like it doesn't carry any useful functionality anymore and just wastes space (especially on a small screens). 
Switching off decoration module or setting title bar height to 0 doesn't do anything visible.
I even can't understand who is responsible for, say, Alacritty decorations, because setting them off in alacritty config file has no effect, either.

ammen99:
One hack you can try is setting the preferred decoration mode to server via https://github.com/WayfireWM/wayfire/wiki/Configuration#corepreferred_decoration_mode

This would make some of the apps switch to using wayfire's decorations, which are actually not there if you disable the decoration plugin.

This however won't work with GTK apps and possibly many others - you'll have to configure them via toolkit specific options. Gtk has some modules to disable headerbars, etc.

kanyck:
Well, it doesn't help with `alacritty`. Decorations are still there even when they're switched off on both sides. 
However, `wcm` crashes wayfire with `preferred_decoration_mode = server`. Whats even worse, the wayfire crash causes video driver to get stuck, so I may either use kernel magic Alt+SysRq+R... to reboot or ssh from another machine (to reboot, yes). I probably should file another bug about it. Please suggest what kind of debug info may I provide to you for investigation, given that screen is unusable after crash, and `tee`-ing debug log to a file will probably loose last, the most important part of it due to bufferization.

ammen99:
Does this happen if wayfire is running nested inside another compositor? If that is the case, then most probably your system won't crash if you can reproduce the crash there and you'll be able to generate a stack trace.

Also, if you have a bug which you can reliably reproduce, please compile wayfire with ASAN. While wayfire prints a stack trace by default, ASAN can detect subtler memory corruption issues so it is better for reports.

kanyck:
Okay, I'll try to investigate it further. (first I want to eventually finish building firefox. It's probably the 4-th start over). I already experienced wayfire random crashes a couple of times, that's probably the chance to catch up.

elenapan:
> Well, it doesn't help with `alacritty`. Decorations are still there even when they're switched off on both sides.

[...]

PR #568, opened 2020-06-23
Feature - Ignore window decoration for certain views matched by criteria

### LINKED ISSUE #563: Support for Floating windows in Window Rules plugin
Migrating to Wayfire from Sway, I was stuck on getting floating windows to work. I use them for minimal scripts, which uses fzf in a floating window. I thought there would be support in Wayfire but I haven't been able to find anything as such neither in wiki nor in window rules code.

[Here's the Archwiki](https://wiki.archlinux.org/index.php/Sway#Floating_windows) doc on Floating windows in Sway.
Can you please help me out if it already exists or if I can help contribute a fix for it?

Thanks for the awesome WM though! Loved the unique experience :smile:

shivanshs9:
Fixes #563 

- [x] Tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,65 @@
+
+# Created by https://www.toptal.com/developers/gitignore/api/c++,meson,ninja,linux
+# Edit at https://www.toptal.com/developers/gitignore?templates=c++,meson,ninja,linux
+
+### C++ ###
+# Prerequisites
+*.d
+
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+
+# Precompiled Headers
+*.gch
+*.pch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Fortran module files
+*.mod
+*.smod
+
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
+
+### Linux ###
+*~
+
+# temporary files which can be created if a process still has a handle open of a deleted file
+.fuse_hidden*
+
+# KDE directory preferences
+.directory
+
+# Linux trash folder which might appear on any partition or disk
+.Trash-*
+
+# .nfs files are created when an open file is removed but is still being accessed
+.nfs*
+
+### Meson ###
+# subproject directories
+/subprojects/*
+!/subprojects/*.wrap
+
+### Ninja ###
+.ninja_deps
+.ninja_log
+
+build/**
+
+# End of https://www.toptal.com/developers/gitignore/api/c++,meson,ninja,linux
--- a/metadata/decoration.xml
+++ b/metadata/decoration.xml
@@ -35,5 +35,10 @@
 			<_long>Sets the color when the window is inactive.</_long>
 			<default>#333333dd</default>
 		</option>
+		<option name="ignore_views" type="string">
+			<_short>Decoration disabled for specified window types</_short>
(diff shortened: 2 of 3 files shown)


ammen99:
Hi, thanks for the patch, but the correct place for checking whether to have a decoration or not is here: https://github.com/WayfireWM/wayfire/blob/master/plugins/decor/decoration.cpp#L41

`deco-theme`'s responsibility is how to draw the decorations, not whether they are drawn at all.

shivanshs9:
Is it possible to disable client-side decorations too?

shivanshs9:
BUMP.

> I also wanted to know how can I disable client-side decorations too? Since I haven't been able to achieve the same for Alacritty since it uses its own decorations.

@ammen99 any ideas for it?

.gitignore
```diff
@@ -0,0 +1,77 @@
...
+/subprojects/*
+!/subprojects/*.wrap
+
+### Ninja ###
+.ninja_deps
+.ninja_log
+
+### VisualStudioCode ###
+.vscode/*
+!.vscode/settings.json
```
ammen99, line 65: We do not use vscode, so I do not think we need these exceptions in the .gitignore ..

ammen99:
LGTM aside from the comment.



> @ammen99 any ideas for it?

Try setting https://github.com/WayfireWM/wayfire/wiki/Configuration#corepreferred_decoration_mode to `server`, and then tell `alacritty` to use SSD. Finally, add an exception for alacritty in the decoration plugin (via the option you have implemented in this PR).

ammen99:
Oh, I just remembered this comment here: https://github.com/WayfireWM/wayfire/issues/551#issuecomment-640328958, maybe it will be helpful too.

.gitignore
shivanshs9: Fixed.

shivanshs9:
> Try setting [WayfireWM/wayfire/wiki/Configuration#corepreferred_decoration_mode](https://github.com/WayfireWM/wayfire/wiki/Configuration#corepreferred_decoration_mode) to `server`, and then tell `alacritty` to use SSD. Finally, add an exception for alacritty in the decoration plugin (via the option you have implemented in this PR).

Ah, I've already been using it this way but I felt it's like a hack since it sets server side decoration globally, when the aim of this plugin was simply to disable window decoration of windows matched by criteria. But I understand if there's no other way to do it without setting it globally.

> Oh, I just remembered this comment here: #551 (comment), maybe it will be helpful too.

Yep I've been using the same config.

ammen99:
> But I understand if there's no other way to do it without setting it globally.

Theory says it is possible to extend the option in core to match preferred mode per view. However, this is complicated by the fact that xdg-decoration objects are per-surface and the initial negotiation happens before the toplevel is mapped, which means that at least initially we cannot use matcher to separate views from each other. One would have to wait until map event and change things then ...

ammen99:
Can you squash your 1th and 4th commits and the 2nd and 3rd?

shivanshs9:
> Can you squash your 1st and 4th commits and the 2nd and 3rd?

Yep done! :+1: 

> Theory says it is possible to extend the option in core to match preferred mode per view. However, this is complicated by the fact that xdg-decoration objects are per-surface and the initial negotiation happens before the toplevel is mapped, which means that at least initially we cannot use matcher to separate views from each other. One would have to wait until map event and change things then ...

Got it, I'll go with the hack for now then! :sweat_smile:

ammen99:
Thanks!

PR #574, opened 2020-07-01
wayfire.ini: Make wlogout easier to reach

alexherbo2:


ammen99:
I am not very sure about this. `wlogout` seems quite unlikely to be installed on a user's machine ... And we're changing defaults in some sense, which I don't like either.

alexherbo2:
The rationale for switching the two is [wlogout] is a menu and [swaylock] a direct access functionality.

[wlogout]: https://github.com/ArtsyMacaw/wlogout
[swaylock]: https://github.com/swaywm/swaylock

ammen99:
Alright, this is the default config, the user xan change it anyway ... Anyway, thanks!

PR #577, opened 2020-07-05
Add code style checks with uncrustify

ammen99:
Currently `uncrustify` 0.71 is needed. The CI itself uses my own fork because I am working on a few upstream fixes, after some time we should be able to switch to the next stable `uncrustify` release.

Note there have been a few small changes to the current code style, because the code style needs to be consistent :)


plugins/single_plugins/fisheye.cpp
```diff
@@ -10,7 +10,8 @@
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
  *
- * The above copyright notice and this permission notice shall be included in all
+ * The above copyright notice and this permission notice shall be included in
+ * all
```
soreau, line 14: This looks wrong..

plugins/single_plugins/fisheye.cpp
```diff
@@ -107,110 +108,120 @@ class wayfire_fisheye : public wf::plugin_interface_t
...
+        box = output->render->get_target_framebuffer().
+            framebuffer_box_from_geometry_box(box);
+        oc.x = box.x;
+        oc.y = box.y;
+
+        static const float vertexData[] = {
+            -1.0f, -1.0f,
+            1.0f, -1.0f,
+            1.0f, 1.0f,
+            -1.0f, 1.0f
```
soreau, line 178: Is there a way to align the `1`'s with each other?

plugins/single_plugins/fisheye.cpp
ammen99: it is in column 81 after all :)

But good point, this looks important enough to increase the comment width to 80.

plugins/single_plugins/fisheye.cpp
ammen99: Another good point, I initially didn't consider the alignment options, but there a few of them left which might be useful. I'll experiment and see whether I can get it to work sensibly

plugins/single_plugins/fisheye.cpp
ammen99: Well actually we can just rewrap the comment appropriately - the MIT license is originally wrapped at 80 columns, but because it is in a comment here we have a few additional spaces in front, so it should be wrapped even more.

plugins/single_plugins/fisheye.cpp
ammen99: I couldn't find a way to make this work without breaking a ton of other things ...

damianatorrpm: maybe a bad idea, but for fisheye prefix also with + (never tried this though).

PR #578, opened 2020-07-08
Scale zoom limit

dkondor:
It is a preference thing, but I don't like the behavior that some views can become zoomed in or scaled up if their original size was smaller than the allocated space. It is easiest to see if there is only one, relatively small view open (e.g. gnome-calculator).

This commit disallows this behavior. This could be easily made into an option, but I'm not sure that level of customization is necessary :)

### DIFF
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -39,5 +39,10 @@
 			<min>0.0</min>
 			<max>1.0</max>
 		</option>
+		<option name="allow_zoom" type="bool">
+			<_short>Allow Zoomed Windows</_short>
+			<_long>Whether scaled windows can have a size larger than their original size.</_long>
+			<default>false</default>
+		</option>
 	</plugin>
 </wayfire>
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -40,8 +40,8 @@
 			<max>1.0</max>
 		</option>
 		<option name="allow_zoom" type="bool">
-			<_short>Allow Zoomed Windows</_short>
-			<_long>Whether scaled windows can have a size larger than their original size.</_long>
+			<_short>Allow Zoomed Views</_short>
+			<_long>Whether scaled views can have a size larger than their original size.</_long>
 			<default>false</default>
 		</option>
 	</plugin>
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -311,7 +311,7 @@ class wayfire_scale : public wf::plugin_interface_t
             return;
         }
 
-        if (!scale_view(view))
+        if (!scale_view(view) && view->role != wf::VIEW_ROLE_TOPLEVEL)
         {
             return;
         }
@@ -371,11 +371,16 @@ class wayfire_scale : public wf::plugin_interface_t
         if (!view)
         {
             view = last_focused_view;
-            fade_in(view);
             fade_out_all_except(view);
+            fade_in(view);
             output->focus_view(view, true);
             return;
         }
+        if (!scale_view(view) && view->role != wf::VIEW_ROLE_TOPLEVEL)
+        {
+            return;
+        }
+
         int row = scale_data[view].row;
         int col = scale_data[view].col;
 
@@ -479,7 +484,7 @@ class wayfire_scale : public wf::plugin_interface_t
         for (auto& e : scale_data)
         {
             auto view = e.first;
-            if (!view || !scale_data[view].transformer)
+            if (!view || !scale_data[view].transformer || !scale_view(view))
             {
                 continue;
             }
@@ -491,6 +496,23 @@ class wayfire_scale : public wf::plugin_interface_t
             scale_data[view].transformer->alpha = scale_data[view].fade_animation;
 
             view->damage();
+
+            for (auto& child : view->children)
+            {
+                view = child;
+                if (!view || !scale_data[view].transformer)
+                {
+                    continue;
+                }
+
+                scale_data[view].transformer->scale_x = scale_data[view].animation.scale_animation.scale_x;
+                scale_data[view].transformer->scale_y = scale_data[view].animation.scale_animation.scale_y;
+                scale_data[view].transformer->translation_x = scale_data[view].animation.scale_animation.translation_x;
+                scale_data[view].transformer->translation_y = scale_data[view].animation.scale_animation.translation_y;
+                scale_data[view].transformer->alpha = scale_data[view].fade_animation;
+
+                view->damage();
+            }
         }
 
         output->render->damage_whole();
@@ -518,8 +540,17 @@ class wayfire_scale : public wf::plugin_interface_t
         {
             return false;
         }
-        return output->workspace->get_view_layer(view) == wf::LAYER_WORKSPACE ||
-            view->role == wf::VIEW_ROLE_TOPLEVEL;
+        bool found = false;
+        auto views = get_views();
+        for (auto& v : views)
+        {
+            if (v == view)
+            {
+                found = true;
+                break;
+            }
+        }
+        return found;
     }
 
     /* compiz scale plugin algorithm */
@@ -527,6 +558,10 @@ class wayfire_scale : public wf::plugin_interface_t
     {
         if (!views.size())
         {
+            if (!all_workspaces && active)
+            {
+                toggle_cb(wf::activator_source_t{}, 0);
+            }
             return;
         }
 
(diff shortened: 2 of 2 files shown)


plugins/single_plugins/scale.cpp
```diff
@@ -529,7 +532,9 @@ class wayfire_scale : public wf::plugin_interface_t
                 auto vg = view->get_wm_geometry();
 
                 double scale_x = width / vg.width;
+                if(scale_x > max_scale_factor) scale_x = max_scale_factor;
```
soreau, line 535: These conditionals need a space between the `if` and the first `(`, they need braces and the first brace should start on the line after the `if`. Also, both of these clauses should be placed directly above the `scale_x = scale_y = std::min(scale_x, scale_y);` line.

soreau:
This is a good idea I think but I don't understand why you have so many commits in this PR. Please rebase on the scale branch and make it so that there is only a single commit.

plugins/single_plugins/scale.cpp
```diff
@@ -612,7 +621,10 @@ class wayfire_scale : public wf::plugin_interface_t
 
             std::vector<size_t> row;
             x = workarea.x + (int) spacing;
-            width = (workarea.width - (n + 1) * (int) spacing) / n;
+            if (i == lines - 1)
+            {
+                x += (grid_cols - grid_last_row_cols) * width / 2.0;
```
soreau, line 626:
```suggestion
                width = (workarea.width - (n + 1) * (int) spacing) / n;
```

I'd rather leave this what it was for the last row.

plugins/single_plugins/scale.cpp
```diff
@@ -652,16 +664,23 @@ class wayfire_scale : public wf::plugin_interface_t
...
-                    scale_y = height / vg.height;
+                    double child_scale_x = width / vg.width;
+                    double child_scale_y = height / vg.height;
+                    child_scale_x = child_scale_y = std::min(
+                        std::min(scale_x, scale_y), max_scale_factor);
+                    if (max_scale_child > 0.0 && child_scale_x > max_scale_child * scale_x)
+                    {
+                        child_scale_x = max_scale_child * scale_x;
+                        child_scale_y = child_scale_y;
+                    }
```
soreau, line 675: `child_scale_y = child_scale_y;` is definitely wrong. If we do limit scaling of children, we should just limit the max to 1.0. I see no reason for this added complexity.

However, I don't think we should limit children from scaling larger. Consider the case where the parent is scaled smaller but the child is smaller than the parent. I would think it would make the most sense to allow the child to become as big as possible so the user has the best chance of seeing it well. This is scale after all, not only-scale-down.

soreau:
On second thought, we probably should adjust child scale too and I think your code probably does the right thing (that being making the parent and child scale consistent while not allowing > 1.0 scale).

plugins/single_plugins/scale.cpp
dkondor: Yes, this is a typo, I'm adding a fix.

dkondor: I wonder how I did not see this as an issue before

soreau: It's probably because you never hit a case where it mattered.

The more I think about it, the more I think we should just calculate parent scale, clamp it to 1.0, and use the same scale for the child. Would this work?

plugins/single_plugins/scale.cpp
dkondor: This is in response to SnooSnoo-ui's comment about windows in the last row being bigger. But this can be a separate issue for discussion, so I'm reverting to the previous behavior for now.

plugins/single_plugins/scale.cpp
```diff
@@ -83,6 +83,14 @@ class wayfire_scale : public wf::plugin_interface_t
     wf::option_wrapper_t<bool> interact{"scale/interact"};
     wf::option_wrapper_t<bool> all_workspaces{"scale/all_workspaces"};
     wf::option_wrapper_t<double> inactive_alpha{"scale/inactive_alpha"};
+    
+    /* maximum scale -- 1.0 means we will not "zoom in" on a view
+     * (this could be set as an option) */
```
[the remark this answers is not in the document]
dkondor: I was thinking there will be too many options already :) But it's easy to add and there is still space for it.

plugins/single_plugins/scale.cpp
dkondor: I think we need to handle the case when the child is bigger than the parent. In that case, the child must have a smaller scale than parent, otherwise it would not fit in the space allocated for them, so that's why I'm calculating a scale for the child first and then limiting it based on the parent.

E.g. with gedit, in the unlikely case that the main window is resized so small that the "Close without saving" dialog is larger.

soreau: Good point. Carry on then :+1:

dkondor:
I've added a commit that makes both parent and child zoom a config option.

For other options there is the zoom desktop plugin (although in some cases, using scale to zoom in one window can be more convenient). Also, compiz scale has an option where right clicking a window zooms in on it, while hiding all others, so the user can better see what's in there. I'm considering trying to add that, but I'm unsure what is the best way to hide the other views and if it would require too big changes to the current logic.

soreau:
> I've added a commit that makes both parent and child zoom a config option.
> 
> For other options there is the zoom desktop plugin (although in some cases, using scale to zoom in one window can be more convenient). Also, compiz scale has an option where right clicking a window zooms in on it, while hiding all others, so the user can better see what's in there. I'm considering trying to add that, but I'm unsure what is the best way to hide the other views and if it would require too big changes to the current logic.

I think the one option you've added is enough for now. It might be relatively easy to add such a right-click option but it should go in a separate PR of course.

dkondor:
Ok, great, I'll see if I can do something about the right click zoom and create a PR for that

soreau:
@dkondor In case you didn't notice, this PR needs to be rebased since your other PR has landed. Also you can click the 'Resolve conversation' buttons for things you've already fixed.

plugins/single_plugins/scale.cpp
dkondor: changed the formatting and the logic slightly, so there is only one check

plugins/single_plugins/scale.cpp
```diff
@@ -83,6 +83,14 @@ class wayfire_scale : public wf::plugin_interface_t
...
+    const double max_scale_factor = 1.0;
+    /* maximum scale for child views (relative to their parents)
+     * zero means unconstrained, 1.0 means child cannot be scaled
+     * "larger" than the parent */
+    const double max_scale_child = 1.0;
```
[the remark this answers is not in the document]
dkondor: Now child is either limited like this, or not limited at all (dependent on the option); let me know i this is OK

plugins/single_plugins/scale.cpp
```diff
@@ -647,16 +657,16 @@ class wayfire_scale : public wf::plugin_interface_t
 
         y = workarea.y + (int) spacing;
         height = (workarea.height - (lines + 1) * (int) spacing) / lines;
-
+
```
soreau, line 660:
```suggestion

```

plugins/single_plugins/scale.cpp
```diff
@@ -82,6 +82,14 @@ class wayfire_scale : public wf::plugin_interface_t
     wf::option_wrapper_t<int> spacing{"scale/spacing"};
     wf::option_wrapper_t<bool> interact{"scale/interact"};
     wf::option_wrapper_t<double> inactive_alpha{"scale/inactive_alpha"};
+    wf::option_wrapper_t<bool> allow_scale_zoom{"scale/allow_zoom"};
+    
+    /* maximum scale -- 1.0 means we will not "zoom in" on a view
```
soreau, line 87:
```suggestion
    /* maximum scale -- 1.0 means we will not "zoom in" on a view */
```

metadata/scale.xml
```diff
@@ -39,5 +39,10 @@
 			<min>0.0</min>
 			<max>1.0</max>
 		</option>
+		<option name="allow_zoom" type="bool">
+			<_short>Allow Zoomed Windows</_short>
```
soreau, line 43:
```suggestion
			<_short>Allow Zoomed Views</_short>
```

metadata/scale.xml
```diff
@@ -39,5 +39,10 @@
...
+			<_long>Whether scaled windows can have a size larger than their original size.</_long>
```
soreau, line 44:
```suggestion
			<_long>Whether scaled views can have a size larger than their original size.</_long>
```

soreau:
@dkondor Since I pushed the close PR, this one needs rebasing.

PR #579, opened 2020-07-08
Separate activator for all workspaces

dkondor:
This is a personal preference, but I find this behavior better, since I can choose to initiate scale including all windows or only the current workspace.

This replaces the separate bool setting 'all_workspaces'. There are two activators, they can also be used to switch between the all workspaces / current workspace view as well.

### DIFF
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -12,7 +12,7 @@
 		<option name="toggle_all" type="activator">
 			<_short>Toggle for all workspaces</_short>
 			<_long>Toggles scale showing windows from all workspaces.</_long>
-			<default>disabled</default>
+			<default></default>
 		</option>
 		<option name="spacing" type="int">
 			<_short>Spacing</_short>
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -9,10 +9,10 @@
 			<_long>Toggles scale with the specified activator.</_long>
 			<default>&lt;super&gt; KEY_P</default>
 		</option>
-		<option name="all_workspaces" type="bool">
-			<_short>Show Views From All Workspaces</_short>
-			<_long>If this option is not set, only views from the current workspace will be scaled.</_long>
-			<default>false</default>
+		<option name="toggle_all" type="activator">
+			<_short>Toggle for all workspaces</_short>
+			<_long>Toggles scale showing windows from all workspaces.</_long>
+			<default>disabled</default>
 		</option>
 		<option name="spacing" type="int">
 			<_short>Spacing</_short>
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -81,8 +81,11 @@ class wayfire_scale : public wf::plugin_interface_t
     std::map<wayfire_view, view_scale_data> scale_data;
     wf::option_wrapper_t<int> spacing{"scale/spacing"};
     wf::option_wrapper_t<bool> interact{"scale/interact"};
-    wf::option_wrapper_t<bool> all_workspaces{"scale/all_workspaces"};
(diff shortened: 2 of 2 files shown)


soreau:
Again, there should only be one commit here.

About this idea, what do you expect to happen if you have scaled with one binding and try to use the other? Unscale? Switch to the new set of views? Nothing?

dkondor:
I've rebased to the latest upstream scale.

If scale is activated by one of the bindings, then the other binding switches to the other view. E.g. if scale is activated by "scale/toggle", only views from the current workspace are shown; if now the user activates "scale/toggle_all", it switches to showing views from all workspaces.

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
     {
         if (active)
         {
-            deactivate();
+            if(all_workspaces) {
```
soreau, line 175:
```suggestion
            if (all_workspaces)
            {
```

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
...
+                all_workspaces = false;
+                all_workspaces_option_changed();
+            }
+            else {
```
soreau, line 179:
```suggestion
            else
            {
```

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
...
+                deactivate();
+            }
+        }
+        else {
```
soreau, line 183:
```suggestion
        else
        {
```

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
...
+        output->render->schedule_redraw();
+        return true;
+    };
+
+    wf::activator_callback toggle_all_cb = [=] (wf::activator_source_t, uint32_t)
+    {
+        if (active)
         {
-            return false;
+            if(!all_workspaces) {
```
soreau, line 199:
```suggestion
            if (!all_workspaces)
            {
```

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
...
+                all_workspaces = true;
+                all_workspaces_option_changed();
+            }
+            else {
```
soreau, line 203:
```suggestion
            else
            {
```

plugins/single_plugins/scale.cpp
```diff
@@ -167,11 +172,44 @@ class wayfire_scale : public wf::plugin_interface_t
...
+                deactivate();
+            }
+        }
+        else {
```
soreau, line 207:
```suggestion
        else
        {
```

metadata/scale.xml
```diff
@@ -9,10 +9,10 @@
...
 			<default>&lt;super&gt; KEY_P</default>
 		</option>
-		<option name="all_workspaces" type="bool">
-			<_short>Show Views From All Workspaces</_short>
-			<_long>If this option is not set, only views from the current workspace will be scaled.</_long>
-			<default>false</default>
+		<option name="toggle_all" type="activator">
+			<_short>Toggle for all workspaces</_short>
+			<_long>Toggles scale showing windows from all workspaces.</_long>
+			<default>disabled</default>
```
soreau, line 15:
```suggestion
			<default></default>
```

soreau:
Thanks!

PR #580, opened 2020-07-08
Scale close

dkondor:
Extra functionality to close a window by middle-clicking on it, similarly how it works in Compiz.

### DIFF
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -39,5 +39,10 @@
 			<min>0.0</min>
 			<max>1.0</max>
 		</option>
+		<option name="middle_click_close" type="bool">
+			<_short>Close windows with middle click</_short>
+			<_long>Use the middle mouse button to close windows in the scale view.</_long>
+			<default>false</default>
+		</option>
 	</plugin>
 </wayfire>
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -40,8 +40,8 @@
 			<max>1.0</max>
 		</option>
 		<option name="middle_click_close" type="bool">
-			<_short>Close windows with middle click</_short>
-			<_long>Use the middle mouse button to close windows in the scale view.</_long>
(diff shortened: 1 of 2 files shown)


plugins/single_plugins/scale.cpp
```diff
@@ -263,7 +264,8 @@ class wayfire_scale : public wf::plugin_interface_t
             input_release_impending = false;
         }
 
-        if (button != BTN_LEFT || state != WLR_BUTTON_PRESSED)
+        if ( (button != BTN_LEFT && (button != BTN_MIDDLE ||
```
soreau, line 267: Since there is more than one possible case now, it should be cased out like in `process_key()`, instead of a complex conditional.

soreau:
Please rebase on scale and make this PR only have one commit.

plugins/single_plugins/scale.cpp
```diff
@@ -359,6 +374,11 @@ class wayfire_scale : public wf::plugin_interface_t
         {
             return;
         }
+        
+        if (button == BTN_MIDDLE) {
```
soreau, line 378:
```suggestion
        if (button == BTN_MIDDLE)
        {
```

metadata/scale.xml
```diff
@@ -39,5 +39,10 @@
 			<min>0.0</min>
 			<max>1.0</max>
 		</option>
+		<option name="middle_click_close" type="bool">
+			<_short>Close windows with middle click</_short>
+			<_long>Use the middle mouse button to close windows in the scale view.</_long>
```
soreau, line 44: The terminology in wayfire is *views* instead of *windows*.

dkondor: Ok, I'm changing it; I wasn't sure if this was used in user-facing parts as well

soreau:
Thanks!

PR #581, opened 2020-07-08
Avoid multiple lookups in scale_data map

dkondor:
This is a code style change, it uses a local reference instead of repeatedly looking up scale_data[view]; in my opinion, this is a bit cleaner and slightly more efficient (depending on the compiler).

### DIFF
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -145,7 +145,6 @@ class wayfire_scale : public wf::plugin_interface_t
         wf_scale *tr = new wf_scale(view);
         scale_data[view].transformer = tr;
         view->add_transformer(std::unique_ptr<wf_scale>(tr), transformer_name);
-        tr->alpha = 1;
         view->connect_signal("geometry-changed", &view_geometry_changed);
     }
 
@@ -623,8 +622,9 @@ class wayfire_scale : public wf::plugin_interface_t
                 auto it = scale_data.find(child);
                 if ((it == scale_data.end()) || !it->second.transformer)
                 {
-                    /* this should not happen?
-                     * child views can show up here before they are attached */
+                    /* note: child views can show up here before they
+                     * should be visible (between they are attached and
+                     * mapped), these should be skipped */
                     continue;
                 }
 
@@ -816,7 +816,6 @@ class wayfire_scale : public wf::plugin_interface_t
 
                     auto& view_data = scale_data[child];
 
-                    scale_x = scale_y = std::min(scale_x, scale_y);
                     view_data.animation.scale_animation.scale_x.set(
                         view_data.transformer->scale_x, active ? scale_x : 1);
                     view_data.animation.scale_animation.scale_y.set(
@@ -1008,7 +1007,7 @@ class wayfire_scale : public wf::plugin_interface_t
                 return;
             }
 
-            layout_slots(views);
+            layout_slots(std::move(views));
         }
     };
 
@@ -1020,7 +1019,7 @@ class wayfire_scale : public wf::plugin_interface_t
                 return;
             }
 
-            layout_slots(views);
+            layout_slots(std::move(views));
         }
     };
 
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -69,7 +69,7 @@ class wf_scale : public wf::view_2D
 struct view_scale_data
 {
     int row, col;
-    wf_scale *transformer;
+    wf_scale *transformer = 0; /* avoid potential UB from uninitialized member */
     wf::animation::simple_animation_t fade_animation;
     wf_scale_animation_attribs animation;
 };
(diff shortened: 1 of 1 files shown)


soreau:
> This is a code style change, it uses a local reference instead of repeatedly looking up scale_data[view]; in my opinion, this is a bit cleaner and slightly more efficient (depending on the compiler).

There should only be one commit in this PR. Also this patch introduces several changes that are inconsistent with the rest of the plugin. Most notably, there should be a space between the `if` and `(` and braces should always be used for conditionals and they should start on the next line. Also there are whitespaces added such as [this](https://github.com/WayfireWM/wayfire/pull/581/files#diff-15c486c4a34bb51e29cc2f4a9d20b4f9R450).

Do this:
```
if (condition)
{
    //code
}
```
Not this:
```
if(condition) {
    //code
}

if(condition) //code
```

soreau:
@dkondor Can you rebase this on latest scale branch?

plugins/single_plugins/scale.cpp
```diff
@@ -142,10 +142,10 @@ class wayfire_scale : public wf::plugin_interface_t
...
         }
 
-        scale_data[view].transformer = new wf_scale(view);
-        view->add_transformer(std::unique_ptr<wf_scale>(
-            scale_data[view].transformer), transformer_name);
-        scale_data[view].transformer->alpha = 1;
+        wf_scale *tr = new wf_scale(view);
+        scale_data[view].transformer = tr;
+        view->add_transformer(std::unique_ptr<wf_scale>(tr), transformer_name);
+        tr->alpha = 1;
```
soreau, line 148: I know it was like this already but I think the transformer values are already initialized, so no need to set alpha = 1 here.

plugins/single_plugins/scale.cpp
```diff
@@ -595,43 +595,52 @@ class wayfire_scale : public wf::plugin_interface_t
...
+                view_data.animation.scale_animation.translation_y;
+            view_data.transformer->alpha = view_data.fade_animation;
 
             view->damage();
 
             for (auto& child : view->children)
             {
-                view = child;
-                if (!view || !scale_data[view].transformer)
+                if (!child)
```
soreau, line 618: Why are we not checking for !scale_data[child].transformer ?

plugins/single_plugins/scale.cpp
```diff
@@ -595,43 +595,52 @@ class wayfire_scale : public wf::plugin_interface_t
...
-                    scale_data[view].animation.scale_animation.translation_y;
-                scale_data[view].transformer->alpha =
-                    scale_data[view].fade_animation;
-
-                view->damage();
+                auto it = scale_data.find(child);
+                if ((it == scale_data.end()) || !it->second.transformer)
+                {
+                    /* this should not happen?
+                     * child views can show up here before they are attached */
```
soreau, line 627: The two lines in this comment seem to contradict each other.

plugins/single_plugins/scale.cpp
```diff
@@ -804,29 +814,30 @@ class wayfire_scale : public wf::plugin_interface_t
...
-                        scale_data[child].transformer->scale_x,
-                        active ? child_scale_x : 1);
-                    scale_data[child].animation.scale_animation.scale_y.set(
-                        scale_data[child].transformer->scale_y,
-                        active ? child_scale_y : 1);
-                    scale_data[child].animation.scale_animation.translation_x.set(
-                        scale_data[child].transformer->translation_x,
+                    auto& view_data = scale_data[child];
+
+                    scale_x = scale_y = std::min(scale_x, scale_y);
```
soreau, line 819: It looks like this line was added by mistake?

plugins/single_plugins/scale.cpp
soreau: I see now, it's checked below.

plugins/single_plugins/scale.cpp
dkondor: Yes, I see and agree; I'll remove the assignment

plugins/single_plugins/scale.cpp
dkondor: Yes, indeed, thanks for noticing!

plugins/single_plugins/scale.cpp
dkondor: I'm not sure. At first I thought this should not happen, but then realized this can happen under normal circumstances (a child view is created, but not visible yet?).
Anyway, I'll have to double check this in debugger and then I'll update the comment.

dkondor: Looking in a debugger, this is the sequence of events (in the case of gedit, for the "close without saving" dialog):
1. view_attached() is called
2. child view shows up here, but does not have a transformer set up, so the continue statement is hit on line 628
3. view_geometry_changed() is called which results in the transformer set up (via layout_slots())
So I think it is normal for this case to be hit

PR #591, opened 2020-07-14
idle: support idle-inhibit v1 protocol

myfreeweb:
As discussed in IRC, we were not actually supporting this protocol.

This whole inhibition thing is quite complex, I'll need to use a model checker to convince myself that all the behavior is correct.

- changing `disable_on_fullscreen` on the fly complicates everything :/
- in `idle_inhibit`, `idle_inhibit_ref++` was before the early return for `!disable_on_fullscreen`.. was this a bug? was some support for *enabling* `disable_on_fullscreen` on the fly with an existing fullscreen surface planned? 
- I'm not sure about the toggle hotkey directly touching `toggle_idle`.. maybe it's still fine but I rewrote it to act like an extra inhibitor for now, just in case
- it's a correct assumption that we only deal with one fullscreen surface, right? (since the event is `fullscreen-layer-focused`)


ammen99:
> * it's a correct assumption that we only deal with one fullscreen surface, right? (since the event is `fullscreen-layer-focused`)

Ehh, not really. There is nothing preventing 2 or more fullscreen surfaces.

plugins/single_plugins/idle.cpp
```diff
@@ -136,15 +138,20 @@ class wayfire_idle
         }
     }
 
-    void idle_inhibit()
+    void idle_inhibit(bool because_fullscreen)
```
ammen99, line 141: why do we need this flag at all ???

myfreeweb: We need to track `idle_inhibited_by_fullscreen` to make sure that `disable_on_fullscreen_changed` doesn't mess with the protocol-based inhibitors.

myfreeweb:
> Ehh, not really. There is nothing preventing 2 or more fullscreen surfaces.

Oh, yeah, but we do have to specifically treat it as one here, since `fullscreen-layer-focused` is emitted on changes between "there is a fullscreen surface" and "there are no fullscreen surfaces".

---

I simplified the code, it's pretty understandable now. All the decision making is in `update()` now. The toggle key is just another inhibitor, like the protocol object would be — this makes sense, it's like a [caffeinate](https://caffeine.en.softonic.com/mac) button. The fullscreen "count" (could be a bool but there's no benefit to that) is counted separately to make the `disable_on_fullscreen` setting work dynamically.

Please test :)

kode54:
I tried this pull, and this makes the Waybar idle inhibitor actually work properly.

plugins/single_plugins/idle.cpp
```diff
@@ -191,7 +228,13 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
             return false;
         }
 
-        get_instance().toggle_idle();
+        if (toggled_off ^= 1)
```
ammen99, line 231: This would make the plugin work per-output, whereas I expect idle-inhibit shortcut to work globally?

plugins/single_plugins/idle.cpp
```diff
@@ -223,16 +266,28 @@ class wayfire_idle_singleton : public wf::singleton_plugin_t<wayfire_idle>
...
         screensaver_timeout.set_callback([=] ()
         {
             create_screensaver_timeout(screensaver_timeout);
         });
         create_screensaver_timeout(screensaver_timeout);
+
+        on_new_inhibitor.set_callback([&] (void *data)
+        {
+            auto inhibitor = static_cast<wlr_idle_inhibitor_v1*>(data);
+            /* will be freed by the destroy request */
```
ammen99, line 286: This will likely crash if an inhibitor is created and in the meantime the idle plugin is unloaded.

ammen99 (commented):
Why not put this stuff into core?

myfreeweb:
Because you usually didn't want to put anything into core? :D Yeah I'll do that

ammen99:
@myfreeweb Well except for things which belong there ;) Unfortunately, I do not have a clear-cut criteria, so it can be confusing sometimes :)

kode54:
Whoops, needs a rebase now.

myfreeweb:
Finally got around to working on this, sorry for the delay.

The new version works like this:

- inhibitor count is stored on the core impl
- `wf::idle_inhibitor_t` is a non-copyable dummy type that increments the inhibitor count when created, decrements when destroyed
  - and calls `wlr_idle_set_enabled` when either operation makes the count zero or non-zero (actually, could do `>=` here I guess)
- `wf::wlr_idle_inhibitor_t` is a non-copyable, non-deletable (private destructor) subclass that ties the lifecycle of an inhibitor to the corresponding wayland protocol object
  - the callback to create these from the wayland protocol manager object is in core
- now in the idle plugin:
  - the activator manages the existence of an `idle_inhibitor_t` on the singleton
  - the fullscreen thing manages the existence of an `idle_inhibitor_t` on the per-output plugin (`fullscreen-layer-focused` is an output signal, the workspace is under the output, etc)
  - as before, the singleton reacts to wlroots's idle timeout to do the screensaver/dpms stuff

src/api/wayfire/idle.hpp
```diff
@@ -0,0 +1,25 @@
...
+class idle_inhibitor_t
+{
+  public:
+    idle_inhibitor_t();
+    ~idle_inhibitor_t();
+    idle_inhibitor_t(const idle_inhibitor_t&) = delete;
+    idle_inhibitor_t& operator =(const idle_inhibitor_t&) = delete;
+};
+
+class wlr_idle_inhibitor_t : public idle_inhibitor_t
```
ammen99, line 16: I don't think we should expose this, it is an implementation detail in core.

src/api/wayfire/idle.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include <wayfire/util.hpp>
+#include <wayfire/nonstd/wlroots-full.hpp>
+
+namespace wf
+{
+class idle_inhibitor_t
```
ammen99, line 7: Please add a public comment explaining what and how this is useful (you already explained it in the comment)

src/core/idle.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+    /* NOTE: inhibited -> NOT enabled */
+    wlr_idle_set_enabled(wf::get_core().protocols.idle, NULL, !state);
+}
+
+wf::idle_inhibitor_t::idle_inhibitor_t()
+{
+    LOGD("creating idle inhibitor %p, previous count: %d", this,
+        wf::get_core_impl().idle_inhibitors);
+    if (++wf::get_core_impl().idle_inhibitors == 1)
```
ammen99, line 15: seems like this should be a public method of core_impl_t

ammen99 (changes requested):
A few small changes, I like the API design +1

src/core/idle.cpp
myfreeweb: The increment/decrement operation?

ammen99: Yes, which also calls enable/disable on the wlr_idle

myfreeweb: hm, my thinking was that I should avoid adding methods to the core_impl because these operations should only ever be done from this class, not from anywhere else

ammen99: then make the counter a static member variable? My main concern is that "friend class" is rarely a good idea, and here it seems like the counter belongs to the idle inhibitor class.

myfreeweb:
^^ also having a whole .cpp file just for `idle_inhibitor_t` impls (after moving `wlr_idle_inhibitor_t` fully into `core.cpp`) is ehh :/

src/core/idle.cpp
myfreeweb: oh yeah that's a good solution

ammen99:
> ^^ also having a whole .cpp file just for `idle_inhibitor_t` impls (after moving `wlr_idle_inhibitor_t` fully into `core.cpp`) is ehh :/

Did you see my last comment? https://github.com/WayfireWM/wayfire/pull/591#discussion_r507204392

myfreeweb:
Yes.. (though not before writing https://github.com/WayfireWM/wayfire/pull/591#issuecomment-711412107 and I did refresh the page, feels like github is having some kind of lag)

ammen99 (approved):

ammen99:
Thanks!

PR #594, opened 2020-07-15
Fix 593

soreau:


soreau:
> +1 from me for the changes so far.
> 
> Maybe we'd want the same for fullscreen state?

Makes sense to me.

> Should we also constrain the size of non-maximized views, so that they are not bigger than the workarea?

It would be polite, yes.

src/core/core.cpp
```diff
@@ -713,19 +713,54 @@ std::string wf::compositor_core_impl_t::get_xwayland_display()
...
     v->set_output(new_output);
     new_output->workspace->add_view(v,
         v->minimized ? wf::LAYER_MINIMIZED : wf::LAYER_WORKSPACE);
     new_output->focus_view(v);
+
+    if (reconfigure)
+    {
+        if (fullscreen)
+        {
+            v->set_geometry(new_output->get_relative_geometry());
```
ammen99, line 754: can't we do a fullscreen_request here?

src/core/core.cpp
```diff
@@ -713,19 +713,54 @@ std::string wf::compositor_core_impl_t::get_xwayland_display()
...
     wf::view_move_to_output_signal data;
     data.view = v;
-    data.old_output = v->get_output();
+    data.old_output = old_output;
     data.new_output = new_output;
     this->emit_signal("view-move-to-output", &data);
 
+    uint32_t edges;
+    bool fullscreen;
+    wf::geometry_t vg, og, ng;
```
ammen99, line 727: since we have a ton of `*g`s here, I think it would make more sense to explicitly write what each one is. `view_g`, `old_output_g`, `new_output_g`.

ammen99:
Thanks!

PR #595, opened 2020-07-16
Bump wlroots version.

ericonr:
There are compilation errors with 0.10.0, so the minimum has been bumped
to the latest wlroots release.


ammen99:
Indeed, this is true. Thanks!

Issue #596, opened 2020-07-16
Wayfire 0.5


ammen99:
Wayfire 0.5 would be compatible with the newly released wlroots 0.11. ETA: before 01.08.2020.

To packagers: if you need to immediately update the Wayfire package (for ex. because you need to update wlroots), please take Wayfire master branch for the time being.

ericonr:
The minimum version of wf-config needs to be updated, it doesn't build with 0.4.1:

```
[66/180] Compiling C++ object src/wayfire.p/core_matcher.cpp.o
FAILED: src/wayfire.p/core_matcher.cpp.o 
g++ -Isrc/wayfire.p -Isrc -I../src -I. -I.. -I../src/api -Isubprojects/wf-utils -I../subprojects/wf-utils -I/usr/include/libdrm -I/usr/include/pixman-1 -I/usr/include/elogind -I/usr/include/libevdev-1.0/ -I/usr/include/libxml2 -flto -fdiagnostics-color=always -DNDEBUG -pipe -D_FILE_OFFSET_BITS=64 -std=c++17 '-DWAYFIRE_VERSION="0.4.0-3ca208c (" __DATE__ ", branch '"'"'master'"'"')"' -DWLR_USE_UNSTABLE -Wno-unused-parameter -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -std=c++17 -DHAS_ADDR2LINE=1 -MD -MQ src/wayfire.p/core_matcher.cpp.o -MF src/wayfire.p/core_matcher.cpp.o.d -o src/wayfire.p/core_matcher.cpp.o -c ../src/core/matcher.cpp
../src/core/matcher.cpp: In lambda function:
../src/core/matcher.cpp:38:48: error: 'using element_type = class wf::config::option_t<std::__cxx11::basic_string<char> >' {aka 'class wf::config::option_t<std::__cxx11::basic_string<char> >'} has no member named 'get_default_value'; did you mean 'std::__cxx11::basic_string<char> wf::config::option_t<std::__cxx11::basic_string<char> >::default_value'? (not accessible from this context)
   38 |             if (option->get_value() != option->get_default_value())
      |                                                ^~~~~~~~~~~~~~~~~
In file included from ../src/api/wayfire/matcher.hpp:4,
                 from ../src/core/matcher.cpp:1:
/usr/include/wayfire/config/option.hpp:246:11: note: declared protected here
  246 |      Type default_value; /* default value */
[...]

PR #597, opened 2020-07-17
Bump wf-config version.

ericonr:
Adresses my comment in #596, depends on https://github.com/WayfireWM/wf-config/pull/28


ammen99:
Thanks, and thanks for the reminder to merge this!

PR #607, opened 2020-07-21
xwayland: Handle minimize actions

soreau:
This fixes xwayland CSD minimize buttons for clients like steam and chrome.

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -311,8 +311,8 @@ class wayfire_unmanaged_xwayland_view : public wayfire_xwayland_view_base
 class wayfire_xwayland_view : public wayfire_xwayland_view_base
 {
     wf::wl_listener_wrapper on_request_move, on_request_resize,
-        on_request_maximize, on_request_fullscreen, on_request_activate,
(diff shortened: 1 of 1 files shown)


damianatorrpm:
well fair enough, I guess I can close my PR than and open two other PR's for the stuff in since I can recreate damianatorrpm/wayfire when no PR depends on it.

This one here seems to fail though:
```
[  124s] ../src/view/xwayland.cpp:341:55: error: ‘data’ was not declared in this scope; did you mean ‘std::data’?
[  124s]   341 |             auto ev = (wlr_xwayland_minimize_event *) data;
[  124s]       |                                                       ^~~~
[  124s]       |                                                       std::data
```

kode54:
Oh, crud, when I was manually fixing that event in the patch file, I forgot to add it as a parameter to the callback. Basically, the minimize callback should be `(void* data)`

ammen99:
LGTM, but I will postpone merging this until we get wayfire 0.5 because the wlroots patch this thing relies on didn't make it to wlroots 0.11 (and wayfire 0.5 should depend on wlroots 0.11, not on wlroots master).

ammen99:
Thanks!

PR #608, opened 2020-07-22
only query keyboard when required

damianatorrpm:


ammen99:
Thanks! Not sure how much of a difference this would make, but it definitely won't make perf worse.

PR #609, opened 2020-07-22
add system bell signal

damianatorrpm:


damianatorrpm:
@ammen99 I am unsure is there a code style problem here?
The checks seem to fail bu I can't see it.

soreau:
@damianatorrpm You should use the following steps to submit an uncrustified PR:

- Make changes and test
- Run `git ls-files | grep "hpp$\|cpp$" | xargs uncrustify -c uncrustify.ini --no-backup` using uncrustify 0.71 or later
- Use git commit, push, etc.

src/output/gtk-shell.cpp
```diff
@@ -167,8 +167,21 @@ static void handle_gtk_shell_set_startup_id(wl_client *client, wl_resource *reso
  */
 static void handle_gtk_shell_system_bell(wl_client *client, wl_resource *resource,
     wl_resource *surface)
-{}
-
+{
+    if (surface)
```
ammen99, line 171: I think I also said it on the other PR, why not emit the signal on core, and jusst pass the wl_surface to it ... The plugin can then decide whether to ignore system bell on null surface or not.

damianatorrpm: it does it now

ammen99: No, you didn't, just look at the code, it begins with an `if (surface)` ...

I mean you could just emit the code with the raw `wl_resource`. I can imagine you could get a system bell for a subsurface (which won't have a view), or a NULL system bell (which should still play sound afaik)

src/output/gtk-shell.cpp
```diff
@@ -167,8 +167,21 @@ static void handle_gtk_shell_set_startup_id(wl_client *client, wl_resource *reso
...
+    {
+        auto wl_surface =
+            static_cast<wl_resource*>(wl_resource_get_user_data(surface));
+        wayfire_view view = wf::wl_surface_to_wayfire_view(wl_surface);
+
+        if (view)
+        {
+            data.view = view;
+        }
```
ammen99, line 181:
```suggestion
        data.view = wf::wl_surface_to_wayfire_view(wl_surface);
```

This is equivalent, since data.view is nullptr anyway if view is null.

src/api/wayfire/signal-definitions.hpp
```diff
@@ -77,6 +77,13 @@ struct view_tiled_signal : public _view_signal
 struct view_self_request_focus_signal : public _view_signal
 {};
 
+/**
+ * view-system-bell signal is emitted whenever the client indicates
+ * the view wants to invoke the system bell if such is available.
```
ammen99, line 82:
```suggestion
 * the view wants to invoke the system bell if such is available.
 *
 * Note: system-bell may be invoked with a nullptr view.
```

ammen99:
Thanks!

PR #619, opened 2020-07-24
meson.build: expand execinfo comment.

ericonr:


ammen99:
Hmm, the CI fails because master fails too ... But It is good otherwise, so merging. Thanks :)

PR #623, opened 2020-07-24
Signal rename

### LINKED ISSUE #108: Ensure consistency in signal naming
Currently all kinds of signal naming is used (`view-unmap`, `attach-view`, etc). We should decide how to name them, so they all follow the same pattern.

ammen99:
Fixes #108 

I am sorry for all of you who have custom plugins (@damianatorrpm  @myfreeweb  @soreau), this will very likely break compilation. At least with this PR all signal names should be consistent, all signal structs also follow the same naming convention and are all in the `wf::` namespace. Also, I have added documentation for all signals emitted from core.

Here is an incomplete porting guide:

1. Most events are the same, the type has changed `view_geometry_changed_signal` -> `wf::view_geometry_changed_signal`
2. `unmap-view`, `pre-unmap-view`, `map-view` are renamed to `view-unmapped`, `view-pre-unmapped` and `view-mapped`
3. `attach-view`, `detach-view`, `layer-attach-view`, `layer-detach-view` are now `view-(layer-)attached` and `view-(layer-)detached`.
4. `view-disappeared` now is emitted together with all the `view-detached`, `view-unmapped` and `view-minimized` signals.
5. `focus-view` -> `view-focused`
6. `view-maximize-request` -> `view-tile-request`

### DIFF
--- a/plugins/animate/animate.cpp
+++ b/plugins/animate/animate.cpp
@@ -185,8 +185,8 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
         grab_interface->name = "animate";
         grab_interface->capabilities = 0;
 
-        output->connect_signal("map-view", &on_view_mapped);
-        output->connect_signal("pre-unmap-view", &on_view_unmapped);
+        output->connect_signal("view-mapped", &on_view_mapped);
+        output->connect_signal("view-pre-unmapped", &on_view_unmapped);
         output->connect_signal("start-rendering", &on_render_start);
         output->connect_signal("view-minimize-request", &on_minimize_request);
     }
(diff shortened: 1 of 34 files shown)


myfreeweb:
Is that consistent enough? Looks like not enough `*ed` to me :D Why not `view-pre-unmapped`, `view-tile-requested`?

ammen99:
> Why not `view-pre-unmapped`, `view-tile-requested`

The `-request` part is a noun, but will change the other one too :)

ammen99:
Also, my main concern was the order. In master, we have: `map-view`, `move-request`, `view-maximize-request` ... I myself need to look up the order each time :)

damianatorrpm:
for the sake of consistency maybe you should rename view-move-to-output to also have a pre-suffix and emit it without the pre suffix at the end of the function with the same data structure when it actually moved

kode54:
You forgot to apply the settings to `src/view/xwayland.cpp`, which also implies that CI is not verifying whether the Xwayland bits build.

ammen99:
@kode54, what do you mean? CI builds the whole project and uncrustify runs on all cpp and hpp files.

kode54:
Oh, hmm, maybe the errors were due to another merge request in my test queue. Oh well, I'll leave whoever authored that merge to fix their thing when this lands.

ammen99:
For people who are trying to fix their plugins, here are the commands I have used for the main plugins: https://pastebin.com/44igfdvH. The commands do some things wrongly at first but if you run all of them you should get correct results. Anyway, do check what changed.

Issue #625, opened 2020-07-25
quarter tiling bad unmaximize


damianatorrpm:
If a view is quarter tiled and the application is un-tiled/maximized from the window buttons the position is weird:
https://drive.google.com/file/d/1M2tzuTaiVvhlnwUrmiBkXt9xFT1GMIEJ/view?usp=sharing

ammen99:
I can see what is causing this, this is also something @soreau reported on IRC awhile ago, although the reproduction steps are different (his reproduction steps were moving a view from one output to the other).

Basically, the view is restored to the last geometry it had before tiling. In your video you can see that the view was offscreen right before it was tiled, so it is restored there. Grid should probably clamp the restored geometry to the workarea so the view is always visible, and it should probably try to adjust the view to be around the mouse pointer.

ammen99:
This is actually more or less a duplicate of #606

Issue #606, opened 2020-07-21
Maximized view wrong size after maximize > fullscreen > oswitch > unfullscreen


soreau:
This happens when the outputs switched from/to are different sizes.

PR #633, opened 2020-07-26
do not break fullscreen views

damianatorrpm:
Note: games that have the option 'desktop friendly fullscreen minimizes on focus loss'
are still broken when they loose focus, without the minimize PR

### DIFF
--- a/plugins/single_plugins/move.cpp
+++ b/plugins/single_plugins/move.cpp
@@ -303,8 +303,22 @@ class wayfire_move : public wf::plugin_interface_t
             return;
         }
 
-        /* Snap the view */
-        if (enable_snap && (slot.slot_id != 0))
+        /**
+         * Snap the view
+         * A view that is fullscreened may break if resized i.e
+         * tiled_vertical or tiled_horizontal, games usually have
+         * their own option to be fullscreen.
+         * If the compositor resizes them and thus removes
+         * for xwayland _NET_WM_STATE fullscreen (there
+         * are not much wayland games around to test what wayland games
+         * do when they have fullscreen in their settings and
+         * the compositor puts them in another state) they break.
+         *
+         * This way fullscreened view can be moved from one
+         * output to another and be restored to fullscreen
+         * on a different output (hopefully with the same resolotion)
+         */
+        if (enable_snap && (slot.slot_id != 0) && !this->view->fullscreen)
         {
             snap_signal data;
             data.view = view;
@@ -320,6 +334,15 @@ class wayfire_move : public wf::plugin_interface_t
         workspace_may_changed.to   = output->workspace->get_current_workspace();
         workspace_may_changed.old_viewport_invalid = false;
         output->emit_signal("view-change-viewport", &workspace_may_changed);
+        /**
+         * Restore the fullscreen state so the window is
+         * not in a weird position afterwards, fullscreen
+         * are by spec the geometry of the output (or span outputs)
+         */
+        if (this->view->fullscreen)
+        {
+            view->fullscreen_request(output, true);
+        }
 
         this->view = nullptr;
     }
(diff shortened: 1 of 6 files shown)


ammen99:
Something which doesn't really work: open chrome, fullscreen, drag to another workspace with expo => chrome is still fullscreen but with invalid coordinates

damianatorrpm:
> Something which doesn't really work: open chrome, fullscreen, drag to another workspace with expo => chrome is still fullscreen but with invalid coordinates

You mean that the geometry of the view is wrong or that it doesn't snap properly?
I think snap to fullscreen should work  properly with expo too ofc. looking into that.

damianatorrpm:
Expos is working properly now as well from what I can tell,
let me know if something else is needed

ammen99:
As I said on IRC, this will have to wait a bit until #606 and #625 are solved. Then we'll be able to have this without any bugs.

damianatorrpm:
Does this fix conflict somehow with something else?
Any chance #625 will be resolved soon?

ammen99:
> Does this fix conflict somehow with something else?
> Any chance #625 will be resolved soon?

Yes, I am working on this atm.

plugins/single_plugins/move.cpp
```diff
@@ -320,6 +334,15 @@ class wayfire_move : public wf::plugin_interface_t
         workspace_may_changed.to   = output->workspace->get_current_workspace();
         workspace_may_changed.old_viewport_invalid = false;
         output->emit_signal("view-change-viewport", &workspace_may_changed);
+        /**
+         * Restore the fullscreen state so the window is
+         * not in a weird position afterwards, fullscreen
+         * are by spec the geometry of the output (or span outputs)
+         */
+        if (this->view->fullscreen)
```
ammen99, line 342: This should be done by the move_snap_helper_t::handle_input_released function.
Also, you should take into account which workspace the mouse position is on: for example if dragging in Expo, and mouse is on the workspace 3,3 and a fullscreen window is released there, then the fullscreen request should be for that workspace.

In case you don't know how to get the workspace where the mouse is(in pseudocode): `floor(last_grabbing_position.x / resolution.width) + current_workspace.x)`

plugins/single_plugins/move.cpp
```diff
@@ -303,8 +303,22 @@ class wayfire_move : public wf::plugin_interface_t
...
+         * for xwayland _NET_WM_STATE fullscreen (there
+         * are not much wayland games around to test what wayland games
+         * do when they have fullscreen in their settings and
+         * the compositor puts them in another state) they break.
+         *
+         * This way fullscreened view can be moved from one
+         * output to another and be restored to fullscreen
+         * on a different output (hopefully with the same resolotion)
+         */
+        if (enable_snap && (slot.slot_id != 0) && !this->view->fullscreen)
```
ammen99, line 321: why do we need to check this here, below you say that the slot for fullscreen views shouldn't be updated, so it will always be 0 anyway for fullscreen views?

ammen99 (changes requested):
Also rebase on top of master, and you can remove the changes to core from this PR because I did the same in #809

plugins/single_plugins/move.cpp
damianatorrpm: removed: new PR upcoming old repo deleted

plugins/single_plugins/move.cpp
damianatorrpm: I'm not sure I understand:
I think you mean : 
```
        if (this->view->fullscreen)
        {
            view->fullscreen_request(output, true);
        }
```
to move-snap-helper.hpp
```
    virtual void handle_input_released()
    {
        for (auto v : enum_views(view))
        {
            end_wobbly(v);
        }
        if (this->view->fullscreen)
        {
            view->fullscreen_request(output, true);
        }
        view->disconnect_signal("geometry-changed", &view_geometry_changed);
    }

```
>Also, you should take into account which workspace the mouse position is on: for example if dragging in Expo, and ??>mouse is on the workspace 3,3 and a fullscreen window is released there, then the fullscreen request should be for that >workspace.

You mean inside the
```
        if (this->view->fullscreen)
        {
            view->fullscreen_request(output, true);
        }
```
do something like 
ws = floor(last_grabbing_position.x / resolution.width) + current_workspace.x)
 view->fullscreen_request(output, true, ws);

Do I get that correct?

ammen99: yes, but you should put that after disconnecting from the signal

ammen99:
Closing in favor of #818

PR #643, opened 2020-07-29
meson: detect address sanitizer via CXXFLAGS

ammen99:
We need to disable our internal backtrace in this case.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,7 +13,6 @@ project(
 	],
 )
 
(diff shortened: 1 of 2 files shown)


ericonr:
Wouldn't it be better to have a build option for this?

ammen99:
> Wouldn't it be better to have a build option for this?

How would that help?

PR #651, opened 2020-07-31
Plugin mashup

### LINKED ISSUE #333: [expo][multi-monitor] Window drag between two expo's doesn't work
Environment:
2+ monitors

Reproduce:
1. Open expo on first monitor
2. Open expo on second monitor
3. Try to drag a window from one grid to the other

Expected result: window can be dragged between monitors

Actual result: window is dropped in original expo view once cursor makes it to monitor boundary

### LINKED ISSUE #404: Allow activating Expo plugin while Move plugin is active
Currently, it is the other way around: we press and release the modifier key to expose workspaces and grab the window to send it.

### LINKED ISSUE #582: Dialog placement with multilple workspaces
Hi,

this is a behavior I noticed while experimenting with the scale plugin and I find it strange. It seems that modal dialogs have a placement that is the same on all workspaces and are hidden / shown according to whether their parent is visible.

Normal windows / views can be placed to span multiple workspaces and they display consistently in the overview. On the other hand, child dialogs have a strange behavior that they are placed at the same position on each workspace the parent is visible on.

Steps to reproduce:
1. Place a window that it spans multiple workspaces (e.g. nautilus in the first screenshot)
2. Open a modal dialog (e.g. properties) and place it that it also spans more than one workspace (see second screenshot)
3. Switch to the other workspace that the views should occupy (the one to the right in this case)
4. Parent view is in expected location, the dialog is however in the same absolute location as in the other workspace (note: third screenshot is in the top middle workspace, 4th screenshot is in the top right one)

Expected behavior:
Modal dialog is displayed as if it spans multiple workspaces, similar to its parent. E.g. in the fourth screenshot, I expect it to be in the left side of the workspace, showing its right half, instead of being in the right side of the workspace, showing its left half

Placement of main window:
wayfire_dialog1

Placement with dialog:
wayfire_dialog2

Top middle workspace:
wayfire_dialog3

Top right workspace, dialog is in the "wrong" location:
wayfire_dialog4

### LINKED ISSUE #650: Figure out a way to reuse workspace-wall objects between plugins
All plugins that use workspace-wall could re-use the same workspace-wall object. This is desirable because a workspace-wall object holds a framebuffer for each workspace, so having 3 workspace wall objects more or less triples the GPU memory consumption (expo, vswitch, vswipe).

ammen99:
The goal of this PR is to increase the possibilities for interaction between some plugins, and to investigate the different possibilities to achieve that goal. Plan is:

- [x] Allow workspace switching when moving a view
- [x] Allow dragging a view between two outputs with:
    - [x] Two expo plugins (would fix #333) 
    - [x] Move & Expo ~~(would fix #404 or at least make it very easy)~~ Works on different outputs, activating on the same output still doesn't.
    - [x] ~~Move & Scale (not sure if this is feasible)~~ Since scale is not merged yet, this will happen later if at all (scale also doesn't support dragging views around for now)
- [x] On the way also fix #650, because we get an increasing amount of plugins which use workspace-wall
- [x] Fixes #582 

Strategy for first part is to make the vswitch code fully reusable, this way plugins can use it to animate workspace switching without involving the vswitch plugin itself.

For second part, it may make the most sense to have a common signal between all mentioned plugins to pass the actively grabbed view around (and the `move_snap_helper` object to make it seamless with wobbly).


ammen99:
This is ready for testing now. The new features are:

1. One can now drag views between two Expos or Move/Expo on two different monitors
2. One can drag views to the  edge of the output, wait a bit, then a workspace switch action happens so the view goes to the next workspace while still being grabbed (NB: disabled by default, https://github.com/WayfireWM/wayfire/pull/651/files#diff-2dff8ed504e693569ccdbf25fad10cb8R38-R44)

ammen99:
Finally finished all the changes necessary for expo. Now it can easily register the same bindings as vswitch and switch workspaces as it wishes. Also, vswitch needs only CUSTOM_RENDERER capabilities when servicing the workspace change request.

Along the way, I also think I fixed #582, @dkondor can you confirm?

PR #656, opened 2020-08-01
Pluralize xkb_option and xkb_rule and rename disable_while_typing

### LINKED ISSUE #402: Pluralize input.xkb_option(s) and input.xkb_rule(s)

### LINKED ISSUE #407: Rename input.disable_while_typing → input.disable_touchpad_while_typing

### LINKED ISSUE #641: xkb_option should be xkb_options
Seems to be correct everywhere internally but not the important config key ><
https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/keyboard.cpp#L59
https://github.com/WayfireWM/wayfire/blob/master/metadata/input.xml#L35
Especially since wayfire does not nag on invalid config ><

kode54:
Fixes #641 
Fixes #402 
Fixes #407

### DIFF
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -141,7 +141,7 @@
 				<_name>On-button down</_name>
 			</desc>
 		</option>
-		<option name="disable_while_typing" type="bool">
+		<option name="disable_touchpad_while_typing" type="bool">
 			<_short>Disable touchpad while typing</_short>
 			<_long>Disables the touchpad while typing.</_long>
 			<default>false</default>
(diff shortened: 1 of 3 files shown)


ammen99:
Can we also add https://github.com/WayfireWM/wayfire/issues/407 ? (as stated in IRC, goal is to do all breaking changes at once).

kode54:
Added #407 to this. Sorry, I didn't notice that one before.

ammen99:
Thanks!

PR #673, opened 2020-08-07
input-method-relay: hopefully appease ASan

myfreeweb:
ASan detected a heap-use-after-free with both the use and the free
pointing at the `remove_if` thing (specifically the use was pointing at `input`,
aka `this->input`, so I guess `this` was freed).

Let's move the `remove_if` thing into a method on `relay`.

### DIFF
--- a/src/core/seat/input-method-relay.cpp
+++ b/src/core/seat/input-method-relay.cpp
@@ -132,6 +132,17 @@ void wf::input_method_relay::disable_text_input(wlr_text_input_v3 *input)
     send_im_state(input);
 }
 
+void wf::input_method_relay::remove_text_input(wlr_text_input_v3 *input)
+{
+    auto it = std::remove_if(text_inputs.begin(),
+        text_inputs.end(),
+        [&] (const auto & inp)
+    {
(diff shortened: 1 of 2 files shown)


ammen99:
Would you mind sharing the full ASAN backtrace? I don't really see how this fixes anything (but I'm not sure what the crash was either).

myfreeweb:
> Would you mind sharing the full ASAN backtrace?

Well.. I did capture it, but only in this awful form:





..yeah, I need to add file logging to Wayfire, since always opening Wayfire under tmux is too much of a hassle. (I think I've had some kind of problem with redirecting the output into a file..? don't remember exactly)

Anyway, I think this should fix it because we're no longer using `this->input` after `this` has been moved by `remove_if`, we've already copied the `input` into a whole new stack frame.

ammen99:
> Anyway, I think this should fix it because we're no longer using `this->input` after `this` has been moved by `remove_if`, we've already copied the `input` into a whole new stack frame.

Ah, I see. Dereferencing elements after `remove_if` is UB, it is not reordering them. Thanks!

PR #674, opened 2020-08-07
Change Arch Linux installation procedure

nenad:
The package is in the AUR and `pacman` can not install packages from there. A popular suggestion for installing from the AUR is `yay`.


ammen99:
That's true, thanks!

PR #677, opened 2020-08-08
proposed fix for #676

### LINKED ISSUE #676: vswitch: multiple set-workspace-request end up on wrong workspace
I've found that in the vswitch plugin, if it receives a set-workspace-request while it is active, the result is inconsistent (it ends up of a different workspace than intended).

This is an edge case, I've found it while playing with cairo-dock (https://github.com/dkondor/cairo-dock-core/tree/foreign_toplevel).

I've describing here a simpler way to reproduce, but this method requires a touchscreen or very fast reflexes :)

Steps to reproduce:
1. Start wayfire. I'm assuming the following layout for workspaces, but the issue is present with a 3x3 grid as well:
```
A B
C D
```
2. Start at least two apps on two different workspaces. E.g. on workspace B and C in the above configuration.
3. Start wf-dock.
4. Switch back to workspace A.
5. Activate the two apps from wf-dock in quick succession (so that both are selected before the workspace switch can happen). I have a touchscreen and found that I'm able to do this by tapping quickly with different fingers on the two icons. I've managed to do this very quickly with the mouse only a few times.

Expected results: wayfire ends up on the workspace with the app that was activated last (either B or C)

Actual results: wayfire switches to workspace D. Also, the animation is weird.

Note: what seems to be happening is that two relative motions are carried out (right and down) which add up to ending up on workspace D, instead of only one of the motions that would go to workspace B or C.

Also, this does not happen with the workspace navigation keyboard shortcuts.

If unclear, I can create a dummy "plugin" that reproduces this issue.

Also, I will create a pull request with a proposed solution.

dkondor:
See https://github.com/WayfireWM/wayfire/issues/676

I'm not completely sure about this since I don't understand the relevance of the old_viewport variable in wf::workspace_change_request_signal struct.

Fixes #676


.gitmodules
```diff
@@ -1,9 +1,3 @@
-[submodule "subprojects/wf-config"]
```
ammen99, line 1: -1 for this change, I believe it was unintentional

dkondor: Yes, it was

plugins/single_plugins/vswitch.cpp
```diff
@@ -197,12 +197,18 @@ class vswitch : public wf::plugin_interface_t
 
     wf::signal_connection_t on_set_workspace_request = [=] (wf::signal_data_t *data)
     {
+        auto ev = static_cast<wf::workspace_change_request_signal*>(data);
         if (is_active())
         {
-            return;
+            auto cws = output->workspace->get_current_workspace();
```
ammen99, line 203: Does this fix it at all? I'd think not.

.gitmodules
dkondor: I reverted this; I don't understand that it still shows those modules removed (I have them in a separate directory).

plugins/single_plugins/vswitch.cpp
dkondor: Yes, at least in the case I could reproduce this issue before. it seems that cws does not change until the animation is finished. But this might not always be the case?
That's why I wrote I'm unsure about the ev->old_viewport parameter as well

dkondor: Oh, I realized a problem here: add_direction should not be called in this case; I wonder why it was still working

ammen99: `add_direction` is always the correct thing to do. The new direction would be `ev->new_viewport.x - animation.dx.end`

.gitmodules
dkondor: Should be fixed

plugins/single_plugins/vswitch.cpp
dkondor: I found that it still needs to take into consideration the old workspace; I've added an updated version

plugins/single_plugins/vswitch.cpp
```diff
@@ -197,14 +197,17 @@ class vswitch : public wf::plugin_interface_t
 
     wf::signal_connection_t on_set_workspace_request = [=] (wf::signal_data_t *data)
     {
+        auto ev = static_cast<wf::workspace_change_request_signal*>(data);
         if (is_active())
         {
-            return;
+            ev->carried_out = add_direction(ev->new_viewport.x -
```
ammen99, line 203: I think for the sake of being 100% correct we should use `get_current_workspace` instead of `old_viewport` here, otherwise lgtm

PR #678, opened 2020-08-08
fix for child views appearing in the wrong location

dkondor:
This is a fix for a small issue where child views do not appear in the right location.

Steps to reproduce:
1. Start at least two apps
2. Start scale with interactive mode enabled
3. Interact with one of the apps in a way that creates a child views (e.g. properties in nautilus or the quit without saving confirmation in gedit)

Expected behavior: child view appears in a position that is relative to the parent (e.g. centered on the parent).

Actual behavior: child view appears somewhere outside and moves to its correct location with a short animation. This is caused by setting scale to 1 and translation to 0 initially for the child view, so it appears in its untransformed location and then moves to its correct location via the scale animation.


dkondor:
Hi,
I agree it's nicer not to have an extra member in view_scale_data. On the other hand, I don't see a way to remove the bool parameter from layout_slots(), since it needs to behave differently when starting scale with child views already present which need to be animated together with the parents.

I realized there is also the edge case where a child view appears when scale is starting or ending (the animation is in progress). The way I tried handling this (conditional on active) looks ugly and I don't think it will work. I could not actually test this, since so far I cannot come up with a way to make a child show up just at the right time. Maybe not handling this case is OK, since its probably not that disturbing among all the other animations running? If you agree, I'll just remove the code related to this.

Also, I found another strange issue when a child is closed and opened quickly. In this case, for a brief period, the scale plugin acts as if was an extra view present and starts the rearrangement animation, then puts everything back in its original place. This is something I can reproduce with gedit, with the following steps:

1. Open gedit and write something without saving (so that the unsaved changes confirmation will show up when closing it).
2. Set the options "middle click to close" and "interact with views"
3. Start scale
4. Middle click on gedit to close it. The confirmation dialog shows up.
5. In quick succession, click on the cancel button in the dialog and then again middle click on gedit. The confirmation dialog is closed and then opened again.

Expected result: views do not move (expect for the dialog).

Actual result: views start to be rearranged (as if one more view is added), but then go back to their original locations.

I'm not sure if this is strictly related here, but I think the main reason is that it is not straightforward to keep track of child views. Mainly, for a short time, the child view is included in the list returned by get_views() -- if layout_slots() is called at this time (which happens when the new child view appears), it tries to rearrange views as if there is one more. This behavior is present in my branch and in the original branch as well, so this might be a separate issue (or ignored for the moment, since this is also an edge case).

soreau:
Does [this](http://ix.io/2tJ5) work?

soreau:
> Also, I found another strange issue when a child is closed and opened quickly. In this case, for a brief period, the scale plugin acts as if was an extra view present and starts the rearrangement animation, then puts everything back in its original place. This is something I can reproduce with gedit, with the following steps:
> 
>     1. Open gedit and write something without saving (so that the unsaved changes confirmation will show up when closing it).
> 
>     2. Set the options "middle click to close" and "interact with views"
> 
>     3. Start scale
> 
>     4. Middle click on gedit to close it. The confirmation dialog shows up.
> 
>     5. In quick succession, click on the cancel button in the dialog and then again middle click on gedit. The confirmation dialog is closed and then opened again.
> 
> 
> Expected result: views do not move (expect for the dialog).
> 
> Actual result: views start to be rearranged (as if one more view is added), but then go back to their original locations.
> 
> I'm not sure if this is strictly related here, but I think the main reason is that it is not straightforward to keep track of child views. Mainly, for a short time, the child view is included in the list returned by get_views() -- if layout_slots() is called at this time (which happens when the new child view appears), it tries to rearrange views as if there is one more. This behavior is present in my branch and in the original branch as well, so this might be a separate issue (or ignored for the moment, since this is also an edge case).

Does [this](http://ix.io/2tJn) help?

dkondor:
> Does [this](http://ix.io/2tJn) help?

Yes, this fixes this problem! Added it as a separate commit to this branch.

I'm wondering if there is a case where this could cause issues though. If layout_slots() now ignores unmapped views, maybe we would need to set up a signal callback for the case when a view is mapped, similarly to the view-attached signal handler?

dkondor:
> Does [this](http://ix.io/2tJ5) work?

I think the problem is that this might start a new animation. I've added an alternate version that tries to copy the whole animation state from the parent. Although this might be too complicated for a case that is likely not noticable unless the user sets animations to a super slow speed.

soreau:
> > Does [this](http://ix.io/2tJn) help?
> 
> Yes, this fixes this problem! Added it as a separate commit to this branch.

Why add it as a separate commit? It should be the only changes needed to close this issue.

> 
> I'm wondering if there is a case where this could cause issues though. If layout_slots() now ignores unmapped views, maybe we would need to set up a signal callback for the case when a view is mapped, similarly to the view-attached signal handler?

We can cross this bridge when we get there, or in another PR. I can't understand why you think you still need such a complicated patch to fix this.

soreau:
> > Does [this](http://ix.io/2tJ5) work?
> 
> I think the problem is that this might start a new animation.

Might? What case might this happen? Do you have a reproducible case? I'd rather focus less on what might be and more on what we know is the case.

> I've added an alternate version that tries to copy the whole animation state from the parent. Although this might be too complicated for a case that is likely not noticable unless the user sets animations to a super slow speed.

Yes it is too complicated, by far. You should only need the two patches I gave you.

dkondor:
> > > Does [this](http://ix.io/2tJn) help?
> > 
> > 
> > Yes, this fixes this problem! Added it as a separate commit to this branch.
> 
> Why add it as a separate commit? It should be the only changes needed to close this issue.
> 
> > I'm wondering if there is a case where this could cause issues though. If layout_slots() now ignores unmapped views, maybe we would need to set up a signal callback for the case when a view is mapped, similarly to the view-attached signal handler?
> 
> We can cross this bridge when we get there, or in another PR. I can't understand why you think you still need such a complicated patch to fix this.

I'm not familiar when these signals are generated, but so far I didn't see any issues. I'll do more testing and see if there are any issues.

dkondor:
> > > Does [this](http://ix.io/2tJ5) work?
> > 
> > 
> > I think the problem is that this might start a new animation.
> 
> Might? What case might this happen? Do you have a reproducible case? I'd rather focus less on what might be and more on what we know is the case.
> 
> > I've added an alternate version that tries to copy the whole animation state from the parent. Although this might be too complicated for a case that is likely not noticable unless the user sets animations to a super slow speed.
> 
> Yes it is too complicated, by far. You should only need the two patches I gave you.

I've tried to simplify everything; so now there is only new behavior if a new child appeared.

I was confused by how the animations are run. Now I realized that layout_slots() will result in all animations restarting anyway, so I agree there is no need to handle the child animation with added complexity.

On the other hand, I feel it might be unnecessary to call layout_slots() from the view_geometry_changed() signal handler, since there is only one affected view whose scale needs to be recalculated. Nevertheless, the current code works well, so it is probably not worth handling that separately for just the sake of optimization.

soreau:
> > > > Does [this](http://ix.io/2tJ5) work?
> > > 
> > > 
> > > I think the problem is that this might start a new animation.
> > 
> > 
> > Might? What case might this happen? Do you have a reproducible case? I'd rather focus less on what might be and more on what we know is the case.
> > > I've added an alternate version that tries to copy the whole animation state from the parent. Although this might be too complicated for a case that is likely not noticable unless the user sets animations to a super slow speed.
> > 
> > 
> > Yes it is too complicated, by far. You should only need the two patches I gave you.
> 
> I've tried to simplify everything; so now there is only new behavior if a new child appeared.
> 
> I was confused by how the animations are run. Now I realized that layout_slots() will result in all animations restarting anyway, so I agree there is no need to handle the child animation with added complexity.
> 
> On the other hand, I feel it might be unnecessary to call layout_slots() from the view_geometry_changed() signal handler, since there is only one affected view whose scale needs to be recalculated. Nevertheless, the current code works well, so it is probably not worth handling that separately for just the sake of optimization.

The reason for calling layout_slots() in view_geometry_changed() is because I view can change its size at any given time, and at that point, we have to recompute everything if we want it to look right. The other time that view_geometry_changed() is called is when switching workspaces. We listen to the workspace-changed signal now but for a different reason (to make sure the focused view stays focused when switching to another workspace). So for now, this is the trigger that allows all_workspaces to function the way it does (views follow you when switching workspaces with all_workspaces).

In general, I think scale is mostly feature complete (please say if there is anything else you see) except vswitch doesn't check if the switch-with-view view is transformed to a different workspace, so you can e.g. switch to another workspace with scale all_workspaces, then do vswitch switch-with-view and get the view to end up in an undesired location, perhaps off screen. I will call this a vswitch bug though.

For these two particular bugs (child originating from somewhere other than the middle of the parent and fixing the child unmapped case) I really think the only two patches you need are the ones I posted here. I no longer think that you should need to make add_transformer() return bool. If you try with *only* these patches, does it not completely fix these two problems?

dkondor:
Hi,

I've tested using only your patches and I see one problem: it assumes that having the translation values for the child equal to the parent is always a good choice. This is not the case for child views that the user moved before starting scale (best demonstrated with moving a child view so that it does not overlap with the parent anymore). For these, setting the translation values to the same as the parent causes them to jump to a different location and move back to the center of their slot every time layout_slots() is called.

For newly created child views, your patch works well. When thinking about this, I feel it is necessary to be able to identify newly-create child views while scale is running and handle them separately. Basically, I feel the problem is that when the "view-layer-attached" signal is emitted, the child-parent relationship is not necessarily set up yet, so it is not possible to set good default values there. Let me know if you see some better way to do this than the current solution.

Regarding scale in more general, I'm thinking of two additional features, but I don't think they belong to core functionality:
 -- Right click to "zoom in" on any of the scaled views; this is a nice feature in compiz, although I use it rarely. This seems easy to add and I'm hoping to create a separate PR for this when I have a bit more time.
 -- Filter views base on title or app id (this was mentioned by damianatorrpm in the beginning as well); I think this is a significant add-on though. On compiz, this provides actually two modes: (1) the user can type in (non-interactive) scale mode and views would be filtered by title; (2) another plugin or even an external program such as a taskbar / dock could request scale to be activated for a group of apps. In my opinion, both modes are very useful (the latter in combination with cairo-dock), but this seems like a big feature with some difficulties (e.g. I have no idea how to text input properly handle intext input  this case). Of course, I'd be interested to work on this, but I'm unsure how fast I would progress, and how serious changes it would mean.

soreau:
> Hi,
> 
> I've tested using only your patches and I see one problem: it assumes that having the translation values for the child equal to the parent is always a good choice. This is not the case for child views that the user moved before starting scale (best demonstrated with moving a child view so that it does not overlap with the parent anymore). For these, setting the translation values to the same as the parent causes them to jump to a different location and move back to the center of their slot every time layout_slots() is called.

I see. This touches on what I was thinking, which is that always setting the translation from the last scale_view isn't the greatest choice. What if there are more than one child view? I think it would work but haven't looked into it much, especially since I don't have a test case with multiple child views.

About the bug, does [this variant](http://ix.io/2u4E) of the first patch fix the problem?

> 
> For newly created child views, your patch works well. When thinking about this, I feel it is necessary to be able to identify newly-create child views while scale is running and handle them separately. Basically, I feel the problem is that when the "view-layer-attached" signal is emitted, the child-parent relationship is not necessarily set up yet, so it is not possible to set good default values there. Let me know if you see some better way to do this than the current solution.
> 
> Regarding scale in more general, I'm thinking of two additional features, but I don't think they belong to core functionality:
> -- Right click to "zoom in" on any of the scaled views; this is a nice feature in compiz, although I use it rarely. This seems easy to add and I'm hoping to create a separate PR for this when I have a bit more time.
> -- Filter views base on title or app id (this was mentioned by damianatorrpm in the beginning as well); I think this is a significant add-on though. On compiz, this provides actually two modes: (1) the user can type in (non-interactive) scale mode and views would be filtered by title; (2) another plugin or even an external program such as a taskbar / dock could request scale to be activated for a group of apps. In my opinion, both modes are very useful (the latter in combination with cairo-dock), but this seems like a big feature with some difficulties (e.g. I have no idea how to text input properly handle intext input this case). Of course, I'd be interested to work on this, but I'm unsure how fast I would progress, and how serious changes it would mean.

There will be plenty of room for extra features. I appreciate your rigorous testing, it's difficult to find every possible test case.

dkondor:
Hi,

I tested the new patch and it works fine in normal usage, but I'm still able to demonstrate buggy behavior in the following contrived example, exploiting the fact that having translation_x == translation_y == 0 can be the correct behavior for child views (with some low probability):

1. Start wayfire without any apps running (or move to an empty workspace)
2. Start nautilus. By default, it displays exactly in the middle of the screen.
3. Open the properties dialog (for anything). Again, this displays exactly in the middle of the screen.
4. Move the main nautilus window (without moving the properties dialog). This is not possible by dragging it by the title bar, but possible with the "move" plugin. So the properties dialog is still in the middle of the screen, but the main window is dragged to the side.
5. Start scale (main window moves back to the middle of the screen, while the dialog does not move at all; this means that translation for the child is 0, and for the parent, it is nonzero).
6. Start any other app. (Of course, I'm running wayfire in a window, so it is easy to do this from a terminal for me; running wayfire full-screen would make this scenario unlikely).

Result is that the child dialog jumps outside and then moves back to its proper location. This only occurs when the child is originally in the middle.

I'm not sure how relevant it is, since it is very unlikely to happen in normal usage and only causes a minor visual glitch. At the same time, it is a know bug in some sense :)


Regarding the case with multiple child views, I don't think there is an issue, view_data is redefined on line 893, which is weird, but valid in C++ AFAIK. It would be cleaner code to use a different variable name for it though.

soreau:
> Hi,
> 
> I tested the new patch and it works fine in normal usage, but I'm still able to demonstrate buggy behavior in the following contrived example, exploiting the fact that having translation_x == translation_y == 0 can be the correct behavior for child views (with some low probability):

Yes you are right, this isn't a great solution either. Would it suffice to use the return value of add_transformer() as the condition as discussed previously, to set translation instead of translation_x == translation_y == 0?

dkondor:
I've added a simplification, that does not add a bool parameter to layout_slots().

It works well as far as I can tell.

What I see as a potential issue is that the scale parameters of new child views are not set (these are 1 by default), but this is not really noticeable and I cannot think of a test case where I could demonstrate a problem with it.

plugins/single_plugins/scale.cpp
```diff
@@ -993,16 +1006,18 @@ class wayfire_scale : public wf::plugin_interface_t
         {
             auto view = get_signaled_view(data);
```
soreau, line 1008: What is this change for? From here to..

plugins/single_plugins/scale.cpp
```diff
@@ -993,16 +1006,18 @@ class wayfire_scale : public wf::plugin_interface_t
...
             }
 
-            if (!scale_view(view) && (view->role != wf::VIEW_ROLE_TOPLEVEL))
+            view->connect_signal("geometry-changed", &view_geometry_changed);
+
+            auto it = scale_data.find(view->parent);
+            if (it != scale_data.end())
             {
+                layout_slots(get_views());
+
```
soreau, line 1020: here. Why did you change this code? Also I don't really want the `geometry-changed` handler moved out of add_transformer function.

dkondor: The problem is that layout_slots() needs to know if it sees a view for the first time or not, since that allows it to correctly place a child view in a correct location.

In the current version, this is handled by checking the result of add_transformer(), i.e. add_transformer() is only called from within layout_slots() and if it returns true, then it is a new view that was not handled before.

At the same time, the geometry-changed signal needs to be connected earlier, since typically that is what triggers the layout_slots() call that will set up the transformer for a newly appeared child view. For this reason, the geometry-changed signal needs to be set up the first time a view is seen.

There could be other ways to keep track if a view was seen or is a potential new child view. But so far, I could not think of any way where it is clearly signaled that a new child view has been added.

The problem is also that the first time a view is attached, the parent relationship is not set up yet, at least in the cases tested view->parent is null here.

dkondor: Now, I'm not sure what was my reason of swapping the two conditions. Is there a case where the first condition (in the original code, i.e. looking up the parent in the map) is true, while the second condition is false?

dkondor: I realized that my original intention was that the "geometry-changed" signal should be connected in all cases except when the second condition (checking if view is a scale_view) is true. Now I'm less sure if this is necessary; I've added an update that just replaces the add_transformer() call. I'm concerned this could result in the "geometry-changed" signal not being connected if a new view is directly handled by the previous cases. 

I'm wondering if it is OK to connect the same signal multiple times? Does that mean we have to disconnect it the same number of times it was connected?

PR #689, opened 2020-08-15
Prepare for wlroots #2240

### LINKED ISSUE #98: Cube's star deformation doesn't work with post effects (invert, zoom, fisheye)
screenshot

### LINKED ISSUE #636: wayfire nested on wayfire shows white screen
Below would fix that and wayfire would work nested on wayland with wayland backend
```
[WL-$WAYLAND_DISPLAYNUMER]
layout = 0,0
mode = 1080x640@60.000000
scale = 1.400000
transform = normal
```

### LINKED ISSUE #703: master: empty window frame launching from gtk terminal here...
Not sure whether distro settings related, or internal...

[empty-wayfire-frame-in-gtk-terminal.zip](https://github.com/WayfireWM/wayfire/files/5114317/empty-wayfire-frame-in-gtk-terminal.zip)

Thks 4 ur attention, interest & support.

ammen99:
wlroots #2240 is a breaking change which would silently break Wayfire (because it changes some assumptions we make about how GL works). This PR attempts to fix that. There should be no code breakage for plugins, although some adjustments may be necessary.

~~On the way I managed to reproduce #636 and I think it should be fixed with this PR.
@damianatorrpm Can you confirm?~~

Fixes #98
Fixes #636
Fixes #703 

Also, this PR is wlroots-backwards-compatible, i.e it should work with current wlroots master too.

### DIFF
--- a/plugins/single_plugins/switcher.cpp
+++ b/plugins/single_plugins/switcher.cpp
@@ -645,8 +645,7 @@ class WayfireSwitcher : public wf::plugin_interface_t
             (float)sv.attribs.rotation, {0.0, 1.0, 0.0});
 
         transform->color[3] = sv.attribs.alpha;
-        sv.view->render_transformed(output->render->get_target_framebuffer(),
-            output->render->get_target_framebuffer().geometry);
+        sv.view->render_transformed(buffer, buffer.geometry);
     }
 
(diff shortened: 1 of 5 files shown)


ammen99:
Ready for testing, if someone is interested :)

PR #692, opened 2020-08-16
fix crash if initial_focus_view becomes invalid

dkondor:
If the view that was initial_focus_view closes, the scale plugin can crash wayfire. Steps to reproduce:

1. Disable interaction with views and enable middle click close.
2. Start at least two apps.
3. Start scale.
4. Close the view with the original focus.
5. Press ESC to exit scale.

This results in a crash. The reason is that ESC is used for "cancel", so scale tries to re-focus the view that was focused originally. Of course, if this view does not exist anymore, this results in a crash.

This patch sets initial_focus_view to the currently active view if the original one closes. Another option could be setting it to NULL.


soreau:
I think it would be better to make sure this is correct in view_minimized, view_detached and view_unmapped handlers. You can keep the null check in the ESC handler. I don't think we need to do anything in pop_transformer.

dkondor:
I've changed the code to move the main check into the signal handlers. I still feel child views are better checked in remove_views(). I'm not sure if initial_focus_view can be set to a child view in the first place though; if there is a guarantee that this cannot happen, checking for child views is not needed at all.

PR #693, opened 2020-08-16
Scale fixes

ammen99:
A few simplifications I noticed, see individual commits for explanations

### DIFF
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -154,10 +154,9 @@ class wayfire_scale : public wf::plugin_interface_t
     {
         for (auto& view : views)
         {
-            add_transformer(view);
-            for (auto& child : view->children)
+            for (auto& toplevel : view->enumerate_views())
             {
-                add_transformer(child);
+                add_transformer(toplevel);
             }
         }
     }
@@ -169,20 +168,16 @@ class wayfire_scale : public wf::plugin_interface_t
             return;
         }
 
-        if (view->get_transformer(transformer_name))
-        {
-            view->pop_transformer(transformer_name);
-        }
+        view->pop_transformer(transformer_name);
     }
 
     void remove_transformers()
     {
         for (auto& e : scale_data)
         {
-            pop_transformer(e.first);
-            for (auto& child : e.first->children)
+            for (auto& toplevel : e.first->enumerate_views(false))
             {
-                pop_transformer(child);
+                pop_transformer(toplevel);
             }
         }
     }
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -184,13 +184,8 @@ class wayfire_scale : public wf::plugin_interface_t
 
     bool all_same_as_current_workspace_views()
     {
-        auto all_workspace_views     = get_all_workspace_views();
-        auto current_workspace_views = get_current_workspace_views();
-
-        std::sort(all_workspace_views.begin(), all_workspace_views.end());
-        std::sort(current_workspace_views.begin(), current_workspace_views.end());
-
-        return all_workspace_views == current_workspace_views;
+        return get_all_workspace_views().size() ==
+            get_current_workspace_views().size();
     }
 
     wf::activator_callback toggle_cb = [=] (wf::activator_source_t, uint32_t)
--- a/plugins/single_plugins/scale.cpp
+++ b/plugins/single_plugins/scale.cpp
@@ -188,71 +188,49 @@ class wayfire_scale : public wf::plugin_interface_t
             get_current_workspace_views().size();
     }
 
-    wf::activator_callback toggle_cb = [=] (wf::activator_source_t, uint32_t)
+    bool handle_toggle(bool want_all_workspaces)
     {
-        if (active)
+        if (active && (all_same_as_current_workspace_views() ||
+                       (want_all_workspaces == this->all_workspaces)))
         {
-            if (all_workspaces)
-            {
-                all_workspaces = false;
-                if (all_same_as_current_workspace_views())
-                {
-                    deactivate();
-
-                    return true;
-                }
-
-                all_workspaces_option_changed();
-            } else
-            {
-                deactivate();
-            }
-        } else
-        {
-            all_workspaces = false;
-            if (!activate())
-            {
-                return false;
-            }
+            deactivate();
+            return true;
         }
 
-        output->render->schedule_redraw();
+        this->all_workspaces = want_all_workspaces;
+        if (active)
+        {
+            all_workspaces_option_changed();
+            return true;
+        } else
+        {
+            return activate();
+        }
+    }
 
-        return true;
+
+    wf::activator_callback toggle_cb = [=] (wf::activator_source_t, uint32_t)
+    {
+        if (handle_toggle(false))
+        {
+            output->render->schedule_redraw();
+
+            return true;
+        }
+
+        return false;
     };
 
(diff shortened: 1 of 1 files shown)


soreau:
This seems to break more things than it fixes. Reportedly, if you scale, close a view and unscale while the close animation is running, it will crash wayfire. Can you rebase and also check if this is the case?

ammen99:
> This seems to break more things than it fixes. Reportedly, if you scale, close a view and unscale while the close animation is running, it will crash wayfire. Can you rebase and also check if this is the case?

I can't repro with these steps.

kode54:
Can no longer reproduce the issue I was having.

soreau:
@dkondor Can you please test this PR?

dkondor:
It seems to be working, I don't see the previous issues after some quick testing.
Two small issues I noticed (both with the case of scale with all workspaces):
 - exiting interactive scale (or non-interactive scale with the activator keybinding) does not switch to the workspace of the selected view (although this is what I'd expect)
 - starting non-interactive scale from an empty workspace and then exiting with the "cancel" key (ESC), it still switches to the view that was first highlighted; I'd expect to stay on the original, empty workspace in this case
I don't remember if these were present before, probably yes, but I didn't notice or care about them.

It still has the issue of child views from other workspaces appearing in "wrong" locations (instead of coming from their workspace, they appear in their location in the current workspace), but I think this is caused by issue #582 and is not a problem with scale itself.

Also, menus (and other popups) appear unscaled. For interactive mode, I think this is OK since it makes it easier to interact with them; for non-interactive scale, my expectation would be for popups to close or at least be scaled when starting scale.

soreau:
> It seems to be working, I don't see the previous issues after some quick testing.

Great, thanks for testing.

> Two small issues I noticed (both with the case of scale with all workspaces):
> 
>     * exiting interactive scale (or non-interactive scale with the activator keybinding) does not switch to the workspace of the selected view (although this is what I'd expect)
 
I don't really want to bomb this PR with unrelated conversation but since scale isn't merged yet, I'm not sure an issue would be appropriate, so I think here is fine.
My question is, how are you exiting interactive scale if not with the activator? In any event, I imagine it shouldn't be too hard to fix this.

>     * starting non-interactive scale from an empty workspace and then exiting with the "cancel" key (ESC), it still switches to the view that was first highlighted; I'd expect to stay on the original, empty workspace in this case

Probably yet another bug, hopefully shouldn't be too difficult to find a solution.

> I don't remember if these were present before, probably yes, but I didn't notice or care about them.
> 
> It still has the issue of child views from other workspaces appearing in "wrong" locations (instead of coming from their workspace, they appear in their location in the current workspace), but I think this is caused by issue #582 and is not a problem with scale itself.

Yes, I think this is a vswitch bug. I don't think it handles child views.
> Also, menus (and other popups) appear unscaled. For interactive mode, I think this is OK since it makes it easier to interact with them; for non-interactive scale, my expectation would be for popups to close or at least be scaled when starting scale.

soreau:
@dkondor [This](http://ix.io/2vYF) fixes the bugs you describe, but might introduce others, I haven't tested thoroughly. Can you try it and see if you notice any other problems?

dkondor:
Hi,
I had limited time to test. So far, it seems it fixes the errors, but I think it introduces a separate issues:
 - Now if I start from an empty workspace, and exit with the activator toggle, it never switches workspaces (either interactive or non-interactive). It does switch workspaces in non-interactive mode if I click with the mouse. What I'd expect to happen is that it switches to a new workspace in all cases, except when I cancel it (i.e. non-interactive mode and press ESC). Alternatively, I could imagine it only switching if I explicitly selected a view.
 - In some cases, it switches to a bad workspace. I haven't found a clear pattern, but two examples that seem to always happen, assuming a 3x3 grid of workspaces and a view on the top left (0,0) and middle (1,1) one:
    1. Starting scale from the middle and selecting the view on the top left: the screen goes blank for a few seconds, and I see the following message in the terminal: "EE 03-09-20 12:14:17.322 - [src/output/workspace-impl.cpp:612] Attempt to set invalid workspace: (-1,-1) workspace grid size is 3x3" After this, it correctly switches to the top left (0,0) workspace. This only happens when selecting the view by clicking or pressing ENTER in non-interactive mode. Using the activator toggle works well.
    2. Starting from the empty bottom left (2,0) workspace, and selecting the view in the middle (1,1), it actually ends up on the top middle (0,1) workspace

dkondor:
> My question is, how are you exiting interactive scale if not with the activator? In any event, I imagine it shouldn't be too hard to fix this.

In non-interactive mode, either ENTER (should switch to the selected view), ESC (to cancel), or clicking on a view (should switch to that view)

soreau:
@dkondor Does [this](http://ix.io/2vYF) work any better?

dkondor:
Hi,
yes, but I see two issues:
 - now if I exit with ESC, it does not refocus the initial view (it keeps focusing the last selected); I've tried looking into this, but i have not idea what is happening
 - if I exit with ENTER, switching workspaces does not work properly; I think this is issue #676 , I will reopen and comment there

soreau:
> Hi,
> yes, but I see two issues:
> 
>     * now if I exit with ESC, it does not refocus the initial view (it keeps focusing the last selected); I've tried looking into this, but i have not idea what is happening

Refresh the patch link and try again. I think I was missing one line to fix the ESC not refocusing initial view.

> 
>     * if I exit with ENTER, switching workspaces does not work properly; I think this is issue #676 , I will reopen and comment there

I noticed this as well. I debugged this and it is indeed the reason why we're seeing black screen sometimes. Funny enough, vswitch has [this comment](https://github.com/WayfireWM/wayfire/blob/master/plugins/single_plugins/vswitch.cpp#L161) before supposedly clamping away any way for this exact case to happen.

dkondor:
> Refresh the patch link and try again. I think I was missing one line to fix the ESC not refocusing initial view.

It still does not work. I've put a breakpoint on all calls to output->focus_view() and saw that after calling this, there is a later call on line 1158 that sets it back.

I've found that adding
```last_focused_view = initial_focus_view;```
on line 536 (before the call to focus_view()) fixes this problem for me and I don't see any bad side effect. But I don't entirely understand the conditions on lines 1141 -- 1144, so this is a guess on my part. It would be great if you could test and reproduce the problem and solution.

Thanks!

soreau:
I see now, I was testing the wrong case. I 'pushed' a new patch to the link. Hopefully this fixes it.

dkondor:
Yes; I just tested, the latest version of the patch fixes this!

soreau:
> It still does not work. I've put a breakpoint on all calls to output->focus_view() and saw that after calling this, there is a later call on line 1158 that sets it back.
> 
> I've found that adding
> `last_focused_view = initial_focus_view;`
> on line 536 (before the call to focus_view()) fixes this problem for me and I don't see any bad side effect. But I don't entirely understand the conditions on lines 1141 -- 1144, so this is a guess on my part. It would be great if you could test and reproduce the problem and solution.

It is clear that this implementation of scale deserves some comments. I will try to add some.

Regarding your findings, this is because without [this](https://github.com/WayfireWM/wayfire/blob/scale/plugins/single_plugins/scale.cpp#L1131-L1155), other code in wayfire will focus a view that isn't the one we are tracking as last_focused_view. (I can't recall the exact cases this happens but you can omit this signal handler (or never connect it) and see what happens.) We have to be careful though because calling focus_view() calls the focus handler and you can possibly get into an infinite loop when calling focus_view from the focus handler. This is why there are so many conditional statements at the beginning of the focus handler.

> 
> Thanks!

Issue #522, opened 2020-05-21
clicking on a surface during it's minimize animation, cancels minimization 


damianatorrpm:
Especially if the animation is long it's pretty easy to double click a minimize button 
and the behavior is unexpected, it finishes the animation than gets un-minimized immediately afterwards, sometimes it gets minimized after being activated again, sometimes not,
it seems it's not even necessary to be clicked on the flying out surface but just its original coordinates

EDIT: minimize is completely cancelled if you drag while it reappears due to original coordinates clicked

ammen99:
I am not sure I understand what happens. So you click on the minimize button twice effectively, and what happens?

damianatorrpm:
https://file.io/u1h5ygwb here a recording

ammen99:
Says "404 not found", wrong link?

damianatorrpm:
It expired so fast...
https://drive.google.com/open?id=11dejyBZglJhGQ-BqJxe5vxXc_z2gWaTM

ammen99:
Ah well you double-clicked on it, but the animation was already running - then it appeared that you are dragging the surface, which in turn means that it gets unminimized because reasons ...

I guess we could make the minimize animation disallow clicks on the view, should be enough to "fix" this.

damianatorrpm:
The "same" issue can be seen for workspace switching -Expo- as well,
but "reversed".

Even with the default animation duration for viewport switching there is ~200 ms when the grid appears already expanded but the animation has not finished. When you click during these 200 ms on any workspace it will just go back to the current.

damianatorrpm:
Another thing that cancels minimize animation is also when you try to minimize a view,
while the minimize animation of a different view is still in progress.

Minimizing a view while the minimize animation of a different view is still in progress,
will cancel the minimization of the different view.

ammen99:
Was fixed only partly, clicking on the view itself may still restore it.

damianatorrpm:
There is still one little issue with normal duration, the animation looks weird when you quickly double click a minimize button of an application (it partly restores)

PR #698, opened 2020-08-22
output: don't consider minimized views for refocus

soreau:


ammen99:
Didn't we discuss we should also prevent these views from having cursor focus?

soreau:
Yes but this particular PR is about fixing #522, not about the unrelated issue discussed on IRC.

soreau:
> You can add a check here in this function: https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/input-manager.cpp#L323

This should go in a different PR.

ammen99:
#522 is exactly what I described, please read the issue more carefully :)

ammen99:
But you're right, this is about a different bug. I'll merge this PR (but it doesn't fix #522)

PR #706, opened 2020-08-24
xwayland: add support for views that change OR status

### LINKED ISSUE #684: Support changing override-redirect views to normal views
When launching Call of duty waw or mw2 in fullscreen mode under proton/wine, it launches correctly in fullscreen mode but it doesn't "grab"(focus?) it, so there is no mouse or keyboard input to the application, and it ends up with the layer on top of everything, being unable to switch to other applications within a workspace. strangely this doesn't happen with other applications under proton/wine using the exact same version and drawing apis (dx9 -> vulkan)

Ive also found that wf-dock doesn't see the application, 
everything works as expected in windowed mode.  .

ammen99:
The easiest way to do that is just destroy the old view and create a new one of the correct type.

Fixes #684 

@ovlx Can you test this and see if it works?

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -713,7 +713,7 @@ void wayfire_xwayland_view_base::recreate_view_with_or_type()
      * Copy xw and mapped status into the stack, because "this" may be destroyed
      * at some point of this function.
      */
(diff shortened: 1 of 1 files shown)


ovlx:
works properly now
@ammen99

PR #715, opened 2020-08-30
keyboard: Populate lock key states after creating context

### LINKED ISSUE #589: [Feature request] add xkb_numlock support / odd behaviour
I have found no way to set Numlock to be enabled by default, and even if numlock is enabled at boot/ in the login manager, it is disabled upon starting the wayfire session.

also it seemingly gets disabled at random even if it is set manually after the session has started , I have been unable to figureout exactly what is causing this , so i cannot say for sure if this is a bug or some other issue caused by an application But i thought i'd mention it incase something can be set to force it globally.

soreau:
Fixes #589

### DIFF
--- a/src/core/seat/keyboard.cpp
+++ b/src/core/seat/keyboard.cpp
@@ -5,6 +5,7 @@ extern "C"
 {
 #include <xkbcommon/xkbcommon.h>
 #include <wlr/backend/session.h>
+#include <wlr/interfaces/wlr_keyboard.h>
 }
 
 #include <wayfire/util/log.hpp>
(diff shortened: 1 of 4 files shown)


src/core/seat/keyboard.cpp
```diff
@@ -294,6 +318,33 @@ bool input_manager::handle_keyboard_key(uint32_t key, uint32_t state)
...
+            kbd->xkb_state,
+            kbd->led_indexes[i]))
+        {
+            leds |= (1 << i);
+        }
+    }
+
+    if (leds & WLR_LED_NUM_LOCK)
+    {
+        wf::get_core_impl().input->locked_mods |= WF_KB_NUM;
```
ammen99, line 334: A small suggestion, 
```cpp
uint32_t& locked_mods = wf::get_core_impl().input->locked_mods;
```

Then you dont' have to type the whole thing each type (don't change if you're too lazy ;)

And please put this in a separate method.

ammen99 (commented):
My requests are mostly: put things in separate methods.

Otherwise LGTM.

ammen99 (approved):
Thanks!

PR #719, opened 2020-09-02
Drop xdg shell v6

ammen99:
In preparation for https://github.com/swaywm/wlroots/pull/2391


kode54:
And the above was merged into wlroots.

PR #723, opened 2020-09-04
[vswitch] fix for requesting the same workspace multiple times

dkondor:
Special case of issue #676 

My comment from there:

I've just found that this issue can persist in the edge case when the same workspace is requested twice in quick succession. I've been able to see it with the scale plugin with this patch by @soreau

I've added a simple print statement and I think what is happening is that for the second request, vswitch::add_direction() is called with x == y == 0, and thus returns false (line 140). This sets carried_out to false in on_set_workspace_request() (line 204).

In workspace-impl.cpp, line 910, then it will test this and call set_workspace() that instantly sets the new workspace. But the vswitch animation is still running, so it ends up on the wrong workspace in the end.

Maybe, add_direction() could return true if x == y == 0, and thus the operation is a no-op? This fixes in the cases I could test.

### DIFF
--- a/plugins/single_plugins/vswitch.cpp
+++ b/plugins/single_plugins/vswitch.cpp
@@ -135,11 +135,6 @@ class vswitch : public wf::plugin_interface_t
 
     bool add_direction(int x, int y, wayfire_view view = nullptr)
     {
-        if (!x && !y)
-        {
-            return false;
-        }
-
         if (!is_active() && !start_switch())
         {
             return false;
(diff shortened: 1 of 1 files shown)


soreau:
Can you force push your branch so that it contains only one commit?
For example, I would do the following:

1) `git format-patch HEAD~1` - produces a patch file for the topmost commit
2) `git fetch origin && git reset --hard origin/master` - reset to latest master branch
3) `git am <name of 0001-vswitch-fix-for-multiple.patch>` - reapply the patch produced in step 1) as a new commit
4) `git push -f <your remote> vswitch-fix2` - force push to your vswitch-fix2 branch

The result should be latest master with only the single relevant patch on top.

plugins/single_plugins/vswitch.cpp
```diff
@@ -137,7 +137,7 @@ class vswitch : public wf::plugin_interface_t
     {
         if (!x && !y)
```
ammen99, line 138: It would be better to move this check to https://github.com/WayfireWM/wayfire/pull/723/files#diff-7746de7cb4788c85484ac7396b90b98bR198 as a whole, it does not belong here (not sure why it was put it here in the first place). Otherwise LGTM.

dkondor:
New version with handling no-op in on_set_workspace_request(). Tested in the same way as before, fixes the issue with multiple requests.

ammen99 (approved):
Thanks!

PR #727, opened 2020-09-09
Activate fast-switcher only through configured key binding

wb9688:


ammen99 (approved):
Fixing bugs by mostly deleting code, this has my +1 :) 
Thanks!

PR #728, opened 2020-09-09
core: Disable max_render_time by default

soreau:
Many users complaining about wayfire performance were able to fix the
problem by setting max_render_time to -1, which disables it. This option
is also supposed to be a per-output setting so it could cause problems
where there is more than one monitor with different refresh rates.


ammen99 (approved):
Thanks!

PR #737, opened 2020-09-13
core: Use GL_DEPTH_COMPONENT with GL_UNSIGNED_INT when setting up dep…

soreau:
…th buffer

Some drivers like VC4 on RPI3 were failing on headless backend with
GL errors because we were using GL_DEPTH_COMPONENT32F with GL_FLOAT,
when setting up the depth buffer. This doesn't currently affect the
drm backend because this code is simply not used in that case. Here
we use a combination that is supported by more drivers.


ammen99:
Seems to work here as well (with swapchain+DRM)

ammen99:
Thanks :)

PR #740, opened 2020-09-16
main: use wayland sockets from wayland-1 to wayland-32

### LINKED ISSUE #738: Avoid wayland-0
This would be to work around libwayland-client's behavior, see https://gitlab.freedesktop.org/wayland/weston/commit/ae69381b6eb4. Relevant code in Wayfire is at: https://github.com/WayfireWM/wayfire/blob/5bdc2b20faa8897b02d7437091be52f0a6ed2174/src/main.cpp#L370-L376

ammen99:
Fixes #738

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -248,6 +248,20 @@ static void signal_handler(int signal)
 
 #endif
 
+static std::optional<std::string> choose_socket(wl_display *display)
+{
+    for (int i = 1; i <= 32; i++)
+    {
+        auto name = "wayland-" + std::to_string(i);
+        if (wl_display_add_socket(display, name.c_str()) >= 0)
+        {
+            return name;
+        }
+    }
+
+    return {};
+}
+
 int main(int argc, char *argv[])
 {
     config_dir = nonull(getenv("XDG_CONFIG_HOME"));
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -367,28 +381,28 @@ int main(int argc, char *argv[])
...
+        LOGE("Failed to initialize backend, exiting");
         wlr_backend_destroy(core.backend);
         wl_display_destroy(core.display);
 
         return -1;
     }
 
-    LOGI("running at server ", server_name);
-    setenv("WAYLAND_DISPLAY", server_name, 1);
+    setenv("_WAYLAND_DISPLAY", core.wayland_display.c_str(), 1);
```
wb9688, line 403: What's the point of this env var?

ammen99: I don't know lol, I copied that part from somewhere else and have not changed it since.

ammen99: Here is where this comes from: https://github.com/swaywm/rootston/blob/master/main.c#L46

However, sway doesn't set it.

ammen99: As emersion said on IRC, this is not really necessary, I removed it.

PR #769, opened 2020-10-06
support for sending the parent with wlr_foreign_toplevel_handle

dkondor:
Implementation for https://github.com/swaywm/wlr-protocols/pull/52

Requires https://github.com/swaywm/wlroots/pull/2419 -- see my description there for more details

### DIFF
--- a/src/view/view-impl.cpp
+++ b/src/view/view-impl.cpp
@@ -520,6 +520,11 @@ void wf::wlr_view_t::toplevel_send_state()
         tiled_edges == TILED_EDGES_ALL);
     wlr_foreign_toplevel_handle_v1_set_activated(toplevel_handle, activated);
     wlr_foreign_toplevel_handle_v1_set_minimized(toplevel_handle, minimized);
+
+    /* update parent as well */
+    wf::wlr_view_t *parent_ptr = dynamic_cast<wf::wlr_view_t*>(parent.get());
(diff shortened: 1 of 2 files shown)


src/api/wayfire/view.hpp
```diff
@@ -491,6 +491,11 @@ class view_interface_t : public surface_interface_t, public wf::object_base_t
      * or activated state changes */
     virtual void desktop_state_updated();
 
+    /**
+     * Called when the parent of a toplevel view is changed.
+     * Needed for the xdg_foreign_toplevel interface. */
+    virtual void parent_updated();
```
ammen99, line 497: Seems this can be grouped together with `desktop_state_updated()` to avoid having multiple functions (after all parent is kind of setting transient-for state)

dkondor: My intention was to only send the parent event if there is a change in parent (a strict interpretation of the protocol). To avoid adding to the interface, this check could be done in the wlroots code as well.
Also, I've noticed that Wayfire / wlroots sends a bunch of state events even if the state did not change; I might submit a separate PR to wlroots for that.

ammen99 (approved):
Looks good, can you also update the wlroots submodule (preferably in the same commit to make sure git bisect doesn't fail)?

ammen99:
Thanks!

PR #777, opened 2020-10-10
Xwayland position fix

### LINKED ISSUE #768: GIMP dialog windows misbehavior on multi-headed setup
This may be difficult to describe, but let's try. I work on a laptop with 2 external displays:

```text
[DP-1]
layout = 0,0
mode = 1920x1080@60.000000
scale = 1.000000
transform = normal

[eDP-1]
layout = 1920,0
mode = 1920x1080@120.000000
scale = 1.000000
transform = normal

[HDMI-A-1]
layout = 3840,0
mode = 1920x1080@60.000000
scale = 1.000000
transform = normal
```

If I open GIMP on `DP-1`, it works correctly on all workspaces. The problem appears on `eDP-1` and `HDMI-A-1`. I use the default 3x3 workspaces matrix. If I open GIMP on one of workspaces in the left column (1, 4 or 7), dialog popups are being moved to the left margin. On all other workspaces, they land somewhere off the screen. I use Waybar with the `wlr/taskbar` module. If I click on the button corresponding to the missing window, I'm being moved to the first left workspace in the same row. 

Sample steps to reproduce:

- launch GIMP on a display which coordinates are not 0,0 and workspace 2, 3, 5, 6, 8 or 9;
- open an image
- select Image / Canvas size
- change width or height, press Center

The dialog popup disappears from the screen.

soreau:
Fixes #768

### DIFF
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -193,28 +193,91 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
         wf::wlr_view_t::destroy();
     }
 
+    /* Translates geometry from X client configure requests to wayfire
+     * coordinate system. The X coordinate system treats all outputs
+     * as one big desktop, whereas wayfire treats the current workspace
+     * of an output as 0,0 and everything else relative to that. This
+     * means that we must take care when placing xwayland clients that
+     * request a configure after initial mapping, while not on the
+     * current workspace.
+     *
+     * @param output    The view's output
+     * @param ws_offset The view's workspace minus the current workspace
+     * @param geometry  The configure geometry as requested by the client
+     *
+     * @return Geometry with a position that is within the view's workarea.
+     * The workarea is the workspace where the view was initially mapped.
+     * Newly mapped views are placed on the current workspace.
+     */
+    wf::geometry_t translate_geometry_to_output(wf::output_t *output,
+        wf::point_t ws_offset,
+        wf::geometry_t g)
+    {
+        auto outputs = wf::get_core().output_layout->get_outputs();
+        auto og   = output->get_layout_geometry();
+        auto from = wf::get_core().output_layout->get_output_at(
+            g.x + g.width / 2 + og.x, g.y + g.height / 2 + og.y);
+        if (!from)
+        {
+            return g;
+        }
+
+        auto lg = from->get_layout_geometry();
+        g.x += (og.x - lg.x) + ws_offset.x * og.width;
+        g.y += (og.y - lg.y) + ws_offset.y * og.height;
+        if (!this->is_mapped())
+        {
+            g.x *= (float)og.width / lg.width;
+            g.y *= (float)og.height / lg.height;
+        }
+
+        return g;
+    }
+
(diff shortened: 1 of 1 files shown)


src/view/xwayland.cpp
```diff
@@ -198,23 +198,51 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
...
+                vg.y += current_workspace.y * og.height;
+                auto center = wf::point_t{vg.x + vg.width / 2, vg.y + vg.height / 2};
+                auto ws     =
+                    wf::point_t{int(center.x / og.width), int(center.y / og.height)};
+                wsg.x += (ws.x - current_workspace.x) * og.width;
+                wsg.y += (ws.y - current_workspace.y) * og.height;
+            }
+
+            if (view_impl->frame)
+            {
```
ammen99, line 232: You are looking for get_output_geometry() here. A badly named function, but it gives you the geometry without decorations.

soreau: I already have the decorationless size, because that's what we compute for the configure geometry. I actually need to know the titlebar height and the border size so that I can clamp the X geometry (recall that's without borders) to the workarea, before the borders are taken into account. Basically I have to shrink the workarea by the decoration sizes on their respective edges. get_output_geometry() will give me an old size anyway.

ammen99: Instead of comparing undecorated geometry to shrunk workspace geometry, you can just as well compare the expanded geometry to the whole workspace (the math works out if you think about it)

src/view/xwayland.cpp
```diff
@@ -193,28 +193,98 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
...
 
+    /* Translates geometry from X client configure requests to wayfire
+     * coordinate system. The X coordinate system treats all outputs
+     * as one big desktop, whereas wayfire treats the current workspace
+     * of an output as 0,0 and everything else relative to that. This
+     * means that we must take care when placing xwayland clients that
+     * request a configure after initial mapping, while not on the
+     * current workspace.
+     *
+     * params:
```
ammen99, line 204: Please follow the format of other functions.

```
@param <param1 name> <Description>
@param <param2 name> <Description>

@return <Return description>
```

src/view/xwayland.cpp
```diff
@@ -193,28 +193,93 @@ class wayfire_xwayland_view_base : public wf::wlr_view_t
...
+            g.x + g.width / 2 + og.x, g.y + g.height / 2 + og.y);
+        if (!from)
+        {
+            return g;
+        }
+
+        auto lg = from->get_layout_geometry();
+        g.x += (og.x - lg.x) + ws_offset.x * og.width;
+        g.y += (og.y - lg.y) + ws_offset.y * og.height;
+        if (!this->self()->is_mapped())
```
ammen99, line 230:
```suggestion
        if (!this->is_mapped())
```

ammen99 (approved):
LGTM overall, see my comments.

ammen99:
Thanks!

PR #793, opened 2020-10-16
Scale vswitch

ammen99:
@dkondor @soreau Please test once again.

The changes this time:

- You can change the current workspace with the bindings from vswitch, however, if scale is not active for all workspaces, scale exits.
- There is basic touchscreen support, however, I don't know if anyone except me can test that
- Changes in how button press works. Before if you were to accidentally click on a view, you can't revert. I changed it so that you can click on the view, but the action is carried out when the button is released (if the mouse is still on the same view).

PS: instead of the license from @soreau I just put you two as original authors. I believe this is more fair since he's not the only one working on the plugin ;)


dkondor:
Hi,

the new functionality seems to be working, I can switch workspaces. It is a bit strange that bringing the views to the new workspace seems to "lag" after the workspace switch animation, but it might also actually be good to have a better sense of what is happening.

I found a new issue with switching to the workspace of a selected view. E.g. in the situation in the screenshot below, if I start scale from one of the workspaces on the right and want to switch to the wcm view, I end up on the top left workspace instead of the top middle one.

I have the following suggestion for a fix:
```
diff --git a/plugins/scale/scale.cpp b/plugins/scale/scale.cpp
index a96dd172..19cae5dc 100644
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -510,8 +510,8 @@ class wayfire_scale : public wf::plugin_interface_t
         auto center = wf::point_t{vg.x + vg.width / 2, vg.y + vg.height / 2};
 
         return wf::point_t{
-            ws.x + ((center.x - ws.x * og.width) / og.width),
-            ws.y + ((center.y - ws.y * og.height) / og.height)};
+            ws.x + (int)std::floor((double)center.x / og.width),
+            ws.y + (int)std::floor((double)center.y / og.height)};
     }
 
     /* Given row and column, return a view at this position in the scale grid,
```
that fixes this issue for me and seems to work in other cases as well; although I don't completely understand the original logic here, so I might be missing something.

scale_layout

dkondor:
Also, Wayfire now regularly crashes on exiting, but this happens on the master branch as well (recently in ~compositor_core_impl_t).

ammen99:
> Also, Wayfire now regularly crashes on exiting, but this happens on the master branch as well (recently in ~compositor_core_impl_t).

Could you perhaps bisect and see what broke it?

ammen99:
> I have the following suggestion for a fix:
> 
> ```
> diff --git a/plugins/scale/scale.cpp b/plugins/scale/scale.cpp
> index a96dd172..19cae5dc 100644
> --- a/plugins/scale/scale.cpp
> +++ b/plugins/scale/scale.cpp
> @@ -510,8 +510,8 @@ class wayfire_scale : public wf::plugin_interface_t
>          auto center = wf::point_t{vg.x + vg.width / 2, vg.y + vg.height / 2};
>  
>          return wf::point_t{
> -            ws.x + ((center.x - ws.x * og.width) / og.width),
> -            ws.y + ((center.y - ws.y * og.height) / og.height)};
> +            ws.x + (int)std::floor((double)center.x / og.width),
> +            ws.y + (int)std::floor((double)center.y / og.height)};
>      }
>  
>      /* Given row and column, return a view at this position in the scale grid,
> ```

This is a correct fix, I'll add it to the commits.

dkondor:
> > Also, Wayfire now regularly crashes on exiting, but this happens on the master branch as well (recently in ~compositor_core_impl_t).
> 
> Could you perhaps bisect and see what broke it?
> 
> EDIT: or at least can you at least say what the last working commit was.

It seems the problem is introduced by b70215026cc6e0421e3302aeb8e61861fabf1aa3; also, the crash does no happen if disabling the window-rules plugin.

ammen99:
> > > Also, Wayfire now regularly crashes on exiting, but this happens on the master branch as well (recently in ~compositor_core_impl_t).
> > 
> > 
> > Could you perhaps bisect and see what broke it?
> > EDIT: or at least can you at least say what the last working commit was.
> 
> It seems the problem is introduced by [b702150](https://github.com/WayfireWM/wayfire/commit/b70215026cc6e0421e3302aeb8e61861fabf1aa3); also, the crash does no happen if disabling the window-rules plugin.

Thanks! I see what is going on, I'll think about what is the best fix.

Issue #256, opened 2019-07-08
Split input-manager class


ammen99:
It is too big.

PR #803, opened 2020-10-19
add compositor stages

### LINKED ISSUE #671: wlr_output_layout_output_at assertion prevents restarting after crash
Maybe it's a FreeBSD/ConsoleKit2/amdgpu/the-kind-of-crashes-I-have/??? thing (most likely consolekit2, it doesn't do any fancy vt/kms/etc recovery that logind might be doing??), but this assertion:

https://github.com/WayfireWM/wayfire/blob/0e157a5c418341d2dfbc2c2265a31761d3d8740c/src/core/output-layout.cpp#L1609

is usually not letting me start wayfire again after a hard crash.
For the first few early `get_output_coords_at` calls (that happen before the modesetting has finished), `wlr_output_layout_output_at` is returning NULL because the output's box is `0,0,0,0`.

Just commenting the assert out has helped.

ammen99:
- uncrustify: fix wrongly added newline between enum class and ident
- core: add information about compositor state to API
- core: refresh device mappings after the backend has started

This should fix #671, @myfreeweb do you have a way to verify the crash is fixed?

### DIFF
--- a/src/api/wayfire/core.hpp
+++ b/src/api/wayfire/core.hpp
@@ -34,6 +34,29 @@ class output_t;
 class output_layout_t;
 class input_device_t;
 
+/** Describes the state of the compositor */
+enum class compositor_state_t
+{
+    /** Not started */
+    UNKNOWN,
+    /**
+     * The compositor core has finished initializing.
+     * Now the wlroots backends are being started, which results in
+     * adding of new input and output devices, as well as starting the
+     * plugins on each output.
+     */
+    START_BACKEND,
+    /**
+     * The compositor has loaded the initial devices and plugins and is
+     * running the main loop.
+     */
+    RUNNING,
+    /**
+     * The compositor has stopped the main loop and is shutting down.
+     */
+    SHUTDOWN,
+};
+
 class compositor_core_t : public wf::object_base_t
 {
   public:
@@ -270,6 +293,11 @@ class compositor_core_t : public wf::object_base_t
      * @return The PID of the started client, or -1 on failure.
      */
     virtual pid_t run(std::string command) = 0;
+
+    /**
+     * @return The current state of the compositor.
+     */
+    virtual compositor_state_t get_current_state() = 0;
 
     /**
      * Shut down the whole compositor.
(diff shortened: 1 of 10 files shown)


myfreeweb:
I can restart a few times in a row if I do it carefully, but looks like it's possible without this patch too.

I can still fail that assertion by moving the mouse while starting Wayfire!

I'm not sure why you decided that it must be the mappings thing rather than real events. I think input processing just should be made to work with no display, including after unplugging an output..

ammen99:
> I can restart a few times in a row if I do it carefully, but looks like it's possible without this patch too.
> 
> I can still fail that assertion by moving the mouse while starting Wayfire!

I thought we don't get input events before the backend is started??

> I'm not sure why you decided that it must be the mappings thing rather than real events. I think input processing just should be made to work with no display, including after unplugging an output..

Yes, it should work with no display, output-layout creates a noop output if no other outputs are connected.

Based on your observations, it seems like Wayfire should simply have some handling to skip any input events before the backends are started.

myfreeweb:
But it's not before backends are started, it's with noop:
https://github.com/WayfireWM/wayfire/issues/671#issuecomment-670510655

> II 07-08-20 16:11:36.218 - [backend/noop/backend.c:16] Starting noop backend
> II 07-08-20 16:11:36.342 - [src/core/seat/input-manager.cpp:75] handle new input: ...
> EE 07-08-20 16:11:36.350 - [types/wlr_output_layout.c:275] output-layout-at 0.000000, 0.000000
> [assert]

ammen99:
> But it's not before backends are started, it's with noop:
> [#671 (comment)](https://github.com/WayfireWM/wayfire/issues/671#issuecomment-670510655)
> 
> > II 07-08-20 16:11:36.218 - [backend/noop/backend.c:16] Starting noop backend
> > II 07-08-20 16:11:36.342 - [src/core/seat/input-manager.cpp:75] handle new input: ...
> > EE 07-08-20 16:11:36.350 - [types/wlr_output_layout.c:275] output-layout-at 0.000000, 0.000000
> > [assert]

The noop backend is started before the noop backend is created, try again with the latest commit.

ammen99:
Note that the crash possibility is still there for touchscreen (for now), but I am working on #256 atm which will change things so that touchscreen events work the same way pointer and touch works.

myfreeweb:
Yes, fixed now.

ammen99:
> Yes, fixed now.

Perfect, thanks for testing.

PR #809, opened 2020-10-21
fix view geometry with multiple outputs and workspaces

### LINKED ISSUE #606: Maximized view wrong size after maximize > fullscreen > oswitch > unfullscreen

### LINKED ISSUE #625: quarter tiling bad unmaximize

ammen99:
- view: consider workspaces for tiling and fullscreening
- output-layout: fix moving views when output is disconnected
- grid: add support for maximizing on a workspace
- view: store windowed geometry before starting move

Fixes #606
Fixes #625

### DIFF
--- a/plugins/common/wayfire/plugins/common/move-snap-helper.hpp
+++ b/plugins/common/wayfire/plugins/common/move-snap-helper.hpp
@@ -138,6 +138,8 @@ class move_snap_helper_t : public wf::custom_data_t
         {
             end_wobbly(v);
         }
+
+        view->disconnect_signal("geometry-changed", &view_geometry_changed);
     }
 
     /** @return Whether the view is freely moving or stays at the same place */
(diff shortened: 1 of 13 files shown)


ammen99:
@damianatorrpm @soreau Please let me know if this PR completely fixes your issues.

@damianatorrpm This is in conflict with your open PR, I'll tell you how to update it once this is merged.

damianatorrpm:
@ammen99 I tested it now and also tested with multiple monitors.
I think the behavior is good, when I start drag from one monitor to tile it on another 
and later un-tile it has the same position on the new monitor like it was on the previous.

Views almost disappearing through the edge when un-tiling is fixed.

The behavior feels natural to me.
(ok, one exception where it felt a bit weird but this wouldn't happen in real work-flow,
when having a view  slightly overlapping with say bottom left corner and than quarter tile it to top right corner than un-tile,
but in a real work flow I would move the view first to the middle at some point first (release mouse button) before deciding to tile it, but I've used it for 5 min now, I am convinced that this would neither happen in real work-flow, and even if it did after a couple of times used, you'd not notice it)

PR #810, opened 2020-10-21
make minimized views per-workspace

### LINKED ISSUE #178: Figure out what to do when restoring a window on another output
Lets say we minimize the window on one output, and then a taskbar on another output wants to restore it. Do we move the window to the other output? Or we deny the request? Or we restore on the original output and change focused output?

ammen99:
- core: make minimized layer behave like the others
- view: make focus_request work better
- vswitch: switch workspace before starting animation

This has been requested in IRC before, and in #178.

With these changes, minimized views are "per-workspace". Clicking on a minimized view in wf-panel will go to the view's original workspace.

Fixes #178 (the view is focused on its original output and we switch to its original workspace).

@AdrianVovk if you want to override this behavior, you can create a plugin to do so:

1. Connect to view-focus-request.
2. If the view is on the focused output, let core handle the request.
3. Otherwise, set carried_out to false to block other plugins and core.
4. Move the view to the output and workspace you want it to be.
5. Call view->focus_request() there.

@damianatorrpm Sorry for the breaking change, but `view-self-focus-request` has been replaced with the more generic `view-focus-request`. You still have a flag to differentiate between a call from wf-panel and a call from gtk-shell.

### DIFF
--- a/plugins/vswitch/vswitch.cpp
+++ b/plugins/vswitch/vswitch.cpp
@@ -132,7 +132,7 @@ class vswitch : public wf::plugin_interface_t
 
         /* Make sure that when we add this direction, we won't go outside
          * of the workspace grid */
-        auto target = algorithm->get_target_workspace();
+        auto target = output->workspace->get_current_workspace();
         auto wsize  = output->workspace->get_workspace_grid_size();
         int tvx     = wf::clamp(target.x + delta.x, 0, wsize.width - 1);
         int tvy     = wf::clamp(target.y + delta.y, 0, wsize.height - 1);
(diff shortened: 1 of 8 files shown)


damianatorrpm:
I have checked what changed and adjusted (merge soon dbus interface).
One question regarding these lines
`https://github.com/WayfireWM/wayfire/pull/810/files#diff-78c74bc64602a8e4e537086adfd7ea3799ca51740c92ef471f5fc40dcc21fb15R496-R510`

You handle the focus request but do not set data->carried_out to true
should't it be set?

ammen99:
> You handle the focus request but do not set data->carried_out to true
> should't it be set to avoid being handled multiple times?

If you look carefully, it is handled after the signal is emitted. So plugins that want to override this behavior can set the carried_out to true, so that core does not handle the request.

Core itself does not need to set carried_out to true because it is the last entity which can handle the request at all.

PR #815, opened 2020-10-22
readme: mention seatd

myfreeweb:
wlroots has added a libseat backend a while ago now


ammen99 (approved):
LGTM, but please also update the wiki :)

PR #818, opened 2020-10-23
Dont break fs 2

damianatorrpm:
continuation on recreated branch from https://github.com/WayfireWM/wayfire/pull/633

### DIFF
--- a/plugins/common/wayfire/plugins/common/move-snap-helper.hpp
+++ b/plugins/common/wayfire/plugins/common/move-snap-helper.hpp
@@ -147,13 +147,19 @@ class move_snap_helper_t : public wf::custom_data_t
          */
         if (view->fullscreen)
         {
-            wf::output_t *output   = view->get_output();
+            wf::output_t *output   = wf::get_core().get_active_output();
+            wf::geometry_t output_geometry = output->get_relative_geometry();
             wf::point_t current_ws = output->workspace;
-            wf::point_t target_ws  =
+            wf::point_t target_ws;
+            target_ws.x =
                 floor(
-                    (last_grabbing_position.x / view->get_output_geometry().width) +
+                    (last_grabbing_position.x / output_geometry.width) +
                     current_ws.x);
-            view->fullscreen_request(view->get_output(), true, target_ws);
+            target_ws.y =
+                floor(
+                    (last_grabbing_position.y / output_geometry.height) +
+                    current_ws.y);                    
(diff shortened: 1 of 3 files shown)


damianatorrpm:
before reviewing => I will need to rebuild OS and test the changes

plugins/single_plugins/move.cpp
```diff
@@ -324,7 +324,21 @@ class wayfire_move : public wf::plugin_interface_t
             return;
         }
 
-        /* Snap the view */
+        /**
+         * Snap the view
```
ammen99, line 328: Shouldn't you move this comment down below? (line 713)

plugins/common/wayfire/plugins/common/move-snap-helper.hpp
```diff
@@ -140,6 +140,27 @@ class move_snap_helper_t : public wf::custom_data_t
...
 
         view->disconnect_signal("geometry-changed", &view_geometry_changed);
+        /**
+         * Restore the fullscreen state so the window is
+         * not in a weird position afterwards, fullscreen
+         * are by spec the geometry of the output (or span outputs)
+         */
+        if (view->fullscreen)
+        {
+            wf::output_t *output = wf::get_core().get_active_output();
```
ammen99, line 150: you should use `view->get_output()`

ammen99 (commented):
This one looks good, I have the following to add: ~~http://ix.io/2BHe~~ http://ix.io/2BHf  add it to the pr and then we can merge.

plugins/common/wayfire/plugins/common/move-snap-helper.hpp
damianatorrpm: I haven't tested it yet, but I'm not sure what's supposed to happen
If I use the views output and I drag the fullscreen view 
from output A (no expo) to output B (expo active),
I would need to include some calculation for the ws on output B?

ammen99: if you drag the view to output B then view->get_output() will already point to output B, this is why I said that the view's output shoukd be used.

ammen99: also don't worry about expo or no expo, the move-snap-helper takes care of that

ammen99 (approved):

ammen99:
Thanks!

Issue #822, opened 2020-10-25
[current master] Popup windows kill Waybar


nwg-piotr:
Opening a popup window (e.g. About in any GTK app, or a dialog in GIMP) makes Waybar disappear.

https://youtu.be/txwnw7q9HWI

ammen99:
Have you updated wlroots to latest master?

nwg-piotr:
Well, I just used the wf-install script.

ammen99:
I just checked, this is a bug in Waybar.

https://github.com/Alexays/Waybar/blob/master/src/modules/wlr/taskbar.cpp#L660

It always binds to newest version (which is 3) but it does not support that version. It should bind to `min(2u, version)` not just `version`.

You should open a bug there.

ammen99:
Actually, Waybar already has a bug opened; https://github.com/Alexays/Waybar/issues/897

PR #823, opened 2020-10-25
wlroots: bump subproject version

ammen99:
It contains fixes related to #822

@nwg-piotr This will not fix the bug in Waybar, but it will fix the issue for other well-behaved clients.


nwg-piotr:
All right, I'll wait on sway for the Waybar issue to be fixed.

PR #828, opened 2020-10-28
seat: restore mod_binding_key support after 8bbbcc72a

myfreeweb:
Currently this is the only way for plugins to e.g. distinguish whether the '<shift>' binding was pressed by the left or right Shift key.

This unbreaks my mod2key plugin after #806, though I'm planning on adding '<lshift>' etc. bindings for that.


ammen99:
why was this closed?

myfreeweb:
> the modifier key is not a really good idea but I suppose you want to have your plugin working with 0.6.0 :)

> eh i don't really care about releases yet

If it's not a good idea, let's not do it.

Issue #572, opened 2020-06-30
Expand the area that triggers a quarter tiling


SnooSnoo-ui:
The area that is calculated before `move.snap_threshold `. 

![resize](https://user-images.githubusercontent.com/67649453/86161416-99dd9680-bae3-11ea-87df-e4f903c22102.jpg)

For comparison this are some settings you have with KDE. Would be nice to have a setting for quarter tiling too. Also before snap  threshold I believe.

![kde](https://user-images.githubusercontent.com/67649453/86161495-b679ce80-bae3-11ea-8ada-9150249d6228.png)

Edit: snap_threshold not snap_off_threshold

ammen99:
So you want two have two separate options for the threshold when snapping to half-screen and when snapping to quarter-screen?

SnooSnoo-ui:
I think all the base thresholds (excluding custom config changes) you have now are good except the one for quarter tilling. It's the only one you need a exact spot I believe it's 10x10 pixels to trigger it. That area, the small red squares from my image, should be expanded similarly to how I illustrated with the blue lines.

I'll leave my suggestion at that because I don't know how it interacts with the other config options, and I might make my suggestion even more confusing.

ammen99:
Oh, I finally understand.

PR #841, opened 2020-10-31
hotspot support

### LINKED ISSUE #298: add hotcorner support
add hot corner support same as kwin.
where we can let cursor touch the corner and a command gets activated 
for eg.
window overview.desktop grid.show desktop.desktop cube.
etc
at top left corner. and other corners.

ammen99:
This PR adds basic support for hotspots. The required changes for wf-config are not merged yet,
see https://github.com/WayfireWM/wf-config/tree/add-hotspots.

Hotspots are currently supported as part of activator bindings. For example,

```ini
[expo]
toggle = hotspot top-left 100x10 1000
```

The meaning of the example: the hotspot is in the top-left corner.
It spawns an area of 100 pixels along both edges, with a "width" of 10 pixels, i.e. the same as explained
here: https://github.com/WayfireWM/wayfire/issues/572

The last parameter means that the cursor needs to stay for `1000` milliseconds in the hotspot to activate it.

Fixes #298


damianatorrpm:
I'm building now with this as well to test, after thinking about it,
it would make more sense to not do special handling here,
but instead for example scale (the only plugin I care w/hotspots for)
to have an option to deny these request if the activator source is hotspot.

damianatorrpm:
works fine with the exception that changing options to an hotspot requires wayfire restart
at runtime changing
```
[scale]
toggle = <super> | hotspot top-left 10x100 100
```

to 

```
[scale]
toggle = <super> | hotspot bottom-left 10x100 100
```
has no effect

ammen99:
> works fine with the exception that changing options to a hotspot requires wayfire restart
> at runtime changing

Ah yes I forgot to recreate hotspots on config reload, I'll fix it.

damianatorrpm:
hotspot reload still doesn't work 
```
[scale]
toggle = <super> | hotspot top-left 10x100 100
```
to

```
[scale]
toggle = <super> | hotspot bottom-left 10x100 100
```
I used branch hotspot-support for build or do other things need to applied (except wf-config)

ammen99:
@damianatorrpm I think I fixed the reloading issue, can you try again? (also update wf-config because the bug was there, the fix was pushed on the same branch)

damianatorrpm:
@ammen99 Yep, work fine

PR #842, opened 2020-10-31
layer-shell: handle commits before map

### LINKED ISSUE #302: Notifications from Mako stop displaying after some time
the mako process is running and the notifications seem to be getting to it, according to 'makoctl list'

this behavior did not replicate on my sway-git r6047.aa8ea2fa-1 installation from the AUR

### LINKED ISSUE #382: set_size is handled incorrectly
When a layer shell surface sets all anchors and sends a `set_size` of (0, 0) Wayfire never responds with an `ack_configure` thus causing a crash of the application. I don't actually know if this considered a protocol error as it only specifies what to do if one of the sizes is set to zero; however, it does function on sway so I'm assuming it is an error.

ammen99:
This is needed to configure the view, otherwise it does not know the correct size it should have.

Fixes #382
Fixes #302 (I think)

### DIFF
--- a/src/view/layer-shell.cpp
+++ b/src/view/layer-shell.cpp
@@ -162,11 +162,8 @@ struct wf_layer_shell_manager
         return result;
     }
 
-    void set_exclusive_zone(wayfire_layer_shell_view *v)
+    int find_exclusive_edge(int edges)
     {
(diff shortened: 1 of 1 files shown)


ammen99:
Actually, this is completely wrong, we already handle these cases in another place ...

PR #845, opened 2020-10-31
Add OpenGL::program_t::uniform3f

myfreeweb:
For plugin usage; maybe also `find_uniform_loc` should be exposed in the outer non-private object but for now I just need this one

### DIFF
--- a/src/api/wayfire/opengl.hpp
+++ b/src/api/wayfire/opengl.hpp
@@ -320,6 +320,8 @@ class program_t : public noncopyable_t
     /** Set the given uniform for the currently used program. */
     void uniform2f(const std::string& name, float x, float y);
     /** Set the given uniform for the currently used program. */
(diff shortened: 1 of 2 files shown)


ammen99 (approved):
Needs a bump in the plugin API/ABI otherwise LGTM.

myfreeweb:
hmmm, why? Adding a new non-virtual member function [doesn't break ABI](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B), and `WAYFIRE_API_ABI_VERSION` is not usable for `>=` checks…

ammen99:
> hmmm, why? Adding a new non-virtual member function [doesn't break ABI](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B), and `WAYFIRE_API_ABI_VERSION` is not usable for `>=` checks…

Hmm, ok, I suppose I am overly cautious.

PR #855, opened 2020-11-06
blur: Add saturation option

soreau:
This option controls the saturation of the blurred image.

### DIFF
--- a/plugins/blur/blur-base.cpp
+++ b/plugins/blur/blur-base.cpp
@@ -26,13 +26,23 @@ static const char *blur_blend_fragment_shader =
 precision mediump float;
 
 @builtin@
+uniform float sat;
 uniform sampler2D bg_texture;
 
 varying mediump vec2 uvpos[2];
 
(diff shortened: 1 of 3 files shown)


plugins/blur/blur-base.cpp
```diff
@@ -232,6 +244,7 @@ void wf_blur_base::render(wf::texture_t src_tex, wlr_box src_box,
     /* Blend blurred background with window texture src_tex */
     blend_program.uniformMatrix4f("mvp", glm::inverse(target_fb.transform));
     /* XXX: core should give us the number of texture units used */
+    blend_program.uniform1f("sat", saturation_opt);
```
ammen99, line 247: please move this before the comment, because the comment is related to the next line

ammen99 (approved):
LGTM with one small nitpick

plugins/blur/blur-base.cpp
soreau: Ah, nice catch.

ammen99:
Thanks!

PR #856, opened 2020-11-08
allow minimized views in scale

damianatorrpm:

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -95,7 +95,8 @@ class wayfire_scale : public wf::plugin_interface_t
     wf::option_wrapper_t<bool> middle_click_close{"scale/middle_click_close"};
     wf::option_wrapper_t<double> inactive_alpha{"scale/inactive_alpha"};
     wf::option_wrapper_t<bool> allow_scale_zoom{"scale/allow_zoom"};
-
+    wf::option_wrapper_t<bool> show_minimized{"scale/show_minimized"};
+    
     /* maximum scale -- 1.0 means we will not "zoom in" on a view */
     const double max_scale_factor = 1.0;
     /* maximum scale for child views (relative to their parents)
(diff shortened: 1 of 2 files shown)


plugins/scale/scale.cpp
```diff
@@ -691,32 +692,59 @@ class wayfire_scale : public wf::plugin_interface_t
...
 
+        uint32_t layer_mask = wf::LAYER_WORKSPACE;
+        if (show_minimized)
+        {
+            layer_mask |= wf::LAYER_MINIMIZED;
+        }
+
         for (auto& view :
-             output->workspace->get_views_in_layer(wf::LAYER_WORKSPACE))
+             output->workspace->get_views_in_layer(layer_mask))
```
damianatorrpm, line 709: It would have been possible to query all views in get_views() but that would require an additional for loop which I think is unnecessary

plugins/scale/scale.cpp
```diff
@@ -691,32 +692,59 @@ class wayfire_scale : public wf::plugin_interface_t
...
             {
                 continue;
             }
 
+            if (show_minimized)
+            {
+                if (output->workspace->get_view_layer(view) == wf::LAYER_MINIMIZED)
+                {
+                    output->workspace->add_view(view, wf::LAYER_WORKSPACE);
+                    view->store_data(
```
damianatorrpm, line 721: it would have been possible to store this as a vector, but I think storing custom data is always better since any plugin or core can always have information about custom view state

damianatorrpm: although on second thought, a vector loop would only iterate over these particular views.

damianatorrpm:
Works fine for me now.

damianatorrpm:
So when animation duration is too long, after starting scale, some views fly out (if you spam scale quickly enough).
That doesn't happen if no animation is used, tbh the only way for animation to look visually appealing is if 
the scale end/re-minimize happens in parallel. Apart from that any other comments/suggestions?

ammen99:
I looked over this, and I have another idea: Scale can simply move the views from the minimized layer to the workspace layer (and then back), preserving their minimized state. In addition, it should animate the alpha value of such views from 0 to whatever they should have.

I think  this would be better because
- the view stays minimized
- we avoid animations of the minimized views, I think that could look weird if you have a lot of minimized views and start all-workspace-scale.

soreau:
> I looked over this, and I have another idea: Scale can simply move the views from the minimized layer to the workspace layer (and then back), preserving their minimized state. In addition, it should animate the alpha value of such views from 0 to whatever they should have.
> 
> I think this would be better because
> 
>     * the view stays minimized
> 
>     * we avoid animations of the minimized views, I think that could look weird if you have a lot of minimized views and start all-workspace-scale.

Yes I agree with this; I had a similar line of thought.
To be clear, there are two parts to minimize, the layer and the state. If the view is in the minimize layer, it simply is not rendered at all. The view minimize state flag is independent of the layer and can be used for cases like these.

ammen99:
Closing since in order to fix this issue, another approach should be taken.

PR #873, opened 2020-11-15
programmable activators

### LINKED ISSUE #655: Programatically callable activators
Would be nice to have a way to programatically invoke activators, maybe even with a custom data struct.

Also, it would make sense to expose the plugin cancelling mechanism to plugins too.

### LINKED ISSUE #833: Make activator bindings receive a struct instead of `uint32_t`
Activator bindings should receive information about what invoked  them with a struct, not as a pair of activator source + a uint. The uint is just not enough in all current and future cases. For example, with #655 plugins could even pass custom data to activators, if needed.

To not make this a breaking change, we should have both the new and old activator callbacks for some time, and probably remove the old support sometime before 1.0

ammen99:
~~Still WIP, but should eventually:~~

fix #655
fix #833

This PR adds the long-awaited support for calling arbitrary plugins via their activator bindings. Example patch to make Expo activate on all workspaces at the same time (although I recommend actually doing an external plugin which activates all Expos on all outputs):

```diff
diff --git a/plugins/single_plugins/expo.cpp b/plugins/single_plugins/expo.cpp
index 2d4e11b6..8092898b 100644
--- a/plugins/single_plugins/expo.cpp
+++ b/plugins/single_plugins/expo.cpp
@@ -35,8 +35,23 @@ class wayfire_expo : public wf::plugin_interface_t
         return wf::point_t{x, y};
     }
 
-    wf::activator_callback toggle_cb = [=] (auto)
+    wf::activator_callback toggle_cb = [=] (auto ev)
     {
+        // Avoid self-loop
+        if (ev.source != wf::activator_source_t::PLUGIN)
+        {
+            wf::activator_data_t data;
+            data.source = wf::activator_source_t::PLUGIN;
+
+            for (auto& wo : wf::get_core().output_layout->get_outputs())
+            {
+                if (wo != output) {
+                    wo->call_plugin("expo/toggle", data);
+                }
+            }
+        }
+
         if (!state.active)
         {
             return activate();
```

### DIFF
--- a/plugins/blur/blur.cpp
+++ b/plugins/blur/blur.cpp
@@ -254,7 +254,7 @@ class wayfire_blur : public wf::plugin_interface_t
         method_opt.set_callback(blur_method_changed);
 
         /* Toggles the blur state of the view the user clicked on */
-        button_toggle = [=] (uint32_t, int, int)
+        button_toggle = [=] (auto)
         {
             if (!output->can_activate_plugin(grab_interface))
             {
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -113,17 +113,17 @@ class wayfire_cube : public wf::plugin_interface_t
 
         reload_background();
 
-        activate_binding = [=] (uint32_t, int, int)
+        activate_binding = [=] (auto)
         {
             return input_grabbed();
         };
 
-        rotate_left = [=] (wf::activator_source_t, uint32_t)
+        rotate_left = [=] (auto)
         {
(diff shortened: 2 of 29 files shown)


ammen99:
@myfreeweb I took your changes from #828 and put them in the appropriate place. Now you can use an activator binding in caps2esc I think.

The reason to use activator bindings is that there you actually can differentiate what triggered the binding - and modifier and keybinding are separate triggers.

damianatorrpm:
Also testing this now inse hotspot is merged

PR #877, opened 2020-11-17
Reload window-rules on config change

marcusbritanicus:
This PR enables the `window-rules` plugin to auto-reload all the rules when config file is modified without having to restart wayfire.

### DIFF
--- a/plugins/window-rules/window-rules.cpp
+++ b/plugins/window-rules/window-rules.cpp
@@ -29,12 +29,41 @@ class wayfire_window_rules_t : public wf::plugin_interface_t
     void setup_rules_from_config();
     wf::lexer_t _lexer;
 
-    wf::signal_callback_t _created;
-    wf::signal_callback_t _maximized;
-    wf::signal_callback_t _unmaximized;
-    wf::signal_callback_t _minimized;
-    wf::signal_callback_t _fullscreened;
-    wf::signal_callback_t reload_config;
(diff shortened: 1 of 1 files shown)


plugins/window-rules/window-rules.cpp
```diff
@@ -26,13 +26,15 @@ class wayfire_window_rules_t : public wf::plugin_interface_t
...
   private:
+    void setup_rules_from_config();
     wf::lexer_t _lexer;
 
     wf::signal_callback_t _created;
     wf::signal_callback_t _maximized;
     wf::signal_callback_t _unmaximized;
     wf::signal_callback_t _minimized;
     wf::signal_callback_t _fullscreened;
+    wf::signal_callback_t reload_config;
```
ammen99, line 37: A small thing: `wf::signal_callback_t` is deprecated, new code should use `wf::signal_connection_t` (and also whenever possible, please initialize lambda as soon as you declare it, I think it should be possible here).

ammen99 (commented):
LGTM otherwise from the comment I left.

plugins/window-rules/window-rules.cpp
marcusbritanicus: Ah I see. I will make changes and make a commit when I'm a little free. I more or less copied the code from commands plugin.

ammen99 (approved):
Thanks!

PR #880, opened 2020-11-17
switcher: remove expression with side effects from assert().

### LINKED ISSUE #879: Building Wayfire with -Db_ndebug=true breaks the switcher plugin
**Describe the bug**

When the switcher plugin is activated, windows move across the screen as expected, but releasing the keyboard completely doesn't quit the switcher plugin; instead, I can interact with one of the windows (only type, no mouse interaction), but I can't move workspaces anymore or open new applications. The only way to solve is by quitting wayfire.

**To Reproduce**

- Build Wayfire with `-Db_ndebug=true`, which adds a `-DNDEBUG` to the compiler command line.
- Open any number of windows and try to use the switcher plugin

**Expected behavior**

The switcher plugin should behave normally, as it does with `-Db_ndebug=false` builds.

**Wayfire version**

0.6.0

ericonr:
Since assert() is a macro, building wayfire with -DNDEBUG leads to the
entire expression being compiled out, which means it's never run. This
made it so that once the switcher plugin was launched, it wouldn't quit.

Fixes #879

### DIFF
--- a/plugins/single_plugins/switcher.cpp
+++ b/plugins/single_plugins/switcher.cpp
@@ -201,7 +201,8 @@ class WayfireSwitcher : public wf::plugin_interface_t
             active = true;
 
             // grabs shouldn't fail if we could successfully activate plugin
-            assert(grab_interface->grab());
+            auto grab = grab_interface->grab();
+            assert(grab);
 
             focus_next(dir);
             arrange();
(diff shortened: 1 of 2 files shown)


ammen99 (approved):
A good catch!

PR #886, opened 2020-11-20
render-manager: attempt to directly scanout a view

### LINKED ISSUE #874: Direct scanout support
Wayfire should try to directly scanout for ex. fullscreen views.

This should help with performance in fullscreen games, or power consumption with a fullscreen terminal, etc.

ammen99:
Fixes #874

Scanning out a view is possible only in a narrow set of cases:

- It must not interfere with plugins
- The view must have the correct size, buffer parameters, etc.


ammen99:
I have been testing this in the last few days, seems to work fine.

PR #889, opened 2020-11-20
wm-actions: Add binding to show desktop

### LINKED ISSUE #340: windows actions like minimize,close,maximize,showdesktop minimize all
so here are some things that i cant find a way to do in wayfire and some changes to defaults.
1.close window default keyboard shortcut key should be alt + KEY_F4 instead of alt+ KEY_FN_F4
2.there is no keyboard shortcut or a way to set it to minimize current active/infocus window.it should be in wcm>general
3.there is no keyboard shortcut to maximize current window.
it should also be in wcm>general
4.show desktop keyboard shortcut
is missing.
and it should work as 
pressing it again should show previous state 
4.minimize all windows.
5.option to open all windows maximized

soreau:
This is a simple implementation that relies on minimize to show the desktop.
Closes #340.

### DIFF
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -8,7 +8,10 @@
 class wayfire_wm_actions_t : public wf::plugin_interface_t
 {
     nonstd::observer_ptr<wf::sublayer_t> always_above;
+    bool showdesktop_active = false;
 
+    wf::option_wrapper_t<wf::activatorbinding_t> toggle_showdesktop{
+        "wm-actions/toggle_showdesktop"};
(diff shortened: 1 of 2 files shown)


plugins/wm-actions/wm-actions.cpp
```diff
@@ -200,11 +238,60 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
...
+        }
+
+        disable_showdesktop();
+
+        return true;
+    };
+
+    void disable_showdesktop()
+    {
+        view_attached.disconnect();
```
ammen99, line 272: no need to disconnect twice, otherwise, LGTM

ammen99 (approved):

PR #890, opened 2020-11-20
vswitch: Offer a way to switch directly to a workspace

soreau:
Use config construct 'select_workspace_# = <binding>' where # is the workspace number.
Heavily inspired by the same feature in expo.

### DIFF
--- a/plugins/vswitch/vswitch.cpp
+++ b/plugins/vswitch/vswitch.cpp
@@ -4,6 +4,16 @@
 #include <wayfire/util/log.hpp>
 
 
+static bool begins_with(std::string word, std::string prefix)
+{
+    if (word.length() < prefix.length())
+    {
+        return false;
+    }
+
+    return word.substr(0, prefix.length()) == prefix;
+}
+
 class vswitch : public wf::plugin_interface_t
 {
   private:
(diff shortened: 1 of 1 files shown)


plugins/vswitch/vswitch.cpp
```diff
@@ -77,6 +88,49 @@ class vswitch : public wf::plugin_interface_t
                 return false;
             }
         });
+
+        std::vector<std::string> workspace_numbers;
+        const std::string select_prefix = "select_workspace_";
+        auto section = wf::get_core().config.get_section("vswitch");
+        for (auto binding : section->get_registered_options())
```
ammen99, line 95: Isn't it simpler to iterate over all available workspaces and then query the config for the appropriate option, instead of trying to parse options?

soreau: Yes, I just copied the principle from expo.

ammen99 (changes requested):
- Do we care enough about dynamically changing the values of the option?
- This code should be in vswitch.hpp, so that these bindings may work in Scale or other plugins which opt-in too
- Switching to a workspace with window?
- Moving a window to a workspace without switching? :P 

If you don't want to work on the last two then it is ok to leave them for later.

To support all of these cases, we should probably expand the control bindings callback here: https://github.com/WayfireWM/wayfire/blob/master/plugins/vswitch/wayfire/plugins/vswitch.hpp#L281

It should have a bitfield parameter which indicates what we want to achieve
- move to particular workspace, to adjacent workspace or not really switch
- move with or without the grabbed window

ammen99:
Closing in favor of #1199, I have implemented the missing functionality there. Thanks for working on this though :)

Issue #891, opened 2020-11-21
Sending keyboard modifier state to inactive and layer-shell views with mouse clicks


dkondor:
Hi,

I'm opening this as a question, since this has become a complicated problem for a simple issue :)

My main problem is detecting if one of the modifier keys are pressed during a mouse click. This does not seem to work for non-focused views, including layer-shell views without the keyboard-interactivity option set.

E.g. cairo-dock has the feature where holding the shift key while clicking on the icon of a running app starts a second instance. This does not work on Wayfire, since the fact that shift is pressed is not reported. Separately, Ctrl-click on a currently unfocused view does not work (e.g. open a link in a new tab on Firefox).

After a bit of investigating, I see two causes:

1. Keyboard focus for layer-shell surfaces is too restricted in my opinion. If a layer-shell surface is on the top layer, it either grabs all keyboard events or nothing (depending whether keyboard interactivity [is set](https://github.com/swaywm/wlr-protocols/blob/992841c2aa731ab02774c9bf8c6cda9f24dfbf10/unstable/wlr-layer-shell-unstable-v1.xml#L206) ). This makes sense for views in the overlay layer, such as a lock screen, but I think there is rationale for layer-shell views that behave similarly to "always-on-top" views, i.e. allow the keyboard focus to be lost. (alternatively, a layer-shell surface could set / unset keyboard interactivity whenever the pointer enters / leaves, but that goes against the typical behavior for keyboard focus). I believe any change here would require changing the protocol definition -- I can open a separate issue for this on wlr-protocols.

2. Non-focused views do not receive active modifiers with mouse clicks and also, setting keyboard focus on click only works if no modifiers are pressed. This is true for non-layer-shell views as well, so left-clicking on a view sets the focus, but e.g. clicking with shift or ctrl pressed will not, since that does not match the focus change binding [here](https://github.com/WayfireWM/wayfire/blob/e1379a4c3ad702325278fcde5b0583037b346489/src/core/wm.cpp#L118). In this case, the view receives the mouse click, but not the state of the modifier (since it never receives keyboard focus). This can be tested e.g. with Firefox, where ctrl + click should open a link in a new tab, but this does not work if Firefox is not the active view. 

[...]

PR #896, opened 2020-11-23
core: Add option to focus a view on mouse click even if modifiers are pressed

### LINKED ISSUE #657: make middle-click also focus the view below the mouse
wayfire supports the middle click (what's its name) secondary paste,
but it is broken in the even of pasting into unfocused window.

the normal kde, gnome, x, etc behavior is when you 
1) mark something in window X and window X is focused
2) middle click into window Y text 
3) window Y becomes focused  and text is pasted

Current behavior:
Nothing happens

EDIT: sway behavior is mouse follows focus so I guess not relevant

dkondor:
This addresses the second part of my issue #891 by allowing a view to be focused by a modifier + mouse click. In my opinion, this behavior is more consistent. Also, this ensures that the clicked view receives the modifier state (since it is always focused).

I've implemented this as an option, with the old behavior being the default.

Depends on https://github.com/WayfireWM/wf-config/pull/36

Currently, click-to-focus only works for mouse clicks without keyboard modifiers. If a modifier is held down, the necessary button binding in the focus handler does not match, so the focus does not change. The view under the mouse receives the mouse click, but not the state of the keyboard modifiers (since it never receives keyboard focus).

This commit adds an option to essentially ignore pressed modifiers for the purpose of focusing the clicked view, so this will work consistently regardless of the state of keyboard modifiers.

Fixes #657

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -58,15 +58,10 @@
 			<_long>Allow focusing the clicked view even if keyboard modifiers are pressed. Without this option, click-to-focus only works if no modifiers are pressed.</_long>
 			<default>false</default>
 		</option>
-		<option name="focus_btn_middle" type="bool">
-			<_short>Focus on click with middle mouse button</_short>
-			<_long>Allow focusing the clicked view with the middle mouse button.</_long>
-			<default>false</default>
-		</option>
-		<option name="focus_btn_right" type="bool">
-			<_short>Focus on click with right mouse button</_short>
-			<_long>Allow focusing the clicked view with the right mouse button.</_long>
-			<default>false</default>
+		<option name="focus_btns" type="activator">
+			<_short>Mouse buttons to focus views</_short>
+			<_long>Clicking on a view with any of the mouse buttons of these will focus it.</_long>
+			<default>BTN_LEFT | BTN_MIDDLE | BTN_RIGHT</default>
 		</option>
 	</plugin>
 </wayfire>
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -53,12 +53,12 @@
 			<_long>Sets the compositor render delay in milliseconds, which allows applications to render with low latency.</_long>
 			<default>-1</default>
 		</option>
-		<option name="focus_btn_mod" type="bool">
+		<option name="focus_button_with_modifiers" type="bool">
 			<_short>Focus on click if keyboard modifiers are pressed</_short>
 			<_long>Allow focusing the clicked view even if keyboard modifiers are pressed. Without this option, click-to-focus only works if no modifiers are pressed.</_long>
 			<default>false</default>
 		</option>
-		<option name="focus_btns" type="activator">
+		<option name="focus_buttons" type="activator">
 			<_short>Mouse buttons to focus views</_short>
 			<_long>Clicking on a view with any of the mouse buttons of these will focus it.</_long>
 			<default>BTN_LEFT | BTN_MIDDLE | BTN_RIGHT</default>
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -53,5 +53,20 @@
 			<_long>Sets the compositor render delay in milliseconds, which allows applications to render with low latency.</_long>
 			<default>-1</default>
 		</option>
+		<option name="focus_btn_mod" type="bool">
+			<_short>Focus on click if keyboard modifiers are pressed</_short>
+			<_long>Allow focusing the clicked view even if keyboard modifiers are pressed. Without this option, click-to-focus only works if no modifiers are pressed.</_long>
+			<default>false</default>
+		</option>
+		<option name="focus_btn_middle" type="bool">
+			<_short>Focus on click with middle mouse button</_short>
+			<_long>Allow focusing the clicked view with the middle mouse button.</_long>
+			<default>false</default>
+		</option>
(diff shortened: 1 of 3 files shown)


ammen99:
I would much rather implement it in a different way. Instead of relying on button bindings, just listen for direct button press signals. This would also solve #657

dkondor:
> I would much rather implement it in a different way. Instead of relying on button bindings, just listen for direct button press signals. This would also solve #657

I was thinking of this (but this needs more changes); I'll look into this.

dkondor:
This is actually simpler than I expected. What I actually don't like with this change is that it splits the focusing implementation between core/wm.cpp and core/seat/pointer.cpp -- I wonder if it would be better to keep the main functionality in wm.cpp and call it from pointer.cpp

ammen99:
> This is actually simpler than I expected. What I actually don't like with this change is that it splits the focusing implementation between core/wm.cpp and core/seat/pointer.cpp -- I wonder if it would be better to keep the main functionality in wm.cpp and call it from pointer.cpp

You don't have to do that - there are events that the focus plugin can listen to.

https://github.com/WayfireWM/wayfire/blob/87d9e5bdc7e1df157e238e607d390a3cbbd81328/src/api/wayfire/signal-definitions.hpp#L82

dkondor:
> You don't have to do that - there are events that the focus plugin can listen to.
> 
> https://github.com/WayfireWM/wayfire/blob/87d9e5bdc7e1df157e238e607d390a3cbbd81328/src/api/wayfire/signal-definitions.hpp#L82

I see; I've created an alternative version with this.

My problem in this case is that there is no way to not change focus if the click was handled by a plugin; this can mess up things if a plugin wants to manipulate focus. E.g. this is a problem for my [mouse gesture plugin](https://github.com/dkondor/wstroke). To mitigate this, I've added options to enable / disable focus behavior separately with the middle and right mouse buttons. This solves the problem in my case (since I use a right click to start gestures -- I could also add an option there to change focus), but it might come up with other plugins that try to change focus.

ammen99:
> My problem in this case is that there is no way to not change focus if the click was handled by a plugin; this can mess up things if a plugin wants to manipulate focus.

How is this different from before?

src/core/wm.cpp
```diff
@@ -109,14 +111,26 @@ void wayfire_focus::init()
...
+            return;
+        }
+
+        if ((!focus_modifiers && wf::get_core().get_keyboard_modifiers()) ||
+            ((ev->event->button == BTN_MIDDLE) && !focus_btn_middle) ||
+            ((ev->event->button == BTN_RIGHT) && !focus_btn_right))
+        {
+            return;
+        }
+
```
ammen99, line 130: There are other buttons, for example BTN_EXTRA, BTN_FORWARD, etc.
I don't think we should change focus with such buttons.

Also, what do you think of creating an activator binding, which is used to match the button? I.e the user would list all the buttons they want to cause focus change, and then this focus plugin here will just match the button against the activator.

dkondor: Yes, that's a good idea; it actually makes the result more flexible. I've made this into an option. There are two small UI issues with this though:
 - any modifiers set are ignored (although I don't know why anyone would want to set them; but this could be easily added)
 - when changing this option in wcm, there is a scary warning message about setting a mouse button binding without any modifiers

dkondor:
> > My problem in this case is that there is no way to not change focus if the click was handled by a plugin; this can mess up things if a plugin wants to manipulate focus.
> 
> How is this different from before?

Before, this issue could not come up, since only the left button was matched (it is unlikely that would be set as a plugin activator). With the default behavior, there shouldn't be any change though.

In my previous version, when the focus logic was in pointer.cpp, it could be skipped if a binding matched.

Anyway, it is working for my use case, and for all the official plugins (I don't think any of them tries to set the focus away from the clicked view) and if anyone else has this issue, that could be solved separately.

metadata/core.xml
```diff
@@ -53,5 +53,15 @@
 			<_long>Sets the compositor render delay in milliseconds, which allows applications to render with low latency.</_long>
 			<default>-1</default>
 		</option>
+		<option name="focus_btn_mod" type="bool">
+			<_short>Focus on click if keyboard modifiers are pressed</_short>
+			<_long>Allow focusing the clicked view even if keyboard modifiers are pressed. Without this option, click-to-focus only works if no modifiers are pressed.</_long>
+			<default>false</default>
+		</option>
+		<option name="focus_btns" type="activator">
```
ammen99, line 61: Can we rename this to `focus_buttons` ?

metadata/core.xml
```diff
@@ -53,5 +53,15 @@
 			<_long>Sets the compositor render delay in milliseconds, which allows applications to render with low latency.</_long>
 			<default>-1</default>
 		</option>
+		<option name="focus_btn_mod" type="bool">
```
ammen99, line 56: `focus_button_with_modifiers`

ammen99 (approved):
I'd prefer the full names of the bindings.

Otherwise, LGTM.

dkondor:
> I'd prefer the full names of the bindings.
Sure, I've adjusted it

> Otherwise, LGTM.
Should I squash everything to one commit?

ammen99:
Don't worry, I can squash the commits when merging.

ammen99:
Thanks!

PR #899, opened 2020-11-25
add custom modeline interlacing support

### LINKED ISSUE #893: Add custom interlaced mode support
Currently wayfire doesn't parse the interlaced flag when using custom modes , this is inherited from the rootston code

I have created this very incompetent patch, because it seemed pretty trivial, and it works correctly so far. 
it needs changes since im not a developer and do not know C or CPP.
_deleted_ (this was stupid)
[interlaced.patchfixed.txt](https://github.com/WayfireWM/wayfire/files/5581351/interlaced.patchfixed.txt) (this one isn't as stupid)

You might be able to  add double-scan mode support the same way, but that's a pretty niche thing, dont know if any modern drivers support it.

ovlx:
hopefully this is acceptable .

Fixes #893

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -126,7 +126,7 @@ static bool parse_modeline(const char *modeline, drmModeModeInfo & mode)
     char hsync[16];
     char vsync[16];
     char interlace[16];
-    interlace[0] = '\0'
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Looks good, thanks!

(will merge once the CI build finishes)

PR #901, opened 2020-11-26
core: put xwayland surfaces on top when giving them pointer focus

### LINKED ISSUE #666: [xwayland] stack order is not sent to Xserver
Iif you scroll the mousewheel on the sub-layered (keep_above) view
but it has not focus, the underlying view receives the scroll event.

e.g Maximize chrome
vlc half size of but on top
focus chrome
mousewheel on vlc => chrome scrolls

The normal behavior is that even if not focused views still act on scroll events.

ammen99:
Fixes #666

Unfortunately, this does not work yet.
The problem is restacking - the `xwayland_bring_to_front` function doesn't seem to do anything.

I'll try to look into it later, but any hints why it doesn't work are much appeciated.

### DIFF
--- a/src/core/seat/pointer.cpp
+++ b/src/core/seat/pointer.cpp
@@ -129,6 +129,8 @@ void wf::pointer_t::update_cursor_focus(
     }
 
     cursor_focus = focus;
+    seat->ensure_input_surface(focus);
+
     wlr_surface *next_focus_wlr_surface = nullptr;
     if (focus && focus->get_wlr_surface())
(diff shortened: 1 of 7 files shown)


ammen99:
Investigation didn't bring much results, but it seems like this could work if the function for restacking is called with the xcb connection from wlroots (hints why this is happening would be appreciated if someone knows the reason).

With the pending PR https://github.com/swaywm/wlroots/pull/2500 it works.

PR #902, opened 2020-11-28
Invert without changing colors

pushqrdx:
This makes the invert plugin only operate on black/white keeping the hue as is.

### DIFF
--- a/metadata/invert.xml
+++ b/metadata/invert.xml
@@ -9,5 +9,10 @@
 			<_long>Inverts the colors of the whole output with the specified activator.</_long>
 			<default>&lt;super&gt; KEY_I</default>
 		</option>
+		<option name="preserve_hue" type="bool">
+			<_short>Preserve Hue</_short>
+			<_long>Only operate on black and white, preserving the hue</_long>
+			<default>false</default>
+		</option>
 	</plugin>
 </wayfire>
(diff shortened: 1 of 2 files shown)


soreau:
The invert plugin inverts colors, it's even in [the description](https://github.com/WayfireWM/wayfire/blob/master/metadata/invert.xml#L5) and [the icon](https://github.com/WayfireWM/wcm/blob/master/icons/plugins/plugin-invert.svg).

I think it might be nicer to have a plugin that installs the shaders as files and loads them, first checking for some location in $HOME and falling back to the installed files. This way, users would be able to write their own shader files, custom tailored to fit their needs. Bonus points to recompile the shaders when they are change on disk. This would be for shaders that are applied to the entire output.

Another plugin could apply shaders based on a match+shader-file-path scheme for per-view shaders.

soreau:
Perhaps we can hide this feature behind an option though, maybe a 'Preserve Hue' bool option.

pushqrdx:
> Perhaps we can hide this feature behind an option though, maybe a 'Preserve Hue' bool option.

Makes more sense to me, as it's a feature that belongs to invert. I have changed the code with the patch you provided, thanks! :D

soreau:
Did you mean to push your fast-switcher commit to another branch? It doesn't really seem to belong in this PR..

ammen99 (approved):
Thanks!

PR #903, opened 2020-11-28
render-manager: add an algorithm for smarter repaint delay

ammen99:
Some time ago, @myfreeweb added the option for repaint delay. While it is good enough for basic scenarios, it is in general very difficult to say what the maximum render time will be (it depends on plugins, etc.).   This is why I thought it might be interesting to experiment with an algorithm which adjusts the repaint delay dynamically. This is off by default, and can be enabled with `workarounds/dynamic_repaint_delay`.

The basic idea is this: the user sets a very low `max_render_time`, maybe even 0. Wayfire's algorithm then tries to increase or decrease delay based on whether the last frames were on time or not.

The goals when designing this algorithm were: The algorithm should periodically try to increase the delay (faster at startup, slower later), however, this should not be often enough to be noticeable.

In case frames are missed, we should rapidly start decreasing the delay, to accomodate for effects like Expo or Cube which suddenly require much higher render time for Wayfire.

I have also explained the algorithm in detail as a source code comment. That being said, this is just an idea and I'd like feedback from people who care about this. I'm also sure the algorithm could be improved and/or simplified, right now it is a bunch of conditions.

### DIFF
--- a/metadata/workarounds.xml
+++ b/metadata/workarounds.xml
@@ -26,5 +26,10 @@
 			<_long>If false, the main window can be focused even if it has a dialog. The dialog is nevertheless kept on top of the main window. Kwin defaults to false.</_long>
 			<default>true</default>
 		</option>
+    <option name="dynamic_repaint_delay" type="bool">
+      <_short>Allow dynamic repaint delay</_short>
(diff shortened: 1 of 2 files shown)


ammen99:
Seems to working according to my tests.

PR #906, opened 2020-11-29
Error no longer logged for move and resize window rules

JordanL2:
Before this change, all move and resize window rules would wrongly log this error:

"View action interface: Unsupported action execution requested. Name: "


ammen99 (approved):
Thanks!

Issue #852, opened 2020-11-03
[fast-switcher] switch views by recently focused instead of by stack order


ammen99:
When switching with fast-switcher, views should be ordered in order of last used, not topmost in the stack.

kontura:
> When switching with fast-switcher, views should be ordered in order of last used, not topmost in the stack.

It seems to already behave in this way, since a used view is always put on top doesn't this order the view stack by last used?

Or is the problem that something else is modifying the view stack and so the plugin should store/get the information about usage order in a different way?

ammen99:
> Or is the problem that something else is modifying the view stack and so the plugin should store/get the information about usage order in a different way?

Yes, for example simple-tile will put multiple views in the same sublayer. This would bring all tiled views to the top when only one of them has been active.

Also there are cases like the follow-focus plugin which can modify the last-used order without changing the stack order.

travankor:
fixed by #907?

ammen99:
ah, yes indeed.

PR #907, opened 2020-11-29
Fast switcher uses focus order instead of stack order

### LINKED ISSUE #569: Add an option for window switcher to cycle back and forth
A suggestion to optionally make it work like common Linux desktops and Windows. Single activation alt + tab goes back and forth between same two windows. To switch to the next you need to press tab again while alt is still pressed.

>On Windows 10, when you use the "Alt + Tab" keyboard shortcut, you'll open the task switcher, which is an experience that allows you to view and switch back and forth between the two most recent applications or any other running task, by pressing and holding the "Alt" key and using the "Tab" key to cycle through.

If possible it should skip animation and preview if key is released quickly to not flicker the screen.

pushqrdx:
this should fix #569 and makes Fast Switcher plugin more intuitive to use

### DIFF
--- a/plugins/single_plugins/fast-switcher.cpp
+++ b/plugins/single_plugins/fast-switcher.cpp
@@ -12,8 +12,9 @@
 class wayfire_fast_switcher : public wf::plugin_interface_t
 {
     wf::option_wrapper_t<wf::keybinding_t> activate_key{"fast-switcher/activate"};
-    size_t current_view_index;
     std::vector<wayfire_view> views; // all views on current viewport
+    size_t current_view_index  = 0;
+    size_t previous_view_index = 0;
 
     bool active = false;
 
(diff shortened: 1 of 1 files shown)


ammen99:
That seems more like a quick workaround than a real fix. You should sort views by their last focused timestamp, and that will give you the result you want for free ;)

ammen99:
Not to mention that previous_view_index could either disappear, a plugin might reorder views, or something else.

ammen99:
I looked through the code and I think it might require a bit more work. The function `update_views` refreshes the list of views, you should sort the views there according to last focused timestamp.

Also, you should make sure that fast-switcher **doesn't** call `focus_view` except when it is exiting, otherwise, this would change the last focus order.

pushqrdx:
@ammen99 what about this, i didn't change any code that calls `focus_view` and it seems to work, the order doesn't get messed up and it toggles correctly

pushqrdx:
@ammen99 I have been using it for a while now, the latest edits i pushed sort views according to their time stamp which does exactly what's expected, should fix #852

ammen99:
LGTM. Can you however extend the function update_views instead of removing it?

ammen99 (approved):
Thanks!

PR #912, opened 2020-11-30
Emit signal on xdg.show_window_menu

### LINKED ISSUE #911: show_window_menu Signal
It'd be nice for there be a hook into xdg-shell's show_window_menu for desktop environments to draw window controls like stay on top, move workspace, etc.

soreau:
Closes #911.

### DIFF
--- a/src/api/wayfire/signal-definitions.hpp
+++ b/src/api/wayfire/signal-definitions.hpp
@@ -504,11 +504,11 @@ using title_changed_signal = _view_signal;
 using app_id_changed_signal = _view_signal;
 
 /**
- * name: show-window-menu
+ * name: view-show-window-menu
  * on: output
  * when: To show a menu with window related actions.
  */
-struct show_window_menu_signal : public _view_signal
+struct view_show_window_menu_signal : public _view_signal
 {
     /** The position as requested by the client, in surface coordinates */
     wf::point_t relative_position;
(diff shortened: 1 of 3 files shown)


src/api/wayfire/signal-definitions.hpp
```diff
@@ -503,6 +503,13 @@ using title_changed_signal = _view_signal;
  */
 using app_id_changed_signal = _view_signal;
 
+/**
+ * name: show-window-menu
+ * on: view
+ * when: To show a menu with window related actions.
+ */
+using show_window_menu_signal = _view_signal;
+
```
ammen99, line 512: Shouldn't this also have the coordinates where the click happened?

src/core/core.cpp
```diff
@@ -735,6 +735,13 @@ void wf::compositor_core_impl_t::erase_view(wayfire_view v)
     views.erase(it);
 }
 
+void wf::compositor_core_impl_t::show_window_menu(wayfire_view view)
```
ammen99, line 738: Why do we need a function in core?

soreau: This is because other things can call this, for example foreign-toplevel could have a show_window_menu request, so that a menu can be shown in response to right-clicking on a button in the window-list of wf-panel, instead of the gtk menu that is there currently, which is suboptimal. Another use-case could be that the window-menu plugin could detect click-on-desktop to open a menu as well.

soreau: We could emit the signal directly and same for foreign-toplevel, then for other things like click-on-desktop menu, we can just run() a menu and match it by app_id.

ammen99: Yes, I think we should emit the signal directly where applicable (note that plugins could emit this too, for click-on-desktop)

ammen99 (approved):
LGTM, and thanks!

PR #913, opened 2020-11-30
Place plugin now uses signal_connection_t API

### LINKED ISSUE #905: Place plugin breaks Window-Rules plugin
If the `place` plugin is enabled, any `on created then move` rules for the `window-rules` plugin won't do anything.

This is because the `place` plugin moves all newly created windows (to the centre by default), and this is triggered *after* the on-create move rule is executed, meaning the window is just immediately moved back to the centre of the screen.

Disabling the `place` plugin fixes this, but it would be nice to have the default placement functionality from `place` with `window-rules` overriding the default position for specific windows, presumably this means the `window-rules` `on created` rules triggered after the `place` rule.

JordanL2:
This fixes #905

### DIFF
--- a/plugins/single_plugins/place.cpp
+++ b/plugins/single_plugins/place.cpp
@@ -8,12 +8,50 @@ class wayfire_place_window : public wf::plugin_interface_t
 {
     wf::signal_connection_t created_cb = [=] (wf::signal_data_t *data)
     {
-        do_created(data);
+        auto ev   = (wf::view_mapped_signal*)(data);
+        auto view = get_signaled_view(data);
+
+        if ((view->role != wf::VIEW_ROLE_TOPLEVEL) || view->parent ||
+            view->fullscreen || view->tiled_edges || ev->is_positioned)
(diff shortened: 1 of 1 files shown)


plugins/single_plugins/place.cpp
```diff
@@ -6,8 +6,14 @@
 
 class wayfire_place_window : public wf::plugin_interface_t
 {
-    wf::signal_callback_t created_cb;
-    wf::signal_callback_t workarea_changed_cb;
+    wf::signal_connection_t created_cb = [=] (wf::signal_data_t *data)
+    {
+        do_created(data);
```
ammen99, line 11: You should put the whole code inside the lambda, this is how we usually do it.

JordanL2: Oh ok, will do. I was looking at the window-rules plugin for comparison, but I guess that was done that way to reuse the same function.

ammen99:
Thanks!

PR #914, opened 2020-11-30
added missing sample call that was missing in #902

pushqrdx:
fixes the issue introduced in #902


ammen99 (approved):
Should have reviewed more carefully :) 

Will test this in a bit.

Issue #797, opened 2020-10-17
scale: better sorting algorithm


ammen99:
Consider the following:

1. Open 2 terminal windows (window 1 and 2)
2. Use grid to tile window 2 on the left
3. Use grid to tile window 1 on the right
4. Start scale

At least here, the pointer to window 1 is smaller as integer value than the pointer to window 2 (if not, just swap left and right window). Because of this when scaling the right view gets column 1 and the left view gets column 2, which doesn't have the optimal visual result (ideally, the columns the views get correspond to their geometry).

This seems to be a problem which is far from trivial to solve, and I do not know what the optimal algorithm is, hence the `help wanted` label.

epsilon-0:
Is this the issue you are mentioning - https://i.imgur.com/FyYbPTk.mp4

ammen99:
Yes this is exactly what I mean.

travankor:
I've noticed that scale will cause wayfire to temporarily freeze if there are too many views to process -- hopefully this can fix that.

ammen99:
> I've noticed that scale will cause wayfire to temporarily freeze if there are too many views to process -- hopefully this can fix that.

What is too many here? 20, 100, or 1000?

travankor:
>20 around when CPU usage becomes quite large
~100 is when wayfire becomes unresponsive for more than a few seconds
And my computer can't handle 1000 :-)

PR #917, opened 2020-12-03
better sorting algorithm for scaling

epsilon-0:
Signed-off-by: epsilonKNOT <floss@bsd.ac>

---

Possibly closes https://github.com/WayfireWM/wayfire/issues/797

---

~~after pondering, it seems like the sorting algo will depend on the final displaying algorithm. `wcnt` = number of windows per row, which is currently set to `ceil(sqrt(views.size()))`, should be changed to reflect the actual number of windows per row.~~
I should spend more time reading rather than getting high and writing code, this was right beside the code where i was tinkering...

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
             target_alpha);
     }
 
+    static bool view_compare_x(const wayfire_view& A, const wayfire_view& B)
+    {
+        auto vg_a = A->get_wm_geometry();
+        std::vector<int> a_coords = {vg_a.x, vg_a.width, vg_a.y, vg_a.height};
+        auto vg_b = B->get_wm_geometry();
+        std::vector<int> b_coords = {vg_b.x, vg_b.width, vg_b.y, vg_b.height};
+        return a_coords < b_coords;
+    }
+
+    static bool view_compare_y(const wayfire_view& A, const wayfire_view& B)
+    {
+        auto vg_a = A->get_wm_geometry();
+        std::vector<int> a_coords = {vg_a.y, vg_a.height, vg_a.x, vg_a.width};
+        auto vg_b = B->get_wm_geometry();
+        std::vector<int> b_coords = {vg_b.y, vg_b.height, vg_b.x, vg_b.width};
+        return a_coords < b_coords;
+    }
+
+    std::vector<std::vector<wayfire_view>> view_sort(
+        std::vector<wayfire_view>& views)
+    {
+        std::vector<std::vector<wayfire_view>> view_grid;
+        std::sort(views.begin(), views.end(), view_compare_y);
+        size_t n = views.size();
+        size_t i = 0, j;
+        while (i < n)
+        {
+            j = std::min(i + grid_cols, n);
+            std::sort(views.begin() + i, views.begin() + j, view_compare_x);
+            view_grid.push_back(std::vector<wayfire_view>(views.begin() + i,
+                views.begin() + j));
+            i += grid_cols;
+        }
+
+        return view_grid;
+    }
+
     /* Compute target scale layout geometry for all the view transformers
      * and start animating. Initial code borrowed from the compiz scale
      * plugin algorithm */
@@ -817,7 +854,7 @@ class wayfire_scale : public wf::plugin_interface_t
         y = workarea.y + (int)spacing;
         height = (workarea.height - (lines + 1) * (int)spacing) / lines;
 
-        std::sort(views.begin(), views.end());
+        view_sort(views);
 
         for (i = 0; i < lines; i++)
         {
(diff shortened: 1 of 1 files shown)


plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
             target_alpha);
     }
 
+    static bool view_compareX(const wayfire_view& A, const wayfire_view& B)
+    {
+        auto vgA = A->get_wm_geometry();
```
soreau, line 792: We try to steer away from camelCase but instead use snake_case. Here I would probably use ...`vg_a = view_a->`...

plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
...
+            return vgA.x < vgB.x;
+        }
+
+        return vgA.y < vgB.y;
+    }
+
+    static void view_sort(std::vector<wayfire_view>& views, int wcnt)
+    {
+        std::sort(views.begin(), views.end(), view_compareY);
+        int n = (int)views.size();
```
soreau, line 817:
```suggestion
        size_t n = views.size();
```

plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
...
+        int i = 0, j;
```
soreau, line 818: No need for casts if we use native types.

```suggestion
        size_t i = 0, j;
```

plugins/scale/scale.cpp
```diff
@@ -817,7 +854,7 @@ class wayfire_scale : public wf::plugin_interface_t
         y = workarea.y + (int)spacing;
         height = (workarea.height - (lines + 1) * (int)spacing) / lines;
 
-        std::sort(views.begin(), views.end());
+        view_sort(views, grid_cols);
```
soreau, line 857: Since grid_cols is a class member variable, if you make view_sort() a class member function, you don't have to pass grid_cols explicitly. This also makes the need for the misleadingly named `wcnt` variable to be tossed out.

```suggestion
        view_sort(views);
```

plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
...
+        auto vgB = B->get_wm_geometry();
+        if (vgA.y == vgB.y)
+        {
+            return vgA.x < vgB.x;
+        }
+
+        return vgA.y < vgB.y;
+    }
+
+    static void view_sort(std::vector<wayfire_view>& views, int wcnt)
```
soreau, line 814:
```suggestion
    void view_sort(std::vector<wayfire_view>& views)
```

plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
...
+    {
+        std::sort(views.begin(), views.end(), view_compareY);
+        int n = (int)views.size();
+        int i = 0, j;
+        while (i < n)
+        {
+            j = std::min(i + wcnt, n);
```
soreau, line 821:
```suggestion
            j = std::min(i + grid_cols, n);
```

plugins/scale/scale.cpp
```diff
@@ -787,6 +787,43 @@ class wayfire_scale : public wf::plugin_interface_t
...
+            std::sort(views.begin() + i, views.begin() + j, view_compareX);
+            i += wcnt;
```
soreau, line 823:
```suggestion
            i += grid_cols;
```

soreau (approved):
Works amazingly well! I only noticed a few things that could be reordered to make the code nicer.

plugins/scale/scale.cpp
epsilon-0: done

plugins/scale/scale.cpp
epsilon-0: done

plugins/scale/scale.cpp
epsilon-0: done

plugins/scale/scale.cpp
epsilon-0: done

plugins/scale/scale.cpp
epsilon-0: don

plugins/scale/scale.cpp
epsilon-0: done

plugins/scale/scale.cpp
epsilon-0: done

ammen99:
I added more bugs and made stuff slower /s

Will merge once CI finishes. Thanks!

Issue #898, opened 2020-11-24
After Idle, all views moved to one output


JordanL2:
This is with wayfire 0.6.0 from AUR on Arch.
I have two monitors connected via DisplayPort, and an AMD GPU (using default Mesa drivers).
Output config:
```
[DP-1]
mode = 3840x2160@60.000000
layout = 0,0
scale = 2

[DP-2]
mode = 3840x2160@60.000000
layout = 1920,0
scale = 2
```

When the idle plugin turns my outputs off, when they turn back on after I move my mouse/press key, all the windows are moved to one monitor.

Below is the log, but a TLDR of what happens is:

1. I reactivate the outputs by pressing key
2. DP-1 disconnects - this moves all windows to DP-2
3. DP-2 disconnects - this moves all windows to NOOP-1
4. DP-1 reconnects and is added back to Wayfire
5. DP-2 reconnects and is added back to Wayfire
6. All windows are moved from NOOP-1 to DP-1

```
12:18:59: II 24-11-20 12:18:59.576 - [backend/drm/drm.c:811] Modesetting 'DP-1' with '3840x2160@59997mHz'
12:18:59: II 24-11-20 12:18:59.592 - [backend/drm/drm.c:811] Modesetting 'DP-2' with '3840x2160@59997mHz'
12:19:01: II 24-11-20 12:19:01.060 - [backend/drm/drm.c:1256] Scanning DRM connectors
12:19:01: II 24-11-20 12:19:01.060 - [backend/drm/drm.c:1431] 'DP-1' disconnected
12:19:01: II 24-11-20 12:19:01.060 - [backend/drm/drm.c:1621] Emitting destruction signal for 'DP-1'
12:19:01: II 24-11-20 12:19:01.130 - [wayfire-0.6.0/src/core/output-layout.cpp:1116] remove output: DP-1
12:19:01: EE 24-11-20 12:19:01.130 - [wayfire-0.6.0/src/core/output-layout.cpp:551] disabling output: DP-1
12:19:01: II 24-11-20 12:19:01.131 - [wayfire-0.6.0/src/core/output-layout.cpp:177] transfer views from DP-1 -> DP-2
12:19:01: II 24-11-20 12:19:01.221 - [backend/drm/drm.c:1256] Scanning DRM connectors
12:19:01: II 24-11-20 12:19:01.221 - [backend/drm/drm.c:1431] 'DP-2' disconnected
[...]

PR #923, opened 2020-12-06
Output layout preserve plugin

JordanL2:
This is to restore the view geometry after DPMS, to fix the issue of all views being moved to one output on wake (#898).

Current issues:

- Calc text is blurry if it's non-tiled after being restored
- Rare crash, details in comment below
- Focusing last output code has arbitrary 10000ms value in the code, make it a config option?


src/api/wayfire/view.hpp
```diff
@@ -168,6 +177,24 @@ class view_interface_t : public surface_interface_t, public wf::object_base_t
      */
     virtual wlr_box get_bounding_box();
 
+    last_output_info_t last_output_info;
+
+    /**
```
ammen99, line 182: This would be better implemented with custom data on the view, see get_data/store_data/etc in object.hpp

ammen99: This way, the struct also doesn't have to be in the view.hpp header file and can be kept private to the plugin

JordanL2: Done.

JordanL2:
@ammen99 - To try to fix the issue with Calc losing its HiDPI, I thought I'd try triggering the restoration with the output's "start-rendering" signal rather than "output-added" to see if moving the window after rendering had started would help. 
```
output->connect_signal("start-rendering", &restore_views_to_output);
```
Problem is, this signal is never actually emitted. It seems it's emitted if the number of inhibits on rendering is raised above zero, then dropped back down to zero. This doesn't happen for me though, `render-manager::impl::add_inhibit` is never called at all. Should this signal be emitted at the end of the `render_manager::impl` constructor or something?

ammen99:
Yeah, start-rendering is a weird signal.

What you see is most likely because the Calc surface somehow loses track of which output it is on. Can you check whether https://github.com/WayfireWM/wayfire/blob/7fc2bd940e9ed08ce2b102dec262a9120dd16046/src/view/surface.cpp#L361 is called with the correct output(s) ?

ammen99:
> * Z-order isn't restored, this would be nice, as well as currently focused view being restored

Yeah, that is one of the more difficult parts to restore. You could try storing an index for each view ... and also mark the view as focused if it was focused.

damianatorrpm:
If you need ideas how to store/restore z-order, you can take a look here:
https://github.com/damianatorrpm/wayfire-plugin_dbus_interface/blob/master/dbus_interface_backend.cpp#L1549-L1671

JordanL2:
Thank you. So am I right in understanding the Z-order is determined by the order of `workspace->get_views_in_layer`?

ammen99:
> Thank you. So am I right in understanding the Z-order is determined by the order of `workspace->get_views_in_layer`?

Yes, this function returns the views from top to bottom in stack order.

JordanL2:
I'm occasionally getting this crash, just after the NOOP-1 output is removed after the screens have been recreated.

```
DD 13-12-20 13:04:27.617 - [src/core/core.cpp:583] focusing layer 0
DD 13-12-20 13:04:27.617 - [src/core/core.cpp:583] focusing layer 0
DD 13-12-20 13:04:27.618 - [src/core/core.cpp:583] focusing layer 0
DD 13-12-20 13:04:27.618 - [src/core/core.cpp:583] focusing layer 0
EE 13-12-20 13:04:27.619 - [src/main.cpp:233] Fatal error: Segmentation fault
EE 13-12-20 13:04:27.671 - #1  signal_handler(int) ../src/main.cpp:235
EE 13-12-20 13:04:27.685 - #2  __restore_rt sigaction.c:?
EE 13-12-20 13:04:27.730 - #3  std::__uniq_ptr_impl<wf::workspace_manager, std::default_delete<wf::workspace_manager> >::_M_ptr() const /usr/include/c++/10.2.0/bits/unique_ptr.h:173
EE 13-12-20 13:04:27.780 - #4  std::unique_ptr<wf::workspace_manager, std::default_delete<wf::workspace_manager> >::get() const /usr/include/c++/10.2.0/bits/unique_ptr.h:422
EE 13-12-20 13:04:27.826 - #5  std::unique_ptr<wf::workspace_manager, std::default_delete<wf::workspace_manager> >::operator->() const /usr/include/c++/10.2.0/bits/unique_ptr.h:417
EE 13-12-20 13:04:27.883 - #6  wf_layer_shell_manager::arrange_unmapped_view(wayfire_layer_shell_view*) ../src/view/layer-shell.cpp:295
EE 13-12-20 13:04:27.934 - #7  wayfire_layer_shell_view::initialize()::{lambda(void*)#5}::operator()(void*) const ../src/view/layer-shell.cpp:377
EE 13-12-20 13:04:27.992 - #8  void std::__invoke_impl<void, wayfire_layer_shell_view::initialize()::{lambda(void*)#5}&, void*>(std::__invoke_other, wayfire_layer_shell_view::initialize()::{lambda(void*)#5}&, void*&&) /usr/include/c++/10.2.0/bits/invoke.h:60
EE 13-12-20 13:04:28.051 - #9  ...) /usr/include/c++/10.2.0/bits/invoke.h:116
EE 13-12-20 13:04:28.098 - #10 std::_Function_handler<void (void*), wayfire_layer_shell_view::initialize()::{lambda(void*)#5}>::_M_invoke(std::_Any_data const&, void*&&) /usr/include/c++/10.2.0/bits/std_function.h:292
EE 13-12-20 13:04:28.132 - #11 std::function<void (void*)>::operator()(void*) const /usr/include/c++/10.2.0/bits/std_function.h:622
EE 13-12-20 13:04:28.169 - #12 wf::wl_listener_wrapper::emit(void*) ../src/util.cpp:524
EE 13-12-20 13:04:28.206 - #13 handle_wrapped_listener(wl_listener*, void*) ../src/util.cpp:460
EE 13-12-20 13:04:28.225 - #14 wlr_signal_emit_safe ../util/signal.c:22
EE 13-12-20 13:04:28.246 - #15 surface_commit ../types/wlr_surface.c:502
EE 13-12-20 13:04:28.262 - #16 ffi_prep_go_closure ??:?
EE 13-12-20 13:04:28.285 - #17 ffi_closure_free ??:?
EE 13-12-20 13:04:28.302 - #18 wl_event_loop_get_destroy_listener ??:?
EE 13-12-20 13:04:28.322 - #19 wl_client_destroy ??:?
EE 13-12-20 13:04:28.341 - #20 wl_event_loop_dispatch ??:?
EE 13-12-20 13:04:28.364 - #21 wl_display_run ??:?
EE 13-12-20 13:04:28.408 - #22 main ../src/main.cpp:397
EE 13-12-20 13:04:28.424 - #23 __libc_start_main ??:?
EE 13-12-20 13:04:28.465 - #24 _start ??:?
(EE) failed to read Wayland events: Broken pipe
```

JordanL2:
@ammen99 I'm trying to solve the problem of focusing the correct output on wakeup. I don't think the info can be stored on the views so I need to store it in some central, singular location. This is the pseudocode:

```
central_place:
    last_focused_output (identifier)
    timestamp

output pre-destroy:
    if output is focused:
        if no central_place.last_focused_output set or central_place.timestamp is more than X ms ago:
            central_place.last_focused_output = this output identifier 
            central_place.timestamp = now 
                                   
output added:                      
    if central_place is this output: 
        focus this output          
        delete central_place.last_focused_output  
```

Any idea where that data could be stored? `output_layout_t` doesn't allow adding custom data to it.

ammen99:
> @ammen99 I'm trying to solve the problem of focusing the correct output on wakeup. I don't think the info can be stored on the views so I need to store it in some central, singular location. This is the pseudocode:
> 
> ```
> central_place:
>     last_focused_output (identifier)
>     timestamp
> 
> output pre-destroy:
>     if output is focused:
>         if no central_place.last_focused_output set or central_place.timestamp is more than X ms ago:
>             central_place.last_focused_output = this output identifier 
>             central_place.timestamp = now 
>                                    
> output added:                      
>     if central_place is this output: 
>         focus this output          
>         delete central_place.last_focused_output  
> ```
> 
> Any idea where that data could be stored? `output_layout_t` doesn't allow adding custom data to it.

Sounds like something you'd store as a custom data on `wf::get_core()`

JordanL2:
Thanks!

JordanL2:
@ammen99 - I think this is ready for actual review. There are a few outstanding possible issues / questions:

1. Calc's text is still blurry after wakeup until it's resized when it's not tiled - this is a problem very specific to Calc
2. The random rare crash I've experienced, log above, I don't actually know if this is related to this plugin, it's happened 2-3 times
3. I have a hardcoded 10000ms value in the code which might want to be a config somewhere

ammen99:
> 1. Calc's text is still blurry after wakeup until it's resized when it's not tiled - this is a problem very specific to Calc

Calc needs to adapt to the output scale. I.e when it gets moved to the new output, it should receive a wl_surface.enter event for the new output. After that, the app itself needs to adjust its scaling to the scaling of the output, otherwise, you'd get blurry apps. Can you check in the output of `WAYLAND_DEBUG=1 <blurry app here>` whether the aforementioned event is really sent from Wayfire?

JordanL2:
I can check, but if it wasn't wouldn't every app be blurry?

ammen99:
> I can check, but if it wasn't wouldn't every app be blurry?

I suggested to verify that this is the case so you could report a bug to the application :) But yes, you are correct, it is very likely that all apps would have been broken if it were a bug in wayfire.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+            // Set current output and geometry in the view's last output data
+            if (!view_has_data(view))
+            {
+                view_store_data(view, output, i);
+            }
+        }
+    };
+
+    wf::signal_connection_t output_removed = [=] (wf::signal_data_t *data)
```
ammen99, line 177: I assume this was for debugging purposes, and we don't need it anymore?

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+ */
+
+class wayfire_preserve_output : public wf::plugin_interface_t
+{
+    std::vector<std::string> outputs_being_removed;
+
+    wf::signal_connection_t output_pre_remove = [=] (wf::signal_data_t *data)
+    {
+        auto signal_data = (wf::output_pre_remove_signal*)data;
+        LOGD("Received pre-remove event: ", signal_data->output->to_string());
```
ammen99, line 137: I think core would emit the pre-remove event, destroy the output, emit the removed event? So why do you need to keep track of which outputs are being removed?

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+            // This event is not for this output
+            return;
+        }
+
+        // This output is being destroyed
+        std::string identifier = make_output_identifier(output);
+
+        // Store this output as the focused one
+        if (wf::get_core().get_active_output()->to_string() == output->to_string())
```
ammen99, line 155: you can directly compare `get_active_output()` with `output`, no need to compare the strings.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+            core_store_focused_output(output);
+        }
+
+        auto views = output->workspace->get_views_in_layer(wf::LAYER_WORKSPACE);
+        for (uint i = 0; i < views.size(); i++)
```
ammen99, line 161: the correct type is `size_t`

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+  public:
+    void init() override
+    {
+        if (wlr_output_is_noop(output->handle))
+        {
+            // Don't do anything for NO-OP outputs
+            return;
+        }
+
+        wf::get_core().output_layout->connect_signal("output-added",
```
ammen99, line 305: you could try doing this on a wl_idle_call (idle calls are executed after control returns to the main loop, so in this case right after all plugins are loaded), waiting for the `output-added` signal on the very same output the plugin is on seems strange to me.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+        // Focus this output if it was the last one focused
+        if (core_get_focused_output() == identifier)
+        {
+            LOGD("This is last focused output, refocusing: ", output->to_string());
+            wf::get_core().focus_output(output);
+            core_erase_focused_output();
+        }
+
+        for (auto& wo : wf::get_core().output_layout->get_outputs())
```
ammen99, line 211: shouldn't you iterate over all views (`wf::get_core().get_all_views()`), build the list, and then sort all the views (instead of doing this per-output?)

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+            // Make a list of views to move to this output
+            auto views = std::vector<wayfire_view>();
+            for (auto& view : wo->workspace->get_views_in_layer(wf::LAYER_WORKSPACE))
+            {
+                if ((view->role != wf::VIEW_ROLE_TOPLEVEL) || !view->is_mapped())
```
ammen99, line 217: what is the point of checking for toplevel, wouldn't you store data only for toplevel views anyway?

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+            {
+                auto last_output_info = view_get_data(view);
+                LOGD("Restoring view: ",
+                    view->get_title(), " to: ", output->to_string());
+
+                wf::get_core().move_view_to_output(view, output, false);
+                view->set_fullscreen(last_output_info->fullscreen);
+                view->set_minimized(last_output_info->minimized);
+                view->set_sticky(last_output_info->sticky);
+                view->set_tiled(last_output_info->tiled_edges);
```
ammen99, line 248: hmm, so why do you need to first set the tiled edges and then emit the tile request? sounds counterintuitive to me.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+                if (last_output_info->tiled_edges != 0)
+                {
+                    view->tile_request(last_output_info->tiled_edges);
+                }
+
+                // This is to ensure the next resize is actually done, helps ensure
+                // apps redraw themselves too
+                view->resize(last_output_info->geometry.width + 1,
+                    last_output_info->geometry.height);
+                view->set_geometry(last_output_info->geometry);
```
ammen99, line 258: This sounds more like a bug in core if you need to write such hacks ... How can I reproduce the bug you're working around here?

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+            outputs_being_removed.end(),
+            signal_data->output->to_string()), outputs_being_removed.end());
+        LOGD("Outputs currently being removed: ");
+        for (auto el : outputs_being_removed)
+        {
+            LOGD("- ", el);
+        }
+    };
+
+    wf::signal_connection_t restore_views_to_output = [=] (wf::signal_data_t *data)
```
ammen99, line 192: I thought the plan was to do that only if the data has not expired yet? And we need a configuration option for the expiration time for sure. It should be easy enough to add, just copy any existing .xml file in `metadata`, and create a new option with type int and default value `10000`.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+            return;
+        }
+
+        wf::get_core().output_layout->connect_signal("output-added",
+            &restore_views_to_output);
+        wf::get_core().output_layout->connect_signal("output-pre-remove",
+            &output_pre_remove);
+        wf::get_core().output_layout->connect_signal("output-removed",
+            &output_removed);
+    }
```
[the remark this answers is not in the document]
JordanL2: If I do that, won't all the data on views be lost when they're in the NOOP-1 output? The plugin is destroyed when the outputs are destroyed, and the data in views and core needs to outlive them.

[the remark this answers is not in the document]
ammen99: the noop output will also have an instance of the plugins, so it will keep the data alive (because there would be at least once instance of the plugin active). core guarantees that this NOOP output will be created (with the plugins on it) before destroying the other outputs.

plugins/single_plugins/preserve-output.cpp
JordanL2: It wasn't tiling properly until I did this. I'll check if I need the set_tiled, I think I do because the tile_request is only emitted if tiled_edges != 0 - and if we emit tile_request when tiled_edges is 0, the grid plugin resizes the window and so stops our resize working.

plugins/single_plugins/preserve-output.cpp
JordanL2: There's code I think the XDG view resize code that checks what the last resize sent was, and doesn't send the resize if so. Possibly this code is getting out of sync with the actual state of the XDG view, I'll try to get more info later on.

plugins/single_plugins/preserve-output.cpp
JordanL2: The data expiration is only for the last output focusing. The pseudocode is:

```
core:
    last_focused_output (identifier)
    timestamp

output pre-destroy:
    if output is focused:
        if no core.last_focused_output set or core.timestamp is more than X ms ago:
            core.last_focused_output = this output identifier 
            core.timestamp = now 
                                   
output added:                      
    if core is this output: 
        focus this output          
        delete core.last_focused_output  
```

As you see, the data expiration is for when an output sets itself as the last focused - it won't set itself as the last focused output if another output has already done so in the last 10 seconds. This is because when we destroy the output that has focus, the other output automatically gets the focus, so when the second output is also destroyed, it will also think it's the one with focus. This way we remember which output actually had focus before the output destruction started. The 10 second timeout is so if someone actually disconnects an output, it isn't stored as the last focused output forever.

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+ */
+
+class last_output_info_t : public wf::custom_data_t
+{
+  public:
+    std::string output_identifier;
+    wf::geometry_t geometry;
+    bool fullscreen = false;
+    bool minimized  = false;
+    bool sticky     = false;
```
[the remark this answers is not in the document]
JordanL2: Ah I didn't know that, will remove this

plugins/single_plugins/preserve-output.cpp
```diff
@@ -0,0 +1,314 @@
...
+    uint32_t tiled_edges = 0;
+    uint z_order;
+    bool focused = false;
+};
+
+std::string make_output_identifier(wf::output_t *output)
+{
+    std::string identifier = "";
```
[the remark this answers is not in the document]
JordanL2: No reason not to, will add

src/output/workspace-impl.cpp
```diff
@@ -793,6 +793,11 @@ class workspace_manager::impl
     {
         auto old_w = output_geometry.width, old_h = output_geometry.height;
         auto new_size = output->get_screen_size();
+        if ((old_w == new_size.width) && (old_h == new_size.height))
```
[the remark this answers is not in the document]
JordanL2: A separate commit in the same PR?

plugins/single_plugins/preserve-output.cpp
JordanL2: Yes definitely - I guess it only works because usually there is only one other output with relevant views on them when this code runs.

plugins/single_plugins/preserve-output.cpp
JordanL2: Good point, will remove this check

src/output/workspace-impl.cpp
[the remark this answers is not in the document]
ammen99: sounds good (or a separate PR if you wish ...)

plugins/single_plugins/preserve-output.cpp
ammen99: Ok. So we'll always restore the views on their orignal output (if they aren't moved) even if the output is plugged after 10 days? (I have nothing against, but I want to know what the desired behavior is)

JordanL2: Yes we will restore the views on their original output no matter how long has passed if they haven't been explicitly moved by the user.

plugins/single_plugins/preserve-output.cpp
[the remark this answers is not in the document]
JordanL2: Got it. Will add this. I'm still a bit unclear - what situation involves all outputs, including NOOP, being destroyed, other than Wayfire shutting down (which would release all memory it's using)?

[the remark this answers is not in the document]
JordanL2: Oh, this is if the user disables the plugin?

plugins/single_plugins/preserve-output.cpp
JordanL2: I've tested without set_tiled and everything still seems to work, so I've removed it.

plugins/single_plugins/preserve-output.cpp
JordanL2: This is because we're listening for move events on the views, to see if the user has explicitly moved them. I was keeping track of whether or not outputs are currently being destroyed because if they are, the move event is probably coming from the view being automatically moved. However this is actually left over from before I changed it to listen for `view_move_request_signal` rather than `view_geometry_changed_signal` and possibly the former is not emitted when the views are being automatically moved?

plugins/single_plugins/preserve-output.cpp
JordanL2: Sorry, don't need what?

plugins/single_plugins/preserve-output.cpp
ammen99: I think you should indeed listen for the view_geometry_changed signal, after all, if the user invokes grid plugin for a view, there won't be any `view_move_request` signal but we don't want to restore to the old output, right?

But also my point was that you don't need to store a vector, a simple `bool output_is_being_removed` should suffice because it is always pre-remove, remove, pre-remove, remove

plugins/single_plugins/preserve-output.cpp
ammen99: hmm, ok. thanks for the explanation

ammen99: however it still should be an option ...

plugins/single_plugins/preserve-output.cpp
JordanL2: Ok, so there won't ever be a situation where there are two outputs being removed at once? Makes sense.

The problem with listening for view_geometry_changed is that rarely an application will actually emit the signal after the output has been removed. Maybe we should think of some other means of invalidating the last_output_info on the views. If the user is sat there using their computer while the output is off that implies the output has been deliberately turned off rather than disconnected temporarily due to DPMS. Then again they could be moving their mouse and pressing keys to wake up the computer so we can't just listen for that.

plugins/single_plugins/preserve-output.cpp
ammen99: Err, I thought we don' tneed to track pre-remove and removed, but we do have to track it so ...

plugins/single_plugins/preserve-output.cpp
ammen99: > Ok, so there won't ever be a situation where there are two outputs being removed at once? Makes sense.

They can be removed one after the other but not "concurrently" so to say.

JordanL2: @ammen99 - I've thought about this and think we can make this simpler and more robust.

We keep track of the last time an output-pre-remove or an output-remove event was emitted. We have a callback on the view geometry changed which removes the view's last output info if an output [pre]-remove event wasn't emitted in the last X ms. Yes this involves another arbitrary timeout, but as I say I've seen Termite emit geometry changed events due to being moved to another output AFTER the output remove event was triggered, so I think this would be more robust.

plugins/single_plugins/preserve-output.cpp
JordanL2: I've made this a config option: last_output_focus_timeout
Clunky name but it's pretty hard to describe what it does in a few words :)

plugins/single_plugins/preserve-output.cpp
JordanL2: In `view-impl.hpp` we have: 
```
wf::dimensions_t last_size_request
```
The `should_resize_client` method uses it to decide if we should send the request, and we only send it if the new request is different to the previous request. For some reason, this is stopping the view->set_geometry in my plugin from actually sending the resize request - and it needs to be sent because the views have been shrunk after their move to the (usually smaller) NOOP-1 output.

I guess this means that moving the views to NOOP-1 causes the views' size to be reduced somehow without setting this last_size_request - which is odd because I thought the view resizing after moving a view to a smaller output used the view::resize method.

Probably the easiest way to reproduce this yourself is to comment out the view->resize line in this code and try it yourself, if your normal screens are bigger than the NOOP it should reproduce it.

JordanL2: @ammen99 I've been trying to get some debug info to shed some light on this, unfortunately it seems in the last week or so a commit to wlroots has broken DPMS for me - my monitors now don't switch back on after trying to wake them up and I have to reboot. I'll raise a wlroots bug for this, in the meantime are there any other changes you want to the code?

ammen99: Hi, unfortunately, I cannot reproduce even though I tried. However, I have a hunch what might be happening, because I have debugged (and thought I had fixed) a very similar bug in the past.

The basic problem is this: we have APIs for just moving, just resizing or moving+resizing a view. Now, a lot of places are using the third operation (`set_geometry()`) when they only need to move the view.

A more detailed explanation why this is problematic: when moving a view with `set_geometry` the usual computation is:

```cpp
auto wm = view->get_wm_geometry();
wm.x += 5;
wm.y += 5;
view->set_wm_geometry();
```

Now consider what would happen if the view had size 100x100, your plugin tries to resize it to 50x50: the `last_size_request` becomes 50x50, but the wm_geometry is still 100x100 because the client has not updated its state.
Now a plugin calls the thing from above, and the view is resized back to 100x100 !

Now, onto the interesting part: how do we fix it? We can't really do something like I have done here: http://ix.io/2JbQ (though that patch might be useful for testing if my hypothesis is correct), because there are cases where two plugins actually compete for setting the view geometry and the second one should win (this for example of simple-tile, which always keeps its views in a fixed layout/geometries). Instead, we should find all incorrect usages of `set_geometry` and replace them with `move`. I'd have done this myself if I could reproduce, but I cannot, so I ask you to do this - print the geometry in `set_geometry`, and also use `wf::print_trace(false/true)` (false is slower, but usually gives more info) to see where set_geometry is called, and to see what the race is.

JordanL2: @ammen99 - I believe I've figured it out. The relevant method is `wayfire_xdg_view::resize`.

First, the view is resized when it's moved to the NOOP. This resize is sent to the client (`wlr_xdg_toplevel_set_size`), because the new size is different to the client's current_size (from `wlr_xdg_surface_get_geometry`).

Second, the preserve-output plugin requests the view be resized back to what it was. It seems the client app (or wlroots?) hasn't actually processed the previous resize, because it's actual current_size hasn't changed yet. So this resize is ignored.

ammen99: The resize does not take effect immediately, the configure request is sent on the next idle, and then the client needs to submit a new buffer. Only after that we have the new size. Anyway, I don't understand why the second resize is ignored?

JordanL2: 1. The view has size 1885x1025
2. During DPMS, the view is moved to NOOP, a resize request is sent to resize it down to 1178x683
3. The preserve-output plugin calls resize to set it to 1885x1025
4. Because the NOOP's resize request in step 2 hasn't been processed yet, `wlr_xdg_surface_get_geometry` returns 1885x1025 for the view - this means the resize call to set it to 1885x1025 isn't sent

In `wayfire_xdg_view::resize` we have this code checking if we should send the resize before sending it:
```
    wf::dimensions_t current_size{current_geometry.width, current_geometry.height};                                    
    if (should_resize_client({w, h}, current_size))                                    
    {                                                                                                          
        this->last_size_request = {w, h};                                                                   
        wlr_xdg_toplevel_set_size(xdg_toplevel->base, w, h);                                                                                                                                   
    } 
```
`should_resize_client` checks the resize request isn't the same as the last resize request sent to the client, and isn't the same as the current size of the client (current_size).

JordanL2: Maybe `should_resize_client` should be changed to do the resize if the new size is different to the last size request *or* the current size? Rather than needing to be different to both as I think is the case now.

JordanL2: Nevermind I misread the should_resize_client method, will dig into it more tomorrow.

JordanL2: Ok, I'm lost now. I've added this debug into the resize method:
```
void wayfire_xdg_view::resize(int w, int h)                                    
{                                    
    LOGD(this->get_title(), " resize called: ", w, "x", h);                                    
    LOGD("This ref: ", this);                                    
    wf::print_trace(true);                                    
    if (view_impl->frame)                                    
    {                                    
        view_impl->frame->calculate_resize_size(w, h);                                    
    }                                    
                                    
    auto current_geometry = get_xdg_geometry(xdg_toplevel);                                    
    wf::dimensions_t current_size{current_geometry.width, current_geometry.height};                                    
    LOGD("- Current size: ", current_size.width, "x", current_size.height);                                    
    LOGD("- last size request: ", this->last_size_request.width, "x", this->last_size_request.height);                                    
    if (should_resize_client({w, h}, current_size))                                    
    {                                    
        LOGD("- YES - Doing resize!");                                    
        this->last_size_request = {w, h};                                    
        LOGD("- NEW last size request: ", this->last_size_request.width, "x", this->last_size_request.height);                                                                                 
        wlr_xdg_toplevel_set_size(xdg_toplevel->base, w, h);                                    
    }                                    
    else {                                    
        LOGD("- NO - Not doing resize.");                                    
    }                                    
}                                    
```

And this is what we get in the logs:
```
 769 DD 24-12-20 11:17:28.073 - [src/view/xdg-shell.cpp:394] Volume Control resize called: 1178x683                                                                                            
 770 DD 24-12-20 11:17:28.073 - [src/view/xdg-shell.cpp:395] This ref: 0x55c45474cef0
 771 EE 24-12-20 11:17:28.078 - #1  wayfire_xdg_view::resize(int, int) at 0x55c451486595
 772 EE 24-12-20 11:17:28.079 - #2  wf::view_interface_t::set_geometry(wlr_box) at 0x55c4514703c8
 773 EE 24-12-20 11:17:28.079 - #3  wf::compositor_core_impl_t::move_view_to_output(nonstd::observer_ptr<wf::view_interface_t>, wf::output_t*, bool) at 0x55c451406ea4
 774 EE 24-12-20 11:17:28.079 - #4  wf::transfer_views(wf::output_t*, wf::output_t*) at 0x55c4513cbf08
 775 EE 24-12-20 11:17:28.079 - #5  wf::output_layout_output_t::destroy_wayfire_output(bool) at 0x55c4513ce97f
 776 EE 24-12-20 11:17:28.079 - #6  wf::output_layout_output_t::apply_state(wf::output_state_t const&, bool) at 0x55c4513d00eb
 777 EE 24-12-20 11:17:28.079 - #7  wf::output_layout_t::impl::apply_configuration(std::map<wlr_output*, wf::output_state_t, std::less<wlr_output*>, std::allocator<std::pair<wlr_output* const     , wf::output_state_t> > > const&) at 0x55c4513d2e57
 778 EE 24-12-20 11:17:28.079 - #8  wf::output_layout_t::impl::remove_output(wlr_output*) at 0x55c4513d1821
 779 EE 24-12-20 11:17:28.079 - #9  wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}::operator()(void*) const at 0x55c4513d14d4
 780 EE 24-12-20 11:17:28.079 - #10 void std::__invoke_impl<void, wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}&, void*>(std::__invoke_other, wf::output_layout_t::impl     ::add_output(wlr_output*)::{lambda(void*)#1}&, void*&&) at 0x55c4513e0f14
 781 EE 24-12-20 11:17:28.079 - #11 std::enable_if<is_invocable_r_v<void, wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}&, void*>, std::enable_if>::type std::__invoke_r     <void, wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}&, void*>(void&&, (wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}&)...) at 0x55c4513dd9     4c
 782 EE 24-12-20 11:17:28.079 - #12 std::_Function_handler<void (void*), wf::output_layout_t::impl::add_output(wlr_output*)::{lambda(void*)#1}>::_M_invoke(std::_Any_data const&, void*&&) at 0     x55c4513d9ca0                                                                                   
 783 EE 24-12-20 11:17:28.079 - #13 std::function<void (void*)>::operator()(void*) const at 0x55c4513cabe1
 784 EE 24-12-20 11:17:28.079 - #14 wf::wl_listener_wrapper::emit(void*) at 0x55c4513ca5a1                    
 785 EE 24-12-20 11:17:28.079 - #15 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(+0x2203d2) [0x55c4513ca3d2]
 786 EE 24-12-20 11:17:28.079 - #16 /usr/lib/libwlroots.so.7(+0x8eab7) [0x7f0506230ab7]          
 787 EE 24-12-20 11:17:28.079 - #17 /usr/lib/libwlroots.so.7(wlr_output_destroy+0x4f) [0x7f050621d614]
 788 EE 24-12-20 11:17:28.079 - #18 /usr/lib/libwlroots.so.7(+0x37c40) [0x7f05061d9c40]
 789 EE 24-12-20 11:17:28.079 - #19 /usr/lib/libwlroots.so.7(+0x37407) [0x7f05061d9407]
 790 EE 24-12-20 11:17:28.079 - #20 /usr/lib/libwlroots.so.7(+0x3227a) [0x7f05061d427a]
 791 EE 24-12-20 11:17:28.079 - #21 /usr/lib/libwlroots.so.7(+0x8eab7) [0x7f0506230ab7]
 792 EE 24-12-20 11:17:28.079 - #22 /usr/lib/libwlroots.so.7(+0x4e751) [0x7f05061f0751]                     
 793 EE 24-12-20 11:17:28.079 - #23 /usr/lib/libwayland-server.so.0(wl_event_loop_dispatch+0xca) [0x7f0506286faa]
 794 EE 24-12-20 11:17:28.080 - #24 /usr/lib/libwayland-server.so.0(wl_display_run+0x27) [0x7f05062854e7]      
 795 EE 24-12-20 11:17:28.080 - #25 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(main+0xc87) [0x55c4513c1d34]
 796 EE 24-12-20 11:17:28.080 - #26 /usr/lib/libc.so.6(__libc_start_main+0xf2) [0x7f0505945152]                 
 797 EE 24-12-20 11:17:28.080 - #27 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(_start+0x2e) [0x55c4513be70e]
 798 DD 24-12-20 11:17:28.080 - [src/view/xdg-shell.cpp:404] - Current size: 1885x1025
 799 DD 24-12-20 11:17:28.080 - [src/view/xdg-shell.cpp:405] - last size request: 1885x1025
 800 DD 24-12-20 11:17:28.080 - [src/view/xdg-shell.cpp:408] - YES - Doing resize!       
 801 DD 24-12-20 11:17:28.080 - [src/view/xdg-shell.cpp:410] - NEW last size request: 1178x683                 

(SNIP)

1102 DD 24-12-20 11:17:28.940 - [src/view/xdg-shell.cpp:394] Volume Control resize called: 1885x1025
1103 DD 24-12-20 11:17:28.940 - [src/view/xdg-shell.cpp:395] This ref: 0x55c45474cef0
1104 EE 24-12-20 11:17:28.941 - #1  wayfire_xdg_view::resize(int, int) at 0x55c451486595
1105 EE 24-12-20 11:17:28.941 - #2  wf::view_interface_t::set_geometry(wlr_box) at 0x55c4514703c8
1106 EE 24-12-20 11:17:28.941 - #3  wayfire_preserve_output::restore_views_to_output() at 0x7f04c0f64195
1107 EE 24-12-20 11:17:28.941 - #4  wayfire_preserve_output::init()::{lambda()#1}::operator()() const at 0x7f04c0f64513
1108 EE 24-12-20 11:17:28.941 - #5  void std::__invoke_impl<void, wayfire_preserve_output::init()::{lambda()#1}&>(std::__invoke_other, wayfire_preserve_output::init()::{lambda()#1}&) at 0x7f0     4c0f699e4
1109 EE 24-12-20 11:17:28.941 - #6  std::enable_if<is_invocable_r_v<void, wayfire_preserve_output::init()::{lambda()#1}&>, std::enable_if>::type std::__invoke_r<void, wayfire_preserve_output:     :init()::{lambda()#1}&>(void&&, (wayfire_preserve_output::init()::{lambda()#1}&)...) at 0x7f04c0f68865
1110 EE 24-12-20 11:17:28.941 - #7  std::_Function_handler<void (), wayfire_preserve_output::init()::{lambda()#1}>::_M_invoke(std::_Any_data const&) at 0x7f04c0f67703
1111 EE 24-12-20 11:17:28.941 - #8  std::function<void ()>::operator()() const at 0x55c4513c3d30
1112 EE 24-12-20 11:17:28.941 - #9  wf::wl_idle_call::execute() at 0x55c4513ca7fe
1113 EE 24-12-20 11:17:28.941 - #10 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(+0x2203f5) [0x55c4513ca3f5]
1114 EE 24-12-20 11:17:28.941 - #11 /usr/lib/libwayland-server.so.0(wl_event_loop_dispatch_idle+0x1b) [0x7f0506286ebb]
1115 EE 24-12-20 11:17:28.941 - #12 /usr/lib/libwayland-server.so.0(wl_event_loop_dispatch+0xeb) [0x7f0506286fcb]
1116 EE 24-12-20 11:17:28.941 - #13 /usr/lib/libwayland-server.so.0(wl_display_run+0x27) [0x7f05062854e7]
1117 EE 24-12-20 11:17:28.941 - #14 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(main+0xc87) [0x55c4513c1d34]
1118 EE 24-12-20 11:17:28.941 - #15 /usr/lib/libc.so.6(__libc_start_main+0xf2) [0x7f0505945152]
1119 EE 24-12-20 11:17:28.941 - #16 /home/jordan/GitRepo/wayfire-fork/build/src/wayfire(_start+0x2e) [0x55c4513be70e]
1120 DD 24-12-20 11:17:28.941 - [src/view/xdg-shell.cpp:404] - Current size: 1885x1025
1121 DD 24-12-20 11:17:28.941 - [src/view/xdg-shell.cpp:405] - last size request: 1885x1025
1122 DD 24-12-20 11:17:28.941 - [src/view/xdg-shell.cpp:414] - NO - Not doing resize.
```

`last_resize_request` is being changed back after it's set by the moving-to-NOOP resize. And there's no other instances of this resize method being called for this view in the log other than these two.

The only other place in the whole codebase where last_size_request is being set that could possibly make something like this happen is in `wayfire_xdg_view::commit`:
```
    auto xdg_g = get_xdg_geometry(xdg_toplevel);
(...)
    this->last_size_request = wf::dimensions(xdg_g);
```
So, it has to be that method causing this somehow.

JordanL2: After the NOOP resize, this is the trace of the call to commit that is changing the last_size_request back, and thus stopping the second resize from working:
```
EE 24-12-20 12:17:32.409 - #1  wayfire_xdg_view::commit() ../src/view/xdg-shell.cpp:337                                                                                        
EE 24-12-20 12:17:32.456 - #2  wf::wlr_surface_base_t::wlr_surface_base_t(wf::surface_interface_t*)::{lambda(void*)#2}::operator()(void*) const ../src/view/surface.cpp:237    
EE 24-12-20 12:17:32.523 - #3  void std::__invoke_impl<void, wf::wlr_surface_base_t::wlr_surface_base_t(wf::surface_interface_t*)::{lambda(void*)#2}&, void*>(std::__invoke_other, wf::wlr_surface_base_t::wlr_surface_base_t(wf::surface_interface_t*)::{lambda(void*)#2}&, void*&&) /usr/include/c++/10.2.0/bits/invoke.h:60        
EE 24-12-20 12:17:32.588 - #4  ...) /usr/include/c++/10.2.0/bits/invoke.h:116    
EE 24-12-20 12:17:32.641 - #5  std::_Function_handler<void (void*), wf::wlr_surface_base_t::wlr_surface_base_t(wf::surface_interface_t*)::{lambda(void*)#2}>::_M_invoke(std::_Any_data const&, void*&&) /usr/include/c++/10.2.0/bits/std_function.h:292                                                                               
EE 24-12-20 12:17:32.677 - #6  std::function<void (void*)>::operator()(void*) const /usr/include/c++/10.2.0/bits/std_function.h:622    
EE 24-12-20 12:17:32.715 - #7  wf::wl_listener_wrapper::emit(void*) ../src/util.cpp:524                                         
EE 24-12-20 12:17:32.753 - #8  handle_wrapped_listener(wl_listener*, void*) ../src/util.cpp:460                                                                        
EE 24-12-20 12:17:32.771 - #9  wlr_signal_emit_safe ../util/signal.c:22                                                    
EE 24-12-20 12:17:32.789 - #10 surface_commit_pending ../types/wlr_surface.c:431                                                   
EE 24-12-20 12:17:32.807 - #11 surface_commit ../types/wlr_surface.c:502                                              
EE 24-12-20 12:17:32.819 - #12 ffi_prep_go_closure ??:?                                                               
EE 24-12-20 12:17:32.836 - #13 ffi_closure_free ??:?                                                                  
EE 24-12-20 12:17:32.854 - #14 wl_event_loop_get_destroy_listener ??:?                                                
EE 24-12-20 12:17:32.876 - #15 wl_client_destroy ??:?                                                                 
EE 24-12-20 12:17:32.895 - #16 wl_event_loop_dispatch ??:?                                                            
EE 24-12-20 12:17:32.907 - #17 wl_display_run ??:?                                                                    
EE 24-12-20 12:17:32.942 - #18 main ../src/main.cpp:397                                                               
EE 24-12-20 12:17:32.964 - #19 __libc_start_main ??:?                                                                 
EE 24-12-20 12:17:33.011 - #20 _start ??:?                                                                  
```

ammen99: Could you try with this patch here: http://ix.io/2Joj ?

JordanL2: @ammen99 That has fixed the issue :)

JordanL2: Are you going to commit that to master, or should I add it as a commit to this PR?

ammen99: Thanks for testing! I'll open a PR with it. Sadly, this will not work for Xwayland clients, because there we don't have any serial we can check against ...

ammen99: #973 has these changes, and also a detailed explanation of what goes wrong in case you're interested in the details :)

JordanL2: Just tested with Minecraft and Steam, this issue doesn't affect Xwayland clients anyway!

damianatorrpm: > Thanks for testing! I'll open a PR with it. Sadly, this will not work for Xwayland clients, because there we don't have any serial we can check against ...

FWIW there is _NET_WM_USER_TIME

ammen99: _NET_WM_USER_TIME is something different from xdg-shell serials. xdg-shell serial is used to count the requests from the compositor to the view, user-time is a timestamp on the last activity, for ex. key press or pointer motion.

ammen99: > Just tested with Minecraft and Steam, this issue doesn't affect Xwayland clients anyway!

Hmm, maybe the race doesn't happen there.. But I wouldn't complain if it works :D

ammen99 (changes requested):
This version looks good! Two things:

- This patch here http://ix.io/2Jph, I think it makes stuff more readable (feel free to commit it with your email, I don't care :)
- ~~Remove the workaround for resizing~~ seems you just did that ..

JordanL2:
I've applied those changes and combined it into the same commit.

ammen99 (approved):
Thanks for working on this! I hope that other people find it useful as well.

JordanL2:
No problem :)

PR #924, opened 2020-12-06
scale: Filter by the title of views when typing in non-interactive mode

dkondor:
This is in a very early stage that I've put together in a few hours. I'm sharing it to hopefully get some initial feedback for the idea :)

What I'd like to achieve is basically a reimplementation of the "Scale Window Title Filter" plugin from Compiz, achieving (at least) the following two things:
 - in (non-interactive) scale, typing filters the views based on their title (case-insensitive, substring match)
 - another plugin can start scale by specifying a filter (based on title or app-id, or maybe even a custom list of views); eventually, using the dbus plugin, scale could be started for a group of views via dbus

This PR currently achieves very basic functionality by converting key presses to ASCII and "hiding" views that do not match. There are multiple areas where I'd like to improve it:

- [x] Make this optional + other options (case-insensitivity, whether to filter title / app-id / both, etc.)
- [x] Feedback on the current filter (text rendered as an overlay effect)
- [x] Better way to hide views (currently they are made transparent and moved above the current workspace) -- I wonder if there is a way to skip rendering a view (without completely taking over the rendering process)
- [x] Transition when hiding / showing filtered views
-  ~~Other types of filters (including app-id or a list of views); I'm not sure if the current, polymorphic design is actually needed.~~
- [x] Ability to add more than one filter (e.g. a filter is set by an external plugin or dbus caller + the user starts typing to filter more)
- [x] Instead of manually converting keycodes to chars, use libxkbcommon with the current layout
- [x] Interface for other plugins to call scale with a filter + example plugin implementation for this


ammen99:
Hi, I think the idea is good. Our plan has been to add a custom signal so that other plugins can add a custom filter in the scale plugin. I believe that should fit your use-case and potentially many others (which is your last suggestion). This way we can put all the handling in a separate plugin, with the intention of keeping scale as simple as possible.

dkondor:
Hi,
thanks for the quick reply; this sounds good. Do you mean that layout_slots() would call a new signal that can be connected to do the filtering?
I'm happy to work on this more if there is a plan on the interface it can follow.

ammen99:
> Hi,
> thanks for the quick reply; this sounds good. Do you mean that layout_slots() would call a new signal that can be connected to do the filtering?
> I'm happy to work on this more if there is a plan on the interface it can follow.

I think it should be it get_views but otherwise sounds fine. A signal with signal data a vector, all subscribers can then modify it as they wish.

dkondor:
Hi,
I've added a second iteration on this, that uses signals.

> I think it should be it get_views but otherwise sounds fine. A signal with signal data containing just a vector, all subscribers can then modify it as they wish.

My problem with this is that scale needs the hidden views in a separate vector to be able to "hide" them by moving them off-screen. Actually, I don't think that this way of hiding them is ideal, but I could not yet think of a better way without completely taking over rendering -- if there is a way to set a view not to be rendered temporarily, that would be great (I think unmapping hidden views would just add more complications)

Some things I'm not satisfied with yet:
 - having a signal forward keypresses feels weird, but I couldn't think of another way if the title filter is a separate plugin
 - needed to add a separate, "activate" binding that filter plugins can use to either start or update scale -- this might be a signal as well though

ammen99:
> if there is a way to set a view not to be rendered temporarily, that would be great (I think unmapping hidden views would just add more complications)

There is such a way :) And its intended use-cases are exactly plugins like that, which need to hide a view for "a bit":https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/view.hpp#L284

> * having a signal forward keypresses feels weird, but I couldn't think of another way if the title filter is a separate plugin

The separate plugin could directly subscribe to the raw `keyboard_key` events from core

> * needed to add a separate, "activate" binding that filter plugins can use to either start or update scale -- this might be a signal as well though

Again, no need for such a signal, see https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/output.hpp#L162

dkondor:
> > if there is a way to set a view not to be rendered temporarily, that would be great (I think unmapping hidden views would just add more complications)
> 
> There is such a way :) And its intended use-cases are exactly plugins like that, which need to hide a view for "a bit":https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/view.hpp#L284

Wow, this makes things a lot simpler :)

> 
> > * having a signal forward keypresses feels weird, but I couldn't think of another way if the title filter is a separate plugin
> 
> The separate plugin could directly subscribe to the raw `keyboard_key` events from core

This makes a lot of sense, I forgot these signals exist :)

> 
> > * needed to add a separate, "activate" binding that filter plugins can use to either start or update scale -- this might be a signal as well though
> 
> Again, no need for such a signal, see https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/output.hpp#L162

Here, I'm less sure, since there are cases where a plugin would want an update and not a toggle. Would it be OK to treat "toggle" as "activate"/"update" instead if it comes from a plugin? Or maybe using the activator_data_t parameter to signal which behavior a plugin wants?

ammen99:
> > > if there is a way to set a view not to be rendered temporarily, that would be great (I think unmapping hidden views would just add more complications)
> > 
> > 
> > There is such a way :) And its intended use-cases are exactly plugins like that, which need to hide a view for "a bit":https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/view.hpp#L284
> 
> Wow, this makes things a lot simpler :)
> 
> > > * having a signal forward keypresses feels weird, but I couldn't think of another way if the title filter is a separate plugin
> > 
> > 
> > The separate plugin could directly subscribe to the raw `keyboard_key` events from core
> 
> This makes a lot of sense, I forgot these signals exist :)
> 
> > > * needed to add a separate, "activate" binding that filter plugins can use to either start or update scale -- this might be a signal as well though
> > 
> > 
> > Again, no need for such a signal, see https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/output.hpp#L162
> 
> Here, I'm less sure, since there are cases where a plugin would want an update and not a toggle. Would it be OK to treat "toggle" as "activate"/"update" instead if it comes from a plugin? Or maybe using the activator_data_t parameter to signal which behavior a plugin wants?

Ah, I see. I misunderstood the problem. You're right, we'll need a signal in scale for that.

src/core/seat/seat.cpp
```diff
@@ -289,6 +289,26 @@ uint32_t wf::seat_t::get_modifiers()
     return current_keyboard ? current_keyboard->get_modifiers() : 0;
 }
 
+std::string wf::seat_t::convert_keycode(uint32_t keycode)
```
ammen99, line 292: This doesn't need to be in core because plugins can access the `wlr_seat` from core (get_current_seat), and then with wlr_seat_get_keyboard() you can find the currently active keyboard.

I wonder why you say it doesn't work in all cases though? Does it work with for ex. umlaut characters like äüö with the compose key?

dkondor: I see, this makes sense.

Now I realized that compose / dead keys are not working on my Wayfire session at all, but working in a nested session, so this might be a problem with my layout settings.

dkondor:
This is now ready for review, with the features I planned working. Some small issues / questions:
 - the scale_test plugin probably does not belong here, it is too simple and is mainly meant to demonstrate filtering; it could be extended to allow a dynamic list of bindings / app-ids, similarly to the command plugin, but that seems too complicated
 - as discussed above compose / "dead key" functionality might not be working (I did not look into libxkbcommon in more detail to see if there is special handling necessary for it)
 - case insensitive comparison only works for ascii characters (I'm wondering if it is worth converting the strings to wstring for the sake of this)

ammen99:
> * the scale_test plugin probably does not belong here, it is too simple and is mainly meant to demonstrate filtering; it could be extended to allow a dynamic list of bindings / app-ids, similarly to the command plugin, but that seems too complicated

Yeah, since we already have a "real" plugin for testing the custom filtering, I think we can remove it from this PR.

> * as discussed above compose / "dead key" functionality might not be working (I did not look into libxkbcommon in more detail to see if there is special handling necessary for it)
> * case insensitive comparison only works for ascii characters (I'm wondering if it is worth converting the strings to wstring for the sake of this)

I'll test the compose keys but I think this is not mandatory in order to merge the PR, if someone cares, they can send a PR to fix it up. I'll review the code a bit later, but thanks even for taking it so far!

plugins/scale/scale.cpp
```diff
@@ -820,6 +872,63 @@ class wayfire_scale : public wf::plugin_interface_t
...
+            auto& view_data = scale_data[v];
+            if (view_data.hidden == view_scale_data::hidden_t::SHOWN)
+            {
+                view_data.hidden = view_scale_data::hidden_t::HIDING;
+                setup_view_transform(view_data, 1, 1, 0, 0, 0);
+            }
+
+            for (auto& child : v->enumerate_views(false))
+            {
+                if (child == v)
```
ammen99, line 917: why do we need special handling for child and view when they seem to have the exact same transform applied to them?

plugins/scale/scale.cpp
```diff
@@ -820,6 +872,63 @@ class wayfire_scale : public wf::plugin_interface_t
             return;
         }
 
+        std::vector<wayfire_view> filtered_views;
```
ammen99, line 875: I think this should be moved in a separate function

plugins/scale/scale.cpp
```diff
@@ -1308,6 +1440,7 @@ class wayfire_scale : public wf::plugin_interface_t
         finalize();
         output->rem_binding(&toggle_cb);
         output->rem_binding(&toggle_all_cb);
+        output->disconnect_signal(&activate_cb);
```
ammen99, line 1443: you don't need to disconnect signal_connection_t

plugins/scale/scale.cpp
```diff
@@ -1265,12 +1395,14 @@ class wayfire_scale : public wf::plugin_interface_t
...
         scale_data.clear();
         grab_interface->ungrab();
+
         disconnect_button_signal();
         view_focused.disconnect();
         view_unmapped.disconnect();
         view_attached.disconnect();
         view_detached.disconnect();
         view_minimized.disconnect();
+        output->emit_signal("scale-end", nullptr);
```
ammen99, line 1405: move this to the very end of the finalize function

ammen99 (changes requested):
Look at the individual comments.

metadata/scale_title_filter.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<wayfire>
+	<plugin name="scale_title_filter">
+		<_short>Scale Title Filter</_short>
+		<_long>Filter the views shown by scale by typing.</_long>
+		<category>Accessibility</category>
```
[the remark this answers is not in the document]
dkondor: I'm OK with moving it elsewhere (e.g. CCSM has this in Utilities); actually I'm unsure if scale belongs to accessibility as well (for me, it's a tool for window management).

plugins/scale/scale_title_filter.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+                std::string tmp(size, 0);
+                xkb_state_key_get_utf8(xkb_state, keycode, tmp.data(), size + 1);
+                char_len.push_back(size);
+                title_filter += tmp;
+            }
+
+            LOGI("Title filter changed: ", title_filter);
+            scale_activate_signal signal;
+            output->emit_signal("scale-activate", &signal);
```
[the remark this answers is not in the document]
dkondor: It is not needed by the title filter, but it would be needed by another plugin that wants to start scale, e.g. the simple test plugin I had included. That can be adapted to use the toggle bindings though.

plugins/scale/scale_title_filter.cpp
```diff
@@ -0,0 +1,322 @@
...
+            std::string title = view->get_title();
+            std::string filter = title_filter;
+            auto transform = [] (unsigned char c)
+            {
+                if (std::isspace(c))
+                {
+                    return ' ';
+                }
+
+                return (char)std::tolower(c);
```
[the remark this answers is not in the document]
dkondor: tolower() should ignore characters that it does not know, but I've added an extra check to make sure to only parse c <= 127, which are the ASCII subset of UTF-8 and cannot appear in multi-byte sequences

plugins/scale/scale_title_filter.cpp
```diff
@@ -0,0 +1,322 @@
...
+                    return;
+                }
+
+                std::string tmp(size, 0);
+                xkb_state_key_get_utf8(xkb_state, keycode, tmp.data(), size + 1);
+                char_len.push_back(size);
+                title_filter += tmp;
+            }
+
+            LOGI("Title filter changed: ", title_filter);
```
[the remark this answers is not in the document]
dkondor: I was meaning to remove this once everything works.

plugins/scale/scale.cpp
dkondor: You're right, there is not need for this.

metadata/scale_title_filter.xml
dkondor, line 6:
```suggestion
		<category>Utility</category>
```

plugins/scale/wayfire/plugins/scale-signal.hpp
```diff
@@ -0,0 +1,79 @@
...
+ *   views_hidden; this will result in views rendered in wrong locations.
+ *
+ * If multiple plugins are connected to this signal, they are called in the
+ * order defined by the logic in signal_provider_t; plugins should not depend
+ * on being called in a predictable order. Specifically, plugins should not
+ * expect views_hidden to be empty (and should not call clear() on it). It is OK
+ * for a plugin to move a view from views_hidden to views_shown, but this will
+ * likely not have predictable results.
+ *
+ * TODO: should there be a way for plugins to specify which order they want to
```
ammen99, line 29: I think this is fine as it is now. I expect all consumers of this API to move from visible to hidden based on some criteria, so it wouldn't make a difference how they are ordered.

plugins/scale/scale_title_filter.cpp
```diff
@@ -0,0 +1,339 @@
...
+        cairo_surface_upload_to_texture(surface, tex);
+        OpenGL::render_end();
+
+        if (!render_active)
+        {
+            output->render->add_effect(&render_hook, wf::OUTPUT_EFFECT_OVERLAY);
+            render_active = true;
+        }
+
+        output->render->damage_whole();
```
ammen99, line 263: can't we damage the correct geometry, not the whole output?

metadata/meson.build
```diff
@@ -29,3 +29,4 @@ install_data('wobbly.xml', install_dir: conf_data.get('PLUGIN_XML_DIR'))
 install_data('workarounds.xml', install_dir: conf_data.get('PLUGIN_XML_DIR'))
 install_data('wrot.xml', install_dir: conf_data.get('PLUGIN_XML_DIR'))
 install_data('zoom.xml', install_dir: conf_data.get('PLUGIN_XML_DIR'))
+install_data('scale_title_filter.xml', install_dir: conf_data.get('PLUGIN_XML_DIR'))
```
ammen99, line 32: sorry, just noticed it, we use `-` instead of `_` for file names

plugins/scale/scale_title_filter.cpp
```diff
@@ -0,0 +1,339 @@
+#include <cctype>
```
ammen99, line 1: also here, it should be `scale-title-filter.cpp` ...

ammen99:
I have tested this and it works quite well! I also have a few suggestions to what we can add. Let me know what you think about them.

- IMO we should have an option to also match the app_id. For example, if I am looking for my terminator window, I don't know what its title is because the title is the current working dir. I could make it an option if you are against such behavior: http://ix.io/2Igz

- Key repeat would be nice I think. I'm currently working on a patch to implement it.

ammen99:
> I have tested this and it works quite well! I also have a few suggestions to what we can add. Let me know what you think about them.
> 
> * IMO we should have an option to also match the app_id. For example, if I am looking for my terminator window, I don't know what its title is because the title is the current working dir. I could make it an option if you are against such behavior: http://ix.io/2Igz
> * Key repeat would be nice ~I think. I'm currently working on a patch to implement it.~ implemented, in #946
EDIT: I'll rebase my PR when this one is merged.

plugins/scale/scale_title_filter.cpp
dkondor: I did an implementation of this; it still damages the whole overlay surface, since I don't want to save the actual size of the text bubble

dkondor:
> I have tested this and it works quite well! I also have a few suggestions to what we can add. Let me know what you think about them.
> 
>     * IMO we should have an option to also match the app_id. For example, if I am looking for my terminator window, I don't know what its title is because the title is the current working dir. I could make it an option if you are against such behavior: http://ix.io/2Igz
> 
>     * Key repeat would be nice ~I think. I'm currently working on a patch to implement it.~ implemented, in #946
>       I'll rebase #946 once we merge this PR.

Sounds good, I think both changes improve this. I'll test #946

ammen99 (approved):
Looks good so far, I will put both my changes in the other PR, please test when you have time.
Thanks!

PR #932, opened 2020-12-10
Screensaver hide cursor

soreau:


soreau:
I'm having trouble hiding the cursor on one setup while it works on another. This attempt fails because grab_input() shows the cursor. It seems like grab() should take a cursor as optional argument or there should be proper cursor disable/enable ref counting API.

soreau:
What about just not setting the cursor in grab_input() by omitting [this line](https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/input-manager.cpp#L185)?

src/core/core-impl.hpp
```diff
@@ -58,9 +58,14 @@ class compositor_core_impl_t : public compositor_core_t
 
     wlr_seat *get_current_seat() override;
     uint32_t get_keyboard_modifiers() override;
+    void warp_cursor(wf::pointf_t pos) override;
+    /**
+     * The functions to hide and set the cursor use reference counting.
+     * A call to hide must come before a matched call to set.
```
ammen99, line 64: this doesn't make sense to me, please add `unhide_cursor` or rename to `set_cursor_visibility(bool is_visible)` which increments/decrements the counter

src/core/core-impl.hpp
```diff
@@ -58,9 +58,14 @@ class compositor_core_impl_t : public compositor_core_t
...
+     */
     void set_cursor(std::string name) override;
     void hide_cursor() override;
-    void warp_cursor(wf::pointf_t pos) override;
+    int hide_ref_counter = 0;
```
ammen99, line 68: this counter belongs in wf::cursor_t

src/core/core.cpp
```diff
@@ -373,12 +373,18 @@ uint32_t wf::compositor_core_impl_t::get_keyboard_modifiers()
 
 void wf::compositor_core_impl_t::set_cursor(std::string name)
```
ammen99, line 374: I'd prefer to put these checks in the wf::cursor_t functions - there are two of them there (one for clients, other for plugins, but both should have these guards).

ammen99 (changes requested):

src/core/core-impl.hpp
soreau: I considered unhide_cursor() but I didn't know what cursor it should show when unhiding. Should it take a string as arg? Assume "default"? Show the last cursor that was displayed?

src/core/core-impl.hpp
soreau: I also thought this at first but the trouble was that init_xcursor() keeps getting called which keeps resetting the variable and it does not work as intended.

src/core/core.cpp
soreau: The checks can go in wf::cursor_t class and I did try this first but settled on this after having to move the counter out of wf::cursor_t class.

src/core/core-impl.hpp
ammen99: I'd display 'default'

src/core/core-impl.hpp
ammen99: I don't understand, why do you need to reset that variable in init_xcursor?

plugins/cube/cube.cpp
```diff
@@ -397,7 +397,7 @@ class wayfire_cube : public wf::plugin_interface_t
     /* Mouse grab was released */
     void input_ungrabbed()
     {
-        wf::get_core().set_cursor("default");
```
ammen99, line 400: why do you need to unhide the cursor here? This will very likely corrupt the counter, because we haven't had a `hide` request before. I think it should still be a simple `set_cursor`

ammen99: Sorry, I mistakenly assumed this is in core.

soreau:
The bug I see is with follow focus. If I grab cube with follow focus enabled and move the cursor to another output and back, the cursor will appear. Maybe the cursor gets recreated at some point?

src/core/seat/cursor.cpp
```diff
@@ -33,7 +33,10 @@ wf::cursor_t::cursor_t(wf::seat_t *seat)
     request_set_cursor.set_callback([&] (void *data)
     {
         auto ev = static_cast<wlr_seat_pointer_request_set_cursor_event*>(data);
-        set_cursor(ev, true);
+        if (!this->hide_ref_counter)
```
ammen99, line 36: please also move this check in the function itself

ammen99 (approved):
LGTM aside from the comment below.

src/core/seat/cursor.cpp
soreau: Ah, I had made the change locally already but only pushed cube.cpp and forgot cursor.cpp

ammen99:
Thanks!

PR #933, opened 2020-12-10
add configurable configuration backend

### LINKED ISSUE #725: Add support for special config-provider plugins
@myfreeweb has his gsettings plugin which has some limitations, which arise from the fact that core has hardcoded support for dynamic (re)loading of the config file. We have discussed various approaches before, but now I have yet another way to do that.
What if we have a new kind of special plugins which are loaded once, at startup, before everything else is initialized. They would be responsible for populating and updating the global config struct.

To set the config backend plugin, we'd need a command line flag (with a build-time-configurable default value).
This would also make it even easier for other config backends if people are interested (for example, having a more sway-like config file, etc.)

Also, I'd give this plugin control over which plugins are loaded, in which order, etc.

What do you think @myfreeweb? Is this an overkill from design perspective? Any potential issues?

ammen99:
This PR adds the ability to change the configuration backend used by Wayfire.

The default configuration backend simply uses wf-config APIs, however, another backend could load
the settings directly from gsettings or whatever other format is desired.

Fixes #725

There are also changes to the output and input-device configuration syntax.
They now use the newly added object type
(and the newly added output types https://github.com/WayfireWM/wf-config/pull/37),
which can then be used to automatically type-check all output options. This should also
make it easier to write an output configuration section in WCM because of the easier parsing.

The only change required in the .ini file is to append `output:` or `input-device:` in front of the corresponding sections.
For example, `[eDP-1]` becomes `[output:eDP-1]`.


ammen99:
Also if someone tries that, keep in mind I have not had much time to test, so there might be some instability (although my tests so far indicate it works just fine).

damianatorrpm:
`For example, [eDP-1] becomes [output:eDP-1].`
With `-B /usr/lib64/wayfire/libdefault-config-backend.so` 
This doesn't seem to be enough. The output configuration is not used.
Note: I have both [eDP-1] and [output:eDP-1] in the config file.

ammen99:
> `For example, [eDP-1] becomes [output:eDP-1].`
> With `-B /usr/lib64/wayfire/libdefault-config-backend.so`
> This doesn't seem to be enough. The output configuration is not used.
> Note: I have both [DP-1] and [output:DP-1] in the config file, same for DP-2, HDMI-A-1 and HDMI-A-2.

As discussed on IRC, this is because of an invalid value for the output mode. The refresh rate is always an integer, either in Hz or mHz.

PR #938, opened 2020-12-13
cube: Pin pointer so it does not move while rotating

soreau:
When using cube with plugins like follow-focus output, grabbing the cube and
moving the pointer to another output resulted in the cube not reotating until
moving back to the grabbed output. This patch makes cube hold the pointer in
place while rotating, so it cannot move to any other output.

### DIFF
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -36,9 +36,6 @@ class wayfire_cube : public wf::plugin_interface_t
     wf::activator_callback rotate_left, rotate_right;
     wf::render_hook_t renderer;
 
-    /* Used to restore the pointer where the grab started */
-    wf::pointf_t saved_pointer_position;
-
     nonstd::observer_ptr<wf::workspace_stream_pool_t> streams;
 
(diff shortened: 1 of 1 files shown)


plugins/cube/cube.cpp
```diff
@@ -566,6 +562,17 @@ class wayfire_cube : public wf::plugin_interface_t
...
     }
 
+    wf::signal_callback_t on_motion_event = [=] (wf::signal_data_t *data)
+    {
+        auto ev = static_cast<
+            wf::input_event_signal<wlr_event_pointer_motion>*>(data);
+
+        pointer_moved(ev->event);
+
+        ev->event->delta_x = 0;
```
ammen99, line 572: there are also other deltas, `unaccel_dx/dy`, they should be 0 as well.

ammen99 (changes requested):
we should be able to drop `saved_pointer_position` with this change, right?

ammen99:
Thanks! (as a note, this won't work on nested backends but I'm fine with that)

ammen99 (approved):

PR #943, opened 2020-12-16
wayfire.ini: update config

travankor:


wayfire.ini
```diff
@@ -132,14 +135,11 @@ outputs = kanshi
...
-# http://jonls.dk/redshift/
-# Note: Requires Redshift with Wayland support.
-# https://github.com/jonls/redshift/pull/663
-redshift = redshift -m wayland
+# https://sr.ht/~kennylevinsen/wlsunset/
+gamma = wlsunset
 
 # Idle configuration
 # https://github.com/swaywm/swayidle
-# https://github.com/swaywm/swaylock
```
wb9688, line 142: Why is this lined removed?

travankor: you can set up swayidle from a config file now, plus there's a separate section for swaylock

ammen99: well the example itself uses swaylock so I think the link should remain there.

ammen99 (approved):
Thanks!

PR #946, opened 2020-12-16
scale filter add repeat

ammen99:
~~Based on #924,~~

This PR adds support for repeatable timers in core.
This is used to implement key repeat in scale-title-filter.

Also, add the ability to match app-id in scale-title-filter as well.


dkondor:
Hi,
I've just tested it a bit again and it works nicely; I think this is a good addition to the title filter.

PR #962, opened 2020-12-18
layer-shell: adjust keyboard-interactivity

dkondor:
Implementation for https://github.com/swaywm/wlr-protocols/pull/100 -- requires https://github.com/swaywm/wlroots/pull/2555 to actually work.

Only a value of 1 is interpreted as a request for exclusive focus. Other > 0 values result in normal focus semantics.

### DIFF
--- a/src/view/layer-shell.cpp
+++ b/src/view/layer-shell.cpp
@@ -303,7 +303,7 @@ struct wf_layer_shell_manager
             for (auto& v : layer)
             {
                 if (v->is_mapped() &&
-                    v->lsurface->client_pending.keyboard_interactive)
+                    (v->lsurface->client_pending.keyboard_interactive == 1))
                 {
                     focus_mask = std::max(focus_mask, (uint32_t)v->get_layer());
                 }


ammen99:
Marking as draft simply because the wlroots PR is not merged yet. Otherwise, LGTM.

wb9688:
>Other > 0 values result in normal focus semantics.

Shouldn't values > 2 result in an error? Or is that already handled by wlroots then?

dkondor:
> > Other > 0 values result in normal focus semantics.
> 
> Shouldn't values > 2 result in an error? Or is that already handled by wlroots then?

Yes, I think it is better to handle this in wlroots (unless other compositor implementations will want to be able to add their custom values):
https://github.com/swaywm/wlroots/blob/b4786c166d0f2eca1df7a1597910f1f9b05022b6/types/wlr_layer_shell_v1.c#L164

ammen99:
Hmm, actually, this would work on the current wlroots master too, right (we could merge this without breaking anything)?

dkondor:
Yes, I agree, it should not break anything, since wlroots always sets it to 0/1 currently.

ammen99 (approved):
Alright, let's merge it then. Thanks!

emersion:
Is the `keyboard_interactivity == 2` case already implemented in Wayfire?

ammen99:
> Is the `keyboard_interactivity == 2` case already implemented in Wayfire?

Not explicitly, but it will "just work" without any special casing.

PR #972, opened 2020-12-24
Emit view-show-window-menu signal from core

soreau:


soreau:
This is needed to support [wf-extra-protos](https://github.com/soreau/wf-extra-protos) which includes wf-menu, a menu app for wayfire using gtkmm4.

ammen99 (approved):
Thanks!

PR #978, opened 2020-12-25
skip deco for splash screens

damianatorrpm:
authored by @soreau


ammen99 (approved):
Thanks! (also thanks @soreau)

PR #979, opened 2020-12-27
grid: unset fullscreen for restore

### LINKED ISSUE #929: [grid] restore should reset decoration
**Describe the bug**
When restoring a full-screened view, the decorations are not reset.

**To Reproduce**
1. Enable grid and wm-actions plugins
2. Full screen view with wm-actions plugin
3. Restore view with grid plugin

**Expected behavior**
The decoration is reset when a full-screened view is no longer full screen.

**Wayfire version**
0.6.0

travankor:
fixes #929 

todo: need to find a way to set fullscreen for restore

### DIFF
--- a/plugins/single_plugins/grid.cpp
+++ b/plugins/single_plugins/grid.cpp
@@ -84,6 +84,7 @@ class wayfire_grid_view_cdata : public wf::custom_data_t
             this->tiled_edges = view->tiled_edges;
         } else
         {
+            this->view->set_fullscreen(0);
             this->tiled_edges = target_edges;
         }
 


soreau:
> todo: need to find a way to set fullscreen for restore

I'm not really sure this makes sense. If you un-fullscreen a view with this binding, then maximize it, then use the binding again, what are you expecting to happen?

travankor:
well, this might not be needed, but:

1) fullscreen view
2) move view
3) restore view

should thr last restore should re-fullscreen the view?

damianatorrpm:
In general that's what wayfire does. But 'grid' plugin doesn't to move operations, instead it does tile operation.
(I don't use grid so I'm not sure about it's details)

ammen99:
> 3\. 
> 
> should the last restore re-fullscreen the view?

I don't think so

ammen99 (approved):
Thanks!

PR #981, opened 2020-12-28
dynamic lists

ammen99:
Use the new compound_list option type in wf-config for easier parsing of options which are generated on-the-fly by matching options
with prefixes in the same section.

### DIFF
--- a/metadata/command.xml
+++ b/metadata/command.xml
@@ -7,11 +7,25 @@
 		<group>
 			<_short>Commands</_short>
 			<_long>Sets the command bindings.</_long>
-			<option name="command" type="dynamic_list">
+			<option name="bindings" type="dynamic-list">
 				<_short>Command</_short>
-				<_long>Sets the command binding.</_long>
-				<type>string</type>
-				<hint>file</hint>
+				<_long>Sets the command bindings.</_long>
+				<entry prefix="command_" type="string"/>
+				<entry prefix="binding_" type="activator"/>
+			</option>
+
+			<option name="repeatable_bindings" type="dynamic-list">
+				<_short>Command</_short>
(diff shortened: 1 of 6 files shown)


metadata/command.xml
```diff
@@ -7,11 +7,25 @@
...
+				<entry prefix="command_" type="string"/>
+				<entry prefix="repeatable_binding_" type="activator"/>
+			</option>
+
+			<option name="always_bindings" type="dynamic-list">
+				<_short>Command</_short>
+				<_long>Sets the bindings which are always available, even when screen is locked.</_long>
+				<entry prefix="command_" type="string"/>
+				<entry prefix="always_binding_" type="activator"/>
 			</option>
```
soreau, line 29: I would like to make a suggestion here. I believe the syntax should appear as follows:
```xml
<plugin>
	<object name="commands" type="dynamic-list">
		<option name="command" prefix="command_" type="string">
			<_short>Command</_short>
			<_long>The command the will be run when the binding is activated</_long>
			<hint>file</hint>
		</option>
		<option name="bindings" prefix="bindings_" type="string">
			<_short>Binding</_short>
			<_long>Regular command binding</_long>
		</option>
		<option name="repeatable_bindings" prefix="repeatable_bindings_" type="string">
			<_short>Repeatable Binding</_short>
			<_long>Repeatable command binding</_long>
		</option>
		<option name="always_bindings" prefix="always_bindings_" type="string">
			<_short>Always Binding</_short>
			<_long>Always command binding</_long>
		</option>
	</object>
</plugin>
```

ammen99:
Ready for merging once https://github.com/WayfireWM/wcm/pull/28 is reviewed.

PR #983, opened 2020-12-28
Various changes

myfreeweb:
Random things I've had in my local changes (hopefully doesn't change any behavior) + the command metadata change that allows gsettings to work with commands right now (will be obsoleted by #981 and its support in wf-gsettings). Feel free to only cherry-pick what you want

### DIFF
--- a/metadata/command.xml
+++ b/metadata/command.xml
@@ -7,9 +7,14 @@
 		<group>
 			<_short>Commands</_short>
 			<_long>Sets the command bindings.</_long>
+			<option name="binding" type="dynamic_list">
+				<_short>Binding</_short>
+				<_long>Sets the binding.</_long>
+				<type>activator</type>
+			</option>
 			<option name="command" type="dynamic_list">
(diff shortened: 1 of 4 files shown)


metadata/command.xml
```diff
@@ -7,9 +7,14 @@
 		<group>
 			<_short>Commands</_short>
 			<_long>Sets the command bindings.</_long>
+			<option name="binding" type="dynamic_list">
+				<_short>Binding</_short>
+				<_long>Sets the binding.</_long>
+				<type>activator</type>
+			</option>
 			<option name="command" type="dynamic_list">
```
soreau, line 15: How do you expect to have more than one dynamic_list in a plugin with this syntax? The dynamic_list options should be grouped in some way.

ammen99:
I think this is fine to merge, aside from the last commit (since, as you said, it will not be necessary very soon).

ammen99:
Submitted without the last commit in #999

Issue #980, opened 2020-12-28
[scale-title-filter] filtered without scale being active


damianatorrpm:
Wayfire latest git.

@dkondor Really like the scale title filter!

I found two small things:
Sometimes even when scale is not active I get the box that shows when typing during scale,
permanently being filled with null characters '[]'. The screen fills up with them.
To fix this I have to disable and re-enable scale-title-filter.

The other thing is if scale is active and you quickly write something and press escape to cancel scale 
the box that shows the characters stays on the display for a few seconds, I think it should be hidden when
scale finishes.

dkondor:
> Wayfire latest git.
> 
> @dkondor Really like the scale title filter!
> 
> I found two small things:
> Sometimes even when scale is not active I get the box that shows when typing during scale,
> permanently being filled with null characters '[]'. The screen fills up with them.
> To fix this I have to disable and re-enable scale-title-filter.

In what conditions does this happen? I have noticed that a key can get "stuck" when exiting scale. The following patch fixes this for me
```
diff --git a/plugins/scale/scale-title-filter.cpp b/plugins/scale/scale-title-filter.cpp
index 4b73f297..419aa62a 100644
--- a/plugins/scale/scale-title-filter.cpp
+++ b/plugins/scale/scale-title-filter.cpp
@@ -192,6 +192,7 @@ class scale_title_filter : public wf::plugin_interface_t
         wf::get_core().disconnect_signal(&scale_key);
         title_filter.clear();
         char_len.clear();
+        keys.clear();
         clear_overlay();
         scale_running = false;
     };
```
Can you test if this is the case?

> 
> The other thing is if scale is active and you quickly write something and press escape to cancel scale
> the box that shows the characters stays on the display for a few seconds, I think it should be hidden when
> scale finishes.

This is because the overlay is only cleared when the scale animation finishes, which takes some time (although it might not feel so). I have another patch that makes the overlay disappear instantly:
```
diff --git a/plugins/scale/scale.cpp b/plugins/scale/scale.cpp
index be021e37..4b3f28f3 100644
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -1308,11 +1308,17 @@ class wayfire_scale : public wf::plugin_interface_t
         }
 
         refocus();
+        output->emit_signal("scale-end", nullptr);
     }
 
     /* Completely end scale, including animation */
     void finalize()
     {
+        if (active)
+        {
[...]

PR #985, opened 2020-12-29
scale: fix title filter being stuck

dkondor:
This is a fix for the first issue in #980  -- key repeat can become stuck if scale ends while a key is pressed.


ammen99 (approved):
nice catch!

PR #986, opened 2020-12-29
scale: emit end signal earlier

dkondor:
With this change, the "scale-end" signal is emitted earlier, when scale starts to exit (whenever possible). This way, filters that display an overlay (such as the title filter) can deactivate earlier instead of only after the animation is finished.

This fixes the second issue in #980 (overlay displayed too long).


ammen99 (approved):
Makes sense, thanks!

PR #987, opened 2020-12-29
fast-switcher: allow switching backwards

dkondor:
Similarly to how it is possible in switcher. This makes fast-switcher more flexible. E.g. I often "overshoot" the view I'd like to switch to, and currently that requires me to go one more circle.


ammen99 (approved):
LGTM, thanks!

Issue #897, opened 2020-11-24
scale: [non-interact mode] show window decoration


damianatorrpm:
It's already possible to close a view on middle click, but a visual distinction, (that provides also a way to close) from between the hovered window and the others would be nice: 
**https://i0.wp.com/9to5linux.com/wp-content/uploads/2020/01/release-2020.1-gnome-shell-overview.jpeg?w=1360&ssl=1**
normal vv          when hovered ^^
https://upload.wikimedia.org/wikipedia/en/c/c2/Shows_Overview_mode_%28%22Activities%22%29_in_GNOME_3.8.png

The text (window title) is not part of this request.

PR #996, opened 2021-01-02
Show views' title as an overlay during scale

### LINKED ISSUE #990: Scale plugin - optionally show window title
It would be nice if the scale plugin could label each window with its title, as sometimes I have two windows open that aren't necessarily immediately distinguishable by their appearance.

dkondor:
This is an implementation for #990 

Some notes:
 - I've made this part of the title filter plugin, since I'm unsure if it makes sense to add and extra plugin and also to share code more easily; maybe the title filter plugin could be renamed to "Scale addons" or something similar?
 - The helper classes: `cairo_text_t` and `view_overlay` could be part of some common tools. This way, other plugins can re-use them to render text and to add overlays to views.
 - The `view_overlay` transformer is not efficient in its current form, since there is no need to copy the view's texture to a new framebuffer. I wonder if there is a way to add a transformer that renders to the latest framebuffer without copying that I missed, or if it would make sense to add this functionality?
 - I could not find an easy way so far to take output scaling into account; this should probably be fixed before merging, so any suggestion on this is welcome :)
 - Also, this would benefit from better damage tracking. For the current case, it is not really a problem, since the transformers added by scale will need to damage the whole view anyway.

Fixes #990

### DIFF
--- a/plugins/common/wayfire/plugins/common/cairo-util.hpp
+++ b/plugins/common/wayfire/plugins/common/cairo-util.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <string>
 #include <wayfire/plugins/common/simple-texture.hpp>
+#include <wayfire/config/types.hpp>
 #include <cairo.h>
 
 namespace wf
@@ -33,4 +35,256 @@ static void cairo_surface_upload_to_texture(
     GL_CALL(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_B, GL_RED));
     GL_CALL(glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA,
         buffer.width, buffer.height, 0, GL_RGBA, GL_UNSIGNED_BYTE, src));
+}
+
+namespace wf
+{
+/**
+ * Simple wrapper around rendering text with Cairo. This object can be
+ * kept around to avoid reallocation of the cairo surface and OpenGL
+ * texture on repeated renders.
+ */
+struct cairo_text_t
+{
+    wf::simple_texture_t tex;
+
+    /* parameters used for rendering */
+    struct params
+    {
+        /* font size */
+        int font_size = 12;
+        /* color for background rectangle (only used if bg_rect == true) */
+        wf::color_t bg_color;
+        /* text color */
+        wf::color_t text_color;
+        /* scale everything by this amount */
+        float output_scale = 1.f;
+        /* crop result to this size (if nonzero);
+         * note that this is multiplied by output_scale */
+        wf::dimensions_t max_size{0, 0};
+        /* draw a rectangle in the background with bg_color */
+        bool bg_rect = true;
+        /* round the corners of the background rectangle */
+        bool rounded_rect = true;
+        /* if true, the resulting surface will be cropped to the
+         * minimum size necessary to fit the text; otherwise, the
+         * resulting surface might be bigger than necessary and the
+         * text is centered in it */
+        bool exact_size = false;
+
+        params()
+        {}
+        params(int font_size_, const wf::color_t& bg_color_,
+            const wf::color_t& text_color_, float output_scale_ = 1.f,
+            const wf::dimensions_t& max_size_ = {0, 0},
+            bool bg_rect_ = true, bool exact_size_ = false) :
+            font_size(font_size_), bg_color(bg_color_),
+            text_color(text_color_), output_scale(output_scale_),
+            max_size(max_size_), bg_rect(bg_rect_),
+            exact_size(exact_size_)
+        {}
+    };
+
+    /**
+     * Render the given text in the texture tex.
+     *
+     * @param text         text to render
+     * @param par          parameters for rendering
+     *
+     * @return The size needed to render in scaled coordinates. If this is larger
+     *   than the size of tex, it means the result was cropped (due to the constraint
+     *   given in par.max_size). If it is smaller, than the result is centered along
+     *   that dimension.
+     */
+    wf::dimensions_t render_text(const std::string& text, const params& par)
+    {
+        if (!cr)
+        {
+            /* create with default size */
+            cairo_create_surface();
+        }
+
+        cairo_text_extents_t extents;
+        cairo_font_extents_t font_extents;
+        /* TODO: font properties could be made parameters! */
+        cairo_select_font_face(cr, "sans-serif", CAIRO_FONT_SLANT_NORMAL,
+            CAIRO_FONT_WEIGHT_BOLD);
+        cairo_set_font_size(cr, par.font_size * par.output_scale);
+        cairo_text_extents(cr, text.c_str(), &extents);
+        cairo_font_extents(cr, &font_extents);
+
+        double xpad = par.bg_rect ? 10.0 * par.output_scale : 0.0;
+        double ypad = par.bg_rect ? 0.2 * (font_extents.ascent +
+            font_extents.descent) : 0.0;
+        int w = (int)(extents.width + 2 * xpad);
+        int h = (int)(font_extents.ascent + font_extents.descent + 2 * ypad);
+        wf::dimensions_t ret = {w, h};
+        if (par.max_size.width && (w > par.max_size.width * par.output_scale))
+        {
+            w = (int)std::floor(par.max_size.width * par.output_scale);
+        }
+
+        if (par.max_size.height && (h > par.max_size.height * par.output_scale))
+        {
+            h = (int)std::floor(par.max_size.height * par.output_scale);
+        }
+
+        if ((w != surface_size.width) || (h != surface_size.height))
+        {
+            if (par.exact_size || (w > surface_size.width) ||
+                (h > surface_size.height))
+            {
+                surface_size.width  = w;
+                surface_size.height = h;
+                cairo_create_surface();
+            }
+        }
+
+        cairo_set_operator(cr, CAIRO_OPERATOR_CLEAR);
+        cairo_paint(cr);
+
+        int x = (surface_size.width - w) / 2;
+        int y = (surface_size.height - h) / 2;
+
+        if (par.bg_rect)
+        {
+            int min_r = (int)(20 * par.output_scale);
+            int r     = par.rounded_rect ? (h > min_r ? min_r : (h - 2) / 2) : 0;
+
+            cairo_move_to(cr, x + r, y);
+            cairo_line_to(cr, x + w - r, y);
+            if (par.rounded_rect)
+            {
+                cairo_curve_to(cr, x + w, y, x + w, y, x + w, y + r);
+            }
+
+            cairo_line_to(cr, x + w, y + h - r);
+            if (par.rounded_rect)
+            {
+                cairo_curve_to(cr, x + w, y + h, x + w, y + h, x + w - r, y + h);
+            }
+
+            cairo_line_to(cr, x + r, y + h);
+            if (par.rounded_rect)
+            {
+                cairo_curve_to(cr, x, y + h, x, y + h, x, y + h - r);
+            }
+
+            cairo_line_to(cr, x, y + r);
+            if (par.rounded_rect)
+            {
+                cairo_curve_to(cr, x, y, x, y, x + r, y);
+            }
+
+            cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
+            cairo_set_source_rgba(cr, par.bg_color.r, par.bg_color.g,
+                par.bg_color.b, par.bg_color.a);
+            cairo_fill(cr);
+        }
+
+        x += xpad;
+        y += ypad + font_extents.ascent;
+        cairo_select_font_face(cr, "sans-serif", CAIRO_FONT_SLANT_NORMAL,
+            CAIRO_FONT_WEIGHT_BOLD);
+        cairo_set_font_size(cr, par.font_size * par.output_scale);
+        cairo_move_to(cr, x - extents.x_bearing, y);
+        cairo_set_source_rgba(cr, par.text_color.r, par.text_color.g,
+            par.text_color.b, par.text_color.a);
+        cairo_show_text(cr, text.c_str());
+
+        cairo_surface_flush(surface);
+        OpenGL::render_begin();
+        cairo_surface_upload_to_texture(surface, tex);
+        OpenGL::render_end();
+
+        return ret;
+    }
+
+    /**
+     * Standalone function version to render text to an OpenGL texture
+     */
+    static wf::dimensions_t cairo_render_text_to_texture(const std::string& text,
+        const wf::cairo_text_t::params& par, wf::simple_texture_t& tex)
+    {
+        wf::cairo_text_t ct;
+        /* note: we "borrow" the texture from what was supplied (if any) */
+        ct.tex.tex = tex.tex;
+        auto ret = ct.render_text(text, par);
+        if (tex.tex == (GLuint) - 1)
+        {
+            tex.tex = ct.tex.tex;
+        }
+
+        tex.width  = ct.tex.width;
+        tex.height = ct.tex.height;
+        ct.tex.tex = -1;
+        return ret;
+    }
+
+    ~cairo_text_t()
+    {
+        cairo_free();
+    }
+
+    /**
+     * Calculate the height of text rendered with a given font size.
+     *
+     * @param font_size  Desired font size.
+     * @param bg_rect    Whether a background rectangle should be taken into account.
+     *
+     * @returns Required height of the surface.
+     */
+    static unsigned int measure_height(int font_size, bool bg_rect = true)
+    {
+        cairo_text_t dummy;
+        dummy.surface_size.width  = 1;
+        dummy.surface_size.height = 1;
+        dummy.cairo_create_surface();
+
+        cairo_font_extents_t font_extents;
+        /* TODO: font properties could be made parameters! */
+        cairo_select_font_face(dummy.cr, "sans-serif", CAIRO_FONT_SLANT_NORMAL,
+            CAIRO_FONT_WEIGHT_BOLD);
+        cairo_set_font_size(dummy.cr, font_size);
+        cairo_font_extents(dummy.cr, &font_extents);
+
+        double ypad = bg_rect ? 0.2 * (font_extents.ascent +
+            font_extents.descent) : 0.0;
+        unsigned int h = (unsigned int)std::ceil(font_extents.ascent +
+            font_extents.descent + 2 * ypad);
+        return h;
+    }
+
+  protected:
+    /* cairo context and surface for the text */
+    cairo_t *cr = nullptr;
+    cairo_surface_t *surface = nullptr;
+    /* current width and height of the above surface */
+    wf::dimensions_t surface_size = {400, 100};
+
+
+    void cairo_free()
+    {
+        if (cr)
+        {
+            cairo_destroy(cr);
+        }
+
+        if (surface)
+        {
+            cairo_surface_destroy(surface);
+        }
+
+        cr = nullptr;
+        surface = nullptr;
+    }
+
+    void cairo_create_surface()
+    {
+        cairo_free();
+        surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, surface_size.width,
+            surface_size.height);
+        cr = cairo_create(surface);
+    }
+};
 }
--- a/plugins/scale/scale-title-overlay.cpp
+++ b/plugins/scale/scale-title-overlay.cpp
@@ -1,0 +1,557 @@
+#include "scale-title-overlay.hpp"
+
+#include <wayfire/opengl.hpp>
+#include <wayfire/util/log.hpp>
+#include <wayfire/plugins/common/cairo-util.hpp>
+#include <wayfire/plugins/common/simple-texture.hpp>
+
+/**
+ * Get the topmost parent of a view.
+ */
+static wayfire_view find_toplevel_parent(wayfire_view view)
+{
+    while (view->parent)
+    {
+        view = view->parent;
+    }
+
+    return view;
+}
+
+/**
+ * Class storing an overlay with a view's title, only stored for parent views.
+ */
+struct view_title_texture_t : public wf::custom_data_t
+{
+    wayfire_view view;
+    wf::cairo_text_t overlay;
+    wf::cairo_text_t::params par;
+    bool overflow = false;
+    wayfire_view dialog; /* the texture should be rendered on top of this dialog */
+
+    /**
+     * Render the overlay text in our texture, cropping it to the size by
+     * the given box.
+     */
+    void update_overlay_texture(wf::dimensions_t dim)
+    {
+        par.max_size = dim;
+        update_overlay_texture();
+    }
+
+    void update_overlay_texture()
+    {
+        auto res = overlay.render_text(view->get_title(), par);
+        overflow = res.width > overlay.tex.width;
+    }
+
+    wf::signal_connection_t view_changed = [this] (auto)
+    {
+        if (overlay.tex.tex != (GLuint) - 1)
+        {
+            update_overlay_texture();
+        }
+    };
+
+    view_title_texture_t(wayfire_view v, int font_size, const wf::color_t& bg_color,
+        const wf::color_t& text_color, float output_scale) : view(v)
+    {
+        par.font_size    = font_size;
+        par.bg_color     = bg_color;
+        par.text_color   = text_color;
+        par.exact_size   = true;
+        par.output_scale = output_scale;
+
+        view->connect_signal("title-changed", &view_changed);
+    }
+};
+
+/**
+ * Class with the overlay hooks, added to scale's transformer.
+ */
+class view_title_overlay_t : public wf::scale_transformer_t::overlay_t
+{
+  public:
+    enum class position
+    {
+        TOP,
+        CENTER,
+        BOTTOM,
+    };
+
+  protected:
+    /* the transformer we are attached to */
+    wf::scale_transformer_t& tr;
+    /* save the transformed view, since we need it in the destructor */
+    wayfire_view view;
+    /* the position on the screen we currently render to */
+    wf::geometry_t geometry;
+    scale_show_title_t& parent;
+    unsigned int text_height; /* set in the constructor, should not change */
+    position pos = position::CENTER;
+    /* Whether we are currently rendering the overlay by this transformer.
+     * Set in the pre-render hook and used in the render function. */
+    bool overlay_shown = false;
+
+    /**
+     * Get the transformed WM geometry of the view transformed by the given
+     * transformer, including the current transform, but not any padding.
+     */
+    wlr_box get_transformed_wm_geometry(wf::scale_transformer_t& tr)
+    {
+        wlr_box box = tr.get_transformed_view()->get_wm_geometry();
+        return tr.trasform_box_without_padding(box);
+    }
+
+    wlr_box get_transformed_wm_geometry()
+    {
+        return get_transformed_wm_geometry(tr);
+    }
+
+    /**
+     * Get the transformed WM geometry of the given view,
+     * including the current transform, but not any padding.
+     */
+    wlr_box get_transformed_wm_geometry(wayfire_view view)
+    {
+        auto tmp =
+            view->get_transformer(wf::scale_transformer_t::transformer_name());
+        if (!tmp)
+        {
+            /* This might happen if view is a newly created dialog that does not
+             * yet have a transformed added to it. In this case, return a zero
+             * size box that will never overlap with the overlay. */
+            return {0, 0, 0, 0};
+        }
+
+        /* TODO: can we use static_cast in this case? */
+        auto tr = dynamic_cast<wf::scale_transformer_t*>(tmp.get());
+        assert(tr);
+
+        return get_transformed_wm_geometry(*tr);
+    }
+
+    /**
+     * Gets the overlay texture stored with the given view.
+     */
+    view_title_texture_t& get_overlay_texture(wayfire_view view)
+    {
+        auto data = view->get_data<view_title_texture_t>();
+        if (!data)
+        {
+            auto new_data = new view_title_texture_t(view, parent.title_font_size,
+                parent.bg_color, parent.text_color, parent.output->handle->scale);
+            view->store_data<view_title_texture_t>(std::unique_ptr<view_title_texture_t>(
+                new_data));
+            return *new_data;
+        }
+
+        return *data.get();
+    }
+
+    /**
+     * Get the bounding box of the topmost parent of this view without and
+     * padding added by scale's transformer.
+     */
+    wlr_box get_parent_box()
+    {
+        auto view = tr.get_transformed_view();
+        auto box  = get_transformed_wm_geometry(find_toplevel_parent(view));
+        if ((box.width == 0) || (box.height == 0))
+        {
+            /* This is the case if the parent does not have a transformer. This
+             * should normally not happen, but might be the case if this view is
+             * assigned as the child of a newly created view that does not yet
+             * have a transformer.
+             * TODO: check if this case is actually possible! */
+            return get_transformed_wm_geometry();
+        }
+
+        return box;
+    }
+
+    /**
+     * Check if this view should display an overlay.
+     */
+    bool should_have_overlay(view_title_texture_t& title)
+    {
+        if (this->parent.show_view_title_overlay ==
+            scale_show_title_t::title_overlay_t::NEVER)
+        {
+            return false;
+        }
+
+        auto parent = find_toplevel_parent(view);
+
+        if ((this->parent.show_view_title_overlay ==
+             scale_show_title_t::title_overlay_t::MOUSE) &&
+            (this->parent.last_title_overlay != parent))
+        {
+            return false;
+        }
+
+        if (view == parent)
+        {
+            /* Check if the overlay overlaps with any dialogs. */
+
+            /* Update maximum possible extents of the overlay */
+            auto max_geom = get_transformed_wm_geometry();
+            switch (pos)
+            {
+              case position::CENTER:
+                max_geom.y += (max_geom.height - text_height) / 2;
+                break;
+
+              case position::TOP:
+                max_geom.y -= (text_height + 1);
+                break;
+
+              case position::BOTTOM:
+                max_geom.y += max_geom.height;
+                break;
+            }
+
+            max_geom.height = text_height + 1;
+
+            title.dialog = view;
+            for (auto dialog : view->enumerate_views(false))
+            {
+                if ((dialog == view) || !dialog->is_visible())
+                {
+                    continue;
+                }
+
+                auto dialog_box = get_transformed_wm_geometry(dialog);
+                if (dialog_box & max_geom)
+                {
+                    title.dialog = dialog;
+                    break;
+                }
+            }
+        }
+
+        return view == title.dialog;
+    }
+
+    /**
+     * Pre-render hook: calculates new position and optionally re-renders the text.
+     */
+    bool pre_render()
+    {
+        bool ret  = false;
+        auto& tex = get_overlay_texture(find_toplevel_parent(view));
+        if (!should_have_overlay(tex))
+        {
+            if (overlay_shown)
+            {
+                ret = true;
+                overlay_shown = false;
+            }
+
+            view_padding = {0, 0, 0, 0};
+            return ret;
+        }
+
+        if (!overlay_shown)
+        {
+            overlay_shown = true;
+            ret = true;
+        }
+
+        auto box = get_parent_box(); // will return our box if there is no parent
+        auto output_scale = parent.output->handle->scale;
+
+        /**
+         * regenerate the overlay texture in the following cases:
+         * 1. Output's scale changed
+         * 2. The overlay does not fit anymore
+         * 3. The overlay previously did not fit, but there is more space now
+         * TODO: check if this wastes too high CPU power when views are being
+         * animated and maybe redraw less frequently
+         */
+        if ((tex.overlay.tex.tex == (GLuint) - 1) ||
+            (output_scale != tex.par.output_scale) ||
+            (tex.overlay.tex.width > box.width * output_scale) ||
+            (tex.overflow &&
+             (tex.overlay.tex.width < std::floor(box.width * output_scale))))
+        {
+            tex.par.output_scale = output_scale;
+            tex.update_overlay_texture({box.width, box.height});
+            ret = true;
+        }
+
+        int w = tex.overlay.tex.width;
+        int h = tex.overlay.tex.height;
+        int y = 0;
+        switch (pos)
+        {
+          case position::TOP:
+            y = box.y - (int)(h / output_scale);
+            break;
+
+          case position::CENTER:
+            y = box.y + box.height / 2 - (int)(h / output_scale / 2);
+            break;
+
+          case position::BOTTOM:
+            y = box.y + box.height;
+            break;
+        }
+
+        geometry = {box.x + box.width / 2 - (int)(w / output_scale / 2),
+            y, (int)(w / output_scale), (int)(h / output_scale)};
+
+        /* We need to ensure that geometry is within our box. */
+        if (view->parent || (pos != position::CENTER))
+        {
+            /* get out own box (previously we might have had the parent's box */
+            if (view->parent)
+            {
+                box = get_transformed_wm_geometry();
+            }
+
+            view_padding = {0, 0, 0, 0};
+            if (geometry.x < box.x)
+            {
+                view_padding.left = box.x - geometry.x;
+            }
+
+            if (geometry.x + geometry.width > box.x + box.width)
+            {
+                view_padding.right = (geometry.x + geometry.width) -
+                    (box.x + box.width);
+            }
+
+            if (geometry.y < box.y)
+            {
+                view_padding.top = box.y - geometry.y;
+            }
+
+            if (geometry.y + geometry.height > box.y + box.height)
+            {
+                view_padding.bottom = (geometry.y + geometry.height) -
+                    (box.y + box.height);
+            }
+
+            /* note: no need to call damage, the transformer will check if the
+             * padding has changed and will damage the view accordingly */
+        } else
+        {
+            view_padding = {0, 0, 0, 0};
+        }
+
+        return ret;
+    }
+
+    void render(const wf::framebuffer_t& fb, const wf::region_t& damage)
+    {
+        if (!overlay_shown)
+        {
+            return;
+        }
+
+        view_title_texture_t& title = get_overlay_texture(find_toplevel_parent(
+            tr.get_transformed_view()));
+
+        GLuint tex = title.overlay.tex.tex;
+
+        if (tex == (GLuint) - 1)
+        {
+            /* this should not happen */
+            return;
+        }
+
+        auto ortho = fb.get_orthographic_projection();
+        OpenGL::render_begin(fb);
+        for (const auto& box : damage)
+        {
+            fb.logic_scissor(wlr_box_from_pixman_box(box));
+            OpenGL::render_transformed_texture(tex, geometry, ortho,
+                {1.0f, 1.0f, 1.0f, tr.alpha}, OpenGL::TEXTURE_TRANSFORM_INVERT_Y);
+        }
+
+        OpenGL::render_end();
+    }
+
+  public:
+    view_title_overlay_t(wf::scale_transformer_t& tr_, position pos_,
+        scale_show_title_t& parent_) : tr(tr_), view(tr.get_transformed_view()),
+        parent(parent_), pos(pos_)
+    {
+        auto parent = find_toplevel_parent(view);
+        auto& title = get_overlay_texture(parent);
+
+        if (title.overlay.tex.tex != (GLuint) - 1)
+        {
+            text_height = (unsigned int)std::ceil(
+                title.overlay.tex.height / title.par.output_scale);
+        } else
+        {
+            text_height =
+                wf::cairo_text_t::measure_height(title.par.font_size, true);
+        }
+
+        /* add padding required by scale */
+        if (pos == position::BOTTOM)
+        {
+            scale_padding.bottom = text_height;
+        } else if (pos == position::TOP)
+        {
+            scale_padding.top = text_height;
+        }
+
+        pre_hook = [this] ()
+        {
+            return pre_render();
+        };
+        render_hook = [this] (
+            const wf::framebuffer_t& fb,
+            const wf::region_t& damage)
+        {
+            render(fb, damage);
+        };
+    }
+
+    ~view_title_overlay_t()
+    {
+        view->erase_data<view_title_texture_t>();
+        if (view->parent && overlay_shown)
+        {
+            auto parent = find_toplevel_parent(view);
+            /* we need to recalculate which dialog should show the overlay */
+            auto tmp = parent->get_transformer(
+                wf::scale_transformer_t::transformer_name());
+            auto tr = dynamic_cast<wf::scale_transformer_t*>(tmp.get());
+            if (tr)
+            {
+                tr->call_pre_hooks(false);
+            }
+        }
+    }
+};
+
+scale_show_title_t::scale_show_title_t() :
+    view_filter{[this] (auto)
+    {
+        update_title_overlay_opt();
+    }},
+
+    scale_end{[this] (wf::signal_data_t*)
+    {
+        show_view_title_overlay = title_overlay_t::NEVER;
+        last_title_overlay = nullptr;
+        mouse_update.disconnect();
+    }
+},
+
+add_title_overlay{[this] (wf::signal_data_t *data)
+    {
+        const std::string& opt = show_view_title_overlay_opt;
+        if (opt == "never")
+        {
+            /* TODO: support changing this option while scale is running! */
+            return;
+        }
+
+        const std::string& pos_opt = title_position;
+        view_title_overlay_t::position pos = view_title_overlay_t::position::CENTER;
+        if (pos_opt == "top")
+        {
+            pos = view_title_overlay_t::position::TOP;
+        } else if (pos_opt == "bottom")
+        {
+            pos = view_title_overlay_t::position::BOTTOM;
+        }
+
+        auto signal = static_cast<scale_transformer_added_signal*>(data);
+        auto tr     = signal->transformer;
+        auto ol     = new view_title_overlay_t(*tr, pos, *this);
+
+        tr->add_overlay(std::unique_ptr<wf::scale_transformer_t::overlay_t>(ol), 1);
+    }
+},
+
+mouse_update{[this] (auto)
+    {
+        update_title_overlay_mouse();
+    }
+}
+
+{}
+
+void scale_show_title_t::init(wf::output_t *output)
+{
+    this->output = output;
+    output->connect_signal("scale-filter", &view_filter);
+    output->connect_signal("scale-transformer-added", &add_title_overlay);
+    output->connect_signal("scale-end", &scale_end);
+}
+
+void scale_show_title_t::fini()
+{
+    mouse_update.disconnect();
+}
+
+void scale_show_title_t::update_title_overlay_opt()
+{
+    const std::string& tmp = show_view_title_overlay_opt;
+    if (tmp == "all")
+    {
+        show_view_title_overlay = title_overlay_t::ALL;
+    } else if (tmp == "mouse")
+    {
+        show_view_title_overlay = title_overlay_t::MOUSE;
+    } else
+    {
+        show_view_title_overlay = title_overlay_t::NEVER;
+    }
+
+    if (show_view_title_overlay == title_overlay_t::MOUSE)
+    {
+        update_title_overlay_mouse();
+        mouse_update.disconnect();
+        wf::get_core().connect_signal("pointer_motion_absolute_post", &mouse_update);
+        wf::get_core().connect_signal("pointer_motion_post", &mouse_update);
+    }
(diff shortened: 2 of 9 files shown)


ammen99:
> * I've made this part of the title filter plugin, since I'm unsure if it makes sense to add and extra plugin and also to share code more easily; maybe the title filter plugin could be renamed to "Scale addons" or something similar?

The two functionalities (filtering based on title and showing the titles themselves) seem related. So, I think it is fine to have them in the same plugin.




> * 
> * The helper classes: `cairo_text_t` and `view_overlay` could be part of some common tools. This way, other plugins can re-use them to render text and to add overlays to views.

`cairo_text_t` seems pretty generic - so it might be a good fit for `plugins/common/...`




> * 
> * The `view_overlay` transformer is not efficient in its current form, since there is no need to copy the view's texture to a new framebuffer. I wonder if there is a way to add a transformer that renders to the latest framebuffer without copying that I missed, or if it would make sense to add this functionality?

There are two ways I can think of, which will do this, with different trade-offs:

- Add a subsurface above the main surface. However, this subsurface will be affected by the scale's transformer, so the text there would get tiny if you have a lot of open windows, though maybe that's desired? This approach has the benefit that the scale's fade-in/fade-out for hidden views will also fade-in/out the text.
- Render the text as an overlay on top of the whole screen - like the filter text is drawn currently. You'll have to iterate through all views, check their scale transformer and visibility to determine whether to draw text (and what alpha to use for fade-in/out). You can use `view->get_bounding_box()` to figure out the extents of each view on the screen.

> * I could not find an easy way so far to take output scaling into account; this should probably be fixed before merging, so any suggestion on this is welcome :)

This is usually handled like this: assume output scale 2. We want to render a 100x100 rectangle with text size 20. So, we actually create a cairo surface (and then a GL texture) of size 200x200 with text size 40. I hope this is insightful enough ;)

> * 
> * Also, this would benefit from better damage tracking. For the current case, it is not really a problem, since the transformers added by scale will need to damage the whole view anyway.

True, I don't think this is a problem at the current stage, the view transformers will consume much more resources anyway.

dkondor:
> > * The helper classes: `cairo_text_t` and `view_overlay` could be part of some common tools. This way, other plugins can re-use them to render text and to add overlays to views.
> 
> `cairo_text_t` seems pretty generic - so it might be a good fit for `plugins/common/...`

ok, I'm moving it to cairo-util.hpp

> > * The `view_overlay` transformer is not efficient in its current form, since there is no need to copy the view's texture to a new framebuffer. I wonder if there is a way to add a transformer that renders to the latest framebuffer without copying that I missed, or if it would make sense to add this functionality?
> 
> There are two ways I can think of, which will do this, with different trade-offs:
> 
>     * Add a subsurface above the main surface. However, this subsurface will be affected by the scale's transformer, so the text there would get tiny if you have a lot of open windows, though maybe that's desired? This approach has the benefit that the scale's fade-in/fade-out for hidden views will also fade-in/out the text.

I think it is better to have the same font size for all views for consistency and readability. It would be nice to fade together with the view, now it can look a bit weird when a view is filtered out.

>     * Render the text as an overlay on top of the whole screen - like the filter text is drawn currently. You'll have to iterate through all views, check their scale transformer and visibility to determine whether to draw text (and what alpha to use for fade-in/out). You can use `view->get_bounding_box()` to figure out the extents of each view on the screen.

This sounds a bit complicated, especially when views overlap during an animation. I'll think of this a bit more if there might be a better way.

> > * I could not find an easy way so far to take output scaling into account; this should probably be fixed before merging, so any suggestion on this is welcome :)
> 
> This is usually handled like this: assume output scale 2. We want to render a 100x100 rectangle with text size 20. So, we actually create a cairo surface (and then a GL texture) of size 200x200 with text size 40. I hope this is insightful enough ;)

Ok, I think I managed this again. My problem was that I'm not sure how to access the output scale beside looking at target framebuffer's scale.

ammen99:
> My problem was that I'm not sure how to access the output scale beside looking at target framebuffer's scale.

Ah, yes, sorry - you should almost never look at `output->handle->scale`, the framebuffer scale is all a plugin should ever care about.

> This sounds a bit complicated, especially when views overlap during an animation. I'll think of this a bit more if there might be a better way.

Okay, you're right, my two ideas aren't ideal. Here's a third one: replace the scale transformer with your own, which is a subclass of view_2D. This way, scale will use the view_2D bits, but in your plugin, you'll be able to draw additional stuff in there. Whatever approach we decide on could also be useful for #897. 

For the actual implementation, I think it might be a bit difficult because scale currently just keeps an std::map for each view and stores a pointer to the transformer. In general, it seems a good idea to expose this data to any scale extensions that we might write, and this way the transformer will be overwriteable. Let me know if you have better ideas, even if they involve a bit more drastic changes to how scale is implemented.

damianatorrpm:
I tested it and it works great, the thing is if you have a terminal the text is unreadable,
gnome solves this by having the view title slightly below the view:
https://help.gnome.org/misc/release-notes/3.6/figures/users-activities-overview.png

dkondor:
> > My problem was that I'm not sure how to access the output scale beside looking at target framebuffer's scale.
> 
> Ah, yes, sorry - you should almost never look at `output->handle->scale`, the framebuffer scale is all a plugin should ever care about.
> 
> > This sounds a bit complicated, especially when views overlap during an animation. I'll think of this a bit more if there might be a better way.
> 
> Okay, you're right, my two ideas aren't ideal. Here's a third one: replace the scale transformer with your own, which is a subclass of view_2D. This way, scale will use the view_2D bits, but in your plugin, you'll be able to draw additional stuff in there. Whatever approach we decide on could also be useful for #897.
> 
> For the actual implementation, I think it might be a bit difficult because scale currently just keeps an std::map for each view and stores a pointer to the transformer. In general, it seems a good idea to expose this data to any scale extensions that we might write, and this way the transformer will be overwriteable. Let me know if you have better ideas, even if they involve a bit more drastic changes to how scale is implemented.

This sounds like a good idea. I created a simpler implementation first where scale's transformer allows adding overlays (similar to how post effects can be added to the render loop -- actually, since this is not specific to scale, so `scale_transformer` might be part of the core view transformer interface). This way, the transformer does not need to be replaced, but only overlays can be added on top of the transformed view. #897 could be implemented by adding another overlay (although I'm unsure how clicks would be handled, since those are currently implemented by scale).

What I could not figure out yet is how to increase the bounding box of the view that would allow rendering outside, as suggested by @damianatorrpm and what might be needed for #897.

ammen99:
> What I could not figure out yet is how to increase the bounding box of the view that would allow rendering outside

Each transformer has a `get_bounding_box` method which calculates the view bounding box, you could just return a larger rectangle.

dkondor:
> > What I could not figure out yet is how to increase the bounding box of the view that would allow rendering outside
> 
> Each transformer has a `get_bounding_box` method which calculates the view bounding box, you could just return a larger rectangle.

I've tried this and seems to work. Although in this case, any box passed will be padded (it does not seem to be a problem in practice though).

What does not work is that scale does not take into account this extra space when positioning view, so the overlays can overlap with other views. I'm not sure how this can be fixed, since plugins that want to add overlays would need to inform scale of the size of the padding that will be added in advance. Of course, the easiest workaround is just to increase the spacing in the settings for scale.

JordanL2:
> What does not work is that scale does not take into account this extra space when positioning view, so the overlays can overlap with other views. I'm not sure how this can be fixed, since plugins that want to add overlays would need to inform scale of the size of the padding that will be added in advance. Of course, the easiest workaround is just to increase the spacing in the settings for scale.

I've been wondering why this functionality wasn't added into the scale plugin itself, with an option to enable it, since people might not want the scale title filter functionality but still want this. Wouldn't adding it into scale also make it easier to fix this issue?

dkondor:
> > What does not work is that scale does not take into account this extra space when positioning view, so the overlays can overlap with other views. I'm not sure how this can be fixed, since plugins that want to add overlays would need to inform scale of the size of the padding that will be added in advance. Of course, the easiest workaround is just to increase the spacing in the settings for scale.
> 
> I've been wondering why this functionality wasn't added into the scale plugin itself, with an option to enable it, since people might not want the scale title filter functionality but still want this. Wouldn't adding it into scale also make it easier to fix this issue?

Mainly to not make the scale plugin code even more bigger and complex, but also since the text rendering code was originally with the title filter plugin (but since I moved that to the common headers, this is less of a concern). I think this is still up to discussion (and the current implementation is modular enough to easily move somewhere else). I like the idea of adding overlays from external plugin though, so I'd still hope to find a way that this can be taken into account by scale's layout.

You're right that there should be a way to have a way to display the titles without enabling the title filtering. I think it is a separate question how to present these options to the user (but there is a limitation that each plugin .so corresponds to a separate panel in WCM). In a way, showing the title logically belongs to the main properties of scale. Renaming the plugin from "title filter" to something more generic could also help.

soreau:
> > > What does not work is that scale does not take into account this extra space when positioning view, so the overlays can overlap with other views. I'm not sure how this can be fixed, since plugins that want to add overlays would need to inform scale of the size of the padding that will be added in advance. Of course, the easiest workaround is just to increase the spacing in the settings for scale.
> > 
> > 
> > I've been wondering why this functionality wasn't added into the scale plugin itself, with an option to enable it, since people might not want the scale title filter functionality but still want this. Wouldn't adding it into scale also make it easier to fix this issue?
> 
> Mainly to not make the scale plugin code even more bigger and complex, but also since the text rendering code was originally with the title filter plugin (but since I moved that to the common headers, this is less of a concern). I think this is still up to discussion (and the current implementation is modular enough to easily move somewhere else). I like the idea of adding overlays from external plugin though, so I'd still hope to find a way that this can be taken into account by scale's layout.

Quality should come first, above coding conveniences. The plugin can be separated into multiple source files if needed.

> 
> You're right that there should be a way to have a way to display the titles without enabling the title filtering. I think it is a separate question how to present these options to the user (but there is a limitation that each plugin .so corresponds to a separate panel in WCM).

This is not necessarily true. WCM creates a separate button for each .xml file, not each .so file. This means that you are able to have multiple .xml files for a single .so file. You can also use the options in a single .xml file from multiple plugin .cpp files. The thing that is currently bound is the enabling of the plugin via the toggle in WCM. If you create a new .xml file and there is no matching .so file, WCM will still show it. Enabling it will add it to the [core] plugins list. However core will basically ignore it. This is also how wf-shell settings work. Currently, the toggle buttons are not shown for certain 'plugins' in WCM based on hard coding which ones shouldn't have it.

I actually run into this situation recently where I am left wondering what to do when you might want multiple .xml files but only one shares the name of the .so. Perhaps we can simply add a flag to tell WCM not to add a toggle checkbox for the .xml containing the flag. This flag would be added to the list of .xml files WCM currently has hard coded to not show the checkbox.

> In a way, showing the title logically belongs to the main properties of scale. Renaming the plugin from "title filter" to something more generic could also help.

ammen99 (commented):
So far I have only tested it, and it kinda works, however, the title is not shown if I rapidly exit and start scale again.

ammen99:
> Mainly to not make the scale plugin code even more bigger and complex, but also since the text rendering code was originally with the title filter plugin (but since I moved that to the common headers, this is less of a concern). I think this is still up to discussion (and the current implementation is modular enough to easily move somewhere else). I like the idea of adding overlays from external plugin though, so I'd still hope to find a way that this can be taken into account by scale's layout.

@dkondor What exactly is it that you are trying to achieve? Do you just want to increase the spacing option? We could add some more hooks, maybe a `scale-compute-spacing` signal if needed.

> You're right that there should be a way to have a way to display the titles without enabling the title filtering. I think it is a separate question how to present these options to the user (but there is a limitation that each plugin .so corresponds to a separate panel in WCM). In a way, showing the title logically belongs to the main properties of scale. Renaming the plugin from "title filter" to something more generic could also help.

AFAIU the implementation of the show-titles and the actual filtering is more or less independent? We could make one `scale-titles` plugin and one `scale-title-filter` plugin.

> I actually run into this situation recently where I am left wondering what to do when you might want multiple .xml files but only one shares the name of the .so. Perhaps we can simply add a flag to tell WCM not to add a toggle checkbox for the .xml containing the flag. This flag would be added to the list of .xml files WCM currently has hard coded to not show the checkbox.
> 

Sure, more metadata can be added, although here (in our particular case) we're speaking of 2 (maybe 3) separate .so files, each with their own XML file?

dkondor:
> @dkondor What exactly is it that you are trying to achieve? Do you just want to increase the spacing option? We could add some more hooks, maybe a `scale-compute-spacing` signal if needed.

Yes, exactly. Actually, when I think of it, this could be handled as part of the `scale-filter` signal, since that is always emitted before doing the new layout. But let me know if you think it is cleaner to have a separate signal for it.

dkondor:
> > 
> 
> This is not necessarily true. WCM creates a separate button for each .xml file, not each .so file. This means that you are able to have multiple .xml files for a single .so file. You can also use the options in a single .xml file from multiple plugin .cpp files. The thing that is currently bound is the enabling of the plugin via the toggle in WCM. If you create a new .xml file and there is no matching .so file, WCM will still show it. Enabling it will add it to the [core] plugins list. However core will basically ignore it. This is also how wf-shell settings work. Currently, the toggle buttons are not shown for certain 'plugins' in WCM based on hard coding which ones shouldn't have it.
> 
> I actually run into this situation recently where I am left wondering what to do when you might want multiple .xml files but only one shares the name of the .so. Perhaps we can simply add a flag to tell WCM not to add a toggle checkbox for the .xml containing the flag. This flag would be added to the list of .xml files WCM currently has hard coded to not show the checkbox.
> 
> >

@soreau thanks for the explanation! I was actually thinking of the opposite: having multiple .so files share just one .xml (and one button in WCM), so that the options for showing the title overlay can be on the same page as the main scale options. But in this case, there is no way to enable / disable the .so file that does not share the name with the plugin defined in the .xml, right?

I think for now I'll just go with @ammen99 's suggestion of moving the title overlay to a separate plugin and this question can be revisited later. From a UI perspective, it is a tradeoff between having more buttons in WCM or more options on each page. I think I'm very not typical in that I like to play with settings, so I cannot really judge which is better for "normal" users :)

dkondor:
> So far I have only tested it, and it kinda works, however, the title is not shown if I rapidly exit and start scale again.

I see; I've added a fix for this.

dkondor:
A new issue I just realized is that if scale-title-filter is loaded before scale, it gives dynamic linking errors since I put some of the implementations of scale_transformer in scale.cpp. I'm thinking of what is the best way to fix this:
 - Just move everything to the header (simplest; also, it is not a lot of code, so it is not really an issue)
 - Move scale_transformer to core (and rename it to something like view_2D_overlay), since it is not specific to scale anymore
 - Bigger idea: define dependence relationships between the plugins. This would allow add-on plugins to depend on their main functionality and also compiling common utilities (like the cairo-util.hpp) to separate "plugins".

soreau:
> @soreau thanks for the explanation! I was actually thinking of the opposite: having multiple .so files share just one .xml (and one button in WCM), so that the options for showing the title overlay can be on the same page as the main scale options. But in this case, there is no way to enable / disable the .so file that does not share the name with the plugin defined in the .xml, right?

Yes this is correct. I think if you're sharing the .xml though, why have multiple .so files at all? I would put everything in the same plugin (maybe separate source files) and have one .xml with multiple `<group><\group>` tags to separate them in different tabs within the plugin in WCM.

dkondor:
> > @soreau thanks for the explanation! I was actually thinking of the opposite: having multiple .so files share just one .xml (and one button in WCM), so that the options for showing the title overlay can be on the same page as the main scale options. But in this case, there is no way to enable / disable the .so file that does not share the name with the plugin defined in the .xml, right?
> 
> Yes this is correct. I think if you're sharing the .xml though, why have multiple .so files at all? I would put everything in the same plugin (maybe separate source files) and have one .xml with multiple `<group><\group>` tags to separate them in different tabs within the plugin in WCM.

I see you point; I'll try how it works this way.

ammen99:
> * Just move everything to the header (simplest; also, it is not a lot of code, so it is not really an issue)

Sounds like a plan :)

> * Move scale_transformer to core (and rename it to something like view_2D_overlay), since it is not specific to scale anymore

I don't like this, it is quite specific to scale, even if it does not reference scale's structs directly.

> * Bigger idea: define dependence relationships between the plugins. This would allow add-on plugins to depend on their main functionality and also compiling common utilities (like the cairo-util.hpp) to separate "plugins".

So far, I have tried to avoid imposing order on plugin loading, and we don't have any dependency so far (except maybe glib-main-loop from w-p-e but that's something special).

> Yes this is correct. I think if you're sharing the .xml though, why have multiple .so files at all? I would put everything in the same plugin (maybe separate source files) and have one .xml with multiple `<group><\group>` tags to separate them in different tabs within the plugin in WCM.

Also sounds sensible, just try to keep the implementation separate (i.e if possible, don't call functions directly between the parts, and keep using signals as we currently do) - the idea being that in this way, we keep scale as simple as possible, and also allow creation of other scale-extension-plugins out-of-tree.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,296 @@
...
+    {
+        show_view_title_overlay = title_overlay_t::NEVER;
+        last_title_overlay = nullptr;
+        if (mouse_update_connected)
+        {
+            wf::get_core().disconnect_signal(&mouse_update);
+            mouse_update_connected = false;
+        }
+    }
+},
```
dkondor, line 138: I could not get uncrustify to be consistent with this formatting (all of these are part of the initializer list).

dkondor:
I've extended the `scale-filter` signal to include padding and moved the title overlays to the main plugin along with its settings. Let me know if there are any other issues.

ammen99:
> I've extended the `scale-filter` signal to include padding

I don't think that padding belongs to the same signal as filtering ...

plugins/scale/scale-title-overlay.cpp
ammen99: alright, I'll push a fix when the pr is ready for merging

dkondor:
> > I've extended the `scale-filter` signal to include padding
> 
> I don't think that padding belongs to the same signal as filtering ...

OK, I'm adding it as a separate signal instead

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,190 @@
...
+     * @param z_order Relative order; overlays are called in order.
+     * @param data    Any custom data to store. It is passed to the
+     *                callback function and is automatically destroyed
+     *                when the overlay is removed.
+     * @param pad     Any extra padding to use. Paddings are not cumulative,
+     *                i.e. if two overlays add a top padding, the larger
+     *                of the two values is used.
+     */
+    void add_overlay(scale_view_overlay *ol, int z_order,
+        std::unique_ptr<custom_data_t> data, const padding& pad = {0, 0, 0, 0})
```
ammen99, line 82: I think padding should be called separately by the caller.

plugins/scale/scale.cpp
```diff
@@ -1370,4 +1374,23 @@ class wayfire_scale : public wf::plugin_interface_t
     }
 };
 
-DECLARE_WAYFIRE_PLUGIN(wayfire_scale);
+class scale_combined : public wayfire_scale
```
ammen99, line 1377: A cool trick ;)

I think it should be fine to just call the constructor and destructor directly from scale though. There shouldn't be any other direct calls ofc.

plugins/scale/wayfire/plugins/scale-signal.hpp
```diff
@@ -73,4 +74,32 @@ void scale_filter_views(scale_filter_signal *signal, pred&& p)
...
+ *   plugins extending its functionality can add their overlays to it.
+ * argument: pointer to the newly added transformer
+ */
+struct scale_transformer_added_signal : public wf::signal_data_t
+{
+    wf::scale_transformer *transformer;
+};
+
+/**
+ * name: scale-padding
```
ammen99, line 90: Looking at it again, don't we already have the padding in the transformer? Why do we need to calculate it once more here?

plugins/scale/scale.cpp
```diff
@@ -878,26 +874,34 @@ class wayfire_scale : public wf::plugin_interface_t
...
-                (workarea.width - (cnt_cols + 1) * spacing) / cnt_cols;
+            const double scaled_width = std::max((double)
+                (workarea.width - (cnt_cols + 1) * spacing) / cnt_cols -
+                pad.left - pad.right, 1.0);
 
             for (size_t j = 0; j < cnt_cols; j++)
             {
-                double x = workarea.x + spacing + (spacing + scaled_width) * j;
-                double y = workarea.y + spacing + (spacing + scaled_height) * i;
+                double x = workarea.x + spacing +
```
ammen99, line 901: Maybe I am misunderstanding something, but shouldn't `pad_left/right` be added only once (not multiplied by `j`)?

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+#include <wayfire/opengl.hpp>
+#include <wayfire/plugins/common/cairo-util.hpp>
+#include <wayfire/plugins/common/simple-texture.hpp>
+
+class view_title_overlay : public wf::custom_data_t
+{
+  protected:
+    wf::cairo_text_t overlay;
+    wf::cairo_text_t::params par;
+    wayfire_view view;
```
ammen99, line 12: don't we have the view stored in `tr` already?

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+         * 3. The overlay previously did not fit, but there is more space now
+         * TODO: check if this wastes too high CPU power when views are being
+         * animated and maybe redraw less frequently or in a way that will not
+         * hold up the rendering loop!
+         */
+        if ((output_scale != par.output_scale) || (w > box.width * output_scale) ||
+            (overflow && (w < std::floor(box.width * output_scale))))
+        {
+            par.output_scale = output_scale;
+            par.max_size     = {box.width, box.height};
```
ammen99, line 67: isn't this basically the same as `render_overlay` ?

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+        CENTER,
+        BOTTOM,
+    };
+
+    position pos = position::CENTER;
+
+    GLuint get_texture(float output_scale, wf::geometry_t& geometry)
+    {
+        wlr_box box = get_transformed_bounding_box();
+        int w = overlay.tex.width;
```
ammen99, line 51: you could assign these after checking for regeneration of the overlay, then you won't have to re-assign the values.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+            view->damage();
+            w = overlay.tex.width;
+            h = overlay.tex.height;
+            overflow = res.width > w;
+        }
+
+        int y = 0;
+        switch (pos)
+        {
+          /* TODO: Add some padding! */
```
ammen99, line 78: add padding where?

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+        par.font_size  = font_size;
+        par.bg_color   = bg_color;
+        par.text_color = text_color;
+        par.exact_size = true;
+        render_overlay();
+        view->connect_signal("title-changed", &view_changed);
+    }
+
+    wf::signal_connection_t view_changed{[this] (auto)
```
ammen99, line 110: you can use `=` to assign directly, and avoid double `{}`

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+    view_filter{[this] (auto)
+    {
+        update_title_overlay_opt();
+    }},
+
+    scale_end{[this] (wf::signal_data_t*)
+    {
+        show_view_title_overlay = title_overlay_t::NEVER;
+        last_title_overlay = nullptr;
+        if (mouse_update_connected)
```
ammen99, line 133: you can just disconnect, it is a no-op if the signal is not connected (and in this case, no need for `mouse_update_connected`)

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+        show_view_title_overlay = title_overlay_t::MOUSE;
+    } else
+    {
+        show_view_title_overlay = title_overlay_t::NEVER;
+    }
+
+    if (show_view_title_overlay == title_overlay_t::MOUSE)
+    {
+        update_title_overlay_mouse();
+        if (!mouse_update_connected)
```
ammen99, line 274: you can just `mouse_update.disconnect()` and then connect.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+            mouse_update_connected = true;
+        }
+    }
+}
+
+void scale_show_title::update_title_overlay_mouse()
+{
+    wayfire_view v;
+
+    /* TODO: better way to determine if scale's grab interface is active! */
```
ammen99, line 287: There isn't a better way, unfortunately.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+        /* we can use normal focus tracking */
+        v = wf::get_core().get_cursor_focus_view();
+    } else
+    {
+        auto& core = wf::get_core();
+        v = core.get_view_at(core.get_cursor_position());
+    }
+
+    if (v)
```
ammen99, line 300: Hmm, if we have a dialog on top of a view, should we not display the title on the frontmost dialog (not on the main view)? The main view will not be visible for the most part.

ammen99 (changes requested):
See individual comments.

plugins/scale/scale-title-overlay.cpp
damianatorrpm: Modal dialogs e.g of dconf-editor commonly use empty titles.

ammen99: @damianatorrpm no, I think you missed my point. The current approach does this:

- render the base view
- render the base view's title
- render the dialog

The dialog is drawn over the title, so no title is visible at all...

dkondor: I haven't really thought of dialogs, but it is true that they will occlude the parent. One solution could be to show the overlay for both parent and dialog with the exception that we don't show a dialog's title if it is empty. This is mainly a problem if the overlay is centered -- if it is on top or bottom, it will still be visible.

plugins/common/wayfire/plugins/common/cairo-util.hpp
```diff
@@ -34,3 +35,263 @@ static void cairo_surface_upload_to_texture(
...
+     *
+     * @param text         text to render
+     * @param par          parameters for rendering
+     *
+     * @return The size needed to render in scaled coordinates. If this is larger
+     *   than the size of tex, it means the result was cropped (due to the constraint
+     *   given in par.max_size). If it is smaller, than the result is centered along
+     *   that dimension.
+     */
+    wf::dimensions_t render_text(const char *text, const params& par)
```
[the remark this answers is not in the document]
dkondor: No, I think string is fine

plugins/scale/scale-title-overlay.cpp
dkondor: Yes, this can be simplified

plugins/scale/scale-title-overlay.cpp
dkondor: Between the view and the title label (if it is above or below; in these cases, it is exactly touching it). I'm not sure if this is actually necessary, it does not look as bad as I expected.

plugins/scale/wayfire/plugins/scale-transform.hpp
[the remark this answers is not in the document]
dkondor: I'm putting the stored callback into an std::shared_ptr. This way, it can store a lambda created on the fly, but also a plugin can add the same callback object to multiple transformers (if it does not need to store extra data with it). If this is too complicated, I could just use a unique_ptr, requiring a new callback for each transformer.

plugins/scale/scale.cpp
dkondor: My assumption here is that each view has the same padding around it. This makes padding less flexible though.

plugins/scale/wayfire/plugins/scale-signal.hpp
dkondor: The main reason for this is that calculating the layout might happen before the transformers are initialized, so scale needs to know about padding before that.
But this means that we are supplying the padding essentially twice.
Now I'm leaning toward keeping the signal, and making it scale's responsibility to add padding to the transformer. Overlays are then not allowed to extend padding (at least without triggering an update to scale that will emit this signal again). I'll add a version with this implimented.

dkondor:
I've addressed most of the comments; I'm not sure what would be the best solution for dialogs and for padding (I've created a version where only the scale plugin adjusts the padding, based on the responses to the scale-padding signal). I've reorganized the commits and moved the old version here: https://github.com/dkondor/wayfire/tree/scale_show_view_title_old

soreau:
> I'm not sure what would be the best solution for dialogs and for padding (I've created a version where only the scale plugin adjusts the padding, based on the responses to the scale-padding signal).

Maybe child views shouldn't get a title overlay, the parent title should be enough?

plugins/scale/scale.cpp
ammen99: Wouldn't it be better if we could do this: for each view, have a fixed size available for view + padding (for example, width for each view is `workarea_width / views_in_the_row`), and then, the width of the view is the `max_width - spacing - padding`?

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,172 @@
...
+        {
+            (*ol->ol)(target_fb, damage);
+        }
+    }
+
+    /**
+     * Add a new overlay that is rendered after this transform.
+     *
+     * @param ol      The callback function that does the rendering.
+     * @param z_order Relative order; overlays are called in order.
```
ammen99, line 73: The docstring needs to be updated as well :)

ammen99:
> Maybe child views shouldn't get a title overlay, the parent title should be enough?

Sounds sensible to me - a single overlay on the topmost dialog (`enumerate_views().front()`) with the title of the main view.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+#include <wayfire/opengl.hpp>
+#include <wayfire/util/log.hpp>
+#include <wayfire/plugins/common/cairo-util.hpp>
+#include <wayfire/plugins/common/simple-texture.hpp>
+
+/**
+ * Class storing an overlay with a view's title, added to scale's transformer.
+ */
+class view_title_overlay
```
ammen99, line 11: Forgot it on the first review, this is also missing `_t` suffix.

plugins/scale/scale-title-overlay.cpp
ammen99: Alright. We can leave it for later if someone complains :D

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,172 @@
...
+class wayfire_scale;
+
+namespace wf
+{
+/**
+ * transformer used by scale -- it is an extension of the 2D transformer
+ * with the ability to add overlays
+ */
+
+class scale_transformer;
```
ammen99, line 18: also missing `_t`. Only signals don't have it.

plugins/scale/wayfire/plugins/scale-signal.hpp
ammen99: > The main reason for this is that calculating the layout might happen before the transformers are initialized, so scale needs to know about padding before that.

It all depends on how we implement it. We could compute the actual view size after adding the transformer (which is the moment overlays are added too).

dkondor:
https://github.com/WayfireWM/wayfire/pull/996#discussion_r552214215 @ammen99 you're right that scale can directly take into account the padding added to transformers and there is no need for a separate signal. I've made changes that work this way. I'm yet unsure what to do in the case when an overlay wants to update their padding (if this is needed at all).

dkondor:
> > Maybe child views shouldn't get a title overlay, the parent title should be enough?
> 
> Sounds sensible to me - a single overlay on the topmost dialog (`enumerate_views().front()`) with the title of the main view.

Sounds good, I'll see if I can implement this. Is there a guarantee that `enumerate_views().front()` gives the topmost dialog? (So that child views are ordered?)

ammen99:
> Sounds good, I'll see if I can implement this. Is there a guarantee that `enumerate_views().front()` gives the topmost dialog? (So that child views are ordered?)

Yes, the docstring is missing that description but the list is ordered.
https://github.com/WayfireWM/wayfire/blob/master/src/view/view.cpp#L157

dkondor:
I've added some modifications to handle dialogs better. I've decided to add a separate callback for calculating the required padding for this (this is required to properly adjust padding when the dialog's size is changed via a relayout).

What works:
 - If the overlay position is top or bottom, then only the parent view displays an overlay
 - If the overlay position is center, only the topmost dialog displays it, adding padding if the overlay is larger than the dialog

Issues:
 - There are sometimes visual glitches during the animation when a dialog is opened / closed while scale is running (most easily reproduced with gedit's "save before closing" dialog and using the middle click close option).
 - If a scaled dialog is vertically larger than the scaled parent, it will still be on top of the overlay. I'm not sure what would be a good way to adjust the position in this case, maybe the parent can calculate the combined bounding box of its dialogs. Or simply scale can restrict dialogs to be always strictly smaller than their parents.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -403,7 +403,7 @@ void scale_show_title_t::update_title_overlay_opt()
     {
         update_title_overlay_mouse();
         mouse_update.disconnect();
-        wf::get_core().connect_signal("pointer_motion_post", &mouse_update);
+        wf::get_core().connect_signal("pointer_motion_absolute_post", &mouse_update);
```
ammen99, line 406: absolute events happen only on Wayland/X11, so this will actually break DRM backend (which is much more important)

dkondor: Is it an issue to connect to both? (Although I agree that having this feature on X11 is probably not that important)

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,452 @@
...
+                              const wf::region_t& damage)
+        {
+            if (show_view_title_overlay == title_overlay_t::NEVER)
+            {
+                return;
+            }
+
+            if (show_view_title_overlay == title_overlay_t::MOUSE)
+            {
+                auto tmp_view = view;
```
ammen99, line 339: seems to me we should have a helper function `get_toplevel_view` which returns the topmost parent.

plugins/scale/scale-title-overlay.cpp
ammen99: You can connect to both, as long as the `mouse_update` handler does not care about its argument (because the `signal_data_t`s passed to the two signals have different types.

plugins/scale/scale-title-overlay.cpp
dkondor: Yes, that makes sense; I've copied the one from [view.cpp](https://github.com/WayfireWM/wayfire/blob/27a5a97bec803c70b175ae6d0f84a2925c5e329a/src/view/view.cpp#L73)

dkondor:
I'm converting this to a draft since I'm not satisfied with how dialogs are handled. I'm now thinking that using an overlay effect would be a better solution, but it could be best if it is integrated with scale better. I'll see if I can come up with something that works that way.

ammen99:
> I'm converting this to a draft since I'm not satisfied with how dialogs are handled. I'm now thinking that using an overlay effect would be a better solution, but it could be best if it is integrated with scale better. I'll see if I can come up with something that works that way.

I believe using visual tricks here would be the easiest way. We could have the titles fade-in very quickly once the scale animation is done, and then have them fade-out very quickly once scale starts exiting. It would require a pair of new signals from scale, but then you won't have to support overlays and what not in the transformer (so I think overall that solution could actually be simpler than what we have now).

dkondor:
> > I'm converting this to a draft since I'm not satisfied with how dialogs are handled. I'm now thinking that using an overlay effect would be a better solution, but it could be best if it is integrated with scale better. I'll see if I can come up with something that works that way.
> 
> I believe using visual tricks here would be the easiest way. We could have the titles fade-in very quickly once the scale animation is done, and then have them fade-out very quickly once scale starts exiting. It would require a pair of new signals from scale, but then you won't have to support overlays and what not in the transformer (so I think overall that solution could actually be simpler than what we have now).
> 
> EDIT: this is just one more idea I had, feel free to do something else if you have better ideas :)

The problem is that another plugin can add a transformer and run an animation and we need to adapt to that as well (e.g. if that results in the scaled size of the view changing.

I feel the better way to deal with that is to always run a pre-render hook that reacts to any changes in state + also make sure we detect any change in the view's transformed bounding box between the pre-render hook and the actual rendering.

I've put together an implementation that now works in all the corner cases that I could think of (including multiple dialogs, etc.). Let me know if you agree with this direction.

Also, since this turned out to be more complex than I originally expected, any more testing is much appreciated :)

@JordanL2 @damianatorrpm @soreau let me know if you have any more feedback on the implementation

JordanL2:
> @JordanL2 @damianatorrpm @soreau let me know if you have any more feedback on the implementation

I don't think I have the experience to provide any valuable insight on the implementation but thanks very much for doing this work :)

damianatorrpm:
@dkondor I really like it a lot and couldn't find any issues with it.

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+  protected:
+    /* extra info stored together with the overlay callbacks */
+    struct overlay
+    {
+        std::shared_ptr<scale_view_overlay> ol; /* callback function */
+        int z_order; /* order in which overalys are applied */
+        padding pad; /* extra padding needed by this overlay */
+        overlay(std::shared_ptr<scale_view_overlay>& ol_, int z_order_,
```
ammen99: you don't need such constructors, you could directly initialize with `{}`

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,181 @@
...
+            const padding& pad_) :
+            ol(ol_), z_order(z_order_), pad(pad_)
+        {}
+    };
+
+    std::list<std::unique_ptr<overlay>> overlays; /* list of active overlays */
```
ammen99: Seems to me that overlays are movable and copyable - do we need to keep them in `unique_ptr`s ?

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    void add_overlay(std::shared_ptr<scale_view_overlay>&& ol, int z_order,
+        const padding& pad = {0, 0, 0, 0})
+    {
+        add_overlay(ol, z_order, pad);
+    }
+
+    /* remove an existing overlay */
+    void rem_overlay(scale_view_overlay *ol)
```
ammen99: IMO this is inconsistent, if we pass shared ptrs for adding, we should use shared ptrs for removing too. (or unique_ptr + observer_ptrs, but maybe shared_ptrs are easier here)

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,309 @@
...
+    /* if text potentially overflows the available space */
+    bool overflow = false;
+
+    /**
+     * Get the transformed bounding box of the view, including the
+     * current transform
+     */
+    wlr_box get_transformed_bounding_box()
+    {
+        wlr_box box = tr.get_transformed_view()->get_wm_geometry();
```
ammen99: You should use the `get_bounding_box(nonstd::observer_ptr<transformer>)` functions, they will give you the bounding box up to the transformer, and then you can apply the last step manually. `wm_geometry` is rarely a good option since you can have scaling, etc.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,309 @@
...
+    void render_overlay(const wlr_box& box)
+    {
+        par.max_size = {box.width, box.height};
+        auto view = tr.get_transformed_view();
+        auto res  = overlay.render_text(view->get_title(), par);
+        overflow = res.width > overlay.tex.width;
+        view->damage();
+    }
+
+    void render_overlay()
```
ammen99: nitpick: a better name would be `update_overlay_texture` or similar. `render_overlay` sounds like you are actually rendering it on the screen.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -0,0 +1,309 @@
...
+{
+    this->output = output;
+    output->connect_signal("scale-filter", &view_filter);
+    output->connect_signal("scale-transformer-added", &add_title_overlay);
+    output->connect_signal("scale-end", &scale_end);
+}
+
+void scale_show_title_t::fini()
+{
+    output->disconnect_signal(&view_filter);
```
ammen99, line 477: no need for manually disconnecting actually, because the signal_connection_t objects will do this automatically.

plugins/scale/scale-title-overlay.cpp
```diff
@@ -19,28 +47,90 @@ class view_title_overlay_t
...
+            return get_transformed_bounding_box();
+        }
+
+        view = find_toplevel_parent(view);
+
+        auto tmp =
+            view->get_transformer(wf::scale_transformer_t::transformer_name());
+        auto tr = dynamic_cast<wf::scale_transformer_t*>(tmp.get());
+        if (!tr)
+        {
```
ammen99, line 174: you could just assert(false), since this should never happen, if it does, at least we'll be able to detect such mistakes, otherwise, we might spent hours trying to figure out what is actually going on :)

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -90,75 +100,103 @@ class scale_transformer_t : public wf::view_2D
...
-                [z_order] (const auto& other)
+        bool view_pad_shrunk  = false;
+        bool view_pad_grown   = false;
+        bool scale_pad_shrunk = false;
+        for (auto& p : overlays)
         {
-            return other->z_order >= z_order;
-        });
+            auto& ol = *(p.second);
+            if (ol.pre_hook)
```
ammen99, line 137: Can't we just start with empty pad, extend padding by going through all hooks, then compare with last padding? Seems like it would be simpler and shorter.

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -90,75 +100,103 @@ class scale_transformer_t : public wf::view_2D
...
+     * Add a new overlay that is rendered after this transform.
+     *
+     * @param ol      The overlay object to be added.
+     * @param z_order Relative order; overlays are called in order.
+     */
+    void add_overlay(std::shared_ptr<overlay> ol, int z_order)
     {
-        add_overlay(ol, z_order, pad);
-    }
+        if (!ol)
```
ammen99, line 183: This should not be here. Callers need to ensure they provide sensible values. I'd assert() if anything.

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -90,75 +100,103 @@ class scale_transformer_t : public wf::view_2D
...
-        std::shared_ptr<padding_callback_t>&& pad = std::shared_ptr<padding_callback_t>())
-    {
-        add_overlay(ol, z_order, pad);
+        auto it = std::find_if(overlays.begin(), overlays.end(),
+            [z_order] (const auto& other)
+        {
+            return other.first >= z_order;
+        });
+
+        view_padding.extend(ol->view_padding);
```
ammen99, line 194: won't padding be updated by the prehook anyway?

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -232,6 +243,21 @@ class scale_transformer_t : public wf::view_2D
      */
     wlr_box get_bounding_box(wf::geometry_t view, wlr_box region) override
     {
+        if (view != last_view_box)
```
ammen99, line 246: Why is this needed? If another transformer has changed, it should call damage() (before rendering). When it calls damage(), this method will be called, so I don't think you need an idle call and special handling here (even if the pre hooks indicate a size change).

ammen99 (changes requested):
Starting to look good, I hope you won't have to reiterate on the design anymore :)

A few small requests, and then we can merge.

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: I was unsure and I switched to using shared_ptr everywhere. Now I feel it should be unique_ptr, since one overlay will go together with one view transformer. If the caller wants to share state / resources between overlays, they can do this by inheriting from the overlay struct and adding shared_ptr members.
Also, I'm not sure if the callbacks should be std::functions, or it might be better to use virtual functions (if users of overlays will typically inherit anyway).

ammen99: I see, I have missed this reorganization - I was reviewing commit-by-commit, so this particular comment is outdated.

In any case, shared_ptr is not semantically correct here. It implies shared ownership, however, an overlay should never remain in the transformer if the entity which added it is destroyed. So a unique_ptr/observer_ptr combo sounds better.
Plus, if you push observer-ptrs, you can make several such pointers from the same unique_ptr, thus sharing overlays between transformers, if you wanted to.

Another note: please rename `overlay` and `padding` with `_t` versions.

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: It would, but this way, there is no need to call `damage()` twice; also, this will work if this call happens between our pre-hooks and rendering (I'm not sure why that would happen, but it would be completely OK to do that, e.g. from another plugin's pre-hook)

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: I see, so there is no need to call damage here, just update our padding

plugins/scale/scale-title-overlay.cpp
dkondor: I'm not completely sure, since this is called from [here](https://github.com/dkondor/wayfire/blob/9f78ba1c08f9706ff0d04ed66db18f29e01893b1/plugins/scale/scale-title-overlay.cpp#L209) and I wonder if that line can be reached before the scale plugin added its transformer to a newly created dialog. Still, the test could be `if(!tmp)` and the `dynamic_cast` failing can be an error.

plugins/scale/scale-title-overlay.cpp
dkondor: The problem is that `get_bounding_box()` includes shadows. See the attached screenshot where the gedit window was maximized, so no shadows, while the other ones were not, so there is an extra gap below the overlay. This is why I found it is better to use `get_wm_geometry()` and transform it ourselves: the `trasform_box_without_padding()` function that calls `transform_region(this)` to do the transform up to our transformer.

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: I'm leaning toward `unique_ptr` -- this would allow to create, add and "forget" an overlay (and anyway, in most cases, there needs to be some extra data that is unique to every view).

plugins/scale/scale-title-overlay.cpp
dkondor: scale_overlay_padding

dkondor:
I've addressed these issues (see my individual comments) -- let me know if there's more

plugins/scale/wayfire/plugins/scale-transform.hpp
ammen99: Yeah, this is what we do for ex. for view transformers and custom data. There are some pitfalls though - you still need to ensure all of these are removed when the plugin is unloading, otherwise, it could happen that the .so file is unloaded (and thus the destructor cannot run anymore).

plugins/scale/scale-title-overlay.cpp
ammen99: I see, makes sense. Can we then rename the function to something that indicates it is not the full bounding box?

plugins/scale/scale-title-overlay.cpp
ammen99: Oh hm, this is a very good point. In this case, could you add a comment about this? i.e how we can reach that point.

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,282 @@
...
+     * @param damage Damaged region to render.
+     */
+    using render_hook_t = std::function<void (const wf::framebuffer_t& fb,
+        const wf::region_t& damage)>;
+
+    /**
+     * Overlays that can be added to this transformer. Hooks are called
+     * similarly to render-manager.
+     */
+    struct overlay_t : public wf::custom_data_t
```
ammen99, line 91: why does this need to derive from custom_data_t ?

plugins/scale/wayfire/plugins/scale-transform.hpp
ammen99: ok, sounds sensible.

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,282 @@
...
+        {
+            auto& ol = *(p.second);
+            if (ol.render_hook)
+            {
+                ol.render_hook(target_fb, damage);
+            }
+        }
+    }
+
+    /* call pre-render hooks; the parameter indicates if it is possible to
```
ammen99, line 126: I find this a bit confusing. It both damages and returns whether damaging is necessary? Looking at the current usages, it seems we could directly damage (if damaging is possible at all).

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: That's true. For now, the title-overlay is compiled in the same library as scale, so this will not be a problem, but for other uses, this means there is a need to keep track of overlays added (but I'm not sure if using `observer_ptr` or `shared_ptr` would solve it).

plugins/scale/scale-title-overlay.cpp
dkondor: Ok, sure. I'm adding comments here. I'm not entirely sure if this will happen though, but I'm not that familiar with the sequence of events when views are created.

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: You're right, it can have its own virtual destructor; I'll change it.

plugins/scale/wayfire/plugins/scale-transform.hpp
dkondor: This is a small optimization to avoid calling `damage()` twice, since this function is called from [here](https://github.com/dkondor/wayfire/blob/8d68476114371820ccb58a301bc70079476e7d23/plugins/scale/scale.cpp#L694) where damage might be necessary anyway.

I could either remove this if it is not worth the code complexity (since the runtime saving will be small), or use an alternate implementation with a second bool parameter that is set to true if the caller needs to damage the view. This would allow to only damage once even if the overlays' total size has shrunk.

dkondor: I've changed the logic here, and made the version with the `can_damage` parameter protected since it is only used when this is called from `get_bounding_box()`

plugins/scale/wayfire/plugins/scale-transform.hpp
```diff
@@ -0,0 +1,296 @@
...
+            left   = std::max(left, other.left);
+            bottom = std::max(bottom, other.bottom);
+            right  = std::max(right, other.right);
+        }
+    };
+
+    uint32_t get_z_order() override
+    {
+        /* ensure that scale comes after fire */
+        return wf::TRANSFORMER_HIGHLEVEL + 2;
```
ammen99, line 58: Unfortunately, this can't be. Scale needs to come before wobbly so that dragging a view works, see #1020 

Also, why should scale come after fire? Makes no sense.

ammen99 (approved):
Thanks for working on this and thanks for the patience during the reviews!

Looks good to me, and I also tested it, only the Z order is wrong. Once you update it, please also squash your commits together, maybe like this:

- First commit, extracts the text helper
- Other changes to scale-title-filter
- Add support to scale for overlays
- Add title overlay

plugins/scale/wayfire/plugins/scale-transform.hpp
ammen99: So, this should be highlevel - 10 or something like that.

dkondor: Ok, that's reasonable. The motivation for putting this after fire was that if an overlay adds padding, then fire will cover the whole (now padded) area -- this can look strange (e.g. if a wide overlay is over a small dialog). But this is a general problem with fire (e.g. fire animation with a 45-degree rotated view will also look strange), so this can be a separate issue.

dkondor:
OK, I've made the remaining changes; let me know if there is something else

PR #998, opened 2021-01-03
grid crossfade

### LINKED ISSUE #991: Better animation when windows are maximizing and go back to the original size
Hi, I was wondering if Wayfire can do a better animation when windows are maximizing (or full screen) and go back to the original size.
If I compare with Gnome the animation look smoother. 

See example here with YouTube.
https://streamable.com/bfoh2g

Thanks,

ammen99:
I have started to experiment a bit with a crossfade animation instead of the default move+resize for grid.
There are still a few things left that should be fixed, but I think it should already be much smoother than the current animation.

~~Opening as a draft because it seems easy to get Grid into a state where it flips between fullscreen and unfullscreen each frame, making the window unusable. Will investigate further on that.~~ Seems a bug in GTK3, it happens on gnome too.

Note if someone is testing this: it requires the dynamic-list branch of wf-config, otherwise, it won't compile. This is necessary for a small addition I made there.

Fixes #991

### DIFF
--- a/metadata/grid.xml
+++ b/metadata/grid.xml
@@ -12,14 +12,14 @@
 		<option name="type" type="string">
 			<_short>Type</_short>
 			<_long>Sets the type of the animation.</_long>
-			<default>simple</default>
+			<default>crossfade</default>
 			<desc>
 				<value>none</value>
 				<_name>None</_name>
 			</desc>
 			<desc>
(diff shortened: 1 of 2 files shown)


damianatorrpm:
Heavily improves behaviour (especially for chrome the white space during resizing is much much less),

PR #1010, opened 2021-01-06
signal-definitions: use correct name for pointer_motion_absolute

dkondor:
It is currently not consistent with what is used in cursor.cpp


ammen99 (approved):
Nice catch, thanks!

PR #1011, opened 2021-01-06
scale-title-filter: fix damage handling on scaled output

### LINKED ISSUE #1008: problem with overlay damage when output scale > 1
**Describe the bug**
Overlays (`OUTPUT_EFFECT_OVERLAY` added to the `render_manager` are sometimes not updated when the screen is damaged if the output scale is > 1. This results in the overlay being (partially) invisible. I've only tested it when running nested (both X11 and Wayland backend). Not tested without nesting.

I'm not sure if this is a bug, or I used the overlay render function in a wrong way when implementing this (e.g. [here](https://github.com/WayfireWM/wayfire/blob/1247f9df045bb0386fbbb74a153619d913f955e0/plugins/scale/scale-title-filter.cpp#L343). Everything works fine if output scale is set to 1.

**To Reproduce**
1. Enable scale-title-filter with overlays shown (`overlay = true`, the default)
2. Set output scale to > 1 (e.g. `scale = 2.000000` in `[output:X11-1]`)
3. Start scale
4. Type a character that does not change the layout (present in all views' title)
5. Overlay is not shown until multiple characters are typed and the layout changes. When deleting characters, the overlay is "erased" when the layout return to showing all views (note that the filter is still present, typing again restores it).

**Expected behavior**
Overlay is shown.

**Screenshots or stacktrace**
Sometimes on X11, the overlay is partially shown. Also, on X11, moving the mouse over the overlay erases parts of it even when it was shown normally (shown in the attached screenshot).
overlay-issue

**Wayfire version**
git master (commit 802927a )

Also happens on #996

dkondor:
Fixes #1008 -- seems that I was using the wrong function to get the damage, since `get_swap_damage()` returns scaled regions

### DIFF
--- a/plugins/scale/scale-title-filter.cpp
+++ b/plugins/scale/scale-title-filter.cpp
@@ -340,7 +340,7 @@ class scale_title_filter : public wf::plugin_interface_t
             (int)(tex.width / output_scale),
             (int)(tex.height / output_scale)
         };
-        auto damage = output->render->get_swap_damage() & geometry;
+        auto damage = output->render->get_scheduled_damage() & geometry;
         auto ortho  = out_fb.get_orthographic_projection();
 
         OpenGL::render_begin(out_fb);


ammen99 (approved):
Nice catch!

PR #1014, opened 2021-01-07
xwayland: add special implementation for Xwayland DnD icons

ammen99:
With this PR, a new kind of Xwayland views are added - drag and drop icons. They are special-cased just like regular DnD icons are.


damianatorrpm:
I think it looks better than before (before you would drag text with a transparent background).
It's an inprovement, even thought the somewhat expectable configuration dnd btw. chrome and wayland doesn't work.

Anything further would be just an incremental improvement, but I already like this looks better

PR #1019, opened 2021-01-10
Scale decor

dkondor:
This is a proof-of-concept for adding decorations to scaled views (#897), based on #996. Currently, the "decoration" is just a red rectangle (that closes the view on click), so this clearly needs improvement :) I don't have much experience in UI design, so I'm happy if someone takes this over; I'm mainly opening this PR to share this idea.

Apart from that, it has one major issue: if there is a view behind the "decoration" (this is the case for dialogs where the main view is typically behind the dialog), the main view will receive the click as normal, and e.g. in non-interactive scale, this will result in the view selected and scale ending. I'm not sure if this can be solved in an easy way without hacks such as replacing the button in any mouse event processed by this plugin.

Also, this is a separate plugin now, but could be easily integrated into scale, similarly to how it is done for the title overlay in #996 

~~Note: only the last commit is actually this implementation; the rest are the same as in #996~~ Rebased on master since #996 was merged

Other possibilities for #897:
 - use a subsurface (similarly to the decor plugin); the main issue with this is that I believe it's best not to scale these extra decors
 - use a compositor_surface_t: this would solve the problem with handling clicks, but makes it more difficult to attach these surfaces to the scaled views, especially maintaining a consistent z-order during animations (though this could be solved by hiding during animations)


ammen99:
Here is what I think we need as a first step:

- A rectangle to highlight the currently hovered view (a border around the transformed wm geometry).
- configurable color for it :)
- optionally change the highlight color if a button is held pressed

For a close button, I can see two possible ways we can do it:

- The first one is more touch-friendly, have a special area the user can drag and drop a view in. This should be quite easy to implement now, since scale already has drag-and-drop support.

- The second one is more mouse-friendly, the same way you have started doing it. Draw a small X icon in the corner. It should be visible for all views (not just the hovered one) to make this solution viable for touchscreens as well.

I think the second option might be more desirable since our primary target is not smartphones :) But I won't have anything against 1) too since you can already middle-click-to-close a view.


That being said, we could just have the highlight-on-hover as a first step.

ammen99:
Closing this PR because it has gotten pretty stale. With 'modern' Wayfire the solution here would be to add a custom node as a sibling (not as a child) of the scale transformer, thus making sure that it is not scaled by the transformer.

PR #1020, opened 2021-01-10
refactor view drag move

### LINKED ISSUE #527: after moving sometimes input is lost
When you move a view quickly (also between multiple outputs) sometimes at the end of the move operation input is permanently lost.

It seems removing the commented lines in move.cpp below fixes the issue:
```
    void fini() override
    {
        // if (grab_interface->is_grabbed())
        // input_pressed(WLR_BUTTON_RELEASED, false);
```

### LINKED ISSUE #534: [move-plugin] join-view create only 1 mirror-view
in the move plugin, if you join views, only the parent view get's a mirror view

### LINKED ISSUE #582: Dialog placement with multilple workspaces

### LINKED ISSUE #736: [bug] Extra wobble when moving workspaces on already wobbling windows
Scenario - 
 - Maximize a small window
 - It maximizes and is still wobbling a bit
 - Move to another workspace - `Ctrl + Win + <left-arrow>`
 - Original window has a LARGE wobble.

Expectations, moving workspaces and wobbles should be independent

Video - https://streamable.com/vg7kb7

### LINKED ISSUE #830: [expo] shows views that aren't there
So I hope I explain this right, I'd make a video but it's related to multi monitor.
Let's take 2 monitors: MLeft and MRight
on MLeft start expo and take a view to drag it onto MRight.
while you have released the mouse on MRight you can still see the view on the second ws on MLeft

ammen99:
This PR basically reimplements move and Expo's view movement functionality. In addition, the scale plugin now supports dragging its views.

Because of the massive changes I expect a few things to be broken here and there. So far, I have tried testing Expo, Scale and Move on two outputs (the two outputs could have a different plugin active) and dragging and dropping views between them, with and without wobbly enabled. The dragged views themselves can be normal, tiled or fullscreen. I'd be grateful if more people can test this.

A list of fixed bugs:

- Fixes #534
- Fixes #830 (haven't really tested but it should be fixed).
- Maybe fixes #527 (@damianatorrpm please test)
- Fixes #736
- Fixes #582
- Fixes the bug where moving a view to another output with expo shows the mirror-view at incorrect places.

New feature:

- You can drag views from Scale to other outputs (or drop them on the same output, which should be no-op).

Note for testers: the PR still contains a lot of debugging output, I'll remove it if there are no bugs (otherwise, it will still be useful for debugging).


damianatorrpm:
Lovely :)
Biggest improvement for a while imho.
Moving views between different scales on different output's is smooth.

Issues:
1) If you drag a view in scal a view militmeters without exceeding the snap off threshould 
the mouse cursor stays with the open hand cursor. 
2) closing a view in scale while the open hand cursor is shown results in wayfire crashing,
3) Dragging a view from output A (scale) to output B (expo) is not smooth,
the view stops moving when arrived at output B, need to release mouse cursor and drag again.
4) Expo misplaces a window slightly if you throw it (maybe due to wobbly settings). 
e.g drag it with high speed on expo and release mouse when entering one workspace in expo.

ammen99:
> * 
> * closing a view in scale while the open hand cursor is shown results in wayfire crashing,

How do you close it? Or it closes itself?

> * 
> * Expo misplaces a window slightly if you throw it (maybe due to wobbly settings).
>   e.g drag it with high speed on expo and release mouse when entering one workspace in expo.

Yes, this is a general wobbly "feature", if you throw it like this it will continue moving.


I think I have fixed the other two issues, check the last two commits.

damianatorrpm:
>How do you close it? Or it closes itself?
Middle click

damianatorrpm:
>Yes, this is a general wobbly "feature", if you throw it like this it will continue moving.

A thing with this wobbly "feature" (not related to this PR)
is that depending on speed and velocity you can throw it out of all visible workspaces (if the throwing happens between outputs).

But this is an issue that was always there

ammen99:
> A thing with this wobbly "feature" (not related to this PR)
> is that depending on speed and velocity you can throw it out of all visible workspaces (if the throwing happens between outputs).

I think this particular problem might be fixed in a separate PR, let's not overload this one (it already has a lot of changes).

damianatorrpm:
middle click during drag operation (scale) that closes the view crashes wayfire

damianatorrpm:
can no longer reproduce, somehow all issues seem fixed

damianatorrpm:
found another thing:
during move now views can cover "top/overlay" layer-shells

ammen99:
> found another thing:
> during move now views can cover "top/overlay" layer-shells

That's a feature: the dragged view is basically a drag icon which is on top of everything. It also simplifies the implementation greatly.

myfreeweb:
> the dragged view is basically a drag icon which is on top of everything

Hopefully there won't be a "dragged window still visible when lockscreen appears" bug :D
I guess `zwlr_input_inhibitor_v1` should cancel all dragging?

ammen99:
> > 
> 
> Hopefully there won't be a "dragged window still visible when lockscreen appears" bug :D
> I guess `zwlr_input_inhibitor_v1` should cancel all dragging?

Yes, you're right. I'll add handling for this. Basically: one the lockscreen starts, all plugins should cancel their action. And once the fullscreen is active, you shouldn't be able to activate any of the window dragging plugins, so whether the dragged view is an overlay or not is not so important :)

Issue #960, opened 2020-12-18
wayfire crashes on csd(close) clicked


damianatorrpm:
Wayfire reproducibly crashes when the close button on the csd of app (com.github.jms55.Sandbox) is clicked.
```
EE 18-12-20 08:29:47.223 - [src/main.cpp:164] Fatal error: Segmentation fault
EE 18-12-20 08:29:47.291 - #1  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > wf::log::detail::format_concat<char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) ??:?
EE 18-12-20 08:29:47.342 - #2  __restore_rt sigaction.c:?
EE 18-12-20 08:29:47.369 - #3  __dynamic_cast ??:?
EE 18-12-20 08:29:47.401 - #4  wf::pointer_t::send_button(wlr_event_pointer_button*, bool) ??:?
EE 18-12-20 08:29:47.433 - #5  wf::pointer_t::grab_surface(wf::surface_interface_t*) ??:?
EE 18-12-20 08:29:47.466 - #6  wf::pointer_t::handle_pointer_button(wlr_event_pointer_button*) ??:?
EE 18-12-20 08:29:47.501 - #7  register_tm_clones ??:?
EE 18-12-20 08:29:47.637 - #8  wlr_signal_emit_safe ../util/signal.c:22
EE 18-12-20 08:29:47.764 - #9  wlr_signal_emit_safe ../util/signal.c:22
EE 18-12-20 08:29:47.887 - #10 handle_pointer_button ../backend/libinput/pointer.c:89
EE 18-12-20 08:29:47.906 - #11 wl_event_loop_dispatch ??:?
EE 18-12-20 08:29:47.924 - #12 wl_display_run ??:?
EE 18-12-20 08:29:47.958 - #13 main ??:?
EE 18-12-20 08:29:48.009 - #14 __libc_start_main ??:?
EE 18-12-20 08:29:48.044 - #15 _sub_I_65535_0.0 ??:?
```

damianatorrpm:
@ammen99 The creash happens exactly here in the dynamic_cast of compositor surface from surface, don't know how to fix that though:
```
diff -ur wayfire/src/core/seat/pointer.cpp wayfire-novos/src/core/seat/pointer.cpp
[...]

PR #1025, opened 2021-01-11
Fix crash in pointer.cpp - check relation between surface and grab surface

damianatorrpm:
Fixes:
https://github.com/WayfireWM/wayfire/issues/960


damianatorrpm:
I hope this is the correct way to fix this, maybe additionally check if grabbed surface is really a subsurface?
Not sure about the side effects of this thought (what if surface is another sibling of a surface, where grab_surface is a sibling as well?)

damianatorrpm:
The updated PR works for me, I couldn't detect any negative side effects.
I could imagine that some application may loose active grab when a subsurface is destroyed?

ammen99:
Thanks for all the info provided, closing this in favor of #1028 because the real bug was subsurface destruction order.

PR #1031, opened 2021-01-13
Invert 3d transformer

dkondor:
While working on #996 I've noticed [this TODO](https://github.com/WayfireWM/wayfire/blob/faab309a9122b758f5edd164a7bc2748ce3e4f52/src/view/view-3d.cpp#L229) and realized that I can provide a solution. This is possible since the "3D" transform in this case is actually mapping a 2D object to a 2D canvas, so it is not losing information (except in the case when the view is turned perpendicular to the screen).

The first commit is actually the inverse of the transform, the second commit adds a simple plugin to demonstrate that it is working -- I'm actually not sure if this plugin should go in the main plugin collection (maybe it could be merged with the 2D "wrot" plugin?).

### DIFF
--- a/plugins/single_plugins/wrot.cpp
+++ b/plugins/single_plugins/wrot.cpp
@@ -3,7 +3,13 @@
 #include "wayfire/view-transform.hpp"
 #include "wayfire/output.hpp"
 #include "wayfire/core.hpp"
+#include <wayfire/workspace-manager.hpp>
 #include <linux/input.h>
+
+#include <glm/gtc/matrix_transform.hpp>
+
+static const char *transformer_3d = "wrot-3d";
+static const char *transformer_2d = "wrot-2d";
 
 static double cross(double x1, double y1, double x2, double y2) // cross product
 {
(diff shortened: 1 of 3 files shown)


ammen99:
Thanks for working on this! The new plugin allows for some very cool effects. Indeed, I think we can merge the two of them. Maybe we could have two bindings, one for 2D rotation and one for 3D.

I am testing this and it seems that input transformation doesn't work quite right. Consider the following screenshot:

slurped

I cannot click on the title bar of chrome. Now, it is quite small, I agree, but the clicks are actually registered some 100-200 pixels "down", which to me indicates some problem in the formula. It works quite well for not-so-extremely transformed views. Maybe it is the numerical instability you have mentioned? Or maybe the formulas are missing something (can we really ignore the depth component)?

ammen99:
Also, might it be a good idea to use doubles for computing the inverse, would this help with rounding errors? Most other computations involving input are done in doubles.

dkondor:
> Thanks for working on this! The new plugin allows for some very cool effects. Indeed, I think we can merge the two of them. Maybe we could have two bindings, one for 2D rotation and one for 3D.

Sure, I'll add this.

> I am testing this and it seems that input transformation doesn't work quite right. Consider the following screenshot:
> 
> slurped
> 
> I cannot click on the title bar of chrome. Now, it is quite small, I agree, but the clicks are actually registered some 100-200 pixels "down", which to me indicates some problem in the formula. It works quite well for not-so-extremely transformed views. Maybe it is the numerical instability you have mentioned? Or maybe the formulas are missing something (can we really ignore the depth component)?

This is strange; in my testing, everything seems reasonably accurate, there might be 1-2 pixel differences at maximum. I wonder if I'm right using the view's `get_wm_geometry()`, or are there some cases when the full geometry should be used. I've added a version which uses double, let me know if it seems to improve things; I'll see if I can make a version that uses a more direct formula to calculate the results. Note that so far I've only tested this running nested on X11, but I cannot imagine if that would matter.

Depth is taken into account when writing the equations to solve (the `tr[3][0]` and `tr[3][1]` components are used that relate the original x and y coordinates to the w component of the transformed vector).

ammen99:
@dkondor I found out the problem. The window I was testing with was my testing Xwayland browser, and Xwayland does not support negative coordinates. The top portion of the window however was actually off-screen without this transform, and thus, it could not get input at all (but that is a problem in Xwayland, not in your impl).

I tested with wayland-native apps and it works fine. The 3D transformer does not operate on the view geometry, however. It uses the bounding box up to that transformer. The current code is actually wrong as well, here is a small fix http://ix.io/2M02 (you could add it to this PR).

So, what I think should be done before merging this PR is combining this and the wrot plugin into a single plugin, maybe with two different buttons for activation. It would be the equivalent of "Freely transformable windows" from compiz :)

src/view/view-3d.cpp
```diff
@@ -223,15 +223,43 @@ wf::pointf_t wf::view_3D::transform_point(
...
     wf::pointf_t point)
 {
-    return {wf::compositor_core_t::invalid_coordinate,
-        wf::compositor_core_t::invalid_coordinate};
+    auto p  = get_center_relative_coords(view->get_wm_geometry(), point);
+    auto tr = calculate_total_transform();
+
+    /* Since we know that our original z coordinates were zero, we can write a
+     * system of linear equations for the original (x,y) coordinates by writing
+     * out the (x,y,w) components of the transformed coordinate.
```
ericonr, line 237: (x,y,z) ?

dkondor: No, in the forward transform, the z coordinate is discarded and w is used for the perspective projection, see [here](https://github.com/WayfireWM/wayfire/blob/d87aba715ff73a45202e3368c0fcde121c220ef0/src/view/view-3d.cpp#L222) -- so we have to use the expression for w here as well.

dkondor:
> @dkondor I found out the problem. The window I was testing with was my testing Xwayland browser, and Xwayland does not support negative coordinates. The top portion of the window however was actually off-screen without this transform, and thus, it could not get input at all (but that is a problem in Xwayland, not in your impl).
> 
> I tested with wayland-native apps and it works fine. The 3D transformer does not operate on the view geometry, however. It uses the bounding box up to that transformer. The current code is actually wrong as well, here is a small fix http://ix.io/2M02 (you could add it to this PR).
> 
> So, what I think should be done before merging this PR is combining this and the wrot plugin into a single plugin, maybe with two different buttons for activation. It would be the equivalent of "Freely transformable windows" from compiz :)

Ok, I've merged the two plugins (with two button activators) and added your patch regarding the geometry; let me know if you see any other issues.

plugins/single_plugins/wrot.cpp
```diff
@@ -15,14 +22,159 @@ static double vlen(double x1, double y1) // length of vector centered at the ori
...
 
     wf::pointf_t last_position;
     wayfire_view current_view;
 
+    mode current_mode;
+
+    void reset_all()
+    {
+        for (auto v : transformed_views)
+        {
```
ammen99, line 49: I have an idea. Instead of storing the views in a separate std::set and having to check whether the view is detached, why not just iterate over all views on the output and reset them?

dkondor: That makes sense, I've changed the logic to this.

ammen99 (approved):
LGTM, Thanks! Can you squash the commits, so that we have one which changes core, and the other, which changes wrot?

PR #1036, opened 2021-01-14
disable event handling

### LINKED ISSUE #640: Don't pass the click event if wayfire processed the event
When I click a window to raise it, and if there is a clickable button in that place,
then the window is raised and the button is clicked.

I don't hope that.

If wayfire raised the window by the click event, then don't pass the event to applications.

### LINKED ISSUE #997: Send button release when input is grabbed by a plugin
Don't know it this is possible, the problem is this:
1) start selecting using rubberband (the blue selection rectangle) in e.g Nautilus file
2) without releasing the mouse start scale or expo and release the mouse
3) stop scale/expo
4) nautilus still has mouse grab on the rubberband (even though the mouse is not clicked.

### LINKED ISSUE #1000: event->handled field in keyboard_key
Allow plugins reacting to events like keyboard_key to tell core that the event was handled and that no further processing should be done

Use case: custom keybinding handling (plugin handles keybinding -> triggers an action, then tells core not to send the keypress that triggered the event to clients/etc)

ammen99:
- api: add handled field for input events
- focus: add option to not pass buttons to focused surface
- pointer: keep track of pressed buttons and send release on ungrab

Fixes #640
Fixes #1000
Fixes #997

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -63,5 +63,10 @@
 			<_long>Clicking on a view with any of the mouse buttons of these will focus it.</_long>
 			<default>BTN_LEFT | BTN_MIDDLE | BTN_RIGHT</default>
 		</option>
+		<option name="focus_buttons_passthrough" type="bool">
+			<_short>Pass through buttons to client</_short>
+			<_long>Whether to pass buttons through to the client when switching the focus via clicking</_long>
+			<default>true</default>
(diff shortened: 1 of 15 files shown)


damianatorrpm:
Definitely fixes https://github.com/WayfireWM/wayfire/issues/997

PR #1045, opened 2021-01-17
Scale filter multi outputs

### LINKED ISSUE #1021: scale-title-filter and multi-output
**Describe the bug**
scale title filter handles keystrokes even when its output is not focused

**To Reproduce**
1. Start scale with scale-title-filter
2. Switch focused output to a different output
3. Type any key
4. That key is applied as a filter, even if the scale-title-filter should not handle it at all.

**Wayfire version**
git

dkondor:
Fixes #1021 -- made it an option if filter should be shared among outputs + keys are ignored if scale is not running on the current output.

Notes:
 - if the sharing option changes while scale is running, the filter will be reset
 - If the active output changes while a key is held down, it is not handled (but it is handled when the key is released)

I think dealing with the above cases in a special way would add a lot of complexity for a very little benefit.

### DIFF
--- a/metadata/scale-title-filter.xml
+++ b/metadata/scale-title-filter.xml
@@ -5,9 +5,14 @@
 		<_long>Filter the views shown by scale by typing.</_long>
 		<category>Utility</category>
 		<option name="case_sensitive" type="bool">
-			<_short>Case Sensitive</_short>
+			<_short>Case sensitive</_short>
 			<_long>Match titles in a case sensitive way.</_long>
 			<default>false</default>
+		</option>
+		<option name="share_filter" type="bool">
+			<_short>Share filter among outputs</_short>
+			<_long>Whether the active filter is shared among all outputs. Set to false to filter independently on each output.</_long>
+			<default>true</default>
 		</option>
 		<option name="overlay" type="bool">
 			<_short>Show overlay</_short>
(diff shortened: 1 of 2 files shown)


metadata/scale-title-filter.xml
```diff
@@ -5,10 +5,15 @@
...
 		<option name="case_sensitive" type="bool">
-			<_short>Case Sensitive</_short>
+			<_short>Case sensitive</_short>
 			<_long>Match titles in a case sensitive way.</_long>
 			<default>false</default>
 		</option>
+		<option name="share_filter" type="bool">
+			<_short>Share filter among outputs</_short>
+			<_long>Whether the active filter is shared among all outputs. Set to false to filter independently on each output.</_long>
+			<default>true</default>
```
ammen99, line 15: should be off by default, our defaults are independent outputs

plugins/scale/scale-title-filter.cpp
```diff
@@ -134,36 +206,25 @@ class scale_title_filter : public wf::plugin_interface_t
         auto xkb_state = keyboard->xkb_state;
         xkb_keycode_t keycode = raw_keycode + 8;
         xkb_keysym_t keysym   = xkb_state_key_get_one_sym(xkb_state, keycode);
+        auto& filter = share_filter ? get_instance() : local_filter;
```
ammen99, line 209: Maybe we can extract this bit in its own short function? Seems it is repeated quite a lot.

ammen99 (commented):
Looks good :) A few small things and we can merge.

dkondor:
> Looks good :) A few small things and we can merge.

OK, I've added these changes

ammen99 (approved):
Thanks!

PR #1050, opened 2021-01-19
view: properly calculate bounding boxes for reverse transforms

dkondor:
I was experimenting and realized that reverse transform of coordinates in a view with 3D rotations does not work in interactive scale. Part of the reason is the FIXME that this PR aims to fix.

Note: this does not fix the original problem, since that is caused by the view_2D transformer always working with the untransformed WM geometry. I'll open a separate PR for that.


ammen99 (approved):
LGTM, thanks for fixing these old TODOs :)

PR #1051, opened 2021-01-19
Make transforms relative to the WM geometry

dkondor:
Depends on #1050 

Transforms receive the view's bounding box to be able to work in a local coordinate system. In my opinion, this is not ideal, since the bounding box will change in "unimportant" ways, e.g. if a tooltip is added that goes outside of the main surface's bounds.

The view_2D transform works this around by using the WM geometry instead, but this only works if it is the first transformer (which is not the case e.g. for scale).

This PR changes the logic so that transforms always receive the (partially transformed) WM geometry to work with. Also, this makes the view_2D transformer use the passed in geometry.

This fixes my issue with 3D rotation and interactive scale.

Note: without the change to use WM geometry everywhere, changing view_2D to use the passed-in geometry resulted in weird artifacts.

Possible alternatives:
 - Pass both the partially transformed WM geometry and the full bounding box (so that a transformer always knows the full bounds -- currently, the bounding box is only passed to the rendering function).
 - Pass the partially transformed coordinates of the view's center (this is what transformers seem to care about anyway).
 - Change the interface so that transformers always work in a coordinate system where the view's center is in the origin; transform all coordinates to this in view.cpp, transformers don't need to receive any extra geometry parameter.

### DIFF
--- a/plugins/animate/fire/fire.cpp
+++ b/plugins/animate/fire/fire.cpp
@@ -102,7 +102,7 @@ class FireTransformer : public wf::view_transformer_t
         p.base_radius = p.radius = random(size * 0.8, size * 1.2);
     }
 
-    void render_box(wf::texture_t src_tex, wlr_box src_box,
+    void render_box(wf::texture_t src_tex, wlr_box src_box, wlr_box wm_geom,
         wlr_box scissor_box, const wf::framebuffer_t& target_fb) override
     {
         OpenGL::render_begin(target_fb);
--- a/plugins/blur/blur.cpp
+++ b/plugins/blur/blur.cpp
@@ -67,7 +67,7 @@ class wf_blur_transformer : public wf::view_transformer_t
         OpenGL::render_end();
     }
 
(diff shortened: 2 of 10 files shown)


ammen99:
Not sure about this change. Transformers do transform the full bounding box, this is how they work. For example wobbly, fire, etc. They change the full bounding box, and the API should reflect that.

What you said about view_2D is also a bug though. The way view_2D is written, you're right that it expects to always be the first transformer which is not optimal wrong. However, I don't think pushing one more parameter through all of these functions (or changing the bbox parameter to be the wm geometry) is a good way forward. Why can't the transformer just transform the wm_geometry() with view->get_bounding_box(view->get_wm_geometry(), {this}) ? The amount of operations is indeed O(N^2), but a view with more than 2-3 transformers will add a ton of overhead anyway.

dkondor:
I see your point and agree that using the workaround you suggest is a much simpler solution for now. I'll create an alternate version that only changes the behavior of the 2D transformer.

ammen99:
Closing since #1056 supersedes this PR.

PR #1052, opened 2021-01-19
Scale: ensure transformed views fit in their slot

### LINKED ISSUE #1046: Scale and transformers
**Describe the bug**
Not sure if it is actually a bug or just behavior not considered yet. But the main point is that scale does not consider any transformers when calculating the scaling to use. This results in strange layout e.g. if a view is rotated with the wrot plugin.

Not sure if this is a big issue (I'm not really using the rotate plugin regularly, but it is cool to have :) ). Also, fixing this would need a lot of special casing, since some transformers are still better to be ignored by scale (e.g. window open / close animations, wobbly, or anything else after scale, etc.).

**To Reproduce**
1. Enable wrot and rotate a view.
2. Start scale
3. Rotated size is not taken into account (e.g. rotated view can overlap with its neighbors).

**Expected behavior**
View is scaled so that it does not overlap with any other views.

**Screenshots or stacktrace**
20210117_22h03m07s_grim

**Wayfire version**
git master

dkondor:
Fixes #1046 

Depends on #1050 and ~~#1051~~ #1056 -- otherwise the results can look strange.

Note: this is not a perfect solution, since the bounding box of the view's WM geometry can grow excessively with multiple transforms (e.g. two 45 degree rotations will result in the bounding box growing twice as wide). A better way could be to transform each of the corner points (but that only works for transforms that preserve straight lines -- so far all transforms do that).

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -968,7 +968,7 @@ class wayfire_scale : public wf::plugin_interface_t
                     (view == current_focus_view) ? 1 : (double)inactive_alpha;
 
                 // Helper function to calculate the desired scale for a view
-                const auto& calculate_scale = [=] (wf::geometry_t vg,
+                const auto& calculate_scale = [=] (wf::dimensions_t vg,
                                                    const wf::scale_transformer_t::
                                                    padding_t& pad)
                 {
@@ -985,7 +985,9 @@ class wayfire_scale : public wf::plugin_interface_t
                 };
 
                 add_transformer(view);
-                double view_scale = calculate_scale(view->get_wm_geometry(),
+                auto geom = view->transform_region(view->get_wm_geometry(),
+                    scale_data[view].transformer);
+                double view_scale = calculate_scale({geom.width, geom.height},
                     scale_data[view].transformer->get_scale_padding());
                 for (auto& child : view->enumerate_views(false))
                 {
@@ -1021,11 +1023,14 @@ class wayfire_scale : public wf::plugin_interface_t
                         continue;
                     }
 
-                    auto vg = child->get_wm_geometry();
+                    auto vg = child->transform_region(child->get_wm_geometry(),
+                        child_data.transformer);
+                    wf::pointf_t center = {vg.x + vg.width / 2.0,
+                        vg.y + vg.height / 2.0};
 
                     // Take padding into account
                     auto pad     = child_data.transformer->get_scale_padding();
-                    double scale = calculate_scale(vg, pad);
+                    double scale = calculate_scale({vg.width, vg.height}, pad);
                     // Ensure child is not scaled more than parent
                     if (!allow_scale_zoom &&
                         (child != view) &&
@@ -1035,10 +1040,8 @@ class wayfire_scale : public wf::plugin_interface_t
                     }
 
                     // Target geometry is centered around the center slot
-                    const double dx = x + pad.left - vg.x +
-                        ((scaled_width - vg.width) / 2.0);
-                    const double dy = y + pad.top - vg.y +
-                        ((scaled_height - vg.height) / 2.0);
+                    const double dx = x + pad.left - center.x + scaled_width / 2.0;
+                    const double dy = y + pad.top - center.y + scaled_height / 2.0;
                     setup_view_transform(child_data, scale, scale,
                         dx, dy, target_alpha);
                 }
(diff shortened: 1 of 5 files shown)


plugins/scale/scale.cpp
```diff
@@ -985,8 +985,13 @@ class wayfire_scale : public wf::plugin_interface_t
                 };
 
                 add_transformer(view);
-                double view_scale = calculate_scale(view->get_wm_geometry(),
-                    scale_data[view].transformer->get_scale_padding());
+                auto geom = view->get_wm_geometry();
+                auto tr_geom = view->transform_region(geom,
+                    scale_data[view].transformer);
+                double view_scale =
+                    calculate_scale({std::max(geom.width, tr_geom.width),
```
ammen99, line 992: why do we need `max` here?

ammen99:
> Fix for #1046
> 
> Depends on #1050 and #1051 -- otherwise the results can look strange.
> 
> Note: this is not a perfect solution, since the bounding box of the view's WM geometry can grow excessively with multiple transforms (e.g. two 45 degree rotations will result in the bounding box growing twice as wide). A better way could be to transform each of the corner points (but that only works for transforms that preserve straight lines -- so far all transforms do that).

Not all transforms are affine (only view_2D/3D are) - examples include fire, wobbly and my rounded-corners transformer (not part of this repo though).

How do you get two 45 degrees rotations though? wrot 2d + 3d? To fix that, we could make wrot 2d just use the 3d transformer, and then fix view_3d to be centered around the wm geometry.

plugins/scale/scale.cpp
dkondor: This only takes into account the transform if the view is grown by it. This is mainly a workaround for open / close animations. E.g. if a new view opens while scale is running, the animations plugin will attach its own animation (and I can't tell if that happens before or after this point). If this is the zoom animation, then we do not want to take it into account here.

This is not a perfect solution, since another animation in the future might make the view bigger, which will result in it scaled too small (this is the case of the fire animation, but luckily, that has a z-order above scale).

dkondor:
> > Fix for #1046
> > Depends on #1050 and #1051 -- otherwise the results can look strange.
> > Note: this is not a perfect solution, since the bounding box of the view's WM geometry can grow excessively with multiple transforms (e.g. two 45 degree rotations will result in the bounding box growing twice as wide). A better way could be to transform each of the corner points (but that only works for transforms that preserve straight lines -- so far all transforms do that).
> 
> Not all transforms are affine (only view_2D/3D are) - examples include fire, wobbly and my rounded-corners transformer (not part of this repo though).
> 
> How do you get two 45 degrees rotations though? wrot 2d + 3d? To fix that, we could make wrot 2d just use the 3d transformer, and then fix view_3d to be centered around the wm geometry.

Yes (in general, applying the 3D rotation after the 2D can result in a too large bounding box). I agree that this is not a big issue, and also that it would be a nice optimization to merge the two transformers for wrot -- I'll look into doing that when I have some time.

plugins/scale/scale.cpp
ammen99: Animations should have a higher Z order anyway (>=HIGHLEVEL), so this is a bug in the animation plugin. My logic is the following:

- At the lowest level, we have transformers which change the view for a long period (force-fullscreen, wrot, forcing a size in simple-tile, etc).
- After that, we have transformers which change the view for a duration (scale)
- After that, we have transformers which have temporary animations
- Finally, we have blur since it needs to know the background

ammen99:
@dkondor A small request I have, when you open a PR which fixes an issue, please write 'Fixes #N' instead of 'Fix for #N', because GitHub does not recognize the latter and does not close the issue automatically.

plugins/scale/scale.cpp
dkondor: I see; so animations need to set a high enough z-order. Would it make sense to change the `view_2D` and `view_3D` transformer classes to allow setting a custom z-order value at creation?

ammen99: > Would it make sense to change the `view_2D` and `view_3D` transformer classes to allow setting a custom z-order value at creation?

Sounds reasonable.

dkondor: I've added these changes (let me know if they should go to a separate PR instead).

This does clash with the title overlay display a bit (in case a dialog is opened / closed, the overlay is animated together with the dialog), but I don't think this is a big issue and could be fixed separately.

ammen99 (approved):
LGTM, thanks!

PR #1056, opened 2021-01-21
view-3d: 2D and 3D transformers should use tansformed WM geometry

dkondor:
Alternative approach to #1051 

For 2D transformer, this allows accurate transform / untransform of coordinates even if it is not the first transform (like in the case of scale).

For 3D transformer, this fixes rendering artifacts such as on the image below (small region near the tooltip has tearing).

transformer-error


ammen99 (approved):
LGTM, thanks!

PR #1064, opened 2021-01-24
workspace fixes

### LINKED ISSUE #861: Window occupying more than one workspace is always on top when switching workplaces
**Describe the bug**

When a window occupies two workspaces (which can happen accidentaly, for example with a mpv window with a resolution higher than the screen's), switching to workspaces that include it makes it always become the top screen, even if I used the switcher to change that. It's a bit annoying, since I'm required to either resize the mpv window or always switch applications when moving from one workspace to another.

video: https://streamable.com/598qnn

**Expected behavior**
The window order should be stored such that switching workspaces didn't change it.

**Wayfire version**
git: 76b0f7d4551b689a181c4fc65b66b284d41fdc58

### LINKED ISSUE #865: Switching workspace doesn't always focus views on the new one because transformed views are considered "visible on a workspace" when they shouldn't be
**Describe the bug**

https://github.com/WayfireWM/wayfire/blob/a36b9313e57e7dad1a1fdd86c987225bbe403e27/src/output/workspace-impl.cpp#L489-L495

^^^ there's a significant difference in how views are considered to be "visible on" a workspace depending on whether they have transformers.

Without transformers, `wm_geometry` is used, which is *correct*. This is what should be used for `wm` purposes, which of course includes focusing when switching workspaces. (Side note, maybe "visible" is not the best name for this?)

With transformers, `intersects_region` checks that any subsurface would visually touch the workspace, even though these subsurfaces can be *outside of `wm_region`* — these should not affect focusing and stuff!

For example `smithay-client-toolkit` draws CSD borders (when unused, they're blank fully transparent subsurfaces) on all sides of the main surface — the invisible border ends up on the next workspace, and when transformers exist, it is considered as if it should be used for focusing :/

**To Reproduce**
1. Run `alacritty` (well, anything using `smithay-client-toolkit`) on the first (left) workspace
2. Add a transformer to it (blur, alpha, whatever)
3. Maximize it (or snap to the right half, whatever — make it touch the right edge)
4. Run something on the second (right) workspace
5. Switch with fast-switcher from left to right, notice the thing on the new workspace is not focused (actually the alacritty from the first workspace is refocused instead)

**Wayfire version**
git

### LINKED ISSUE #930: [wm-actions] on_top moved with vswitch
**Describe the bug**
When moving a view with the vswitch plugin, vswitch instead moves a secondary view with `always_on_top` set even if the secondary view is not focused.

**To Reproduce**
1. Enable wm-actions and vswitch plugins
2. Toggle `always_on_top` view with wm-actions plugin
3. Focus a different view
4. Move focused view with vswitch plugin

**Expected behavior**
The focused view is moved with vswitch.

**Wayfire version**
0.6.0

### LINKED ISSUE #970: Fullscreen applications steal focus while moving windows between workspaces
**Describe the bug**
If there's a fullscreen application (using WM Actions) on a workspace, moving a window using the vswitch plugin is broken as the full screened window will steal focus from the window currently being moved 

**To Reproduce**
1. Enable vswitch, WM actions plugins
2. Toggle full screen of a window on a certain workspace
3. go to another workspace and try to move a window around using <super> <shift> arrows

**Expected behavior**
Grabbed window keeps in focus since the modifier keys are held down, so you can continue moving it to another workspace

**Actual behavior**
Full screened window steals focus and is grabbed instead

**Wayfire version**
git

ammen99:
This PR contains a few small fixes and heuristics to make the workspaces in Wayfire work better. I think it:

Fixes #861
Fixes #865
Fixes #930
Fixes #970

Let me know what you think about these changes (I see only the last two commits as a bit controversial but it makes things better IMO).

cc @pushqrdx @myfreeweb @travankor @ericonr

### DIFF
--- a/plugins/vswitch/wayfire/plugins/vswitch.hpp
+++ b/plugins/vswitch/wayfire/plugins/vswitch.hpp
@@ -306,19 +306,19 @@ class control_bindings_t
 
         callback_win_left = [=] (const wf::activator_data_t&)
         {
-            return handle_dir({-1, 0}, get_top_view(), callback);
+            return handle_dir({-1, 0}, get_target_view(), callback);
         };
         callback_win_right = [=] (const wf::activator_data_t&)
         {
-            return handle_dir({1, 0}, get_top_view(), callback);
+            return handle_dir({1, 0}, get_target_view(), callback);
         };
         callback_win_up = [=] (const wf::activator_data_t&)
         {
-            return handle_dir({0, -1}, get_top_view(), callback);
+            return handle_dir({0, -1}, get_target_view(), callback);
         };
         callback_win_down = [=] (const wf::activator_data_t&)
         {
-            return handle_dir({0, 1}, get_top_view(), callback);
+            return handle_dir({0, 1}, get_target_view(), callback);
         };
 
         wf::option_wrapper_t<wf::activatorbinding_t> binding_left{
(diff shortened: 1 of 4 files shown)


travankor:
Views that span multiple workspaces could break the minimally overlapping heuristic. As an alternative, views that are only in the outer perimeter of the workspace stay at the bottom of stacking order.

ammen99:
> Views that span multiple workspaces could break the minimally overlapping heuristic. As an alternative, views that are only in the outer perimeter of the workspace stay at the bottom of stacking order.

I don't get it, how do they break the heuristic?

PR #1071, opened 2021-01-30
simple tile improvements

### LINKED ISSUE #680: Improve animations in tiling mode
First off, let me say this: WAYFIRE IS AWESOME! Although the setup process was a little rough because I made the foolish decision to buy an nvidia card several years ago, everything else was impressively easy. The config file is simple, intuitive and has good defaults. Heck, it even reloads automatically when saved! Thank you for this awesome project!

Now, to get to the actual issue: I think the animations in tiling mode could be improved. When opening/closing windows, the animations work just like in stacking mode. There is no resizing animation for the already existing windows, however. This looks a little wrong IMO, especially when closing windows: the remaining windows only resize after the closing animation has finished. I would suggest running the closing and resizing animations simultaneously.

### LINKED ISSUE #681: Allow windows to be swapped in tiling mode
When moving a window in tiling mode, a blue box indicating where the window will go appears. This box can range over a single different window, in which case nothing happens after the mouse button has been released. I think it would be useful if the two windows would be swapped in that case, instead of doing nothing.

### LINKED ISSUE #771: Master and stack layout for tiling mode
Hi! I love how you added some tiling functionality. It's possible to add an automatic arrangement layout, like on dwm, xmonad or awesome? Something like this. Thanks!

image

ammen99:
This PR is aimed at improving simple-tile's behavior.

Goals:

- [x] Allow swapping windows (fixes #681)
- [x] Allow animations (crossfade) when a view changes its size (fullscreen, or a view is added).
- [x] Fix #680
- [ ] ~~Add plugin API for simple-tile so that users can make custom plugins which modify the tiling layout~~
- [ ] ~~Fix #771 by using the tile API to build a simple plugin for setting up master/stack (and fat) layouts~~


ammen99:
> * [ ]  Add plugin API for simple-tile so that users can make custom plugins which modify the tiling layout
> * [ ]  Fix #771 by using the tile API to build a simple plugin for setting up master/stack (and fat) layouts

@Javyre as far as I understand, these things would be eventually available in swayfire, right?

Javyre:
> @Javyre as far as I understand, these things would be eventually available in swayfire, right?

Yup

I intend to expose a public API for swayfire to modify the layout and query the node trees. (This won't come until at least 0.2)
It should also eventually support custom rules for where to place new opened nodes. (I'm still thinking about how best to achieve this)

I'm going to start a lua-wayfire integration experiment soon to see if these custom rules can be implemented by users in lua. That way a master/stack layout can be achieved by setting a custom function to execute for where to place new opened views as well as custom bindings for what to do when you move nodes up/down/left/right.

ammen99:
> > @Javyre as far as I understand, these things would be eventually available in swayfire, right?
> 
> Yup
> 
> I intend to expose a public API for swayfire to modify the layout and query the node trees. (This won't come until at least 0.2)
> It should also eventually support custom rules for where to place new opened nodes. (I'm still thinking about how best to achieve this)
> 
> I'm going to start a lua-wayfire integration experiment soon to see if these custom rules can be implemented by users in lua. That way a master/stack layout can be achieved by setting a custom function to execute for where to place new opened views as well as custom bindings for what to do when you move nodes up/down/left/right.

Sounds perfect ;)

In this case, I don't think it makes sense for me to duplicate the same functionality in Wayfire.

Issue #704, opened 2020-08-23
double clicking on decoration doesn't maximize


nodaba:
double clicking on decoration doesn't maximize the window.
works with gtk programs though.

Thks 4 ur attention, interest & support.

Arnaudv6:
I gave this a quick look. Double-click event is not used anywhere yet in wayfire it seems?
Also it requires delaying "move" event with a (x,y) threshold on [current_input](https://github.com/WayfireWM/wayfire/blob/master/plugins/decor/deco-layout.cpp#L225).
Time-wise, is there a ready to use function, a la wlr_keyboard_set_repeat_info(), that we should rely on in both:
https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/pointer.cpp
https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/touch.cpp
I'd gladly help on that one, but I **do** understand if you don't want to babysit me.
I don't find it that easy, or am I making here harder than it needs to be?

travankor:
>double clicking on decoration doesn't maximize the window.

Shouldn't you use the maximize button instead? Does KDE do this? Qt with CSD doesn't do this behavior.

Arnaudv6:
> Shouldn't you use the maximize button instead? Does KDE do this? Qt with CSD doesn't do this behavior.

That's a classic behavior in openbox, gnome and all. Rationale is you don't need to aim for a (small and distant) button.

ammen99:
@Arnaudv6 There is no such thing as double click handling in Wayfire, we get only a single click and release events and forward them to the clients. To implement this, you'll basically have to detect double click in the decoration plugin. A few starting points:

- Button press and release events coming to the decoration (probably not overly interesting here): https://github.com/WayfireWM/wayfire/blob/master/plugins/decor/deco-subsurface.cpp#L170
- Where the actual logic is: https://github.com/WayfireWM/wayfire/blob/master/plugins/decor/deco-layout.cpp#L225

How I'd go about implementing this: In the decoration layout, have a `wf::wl_timer`. Once the first click happens, start that timeout, if a second click comes before the timeout expires, well, you have a double-click event, return the action to the deco-subsurface.

Regarding move: I would change the mechanics a bit: make it so that clicking on the view once does not start moving it. Instead, make it so that if the user presses down a button and moves say 5 pixels while holding the mouse, then the move action is started.


Hope this helps :) 

[...]

PR #1072, opened 2021-01-30
[do not merge] attempt at #704

Arnaudv6:
This is my first lambda function: hope I'm not messing anything.
I spotted an issue: maximized window sometimes receives the second button press.
Launch firefox with multiple (unloaded) tabs, double-click the title-bar.
Effect: firefox is maximized (cool) but the tab most close to cursor sometimes gets loaded...

### DIFF
--- a/plugins/decor/deco-layout.cpp
+++ b/plugins/decor/deco-layout.cpp
@@ -236,13 +236,7 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
                 return {DECORATION_ACTION_TOGGLE_MAXIMIZE, 0};
             } else
             {
-                timer.set_timeout(300, [=] ()
-                {
-                    timer.disconnect();
-                    return false;
-                });
-                // hardcoded 300ms, to be replaced by option in config:
-                // wf::option_wrapper_t<int> delay{"input/double_click_timeout"};
+                timer.set_timeout(300, [] () { return false; });
                 return {DECORATION_ACTION_MOVE, 0};
             }
         }
--- a/plugins/decor/deco-layout.cpp
+++ b/plugins/decor/deco-layout.cpp
@@ -234,7 +234,7 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
             if (timer.is_connected())  // we have a double-click (on press, not release)
                 return {DECORATION_ACTION_TOGGLE_MAXIMIZE, 0};
             else{
-                timer.set_timeout(300, &(this->timer.disconnect))
+                timer.set_timeout(300, [=] () {timer.disconnect();});
                 // hardcoded 300ms, to be replaced by option in config:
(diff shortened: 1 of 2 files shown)


soreau:
For what it's worth, there is a 'Convert to draft' button toward the top right you can use to disallow merging.

Arnaudv6:
Thanks @soreau. Though I just don't see it. I'll keep looking.

ammen99:
> Thanks @soreau. Though I just don't see it. I'll keep looking.

It is located on the right...

Arnaudv6:
Spotted it, with ^F. Thanks

Arnaudv6:
OK, I have one question left:
do you wish a new option input/double_click_timeout, or do I leave hard-coded value of 300ms?

ammen99:
> OK, I have one question left:
> do you wish a new option input/double_click_timeout, or do I leave hard-coded value of 300ms?

If anything, the option should be in the decoration plugin because only it is affected by it.

However, I don't think an option here is required, unless someone requests it.

plugins/decor/deco-layout.cpp
```diff
@@ -230,7 +231,20 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
...
         {
-            return {DECORATION_ACTION_MOVE, 0};
+            if (timer.is_connected()) /* we have a double-click (before release) */
+            {
+                return {DECORATION_ACTION_TOGGLE_MAXIMIZE, 0};
+            } else
+            {
+                timer.set_timeout(300, [=] ()
+                {
+                    timer.disconnect();
```
ammen99, line 241: you don't need a disconnect here, `return false` is enough.

plugins/decor/deco-layout.cpp
```diff
@@ -230,7 +231,20 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
...
+                    return false;
+                });
+                // hardcoded 300ms, to be replaced by option in config:
+                // wf::option_wrapper_t<int> delay{"input/double_click_timeout"};
+                return {DECORATION_ACTION_MOVE, 0};
```
ammen99, line 246: I don't think this is very nice, we start a move action (potentially triggering wobbly & co) only to immediately exit it. Can we instead modify the code a bit more: trigger move on pointer motion if we have clicked on an area suitable for moving. (you'd also need to change the return type of the `handle_motion` method and handle it in deco-subsurface.cpp)

ammen99 (changes requested):

plugins/decor/deco-layout.cpp
Arnaudv6: agreed. (did not see as I don't use wobbly).
I just filled #1074 which is related but more complicated.

Arnaudv6 (commented):
closing this and opening another pull request to modify deco-subsurface.cpp also.

PR #1075, opened 2021-02-01
Implement #704, and prevent wobbly on simple click on tittle-bar

### LINKED ISSUE #704: double clicking on decoration doesn't maximize

Arnaudv6:
Fixes #704

### DIFF
--- a/plugins/decor/deco-layout.cpp
+++ b/plugins/decor/deco-layout.cpp
@@ -3,6 +3,7 @@
 #include <wayfire/core.hpp>
 #include <wayfire/nonstd/reverse.hpp>
 #include <wayfire/nonstd/wlroots-full.hpp>
+#include <wayfire/util.hpp>
 
 #define BUTTON_ASPECT_RATIO (25.0 / 16.0)
 #define BUTTON_HEIGHT_PC 0.8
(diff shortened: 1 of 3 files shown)


Arnaudv6:
[handle_action(layout.handle_motion)
](https://github.com/WayfireWM/wayfire/pull/1075/files#diff-e83db7f18f3b3aa510185fd64c195a41035824eb5ce79ccf22395da05b29267fR167) makes for not so intuitive code to read I concede.

plugins/decor/deco-layout.cpp
```diff
@@ -230,7 +243,17 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
...
         {
-            return {DECORATION_ACTION_MOVE, 0};
+            if (timer.is_connected())
+            {
+                double_click_at_release = true;
+            } else
+            {
+                timer.set_timeout(300, [=] ()
+                {
+                    timer.disconnect();
```
ammen99, line 253: So you say that without this disconnect Wayfire crashes?

ammen99 (commented):
I don't understand why you need to disconnect the timer manually, it should just work.
Otherwise, LGTM.

plugins/decor/deco-layout.cpp
Arnaudv6: It crashes not, but stays connected: every isolated click after the first one then toggles maximization.

Arnaudv6:
fixes #704. Thanks @ammen99 & @soreau for given assistance

plugins/decor/deco-layout.cpp
ammen99: The timer is supposed to automatically disconnect if the handler returns false. There was a bug in core - can you try #1086 and without the  manual disconnect here?

Arnaudv6: yes, tested (on 0.7.0) works fine. Thanks!

ammen99 (approved):
LGTM, thanks!

PR #1105, opened 2021-03-02
Unbreak CI

ammen99:
Seems like something changed and /etc/pacman.conf is not there, so try to create it before adding stuff there.


ammen99:
Seems like stuff is working again without any changes.

PR #1107, opened 2021-03-05
Only inotify the config file

diamondburned:
This pull request makes Wayfire only watch the config directory for changes related to the Wayfire config file. This prevents sudden config reloads when using other screen managers like `kanshi` when an application modifies the config folder.

### DIFF
--- a/src/default-config-backend.cpp
+++ b/src/default-config-backend.cpp
@@ -9,29 +9,72 @@
 #include <sys/inotify.h>
 #include <unistd.h>
 
-#define INOT_BUF_SIZE (1024 * sizeof(inotify_event))
-static char buf[INOT_BUF_SIZE];
-
+#define INOT_BUF_SIZE (sizeof(inotify_event) + NAME_MAX + 1)
 
 static std::string config_dir, config_file;
 wf::config::config_manager_t *cfg_manager;
 
-static void reload_config(int fd)
+static void readd_watch(int fd)
 {
-    wf::config::load_configuration_options_from_file(*cfg_manager, config_file);
     inotify_add_watch(fd, config_dir.c_str(), IN_CREATE);
     inotify_add_watch(fd, config_file.c_str(), IN_MODIFY);
 }
 
+static void reload_config(int fd)
+{
+    wf::config::load_configuration_options_from_file(*cfg_manager, config_file);
+    readd_watch(fd);
+}
+
 static int handle_config_updated(int fd, uint32_t mask, void *data)
 {
+    if ((mask & WL_EVENT_READABLE) == 0)
+    {
+        return 0;
+    }
+
     LOGD("Reloading configuration file");
 
-    /* read, but don't use */
-    read(fd, buf, INOT_BUF_SIZE);
-    reload_config(fd);
+    char buf[INOT_BUF_SIZE] __attribute__((aligned(alignof(inotify_event))));
 
-    wf::get_core().emit_signal("reload-config", nullptr);
+    bool should_reload = false;
+    inotify_event *event;
+
+    while (true)
+    {
+        auto len = read(fd, buf, INOT_BUF_SIZE);
+        if (len < 0)
+        {
+            break;
+        }
+
+        for (char *ptr = buf; ptr < (buf + len);
+             ptr += sizeof(inotify_event) + event->len)
+        {
+            event = reinterpret_cast<inotify_event*>(ptr);
+
+            if (event->len == 0)
+            {
+                // is file, probably the config file itself
+                should_reload = true;
+                continue;
+            }
+
+            if (config_file == event->name)
+            {
+                should_reload = true;
+            }
+        }
+    }
+
+    if (should_reload)
+    {
+        reload_config(fd);
+        wf::get_core().emit_signal("reload-config", nullptr);
+    } else
+    {
+        readd_watch(fd);
+    }
 
     return 0;
 }


src/default-config-backend.cpp
```diff
@@ -9,29 +9,56 @@
...
 
-    /* read, but don't use */
-    read(fd, buf, INOT_BUF_SIZE);
-    reload_config(fd);
+    char buf[INOT_BUF_SIZE]
+    __attribute__((aligned(alignof(inotify_event))));
+
+    bool should_reload = false;
+
+    while (read(fd, buf, INOT_BUF_SIZE) > 0)
```
ammen99, line 38: Read is not guaranteed to read INOT_BUF_SIZE bytes, see https://man7.org/linux/man-pages/man2/read.2.html
Thus, we need to handle this situation. Perhaps we need a generic utility function for reading exactly N bytes? https://github.com/WayfireWM/wayfire/blob/ipc/src/ipc/ipc.cpp#L172

diamondburned: Do you mean `read` might not read `sizeof(inotify_event)` or `INOT_BUF_SIZE`? `sizeof(inotify_event)` is the size needed, and the rest is just according to the `len` field.

ammen99: I mean that, read might read just a single byte, then get interrupted, right?

diamondburned: Right. Would checking the bytes read to be larger than `sizeof(inotify_event)` suffice then? Or would checking `errno` be a better idea?

diamondburned: The change was trivial enough that I did it anyway. It now checks for `while (read(...) >= sizeof(inotify_event))`.

src/default-config-backend.cpp
```diff
@@ -9,29 +9,56 @@
...
     inotify_add_watch(fd, config_file.c_str(), IN_MODIFY);
 }
 
+static void reload_config(int fd)
+{
+    wf::config::load_configuration_options_from_file(*cfg_manager, config_file);
+    readd_watch(fd);
+}
+
 static int handle_config_updated(int fd, uint32_t mask, void *data)
```
soreau, line 29: We might want to also go ahead and check that `mask` has WL_EVENT_READABLE flag set.

diamondburned: Is the preferred way to do this `(mask & WL_EVENT_READABLE) != 0`?

soreau: I'd probably return early if it's not set.

diamondburned: Added, though untested.

diamondburned: I added a check for `(mask & WL_EVENT_READABLE) == 0` for checking that it's not set. I'll leave this open until further review, because I did the check incorrectly for a split second there.

ammen99: This part looks ok now.

src/default-config-backend.cpp
```diff
@@ -9,29 +9,61 @@
...
-#define INOT_BUF_SIZE (1024 * sizeof(inotify_event))
-static char buf[INOT_BUF_SIZE];
-
+#define INOT_BUF_SIZE (sizeof(inotify_event) + NAME_MAX + 1)
 
 static std::string config_dir, config_file;
 wf::config::config_manager_t *cfg_manager;
 
-static void reload_config(int fd)
+static void readd_watch(int fd)
```
ammen99, line 17: why is this `readd_watch`? can we have a better name?

src/default-config-backend.cpp
```diff
@@ -9,29 +9,61 @@
...
-    /* read, but don't use */
-    read(fd, buf, INOT_BUF_SIZE);
-    reload_config(fd);
+    char buf[INOT_BUF_SIZE]
+    __attribute__((aligned(alignof(inotify_event))));
+
+    bool should_reload = false;
 
-    wf::get_core().emit_signal("reload-config", nullptr);
+    while (read(fd, buf, INOT_BUF_SIZE) >= sizeof(inotify_event))
```
ammen99, line 43: Unfortunately, still wrong. You'd need something like this:

```cpp
read_exact(fd, buf, sizeof(inotify_event));
auto event = reinterpret_cast<inotify_event*>(buf);
read_exact(fd, buf + sizeof(inotify_event), event->len);
```

with read_exact from [here](https://github.com/WayfireWM/wayfire/blob/ipc/src/ipc/ipc.cpp#L175)

ammen99: The reason is: read can be interrupted at any time, so in your loop read might just get 1 byte the first time, but read much more bytes the second time. This would mean that you're reading mis-aligned data.

diamondburned: I don't think that would work. According to the manual:

```
       The behavior when the buffer given to read(2) is too small to
       return information about the next event depends on the kernel
       version: in kernels before 2.6.21, read(2) returns 0; since
       kernel 2.6.21, read(2) fails with the error EINVAL.
```

This suggests that reading a small buffer the size of exactly `inotify_event` will not work if the event comes with a name after it. Instead, the `read` will error out with `EINVAL`.

I don't think my implementation is incorrect in that `read` might return an incomplete event that is unaccounted for.

The implementation does seem to not account for multiple `inotify_event`s in the same buffer, however. The C implementation does this by incrementing the buffer pointer by `sizeof(inotify_event) + event->len`.

soreau:
> Unfortunately, still wrong. You'd need something like this:

> read_exact(fd, buf, sizeof(inotify_event));
> auto event = reinterpret_cast<inotify_event*>(buf);
> read_exact(fd, buf + sizeof(inotify_event), event->len);

Is it possible that the bytes can traverse the `inotify_event` struct boundary? In other words, broken into two or more calls made by libwayland-server.

src/default-config-backend.cpp
ammen99: Ah, so it seems that inotify guarantees we'll get a full event when we read from the FD, which is quite nice. I should have read the inotify docs too, not just the ones for read.

```
Each successful read(2) returns a buffer containing one or more of the following structures:
```



> The implementation does seem to not account for multiple `inotify_event`s in the same buffer, however. The C implementation does this by incrementing the buffer pointer by `sizeof(inotify_event) + event->len`.

Would it be hard to handle multiple events here too?

ammen99:
> > 
> 
> Is it possible that the bytes can traverse the `inotify_event` struct boundary? In other words, broken into two or more calls made by libwayland-server.

The inotify_event can be bigger than `sizeof(inotify_event)` simply because of the variable-length array in the struct (for the file name). But we don't read any bytes via libwayland-server, it only does the polling for us.

src/default-config-backend.cpp
diamondburned: I don't think it'll be hard, no. I just have to do it.

diamondburned: I've rebased in a fix for this. It's untested though.

src/default-config-backend.cpp
diamondburned: I'm not sure, I think the name accurately describes what it does, and it's a function not used elsewhere, so I don't think it's a huge deal. I can change it if you have a better suggestion.

ammen99: oh, you're right, I just misread the name, I thought it was `read` + `d` + `_watch` but it is more like `re` + `add` + `watch`, which makes perfect sense indeed.

ammen99 (approved):
Looks good, thanks!

soreau (commented):
I think this needed more time to review before it was merged.

src/default-config-backend.cpp
```diff
@@ -9,29 +9,72 @@
...
+            event = reinterpret_cast<inotify_event*>(ptr);
+
+            if (event->len == 0)
+            {
+                // is file, probably the config file itself
+                should_reload = true;
+                continue;
+            }
+
+            if (config_file == event->name)
```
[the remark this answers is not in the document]
ammen99: Yes, it should work. `std::string` has a constructor from a char* which will be invoked so that the correct == operator is used.

src/default-config-backend.cpp
```diff
@@ -9,29 +9,72 @@
...
+        for (char *ptr = buf; ptr < (buf + len);
+             ptr += sizeof(inotify_event) + event->len)
+        {
+            event = reinterpret_cast<inotify_event*>(ptr);
+
+            if (event->len == 0)
+            {
+                // is file, probably the config file itself
+                should_reload = true;
+                continue;
```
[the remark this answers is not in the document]
diamondburned: It's `continue` instead of `break` to drain all the `inotify_event` in the buffer.

All files are assumed to be the Wayfire configuration file because the only 2 items being subscribed are the configuration folder and the configuration file itself.

src/default-config-backend.cpp
```diff
@@ -9,29 +9,72 @@
...
+        }
+    }
+
+    if (should_reload)
+    {
+        reload_config(fd);
+        wf::get_core().emit_signal("reload-config", nullptr);
+    } else
+    {
+        readd_watch(fd);
```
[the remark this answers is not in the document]
diamondburned: I only kept this here because the previous code had it.

diamondburned:
I've made pull request #1133 to address the potential fixes in this one.

PR #1114, opened 2021-03-10
build: make stacktrace configureable

travankor:
This should help users that want coredump generation to work correctly.

Systemd-coredump should work when this option is disabled, but I haven't
tested systemd-coredump specifically.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -116,7 +116,7 @@ static void wlr_log_handler(wlr_log_importance level,
     wf::log::log_plain(wlevel, buffer);
 }
 
-#ifndef ASAN_ENABLED
+#ifndef ASAN_ENABLED || COREDUMP_ENABLED
 static void signal_handler(int signal)
 {
     std::string error;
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -116,7 +116,7 @@ static void wlr_log_handler(wlr_log_importance level,
     wf::log::log_plain(wlevel, buffer);
 }
 
-#ifndef ASAN_ENABLED
+#ifndef ASAN_ENABLED || COREDUMP_ENABLED
```
ammen99, line 119: is ifndef the appropriate thing to use here?

ammen99 (changes requested):
I'd rather have a single option which controls stracktrace, not one for asan and one for coredump, it seems confusing to me, especially because both are used for the same thing.

src/meson.build
```diff
@@ -61,9 +61,12 @@ endif
 
 
 cxx_flags_asan = run_command('/bin/sh', '-c', 'echo $CXXFLAGS $CPPFLAGS | grep fsanitize')
-if get_option('b_sanitize').contains('address') or cxx_flags_asan.returncode() == 0
-  message('Address sanitizer enabled, disabling internal backtrace')
-  debug_arguments += ['-DASAN_ENABLED']
+if (not (get_option('b_sanitize').contains('address') or cxx_flags_asan.returncode() == 0)
```
ammen99, line 64: I find this a bit confusing, took me awhile to understand what the ifs do exactly. Can't we just set `print_trace` to false if asan is enabled, and then add the debug arguments in a second if?

ammen99 (changes requested):

src/meson.build
travankor: Sorry for the delay

ammen99 (approved):
LGTM now, thanks!

PR #1126, opened 2021-03-21
wayfire_xdg_popup: calculate position during initialization

### LINKED ISSUE #1125: xdg_popup::configure sends wrong coordinates if popup was moved
**Describe the bug**
When positioning a popup that is [constrained](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/master/stable/xdg-shell/xdg-shell.xml#L238), the coordinates sent with the configure event are the ones before the constraint adjustment was performed.

**To Reproduce**
1. Create an xdg_toplevel and an xdg_popup
2. Position the xdg_popup in a way that it would partially go off screen
3. Set `constraint_adjustment` to allow sliding
4. The popup's position is adjusted so that it is displayed fully on screen
5. The configure event still contains the coordinates before adjustment

**Expected behavior**
Coordinates after adjustment are sent with the configure event.

**Concrete example**
https://github.com/dkondor/xdg-popup-test

1. Compile and run the code from the above repository
2. Clicking on any location in the main window will create a popup, horizontally centered on the clicked point; it will print the coordinates reported by the configure event in the terminal
3. Position the main window (e.g. with the move plugin) so it is next to a screen edge
4. Click in a location close to the edge
5. Popup is positioned in a way that it is fully visible
6. Location reported in the configure event has coordinates that go outside the main window (and the screen); e.g. if the main window is at the left screen edge, the x coordinate of the popup will be reported as a negative number even if it is actually positioned inside the main window.

This also happens with gtk menus (seen e.g. with `WAYLAND_DEBUG=1`)

**Screenshots or stacktrace**
configure_layout

E.g. the coordinates reported in the above case are: (-101, 166, 300, 300) -- I clicked in a location close to the screen edge; the y coordinate is OK, but the x should be positive.

**Wayfire version**
git

Works properly on sway.

dkondor:
Fixes #1125 

It seems that for xdg_popups, wlroots will immediately send a configure event in response to a commit request (see [here](https://github.com/swaywm/wlroots/blob/c740fccc9dd0913908c0632c10f8c6d10b2b1ca4/types/xdg_shell/wlr_xdg_surface.c#L372) and [here](https://github.com/swaywm/wlroots/blob/c740fccc9dd0913908c0632c10f8c6d10b2b1ca4/types/xdg_shell/wlr_xdg_positioner.c#L154)), so there is no chance to update the position then (the configure event might be sent with the wrong position).

This can be avoided by adjusting the position while creating the popup. This is what [sway is doing](https://github.com/swaywm/sway/blob/152a559e30244e64d4d61b9c87db442dfa04ee52/sway/desktop/xdg_shell.c#L107) if I understand correctly. Since the client is required to supply the requested size in the xdg_positioner object when creating the popup, this should be OK to do.

### DIFF
--- a/src/view/xdg-shell.cpp
+++ b/src/view/xdg-shell.cpp
@@ -65,6 +65,8 @@ void wayfire_xdg_popup::initialize()
         &this->parent_app_id_changed);
     popup_parent->connect_signal("title-changed",
         &this->parent_title_changed);
+
+    unconstrain();
 }
 
 void wayfire_xdg_popup::map(wlr_surface *surface)
@@ -82,7 +84,6 @@ void wayfire_xdg_popup::map(wlr_surface *surface)
 
     wlr_view_t::map(surface);
     update_position();
-    unconstrain();
 }
 
 void wayfire_xdg_popup::commit()


ammen99 (approved):
LGTM, also just tested and seems to not break popup constraining.

ammen99:
Thanks!

PR #1133, opened 2021-03-29
Fix default-config-backend PR #1107

### LINKED ISSUE #1120: Crash on config reload
**Describe the bug**
Wayfire crashed on config reload

**To Reproduce**
1. Make changes to config
2. Save file
3. ???
4. Wayfire freezes for several seconds before crashing

**Expected behavior**
Don't crash

**Screenshots or stacktrace**
http://ix.io/2SUj

**Wayfire version**
0.7.0

diamondburned:
This pull request fixes pull request #1107 and closes #1120.

### DIFF
--- a/src/default-config-backend.cpp
+++ b/src/default-config-backend.cpp
@@ -33,42 +33,40 @@ static int handle_config_updated(int fd, uint32_t mask, void *data)
         return 0;
     }
 
-    LOGD("Reloading configuration file");
-
     char buf[INOT_BUF_SIZE] __attribute__((aligned(alignof(inotify_event))));
 
     bool should_reload = false;
     inotify_event *event;
 
-    while (true)
+    auto len = read(fd, buf, INOT_BUF_SIZE);
+    if (len < 0)
     {
-        auto len = read(fd, buf, INOT_BUF_SIZE);
-        if (len < 0)
+        return 0;
+    }
+
+    for (char *ptr = buf;
+         ptr < (buf + len);
+         ptr += sizeof(inotify_event) + event->len)
+    {
+        event = reinterpret_cast<inotify_event*>(ptr);
+
+        if (event->len == 0)
         {
-            break;
+            // is file, probably the config file itself
+            should_reload = true;
+            continue;
         }
 
-        for (char *ptr = buf; ptr < (buf + len);
-             ptr += sizeof(inotify_event) + event->len)
+        if (config_file == event->name)
         {
-            event = reinterpret_cast<inotify_event*>(ptr);
-
-            if (event->len == 0)
-            {
-                // is file, probably the config file itself
-                should_reload = true;
-                continue;
-            }
-
-            if (config_file == event->name)
-            {
-                should_reload = true;
-            }
+            should_reload = true;
         }
     }
 
     if (should_reload)
     {
+        LOGD("Reloading configuration file");
+
         reload_config(fd);
         wf::get_core().emit_signal("reload-config", nullptr);
     } else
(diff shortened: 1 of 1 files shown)


src/default-config-backend.cpp
```diff
@@ -98,7 +98,7 @@ class dynamic_ini_config_t : public wf::config_backend_t
         config = wf::config::build_configuration(
             get_xml_dirs(), SYSCONFDIR "/wayfire/defaults.ini", config_file);
 
-        int inotify_fd = inotify_init1(IN_CLOEXEC);
+        int inotify_fd = inotify_init1(IN_CLOEXEC | IN_NONBLOCK);
```
soreau, line 101: NAK from me. This should be made to work with blocking reads.

soreau (changes requested):
The bug this fixes was introduced with #1107 and this patch doesn't seem like the proper fix.

src/default-config-backend.cpp
ammen99: How to handle multiple inofity events though?

soreau: If the event fires, it is readable and we can read at least one event. After that, I would probably use something like `select` to check if there are subsequent events to read from the fd.

diamondburned: How would that be any better? Isn't Wayland's event loop callback API supposed to be used like this, as in to return when there's nothing left in the `fd`? That's basically non-blocking, isn't it?

soreau: > How would that be any better? Isn't Wayland's event loop callback API supposed to be used like this, as in to return when there's nothing left in the `fd`? That's basically non-blocking, isn't it?

Yes you're right, after reading some man pages it's probably better to set the fd nonblocking in this case.

soreau (approved):

soreau:
I still think on IN_CREATE the `event->name` should be compared to `config_file`, since otherwise it fires every time a new file is created in the directory containing wayfire.ini

src/default-config-backend.cpp
soreau: but actually, libinotify might be writing while we are reading and this can cause a problem.

droc12345:
I've been in  #1134 trying to ease this problem.
I have a patch (also mentioned in 1134) https://dpaste.com/GTMRHZRBV that (at least to me) solves the need for watching the directory. 

My reasoning is, the config file needs to exist, and if it doesn't then we create an empty one, whether it's the default name or one passed in. Then you don't need to watch add the directory for create.

src/default-config-backend.cpp
diamondburned: Are we using `libinotify`?

droc12345: > Are we using `libinotify`?

You're using the linux kernel inotify tools.

soreau: libinotify is used in FreeBSD IIRC.

ammen99:
So, after reading inotify docs and some forums, it seems that inotify supports reading **whole events only**. Which means we cannot avoid reading more events at the same time. On the other hand, it seems we don't really need nonblock. I pushed a commit to this PR which adds some more documentation, and some improvements as suggested by @droc12345 

I also tested with these changes: dynamic reload works when the config file exists, no config reload is done for new files in the config dir (only if the config file itself is created). I tested with VIM and nano.

I also checked, and we do need to readd the watches each time, otherwise it just doesn't work on my system. Unfortunately, I have not been able to find out why.

@droc12345 @soreau @diamondburned please, test this and let me know if any of you still have remaining issues with the latest version.

ammen99 (approved):
Alright, seems like this should work well enough. Thanks everyone for the suggestions and testing.

PR #1135, opened 2021-04-01
Fix a permanent freeze after saving the config file

bodqhrohro:


bodqhrohro:
Oh, sorry, this seems to be a duplicate of #1133.

PR #1143, opened 2021-04-10
Drop gtk-primary-selection which was dropped in wlroots

unrelentingtech:
https://github.com/swaywm/wlroots/pull/2460


ammen99:
Thanks!

PR #1144, opened 2021-04-11
readme: update install instructions on FreeBSD

jbeich:
For new users `pkg install wayfire` is not enough as `wayfire` unless configured would just show blank/black screen. Expand the command line to include recommended addons in order to have something usable out of the box.

I'll package if something is recommended but missing.


ammen99:
Makes sense, thanks!

PR #1146, opened 2021-04-15
core API: Add get_snapshot

soreau:
Allow plugins to get the framebuffer containing the snapshot of the view.


ammen99 (approved):
LGTM, thanks!

PR #1147, opened 2021-04-18
Split input.xml into multiple sections

### LINKED ISSUE #824: Divide input.xml into sections
Currently, all the options in [input.xml](https://github.com/WayfireWM/wayfire/blob/master/metadata/input.xml) are grouped together which means all of the options are in the default 'General' tab in wcm. This issue is a call to create subgroups to organize the options in input.xml. The options should be divided into four categories: Keyboard, Mouse, Touch and General. It should be obvious where to place each option; if it does not fit in the first three groups, do not place it in a \<group>, so that it will be placed under the 'General' tab.

The convention is, by default, options are placed under a 'General' tab in wcm. If the xml file contains options in a \<group> section, a new tab is created for those options. The name of the tab is the \<_short> name in the \<group> section. An example of how to do this can be found in [cube.xml](https://github.com/WayfireWM/wayfire/blob/master/metadata/cube.xml#L8-L26).

xbjfk:
Closes #824

### DIFF
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -107,6 +107,7 @@
 				<min>0.0</min>
 			</option>
 		</group>
+		<!-- Touchpad -->
 		<group>
 			<_short>Touchpad</_short>
 			<_long>Configure the touchpad.</_long>
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -220,15 +220,19 @@
 			</option>
 		</group>
 		<!-- Cursor configuration -->
-		<option name="cursor_theme" type="string">
-			<_short>Cursor theme</_short>
-			<_long>Overrides the system default `XCursor` theme.</_long>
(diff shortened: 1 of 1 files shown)


soreau (changes requested):
Looks great in WCM, thanks for this! The only nitpick I have is that I'd like to see the two cursor options under its own 'Cursor' tab rather than leaving it to default to General.

soreau (approved):
👍

soreau:
Hm, now that I think again, maybe the cursor stuff should go under the mouse tab? What do you think? Either way it's fine with me.

xbjfk:
Maybe, but I think that would confuse touchpad-only users who won't be able to find the cursor options

ammen99:
Seems like this is ready to merge, thanks!

PR #1152, opened 2021-04-21
update meson.build and wlroots submodule

ammen99:
Make wayfire compile with wlroots-master

@soreau Some time ago you asked whether we should update the version number immediately after release ... what about having a version 0.7.999 to indicate we're in development?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,11 +15,22 @@ jobs:
   test_glibc_llvm:
     name: "Test with clang/glibc/libc++/lld on Arch Linux"
     runs-on: ubuntu-latest
-    container: archlinux:latest
+    container:
+        image: archlinux:latest
     steps:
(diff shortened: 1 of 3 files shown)


soreau:
> Make wayfire compile with wlroots-master
> 
> @soreau Some time ago you asked whether we should update the version number immediately after release ... what about having a version 0.7.999 to indicate we're in development?

I'd just make it simple and bump to 0.8.0 so 1) if a user has this and it's not released, we know they are using a git version 2) we don't have to bump again before release, which would halve the bump commit work.
But if we need to make 0.7.2 after bumping though, how does that work? Do we copy branch 0.7.1 to a new 0.7.2 branch, then apply whatever patches on top and tag it?

ammen99:
> I'd just make it simple and bump to 0.8.0 so 1) if a user has this and it's not released, we know they are using a git version 2) we don't have to bump again before release, which would halve the bump commit work.

Ok, fixed.

> But if we need to make 0.7.2 after bumping though, how does that work? Do we copy branch 0.7.1 to a new 0.7.2 branch, then apply whatever patches on top and tag it?

When a 0.N.X release happens, I have always created a new branch so far. Development is on master, then if there are enough important bugfixes I would cherry-pick them to the 0.N.X branch and tag a new release there, instead of on master.

Note that releases generally cannot happen on master: we track latest wlroots in master, but wlroots releases do not necessarily align with Wayfire releases, so we do need a separate branch with the necessary commits to make the release compatible with the latest wlroots release.

PR #1156, opened 2021-04-23
Expo: add keyboard interaction

### LINKED ISSUE #1142: Escape from expo plugin
Expo plugin ends only with preconfigured key (toggle option). It will be nice to end it with ESC key too. I guess all plugins end with ESC key, except expo.

dkondor:
Fixes #1142

Arrows keys can be used to select a target workspace, enter to change, esc to cancel.

Also added shading to unselected workspaces.

TODO:

- [x] Make this optional (add a bool option)
- [x] Option for the shading of unselected workspaces instead of fixed value of 70%
- [x] Animation when the selected workspace changes
- [x] Handle key repeat (when a key is held down)
- [ ] Make keys configurable? Postponed for later
- [x] Decide what should happen if multiple keys are pressed / keys are pressed while dragging a view? Only the last pressed key is used, keys are disabled while dragging a view.
- [x] Test for regressions in other places where `workspace-wall` is used
- [x] Test the case when the grid size changes while expo is running

### DIFF
--- a/metadata/expo.xml
+++ b/metadata/expo.xml
@@ -29,6 +29,19 @@
 			<_long>Allow the use of arrow keys to select a workspace, ESC to cancel and ENTER to accept.</_long>
 			<default>true</default>
 		</option>
+		<option name="inactive_brightness" type="double">
+			<_short>Brightness of inactive workspaces</_short>
+			<_long>Decrease the brightness of all workspaces except the currently selected one to this value.</_long>
+			<default>0.7</default>
+			<min>0.0</min>
+			<max>1.0</max>
+			<precision>0.01</precision>
+		</option>
+		<option name="transition_len" type="int">
+			<_short>Brightness transition duration</_short>
+			<_long>Duration of the transition of brightness when a new workspace is selected in milliseconds.</_long>
+			<default>200</default>
+		</option>
 		<option name="workspace_bindings" type="dynamic-list">
 			<_short>Select workspace</_short>
 			<_long>When the binding is triggered while expo is active, the corresponding workspace will be focused and Expo will exit.</_long>
--- a/metadata/expo.xml
+++ b/metadata/expo.xml
@@ -24,6 +24,11 @@
 			<_long>Sets the delimiter offset (in pixels) between workspaces.</_long>
 			<default>10</default>
 		</option>
+		<option name="keyboard_interaction" type="bool">
+			<_short>Keyboard interaction</_short>
+			<_long>Allow the use of arrow keys to select a workspace, ESC to cancel and ENTER to accept.</_long>
+			<default>true</default>
+		</option>
 		<option name="workspace_bindings" type="dynamic-list">
 			<_short>Select workspace</_short>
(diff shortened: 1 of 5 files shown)


metadata/expo.xml
```diff
@@ -24,6 +24,24 @@
...
+		</option>
+		<option name="inactive_brightness" type="double">
+			<_short>Brightness of inactive workspaces</_short>
+			<_long>Decrease the brightness of all workspaces except the currently selected one to this value.</_long>
+			<default>0.7</default>
+			<min>0.0</min>
+			<max>1.0</max>
+			<precision>0.01</precision>
+		</option>
+		<option name="transition_len" type="int">
```
ammen99, line 40: I'd prefer `transition_length`, the option name is not that long.

plugins/single_plugins/expo.cpp
```diff
@@ -71,8 +74,18 @@ class wayfire_expo : public wf::plugin_interface_t
     } state;
 
     int target_vx, target_vy;
+    int initial_vx, initial_vy;
```
ammen99, line 77: use `wf::point_t` in new code :) makes it easier to use if we need to pass the points around.

plugins/single_plugins/expo.cpp
```diff
@@ -71,8 +74,18 @@ class wayfire_expo : public wf::plugin_interface_t
...
     std::unique_ptr<wf::workspace_wall_t> wall;
 
+    wf::option_wrapper_t<int> delay{"input/kb_repeat_delay"};
```
ammen99, line 80: Feels like we're duplicating code already written here: https://github.com/WayfireWM/wayfire/blob/master/plugins/scale/scale-title-filter.cpp#L19

Maybe we can extract that struct in `plugins/common/...` and re-use it whenever we need key repeat?

plugins/single_plugins/expo.cpp
```diff
@@ -403,6 +440,153 @@ class wayfire_expo : public wf::plugin_interface_t
...
+                });
+
+                return false; // no more repeat
+            });
+
+            key_pressed = key;
+        }
+
+        auto dim = output->workspace->get_workspace_grid_size();
+        if (target_vx >= dim.width)
```
ammen99, line 522: What about using `(target_vx + dim.width) % dim.width` instead of this series of ifs?

ammen99 (commented):
I know this isn't ready but had some time for reviews today so here you go :)

ammen99:
> [ ] Make keys configurable?

If you're not particularly interested in this, I think it should be fine to just handle HJKL and leave this for later.

plugins/single_plugins/expo.cpp
dkondor: Makes sense, I changed for both (didn't make sense to have separate x/y for target if initial workspace is wf::point_t)

plugins/single_plugins/expo.cpp
dkondor: Yes, that makes sense; I've moved it there

dkondor:
I've addressed the code review comments and did some more testing, so I think this is now ready :)

ammen99 (approved):
Looks good, thanks!

PR #1157, opened 2021-04-23
build: remove redundant get_option

travankor:


src/meson.build
```diff
@@ -64,10 +64,10 @@ cxx_flags_asan = run_command('/bin/sh', '-c', 'echo $CXXFLAGS $CPPFLAGS | grep f
 if get_option('b_sanitize').contains('address') or cxx_flags_asan.returncode() == 0
   print_trace = false
   message('Address sanitizer enabled, disabling internal backtrace')
-  message('Overriding print_trace build option!')
+  message('Overriding print_trace build option')
 endif
 
-if get_option('print_trace')
```
soreau, line 70: There is an option in meson_options.txt that has this option. If it's not used anywhere (is this the only place?) then the option should be removed too.

travankor: No, it's in meson.option as well (if I understood correctly).

ammen99 (approved):
Looks good, thanks!

PR #1159, opened 2021-04-24
Prevent multiple events being fired by hotspot area

### LINKED ISSUE #1141: Scale plugin go crazy
**Describe the bug**
Scale plugin go crazy

**To Reproduce**
1. Enable scale plugin
2. set 'duration = 300'
2. set 'toggle_all = <super> KEY_W | hotspot top-left 10x10 200'
3. When you go to top left corner and stay there with mouse, very often scale plugin start for about 100ms, return back, than again start, than return back ... sometimes it ends with with scale plugin on, sometimes off.

**Expected behavior**
Scale plugin should start

**Wayfire version**
0.7.0-30f13ab3 (Apr  4 2021, branch 'HEAD')

JordanL2:
Fixes #1141 

Using a boolean flag to determine whether or not an event has already been fired by the cursor entering the hotspot area. Cursor must leave the area before another event can be fired.

### DIFF
--- a/src/core/seat/hotspot-manager.hpp
+++ b/src/core/seat/hotspot-manager.hpp
@@ -49,6 +49,12 @@ class hotspot_instance_t : public noncopyable_t
     /** Timer for hotspot activation */
     wf::wl_timer timer;
 
+    /**
+     * Only one event should be triggered once the cursor enters the hotspot area.
+     * This prevents another event being fired until the cursor has left the area.
+     */
+    bool armed = true;
+
     /** Timeout to activate hotspot */
     uint32_t timeout_ms;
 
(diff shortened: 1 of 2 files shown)


src/core/seat/hotspot-manager.hpp
```diff
@@ -49,6 +49,11 @@ class hotspot_instance_t : public noncopyable_t
     /** Timer for hotspot activation */
     wf::wl_timer timer;
 
+    /** Only one event should be triggered once the cursor enters the hotspot area.
```
ammen99, line 52: Small nitpick: if comment is not single-line, text should begin on a new line

ammen99 (commented):
LGTM aside from the comment

src/core/seat/hotspot-manager.hpp
JordanL2: Like this?

```
    /**
     *  Only one event should be triggered once the cursor enters the hotspot area.
     *  This prevents another event being fired until the cursor has left the area.
     */
```

ammen99: Yes, but then you'd need one less identation space too :)

JordanL2: Fixed :)

ammen99 (approved):
Perfect, will merge once build is ready, thanks!

PR #1168, opened 2021-05-03
blur: Ignore empty regions when painting

soreau:
When doing region math, sometimes we get an empty region which
means we try to allocate 0 sized buffer in both dimensions. Fix
this by checking the region is not empty before painting.


ammen99 (approved):
LGTM, thanks!

PR #1169, opened 2021-05-05
view: clamp damage to framebuffer and document the behavior

### LINKED ISSUE #1090: [BISECTED] Blur crashes wayfire when dragging to grid left
I get [this crash](http://ix.io/2XaR) when dragging a view to the left edge of the screen with blur and grid enabled. This only happens with dual outputs and mesa debug build (release builds are seemingly unaffected). Bisected to 60d3528, this commit also makes the views flicker around the edges when dragging offscreen other than x<=0.

ammen99:
Might fix #1090, @soreau please test.


soreau:
> Might fix #1090, @soreau please test.

I have tested and it does indeed fix the crash as well as the flickering. Thanks!

PR #1170, opened 2021-05-06
CI: Add xwayland test

soreau:
This builds the xwayland code paths that rely on WF_HAS_XWAYLAND being set.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     container: alpine:edge
     steps:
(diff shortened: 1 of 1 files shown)


ammen99:
I dont' think we need 3 CIs, replacing the current alpine tests should be enough (arch builds without xwayland anyway, so we do test that branch as well)

soreau:
```
wlroots| Run-time dependency xwayland found: NO (tried pkgconfig and cmake)
...
wayfire 0.8.0
       xwayland: true
...
wlroots 0.14.0
    xwayland      : YES
```

As far as I can tell, no released version has the xwayland.pc file pkg-config expects to find (hence the initial message) so currently wlroots looks for the Xwayland binary instead. In future releases of xwayland, wlroots will change and we can drop the xwayland dep in the CI.

ammen99 (approved):
Thanks!

PR #1173, opened 2021-05-07
workspace-impl: Damage all views in sublayer

Javyre:
This avoids glitchy rendering of the other views on the sublayer when
bring_to_front raises it.

### DIFF
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -230,10 +230,14 @@ class output_layer_manager_t
     /** Precondition: view is in some sublayer */
     void bring_to_front(wayfire_view view)
     {
-        view->damage();
-
         auto sublayer = get_view_sublayer(view);
         assert(sublayer);
+
+        for (auto view : sublayer->views)
+        {
+            view->damage();
+        }
+
         if (sublayer->mode == SUBLAYER_FLOATING)
         {
             raise_to_front(sublayer->layer->floating, sublayer);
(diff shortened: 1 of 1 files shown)


soreau (approved):
LGTM 👍

src/output/workspace-impl.cpp
```diff
@@ -230,10 +230,14 @@ class output_layer_manager_t
...
     void bring_to_front(wayfire_view view)
     {
-        view->damage();
-
         auto sublayer = get_view_sublayer(view);
         assert(sublayer);
+
+        for (auto view : sublayer->views)
+        {
+            view->damage();
```
ammen99, line 238: Now that we're on the topic, we should also damage child views (typically these are dialogs). And we should do that in each function in this file (instead of just view->damage()) ...

Javyre: done.

src/output/workspace-impl.cpp
```diff
@@ -171,6 +171,10 @@ class output_layer_manager_t
         }
 
         view->damage();
+        for (auto ch : view->children)
```
ammen99, line 174: Uhh sadly not good enough, this won't work for children's children. Usually, we use `view->enumerate_views()` for such purposes, and it would also include the view itself .. And maybe we can separate this in its own function (albeit a short one)?

Javyre: yup was gonna ask about creating a function actually.

Javyre: Alright I created a small `damage_views()` fn.

ammen99 (approved):
LGTM now, thanks!

PR #1184, opened 2021-05-17
zoom: Warp fix and Interpolation method option

Javyre:
See the commit description for the warp fix.


Javyre:
I added a comment to the warping fix commit to avoid breaking it in the future by accident. It's trivial enough to forget.

ammen99 (approved):
LGTM, thanks!

PR #1185, opened 2021-05-17
default-config-backend: Fix out of range error

Javyre:
When config_file is already a basename and doesn't contain a slash, the
return value of find_last_of is std::string::npos which is out of
bounds.


ammen99 (approved):
Nice catch!

PR #1187, opened 2021-05-18
view: Set geo to expanded wm geo on undecorate

Javyre:
This is the reverse of what happens later in the function (L842):
```
    if (!fullscreen && !this->tiled_edges) ...
```

### DIFF
--- a/src/view/view.cpp
+++ b/src/view/view.cpp
@@ -793,6 +793,10 @@ void wf::view_interface_t::set_decoration(surface_interface_t *frame)
     if (!frame)
     {
         damage();
+
+        // Take wm geometry as it was with the decoration.
+        const auto wm = get_wm_geometry();
+
         if (view_impl->decoration)
         {
             this->remove_subsurface(view_impl->decoration);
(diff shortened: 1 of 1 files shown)


src/view/view.cpp
```diff
@@ -793,13 +793,23 @@ void wf::view_interface_t::set_decoration(surface_interface_t *frame)
...
+
         if (view_impl->decoration)
         {
             this->remove_subsurface(view_impl->decoration);
         }
 
         view_impl->decoration = nullptr;
         view_impl->frame = nullptr;
+
+        if ((fullscreen || this->tiled_edges) && (wm != get_wm_geometry()))
```
ammen99, line 808: I think it is safe to only check whether geometry changes (after all, if we're fullscreen, the decoration should have size 0, in which case `wm == get_wm_geometry` anyway)

Javyre: This would expand normal floating views though. Non-tiling/fullscreen views don't get shrunken down on decorate in L842.. Or am I reading the code wrong?

ammen99: > This would expand normal floating views though. Non-tiling/fullscreen views don't get shrunken down on decorate in L842.. Or am I reading the code wrong?

Hmm, I can see where you're coming from. Well, for fullscreen views it doesn't make sense to change anything at all, so maybe `!fullscreen && tiled_edges && ...`?

ammen99: And if this is good for you, do leave a comment with the rationale for these conditions :)

Javyre: Alright so if I understand correctly, fullscreen views should not be decorated? (or at least their expanded wm geo would be the same as their view geo)

ammen99: Yes

PR #1188, opened 2021-05-19
Focus the main dragged view when a drag ends

### LINKED ISSUE #1186: Moving a dialog focuses parent
**Describe the bug**
When moving a dialog, the parent view gets focused.

**To Reproduce**
1. Disable the "Make all dialogs modal option" (`workarounds/all_dialogs_modal`)
2. Open a view and a modal dialog (tested: Search and replace dialog in [Geany](https://github.com/geany/geany) (Ctrl-F / Ctrl-H), Save changes confirmation in gedit)
3. Move the dialog (either by dragging the titlebar or with the keybinding, e.g. Super + dragging)
4. After the move, the parent view gets focused (titlebar turns active, starts to receive keyboard events for non-modal dialogs)

**Expected behavior**
Dialog stays focused

**Screenshots or stacktrace**
If applicable, add screenshots to help explain your problem. N/A

**Wayfire version**
git 30f13ab31cad83b9e3773c26cd64fa885e7a0d9b + bug fixes (#1155 and e1f8f78a49108e4a820f1b5735f2284796239a03 reverted)

dkondor:
Fixes #1186 (only the case with the move plugin)

Tested with regular move, drag with expo active and setting `workarounds/all_dialogs_modal` both on and off. Did not test with multiple outputs. Does not work if `move/join_views = true`, in that case, it seems that the clicked view is "forgotten" when the drag starts.

### DIFF
--- a/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
+++ b/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
@@ -660,6 +660,9 @@ inline void adjust_view_on_output(drag_done_signal *ev)
     target_ws.x = wf::clamp(target_ws.x, 0, gsize.width - 1);
     target_ws.y = wf::clamp(target_ws.y, 0, gsize.height - 1);
 
+    // view to focus at the end of drag
+    auto focus_view = ev->main_view;
+
     for (auto& v : ev->all_views)
     {
         if (!v.view->is_mapped())
@@ -684,6 +687,12 @@ inline void adjust_view_on_output(drag_done_signal *ev)
         {
             v.view->tile_request(v.view->tiled_edges, target_ws);
         }
+
+        // check focus timestamp and select the last focused view to (re)focus
+        if (v.view->last_focus_timestamp > focus_view->last_focus_timestamp)
+        {
+            focus_view = v.view;
+        }
     }
 
     // Ensure that every view is visible on parent's main workspace
@@ -692,7 +701,7 @@ inline void adjust_view_on_output(drag_done_signal *ev)
         ev->focused_output->workspace->move_to_workspace(v, target_ws);
     }
 
-    ev->focused_output->focus_view(parent, true);
+    ev->focused_output->focus_view(focus_view, true);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


dkondor:
> > Does not work if `move/join_views = true`, in that case, it seems that the clicked view is "forgotten" when the drag starts.
> 
> I'm wondering whether we could use the last focused timestamp to find out the last focused dragged view in this case?

Yes, I think that should work; I'll experiment with it and update the PR

dkondor:
Added logic for selecting the most recent view to focus. Now it works well for me even with `move/join_views = true`

ammen99:
After some thinking I think we need to update the clicked-on view's time stamp in `move.cpp`. Consider the following scenario:

- Main view X
- X has two dialogs A and B
- Dialog A is focused
- Click and drag B (with `all_dialogs_modal = false`)
- After releasing, A will be refocused because it had the latest timestamp

So basically, we should be updating the time stamp before/after https://github.com/WayfireWM/wayfire/blob/master/plugins/single_plugins/move.cpp#L296

dkondor:
I understand what you're saying, but cannot reproduce the issue. I think that the wayfire_focus handler already takes care of updating the timestamp before the drag starts (focus_view is called from [here](https://github.com/WayfireWM/wayfire/blob/9541570c29965581612c8ff69b66196107d70ad2/src/core/wm.cpp#L174)). Is there a case where this would not be true?

ammen99:
> I understand what you're saying, but cannot reproduce the issue. I think that the wayfire_focus handler already takes care of updating the timestamp before the drag starts (focus_view is called from [here](https://github.com/WayfireWM/wayfire/blob/9541570c29965581612c8ff69b66196107d70ad2/src/core/wm.cpp#L174)). Is there a case where this would not be true?

Did you try starting the drag with the move binding and not by dragging the titlebar?

dkondor:
Yes, I did. But I have `core/focus_button_with_modifiers = true`. Now I tried setting it to false, and I can see the issue. I'll add the code to update the timestamp where you suggest and test more.

dkondor:
Added the extra code to update the focus to the grabbed view. Will test more, but it seems to be working in all combinations I can think of :)

ammen99:
LGTM now, thanks!

PR #1194, opened 2021-05-21
core-impl: Fix ODR violation warnings with b_lto

Javyre:
compiling with link-time optimizations turned on generated One
Definition Rule violation warnings due to input_method_relay being
referred to in the global namespace and being defined in wf::

This line was probably placed outside of wf:: by mistake.


ammen99 (approved):
Yes, that is a mistake, thanks!

PR #1205, opened 2021-06-03
Unbreak build with wlroots 0.14.0 (2021-06-03 snapshot)

jbeich:
Can be worked around via `-Duse_system_wlroots=disabled`.


ammen99:
Thanks!

PR #1207, opened 2021-06-04
surface: Return owned surface on remove_subsurface

Javyre:
As discussed here:
https://github.com/Javyre/swayfire/pull/24#issuecomment-854393828

This isn't too useful yet as the subsurfaces still get dropped on `set_decoration()`. I feel like `set_decoration()` should either not `remove_subsurface` at all or at least return the pointer as well. (although that would be messier since there's less guarantee you're getting back your own surface form the `set_decoration(nullptr)`)

### DIFF
--- a/src/view/surface.cpp
+++ b/src/view/surface.cpp
@@ -35,14 +35,22 @@ void wf::surface_interface_t::add_subsurface(
     this->emit_signal("subsurface-added", &ev);
 }
 
-void wf::surface_interface_t::remove_subsurface(
+std::unique_ptr<wf::surface_interface_t> wf::surface_interface_t::remove_subsurface(
     nonstd::observer_ptr<surface_interface_t> subsurface)
 {
     auto remove_from = [=] (auto& container)
     {
-        auto it = std::remove_if(container.begin(), container.end(),
+        auto it = std::find_if(container.begin(), container.end(),
             [=] (const auto& ptr) { return ptr.get() == subsurface.get(); });
-        container.erase(it, container.end());
+
+        std::unique_ptr<surface_interface_t> ret = nullptr;
+        if (it != container.end())
+        {
+            ret = std::move(*it);
+            container.erase(it);
+        }
+
+        return ret;
     };
 
     wf::subsurface_removed_signal ev;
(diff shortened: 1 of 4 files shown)


ammen99:
> I feel like `set_decoration()` should either not `remove_subsurface` at all or at least return the pointer as well. (although that would be messier since there's less guarantee you're getting back your own surface form the `set_decoration(nullptr)`)

I agree, the best solution is to completely separate the decorator and the decoration subsurface. Former is added/removed by `set_decoration`, and does not prevent scanout, while the decoration subsurface(s) are managed via `add/remove_subsurface()`. PRs of course welcome :) Otherwise, open an issue so this gets eventually done.

src/view/surface.cpp
```diff
@@ -35,23 +35,36 @@ void wf::surface_interface_t::add_subsurface(
...
+        auto it = std::find_if(container.begin(), container.end(),
             [=] (const auto& ptr) { return ptr.get() == subsurface.get(); });
+
+        std::unique_ptr<surface_interface_t> ret = nullptr;
+        if (it != container.end())
+        {
+            ret = std::move(*it);
+        }
+
         container.erase(it, container.end());
```
ammen99, line 52: This would delete all subsurfaces until the end of the vector, right? We definitely do not want this.

ammen99 (changes requested):

src/view/surface.cpp
Javyre: woops sorry for that. Fixed now.

Javyre:
> I agree, the best solution is to completely separate the decorator and the decoration subsurface. Former is added/removed by set_decoration, and does not prevent scanout, while the decoration subsurface(s) are managed via add/remove_subsurface(). PRs of course welcome :) Otherwise, open an issue so this gets eventually done.

I'll look into this tomorrow. Would the decorator and decoration surface sill be the same object? What do you mean by prevent scanouts?

ammen99:
> I'll look into this tomorrow.

Great, thanks!

> Would the decorator and decoration surface sill be the same object?

No, they wouldn't be, this is the whole point of the change :)

> What do you mean by prevent scanouts?

Huh, I am stupid, sorry. Decorations do not prevent scanouts in fullscreen because I have written a workaround for that, which I had completely forgotten up to now - https://github.com/WayfireWM/wayfire/blob/master/src/output/render-manager.cpp#L844

ammen99 (approved):
Thanks!

PR #1209, opened 2021-06-06
output-layout: do not check for noop's existence when ensuring it exists

unrelentingtech:
`ensure_noop_output` truly is an "ensure" function, not a "create" one. Among other things, it cancels the `timer_remove_noop`, which always needs to be done.

Otherwise, rapid changes in output configuration (that do happen for me currently with turning my monitor back on after turning it off) cause a situation where the `timer_remove_noop` fires right after removing the last output, causing the noop to be removed too, which results in a crash.


ammen99 (approved):
Nice catch, thanks!

PR #1210, opened 2021-06-06
WIP: deco: Separate decoration surface from frame.

Javyre:
This doesn't work yet as I'm still working on it.

I'm not sure how you want to split the deco plugin's `simple_decoration_surface`. Do you want to store the surface state in the surface itself or in the decoration frame? The new `simple_decoration_frame` could also just be empty and forward all the events to the surface.. I'm not sure how you want to approach this.

### DIFF
--- a/plugins/decor/deco-subsurface.cpp
+++ b/plugins/decor/deco-subsurface.cpp
@@ -20,14 +20,12 @@
 #include <cairo.h>
 
 class simple_decoration_surface : public wf::surface_interface_t,
-    public wf::compositor_surface_t, public wf::decorator_frame_t_t
+    public wf::compositor_surface_t
 {
     bool _mapped = true;
-    int current_thickness;
-    int current_titlebar;
 
     wayfire_view view;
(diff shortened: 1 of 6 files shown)


ammen99:
Thanks! The core changes look good, I'll try to do the decorarion part you mentioned myself tomorrow.

Javyre:
alright, so should I just remove the deco plugin changes from this pr then? Or you can add them in a commit to this same pr if that possible?

ammen99:
> alright, so should I just remove the deco plugin changes from this pr then? Or you can add them in a commit to this same pr if that possible?

Yes, I can push to your branch, don't bother removing the changes, they might turn out useful to me too ;)

Javyre:
Oh alright then perfect! :)

I just force-pushed the formatted code so the commit is somewhat clean at least. I'll leave it to you now then I guess.

ammen99 (approved):
Thanks :)

P.S. I also updated the Wayfire plugin API. I actually had a strange crash that I couldn't figure out, turned out this was because of a plugin I had not recompiled ... so yeah we should update API version properly.

PR #1214, opened 2021-06-15
Improvements to decoration plugin

### LINKED ISSUE #1066: Improve default decorations
Many users have complained that the default SSDs are ugly, and that may very well be true :)

Unfortunately, I'm not so much into GUI design so I don't think I have that many ideas how to make them better. Thus, help wanted to fix this issue.

marcusbritanicus:
A picture speaks a thousand words.
snapshot
Foot terminal with new and improved decoration plugin buttons. 
Note the close button lights up when mouse enters the button.

Fixes #1066


ammen99:
Just tested this, looks amazing, a clear improvement! Thanks! I even believe this is enough to fix #1066, IMO default decorations are no longer ugly :)

marcusbritanicus:
Thanks @ammen99 :)

PR #1217, opened 2021-06-16
Transactions API

ammen99:
Work-in-progress PR, probably won't be ready for testing in the next few weeks.

Plan:

- [ ] Add basic transactions infrastructure for setting maximized/fullscreen states
- [ ] Add window size
- [ ] Add setting output via transactions
- [ ] Use transactions for switching workspaces
- [ ] Add moving/resizing state to views
- [ ] Make grid animate any geometry change aside for moving/resizing views
- [ ] Make simple tile use transactions instead of enforcing geometry
- [ ] Make transaction timeouts configurable


ammen99:
Closing in favor of #1704

Issue #1218, opened 2021-06-16
Invert plugin vanishes my mouse pointer


marcusbritanicus:
**Describe the bug**
Switching on the color inversion using the invert plugin makes mouse pointer disappear. Switching it off, brings back the pointer.
If it matters (I don't think it should), I'm using [bloom](https://github.com/linuxdeepin/deepin-icon-theme) cursors.

**To Reproduce**
1. Enable plugin invert
2. Invoke the invert plugin
3. Mouse pointer disappears

**Expected behavior**
The mouse pointer should be visible, and should be rendered inverted too.

**Screenshots or stacktrace**
![2021-06-16_14:33:43](https://user-images.githubusercontent.com/3350667/122191238-2e393880-ceb0-11eb-8fec-3e1222a07ff5.png)

**Wayfire version**
git

soreau:
I can't reproduce this. I did install bloom icon theme for xwayland and gtk apps, didn't seem to make a difference (though it is a nice cursor theme). Here, invert works without hiding the cursor at all.

ammen99:
@marcusbritanicus On irc you mentioned another cursor related bug oover translucent windows ... Maybe this is really a bug with your drivers? Cursors generally use hardware planes so they are special. You should try the damage debug mode (see --help) to see whether cursor causes damage or not.

marcusbritanicus:
@ammen99 The other issue is not a cursor related issue. easiest to see with cursors. But I will investigate further. My gpu driver is radeon. And now that I think about it, this invert problem has come up recently. I remember seeing the properly inverted cursor long ago.

I'll also check on the other AMD system I have. That uses amdgpu driver. Let's see if this is AMD specific issue.

marcusbritanicus:
I just tested the invert plugin on my amdgpu PC. The mouse cursor does not get inverted, but it does not disappear.
Given the fact that it used to work properly long ago with my radeon driver, and that it works with amdgpu driver, I think the problem might be my gpu/gpu driver (R4/radeon)

soreau:
FWIW, this was fixed with #1221

PR #1221, opened 2021-06-17
render-manager: Render software cursors last

soreau:
This is most consistent with hardware cursor plane behavior.
Without this, there are bugs such as #1218.


ammen99 (approved):
Thanks, good catch!

PR #1229, opened 2021-07-03
workspace-impl: Ensure output is set on add_view_*

Javyre:
It doesn't make sense for views to be added to a workspace without also
being on that output.

I think we should also assert that `damage()` isn't being called when `!get_output()`, but I don't know exactly where best to assert this...

### DIFF
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -1101,6 +1101,7 @@ class workspace_manager::impl
 
     void add_view_to_layer(wayfire_view view, layer_t layer)
     {
+        assert(view->get_output() == output);
         bool first_add = layer_manager.get_view_layer(view) == 0;
         layer_manager.add_view_to_layer(view, layer);
         update_promoted_views();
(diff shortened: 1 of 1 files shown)


Javyre:
Ok so apparently asserting is a bad idea. :( Too many things break and it seems that damage being a noop if output is null is an assumption that is used a lot in wayfire.

ammen99:
> Ok so apparently asserting is a bad idea. :( Too many things break and it seems that damage being a noop if output is null is an assumption that is used a lot in wayfire.

Yes, this makes things much simpler. Otherwise, we simply need a ton of checks all over the place.

ammen99:
I'd also rather assert that the view is on the current output already.

Javyre:
> I'd also rather assert that the view is on the current output already.

Yeah that works too. Done

src/output/workspace-impl.cpp
```diff
@@ -1253,7 +1255,7 @@ void workspace_manager::destroy_sublayer(nonstd::observer_ptr<sublayer_t> sublay
 void workspace_manager::add_view_to_sublayer(wayfire_view view,
     nonstd::observer_ptr<sublayer_t> sublayer)
 {
-    return pimpl->layer_manager.add_view_to_sublayer(view, sublayer);
+    return pimpl->add_view_to_sublayer(view, sublayer);
```
ammen99, line 1258: ha, I wonder how this has survived so long without being noticed :) nice catch!

ammen99 (approved):
Thanks!

PR #1231, opened 2021-07-03
Bump dep versions

soreau:
Bump wlroots and wf-config maximum required versions.


ammen99 (changes requested):
I think we should bump the min version too.. At some point, we'll have breaking changes in wlroots, and it would be very easy to forget bumping the version here.

ammen99 (approved):
LGTM now, thanks!

PR #1236, opened 2021-07-07
wrot: stop interactive rotate if the target view is unmapped

### LINKED ISSUE #1234: Crash when a window closes while in the process of being rotated.
**Describe the bug**
wayfire crashes if a window that's in the process of being rotated (`super+shift+rmb`) gets closed (by the application or otherwise)

**To Reproduce**
1. Enable plugin wrot (enabled by default)
2. Open a terminal window and execute `sleep 3; exit` (to simulate the window-closes-while-being-rotated behavior)
3. Rotate the terminal using `super shift right-mouse`. do *not* release!!
4. Wait for the `sleep` to finish

**Expected behavior**
The window should close normally.

**Wayfire version**
`0.7.2` from the AUR

dkondor:
Fixes #1234

### DIFF
--- a/plugins/single_plugins/wrot.cpp
+++ b/plugins/single_plugins/wrot.cpp
@@ -4,6 +4,7 @@
 #include "wayfire/output.hpp"
 #include "wayfire/core.hpp"
 #include <wayfire/workspace-manager.hpp>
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Thanks!

PR #1237, opened 2021-07-09
workspace-impl: preserve last-focused order

Javyre:
Preserve last-focused order on workspace changed.

This solved the following bug that I described on IRC:
1. open view # 1 and view # 2
2. click on view # 2 to bring it to the front
3. move mouse over view # 1 to give it keyboard focus (with follow-focus plugin and raise view option false)
4. use viewport switcher or expo to go to another workspace
5. switch back to the original workspace and the view # 2 on top is given keyboard focus instead of view # 1

### DIFF
--- a/src/api/wayfire/output.hpp
+++ b/src/api/wayfire/output.hpp
@@ -189,15 +189,15 @@ class output_t : public wf::object_base_t
     bool ensure_visible(wayfire_view view);
 
     /**
-     * Force refocus the topmost view in one of the layers marked in layers
-     * and which isn't skip_view.
+     * Force refocus the most recently focused view in one of the layers marked
+     * in layers and which isn't skip_view.
      *
      * The stacking order is not changed.
      */
     virtual void refocus(wayfire_view skip_view, uint32_t layers) = 0;
 
     /**
-     * Refocus the topmost focuseable view != skip_view, preferring regular
+     * Refocus the most recently focused view != skip_view, preferring regular
      * views. The stacking order is not changed.
      */
     void refocus(wayfire_view skip_view = nullptr);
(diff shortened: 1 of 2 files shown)


ammen99 (commented):
I think this will break the exact case which the existing code is trying to fix:

- Open view A on workspace 1, move it so that it is visible on workspace 2 too.
- Open view B on workspace 1
- Go to workspace 2
- Go to workspace 1

Expected: view A is focused
Actually: view B is focused, because its timestamp was updated when switching to workspace 2

Javyre:
That seems to be an instance of the issue that this pr doesn't solve. When a view is in a single workspace, however, this does solve the issue.

A better way to solve for all cases would be to have per-workspace view focus timestamps as discussed on IRC. Each view would hold a timestamp for each workspace in the grid. This way a view that is either sticky or is visible on multiple workspaces won't interfere with the refocus on the other workspaces it is on.

about your concern about sticky views (not sure if you saw my msg on IRC):
> 2) per-workspace timestamps would be nice, probably, but I'm not sure that this is the behavior we want in all cases. Like if the sticky view is focused on workspace A, then you go to workspace B, why should it lose focus all of a sudden? Sounds counterintuitive to me.

That can be worked around by having sticky views sync the timestamp across all the workspaces. So when the timestamp of a sticky view is set on workspace A, it actually sets all of the view's timestamps.

soreau:
> Preserve last-focused order on workspace changed.
> 
> This solved the following bug that I described on IRC:
> 
>     1. open view # 1 and view # 2
> 
>     2. click on view # 2 to bring it to the front
> 
>     3. move mouse over view # 1 to give it keyboard focus (with follow-focus plugin and raise view option false)


In my opinion, this is a bug in follow-focus plugin and should be fixed there. It can listen for workspace switch and call it's own refocus routine.

PR #1239, opened 2021-07-09
build: Add -fno-gnu-unique compiler option

soreau:
dlclose POSIX docs say "Although a dlclose() operation is not required
to remove structures from an address space, neither is an implementation
prohibited from doing so."[1]. The fno-gnu-unique option makes it so the
plugin is unloaded when dlclose is called[2]. This makes wayfire use less
resident memory when unloading plugins. It also allows developers to make
changes to wayfire plugins without restarting wayfire.

1. The Open Group Base Specifications Issue 6.
http://pubs.opengroup.org/onlinepubs/007904975/functions/dlclose.html
2. GCC Code Gen Options.
https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -87,13 +87,20 @@ else
   conf_data.set('DEFAULT_CONFIG_BACKEND', get_option('default_config_backend'))
 endif
 
+cpp = meson.get_compiler('cpp')
+
 # needed to dlopen() plugins
 # -E is for RTTI/dynamic_cast across plugins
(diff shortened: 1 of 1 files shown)


ericonr:
For the record, `dlopen` on musl is a no-op always. This might fix the "multiple dlopen calls" case, though.

Idk about FreeBSD or others.

ericonr:
Mail thread with the discussion on musl: https://inbox.vuxu.org/musl/20120820004803.GA27715@brightrain.aerifal.cx/

ammen99:
Thanks :)

PR #1245, opened 2021-07-13
output: preserve last-focused order (take 2)

Javyre:
Preserve last-focused order on workspace changed.

This is an alternative, simpler solution for the issue described in #1237. The solution is to simply not meddle with the focus timestamps on `set_workspace()`.

This should also cover the cases described in the previous attempt where if a view is sticky or spans multiple workspaces it could lead to unintuitive behaviour.

### DIFF
--- a/src/api/wayfire/output.hpp
+++ b/src/api/wayfire/output.hpp
@@ -189,15 +189,15 @@ class output_t : public wf::object_base_t
     bool ensure_visible(wayfire_view view);
 
     /**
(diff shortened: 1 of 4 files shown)


soreau:
I can confirm this fixes the bug as described in #1237 and also solves the case mentioned by @ammen99 in the first comment of #1237.

ammen99 (approved):
Alright, looks like it should work. Thanks :)

Issue #1227, opened 2021-06-30
nonstd::noncopyable_t makes objects also non-movable


Javyre:
**Describe the bug**
Any object deriving from `nonstd::noncopyable_t` is implicitly forced to be unmovable because the deleted copy-constructor implicitly deletes the default move constructor. This means that I cannot make for example a `std::vector<wf::simple_texture_t>` because `wf::simple_texture_t` is not move-constructible.

**To Reproduce**
See this playground: https://cplayground.com/?p=cormorant-chinchilla-walrus

**Expected behavior**
For `nonstd::noncopyable_t` not to make objects also non-movable.
See the above playground for a possible fix I found.

**Wayfire version**
at least 0.7.0 to git

If you think my solution is good enough, I can submit a PR to convert uses of `noncopyable_t` to use the fix. Another possible solution is to not use this parent class at all and manually make things non-copyable and still movable with a bit more boilerplate.

Javyre:
Here is a better code example that explains the problem when destructors are present (as is the case with `simple_texture`)
https://cplayground.com/?p=crab-gerbil-eel

Javyre:
A simpler explanation is that wayfire isn't following the rule of 5:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-special-member-functions.html

To check for it automatically you can for example run:
```sh
cd build
clang-tidy ../plugins/common/wayfire/plugins/common/* -checks=-*,cppcoreguidelines-special-member-functions
```

(Ideally you'd run this on all files in the project but that takes some meson setup. I set it up in swayfire even though I'm not really using it)

PR #1246, opened 2021-07-15
Add clang-tidy to CI

Javyre:
Towards fixing: #1227

I added it under the clang build task since clang-tidy mainly only recognizes clang compile args.


Javyre:
Apparently, meson's built-in clang-tidy target is hardcoded to check subprojects... https://github.com/mesonbuild/meson/issues/7200

Javyre:
Some new warnings appear now on clang because of implicitly deleted explicitly defined move constructors. I left certain instances of these issues since I think the real problem is that `option_wrapper_t` and `option_wrapper_t` are non-movable and they should be movable.

src/api/wayfire/util.hpp
```diff
@@ -148,6 +153,11 @@ class wl_idle_call : public noncopyable_t
     /** Will disconnect if connected */
     ~wl_idle_call();
 
+    wl_idle_call(const wl_idle_call &) = delete;
+    wl_idle_call(wl_idle_call &&) = default;
```
ammen99, line 157: Nope, this one isn't trivially movable either. The wayland event loop holds a pointer to this object. Same applies to wl_timer.

src/core/matcher.cpp
```diff
@@ -71,10 +71,16 @@ class wf::view_matcher_t::impl
...
     }
 
+    impl() = default;
     ~impl()
     {
         disconnect_updated_handler();
     }
+
+    impl(const impl &) = default;
+    impl(impl &&) = default;
```
ammen99, line 81: Is here the default that the constructor is deleted? If yes, i'd rather have an explicit `=delete`

plugins/animate/animate.cpp
```diff
@@ -128,6 +134,11 @@ struct animation_hook : public animation_hook_base
             view->unref();
         }
     }
+
+    animation_hook(const animation_hook &) = default;
```
ammen99, line 138: Copy is deleted anyway because of unique_ptr, let's mark it like that. Same applies to all other places in this PR.

ammen99 (changes requested):

src/core/matcher.cpp
Javyre: No this default isn't "implicitly deleted" by anything.

Javyre:
Ok so the last few warnings related to this pr are either solved by keeping option_wrapper_t nonmovable and marking objects containing option_wrapper_ts as nonmovable or by merging https://github.com/WayfireWM/wf-config/pull/50

plugins/common/wayfire/plugins/common/cairo-util.hpp
```diff
@@ -221,11 +221,18 @@ struct cairo_text_t
...
 
+    cairo_text_t() = default;
+
     ~cairo_text_t()
     {
         cairo_free();
     }
 
+    cairo_text_t(const cairo_text_t &) = delete;
+    cairo_text_t(cairo_text_t &&) = default;
```
dkondor, line 232: I believe the move constructor and assignment should be deleted as well or implemented in a way that invalidates the members `cr` and `surface`, since they are raw pointers to object allocated by the Cairo C API (see the destructor below).

dkondor: (the same probably applied to `simple_texture_t` as well, where the destructor frees up the `tex` member)

Javyre:
I gave up on making a lot of these classes movable since there are a lot of pointers being held everywhere and it would be way too much work for this pr. It already took me this long to triple-check all the new constructors for correctness, I'm starting to lose energy for this pr.

I do think that an eventual refactor of `object` could be interesting. It would be nice to have objects and signal-connections be movable. Just.. not now..

src/api/wayfire/object.hpp
```diff
@@ -55,10 +58,13 @@ class signal_connection_t : public noncopyable_t
     void disconnect();
 
     class impl;
-    std::unique_ptr<impl> priv;
+    std::shared_ptr<impl> priv;
```
ammen99, line 61: This should remain unique_ptr .

ammen99 (changes requested):
LGTM aside from the comment.

ammen99:
> I gave up on making a lot of these classes movable since there are a lot of pointers being held everywhere and it would be way too much work for this pr. It already took me this long to triple-check all the new constructors for correctness, I'm starting to lose energy for this pr.

It is a lot of boring boilerplate :/

> I do think that an eventual refactor of `object` could be interesting. It would be nice to have objects and signal-connections be movable. Just.. not now..

Making objects movable is mostly useless, because right now we're utilizing polymorphism heavily which means that objects are all accessed via pointers .. moving such objects rarely makes a lot of sense.

src/api/wayfire/object.hpp
Javyre: woops yeah that's leftover from my movable attempt. Basically unique_ptr still needs to know the sizeof() when destroying the object in some cases, but shared_ptr doesn't. Anyways, I'll remove this now.

Javyre:
> Making objects movable is mostly useless, because right now we're utilizing polymorphism heavily which means that objects are all accessed via pointers .. moving such objects rarely makes a lot of sense.

The problem is that not all object_ts are created/managed by wayfire. In swayfire for example, every tree node is an object_t. Having the object_t class non-movable is intrusive in that it forces polymorphism through indirection and makes things like value types with std::variant impossible. Only time so far this has been an issue I actually ran into was with simple_texture_t though.

ammen99 (approved):
Thanks!

PR #1251, opened 2021-07-29
Add settings to fire effects

yonatan8070:
This PR adds 2 settings, `fire_color` (color) and `random_fire_color` (boolean).

If `random_fire_color` is false, the fire will be colored according to `fire_color`. If it is true then the color of each particle will be randomized.

### DIFF
--- a/plugins/animate/fire/fire.cpp
+++ b/plugins/animate/fire/fire.cpp
@@ -116,17 +116,6 @@ class FireTransformer : public wf::view_transformer_t
             g = random(0, 1);
             b = random(0, 1);
 
-            // Squrae the values several times to make them less of a blinding light and more of a spray of colors
-            // for (int i = 0; i < 4; i++) {
-            //     r *= r;
-            //     g *= g;
-            //     b *= b;
-            // }
-
-            // r *= 2;
-            // g *= 2;
-            // b *= 2;
-
             r = 2 * pow(r, 16);
             g = 2 * pow(g, 16);
             b = 2 * pow(b, 16);
--- a/plugins/animate/fire/fire.cpp
+++ b/plugins/animate/fire/fire.cpp
@@ -91,27 +91,29 @@ class FireTransformer : public wf::view_transformer_t
         p.fade = random(0.1, 0.6);
 
         wf::color_t color_setting = fire_color;
-        
+
         float r;
         float g;
         float b;
 
-        if (!random_fire_color) {
+        if (!random_fire_color)
+        {
             // The calculation here makes the variation lower at darker values
             float randomize_amount_r = (color_setting.r * 0.857) / 2;
             float randomize_amount_g = (color_setting.g * 0.857) / 2;
             float randomize_amount_b = (color_setting.b * 0.857) / 2;
 
-            r = random(color_setting.r - randomize_amount_r, 
-                       std::min(color_setting.r + randomize_amount_r, 
-                       1.0));
-            g = random(color_setting.g - randomize_amount_g, 
-                       std::min(color_setting.g + randomize_amount_g, 
-                       1.0));
-            b = random(color_setting.b - randomize_amount_b, 
-                       std::min(color_setting.b + randomize_amount_b, 
-                       1.0));
-        } else {
+            r = random(color_setting.r - randomize_amount_r,
+                std::min(color_setting.r + randomize_amount_r,
+                    1.0));
+            g = random(color_setting.g - randomize_amount_g,
+                std::min(color_setting.g + randomize_amount_g,
+                    1.0));
+            b = random(color_setting.b - randomize_amount_b,
+                std::min(color_setting.b + randomize_amount_b,
+                    1.0));
+        } else
+        {
             r = random(0, 1);
             g = random(0, 1);
             b = random(0, 1);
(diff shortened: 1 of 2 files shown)


plugins/animate/fire/fire.cpp
```diff
@@ -88,7 +90,46 @@ class FireTransformer : public wf::view_transformer_t
...
+            b = random(color_setting.b - randomize_amount_b, 
+                       std::min(color_setting.b + randomize_amount_b, 
+                       1.0));
+        } else {
+            r = random(0, 1);
+            g = random(0, 1);
+            b = random(0, 1);
+
+            // Squrae the values several times to make them less of a blinding light and more of a spray of colors
+            for (int i = 0; i < 4; i++) {
```
ammen99, line 121: this seems more like `r = 2 * pow(r, 16)` which is much shorter and clearer imo.

plugins/animate/fire/fire.cpp
```diff
@@ -88,7 +90,46 @@ class FireTransformer : public wf::view_transformer_t
         p.life = 1;
         p.fade = random(0.1, 0.6);
 
-        p.color = {random(0.4, 1), random(0.08, 0.2), random(0.008, 0.018), 1};
+        wf::color_t color_setting = fire_color;
+        bool random_colors = random_fire_color;
```
ammen99, line 94: why can't you use `random_fire_color` directly in the if-statement below?

ammen99 (changes requested):
Randomly colored particles don't look especially good (though that's a matter of taste), but I do like cyan flames ;)

A few small things, otherwise LGTM.

plugins/animate/fire/fire.cpp
```diff
@@ -88,7 +90,46 @@ class FireTransformer : public wf::view_transformer_t
...
+        
+        float r;
+        float g;
+        float b;
+
+        if (!random_colors) {
+            // The calculation here makes the variation lower at darker values
+            float randomize_amount_r = (color_setting.r * 0.857) / 2;
```
ammen99, line 102: By the way, where does 0.857 come from?

plugins/animate/fire/fire.cpp
yonatan8070: It can be moved there, no real reason for doing it this way. I'll change it in a sec

plugins/animate/fire/fire.cpp
yonatan8070: 0.857 comes from the original values. I checked what the average for each of the original ranges is, along with how large the difference between them was, and the ratio between them came out to this.

plugins/animate/fire/fire.cpp
yonatan8070: Fixed it now

plugins/animate/fire/fire.cpp
ammen99: Ah, this is why the difference between the old and the new algorithm is negligible .. thanks for putting in the effort!

plugins/animate/fire/fire.cpp
```diff
@@ -88,7 +90,49 @@ class FireTransformer : public wf::view_transformer_t
...
+                       1.0));
+            b = random(color_setting.b - randomize_amount_b, 
+                       std::min(color_setting.b + randomize_amount_b, 
+                       1.0));
+        } else {
+            r = random(0, 1);
+            g = random(0, 1);
+            b = random(0, 1);
+
+            // Squrae the values several times to make them less of a blinding light and more of a spray of colors
```
ammen99, line 119: we don't need these comments ..

yonatan8070: Sorry, forgot to commit the removal

ammen99:
I forgot to allow the CI checks, as you can see, the code is not properly formatted, https://github.com/WayfireWM/wayfire/blob/master/CONTRIBUTING.md explains how to do it with uncrustify.

yonatan8070:
I just ran the formatter, it changed a whole lot of files I didn't modify, should I commit those as well?

ammen99:
> I just ran the formatter, it changed a whole lot of files I didn't modify, should I commit those as well?

Huh, did you run it with the .ini file from the repo? Everything else has already been formatted correctly.

If yes, then can you attach a diff of the changes that uncrustify proposes? I can't imagine why it behaves like that.

yonatan8070:
It seems like it only changed the line wrapping in block comments

ammen99:
Which version of uncrustify are you using? Here it is `Uncrustify_d-0.71.0-48-968c2132`

yonatan8070:
I have Uncrustify_d-0.73.0_f

ammen99:
Well anyway, here's a diff which should make uncrustify happy. http://ix.io/3uoA

yonatan8070:
Fixed it, btw is that a rickroll in the diff's favicon?

plugins/animate/fire/fire.cpp
yonatan8070: What do you mean by old and new algorithm?

ammen99:
> Fixed it, btw is that a rickroll in the diff's favicon?

I don't know ;)

yonatan8070:
It's saying "Changes requested" on my end, but I can't seem to find what changes are requested. What am I missing?

ammen99 (approved):

ammen99:
Thanks!

PR #1255, opened 2021-07-31
Implement reverse animations

soreau:
Requires wf-config #51.

### DIFF
--- a/plugins/animate/animate.cpp
+++ b/plugins/animate/animate.cpp
@@ -18,19 +18,36 @@ bool animation_base::step()
     return false;
 }
 
+void animation_base::reverse()
+{}
+
+int animation_base::get_direction()
+{
+    return 1;
+}
+
 animation_base::~animation_base()
 {}
 
-static constexpr const char *animate_custom_data_id = "animation-hook";
+static constexpr const char *animate_custom_data_fire     = "animation-hook-fire";
+static constexpr const char *animate_custom_data_zoom     = "animation-hook-zoom";
+static constexpr const char *animate_custom_data_fade     = "animation-hook-fade";
+static constexpr const char *animate_custom_data_minimize =
+    "animation-hook-minimize";
+
+static constexpr int HIDDEN = 0;
+static constexpr int SHOWN  = 1;
 
 /* Represents an animation running for a specific view
  * animation_t is which animation to use (i.e fire, zoom, etc). */
 struct animation_hook_base : public wf::custom_data_t
 {
     virtual void stop_hook(bool) = 0;
+    virtual void reverse(wf_animation_type) = 0;
+    virtual int get_direction() = 0;
+
     animation_hook_base() = default;
     virtual ~animation_hook_base() = default;
-
     animation_hook_base(const animation_hook_base &) = default;
     animation_hook_base(animation_hook_base &&) = default;
     animation_hook_base& operator =(const animation_hook_base&) = default;
@@ -45,6 +62,7 @@ struct animation_hook : public animation_hook_base
 
     wf_animation_type type;
     wayfire_view view;
+    std::string name;
     wf::output_t *current_output = nullptr;
     std::unique_ptr<animation_base> animation;
 
@@ -85,15 +103,12 @@ struct animation_hook : public animation_hook_base
         { set_output(view->get_output()); }
     };
 
-    animation_hook(wayfire_view view, int duration, wf_animation_type type)
+    animation_hook(wayfire_view view, int duration, wf_animation_type type,
+        std::string name)
     {
         this->type = type;
         this->view = view;
-        if (type == ANIMATION_TYPE_UNMAP)
-        {
-            view->take_ref();
-            view->take_snapshot();
-        }
+        this->name = name;
 
         animation = std::make_unique<animation_t>();
         animation->init(view, duration, type);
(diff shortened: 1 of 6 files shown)


plugins/animate/animate.cpp
```diff
@@ -247,6 +255,19 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
...
+        if (type == ANIMATION_TYPE_UNMAP)
+        {
+            view->take_ref();
+            view->take_snapshot();
+        }
+
+        if (view->has_data(animate_custom_data_id))
+        {
+            view->get_data<animation_hook_base>(
+                animate_custom_data_id)->reverse(type);
```
ammen99, line 267: What happens if these are different animations, for ex. we had unminimize animation running, then the view is unmapped and it should start a fire animation?

soreau: You mean something like this?


https://user-images.githubusercontent.com/1450125/127919341-dcaebd7e-c042-4139-b539-0cb2df35301b.mp4

ammen99: Yep, this is how it should be.

plugins/animate/animate.cpp
```diff
@@ -243,13 +292,78 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
         return {"none", 0};
     }
 
+    bool reversed(wayfire_view view, wf_animation_type type, std::string name,
```
ammen99, line 295: IMO it would be a bit clearer if the function was called `try_reverse` and the direction which is given is the desired direction (currently, it is the 'wrong' direction)

plugins/animate/animate.cpp
```diff
@@ -243,13 +292,78 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
...
-            animate_custom_data_id);
+        if (type == ANIMATION_TYPE_UNMAP)
+        {
+            view->take_ref();
+            view->take_snapshot();
+        }
+
+        if (type == ANIMATION_TYPE_MAP)
+        {
+            if (reversed(view, type, animate_custom_data_fire, 0) ||
```
ammen99, line 323: We should introduce constants for this, maybe

```
static constexpr int HIDDEN = 0;
...                  SHOWN  = 1;
```

plugins/animate/animate.cpp
```diff
@@ -18,19 +18,31 @@ bool animation_base::step()
...
+{}
+
+int animation_base::get_direction()
+{}
+
 animation_base::~animation_base()
 {}
 
-static constexpr const char *animate_custom_data_id = "animation-hook";
+static constexpr const char *animate_custom_data_fire     = "animation-hook-fire";
```
ammen99, line 30: Is there a reason why we need to differentiate the different types of animation? It seems like it should be enough to differentiate between custom data for map/unmap and for minimize/restore.

plugins/animate/animate.cpp
```diff
@@ -112,8 +122,29 @@ struct animation_hook : public animation_hook_base
             view->set_minimized(true);
         }
 
-        /* Will also delete this */
-        view->erase_data(animate_custom_data_id);
+        if (type & MAP_STATE_ANIMATION)
```
ammen99, line 125: Why do we need this differentiation, isn't the correct name passed to this struct already in the constructor?

ammen99 (changes requested):
The patch looks sensible, there are a few issues around code organizations but I think nothing which should require too much work :)

plugins/animate/animate.cpp
soreau: This seems like it belongs in an enum in wf-config but I can add some definitions for them in animate.

plugins/animate/animate.cpp
soreau: > Is there a reason why we need to differentiate the different types of animation? It seems like it should be enough to differentiate between custom data for map/unmap and for minimize/restore.

Yes, the reason is to support multiple simultaneous animations. Consider the following:
An application maps a surface but unmaps it before the open animation completes. If there are different animations set for open and close, it can only know that 'a close animation is requested' and without knowing which one is requested, it doesn't know whether to reverse the current animation or start a new one. If the open and close animations are always the same, then it wouldn't really matter because it would always reverse.

plugins/animate/animate.cpp
ammen99: Well the 0 does not necessarily indicate HIDDEN (in wf-config) so i'd say here's the place to put them.

plugins/animate/animate.cpp
ammen99: I see, thanks for the explanation.

plugins/animate/animate.cpp
```diff
@@ -243,13 +286,78 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
...
+            {
+                data->reverse(type);
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+#define FORWARD 1
```
ammen99, line 306: 1) please don't use macros when constexpr would suffice
2) nitpicking but: forward seems less clear than 'SHOWN', after all, the forward state is that the view is SHOWN

ammen99 (approved):
LGTM now, thanks!

PR #1265, opened 2021-08-22
Turkish language support added.

mamoo81:
Turkish language support added.

### DIFF
--- a/metadata/alpha.xml
+++ b/metadata/alpha.xml
@@ -2,16 +2,22 @@
 <wayfire>
 	<plugin name="alpha">
 		<_short>Alpha</_short>
+		<_short xml:lang="tr">Alfa</_short>>
 		<_long>A plugin to change the opacity of windows.</_long>
+		<_long xml:lang="tr">Pencerelerin opaklığını değiştirmek için bir eklenti.</_long>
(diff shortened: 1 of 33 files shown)


soreau:
If we dump translations into the main xml files, it will become unmaintainable in short time. Instead we should use something like [this](https://mesonbuild.com/i18n-module.html). The idea is to keep each language translations in separate files and have them joined into one big xml for installation at build time.

soreau:
[This patch](http://ix.io/3x2H) shows a basic translation setup. The .xml files are renamed to .xml.in and they are combined with the .po files for each language in the po/ directory. The po directory can be subprojects/po where po is WayfireWM/po. This way, we can isolate translation commits to their own repo. In this patch, the po directory is included.

There is only one translation I put in the `tr.po` file for testing. The next work would be populating this file with the translations found in this PR.

In addition, wcm needs to be patched to support choosing language by setting LANG. Currently, it always chooses the last _short or _long for the element. [Here](http://ix.io/3xeg) is a preliminary patch for wcm.

ammen99:
Seems quite stale, if someone wants to work on this, please use what @soreau suggested.

Issue #1269, opened 2021-08-24
Send view to back / bring forward


JordanL2:
I'm not sure if this functionality exists and I'm just failing to find it.

I'd like to bind hotkeys / mouse buttons to bringing a focused view to the top, or send to the back. If this doesn't exist yet it would probably be most at home in the `wm-actions` plugin.

travankor:
Maybe `always_on_top`?

JordanL2:
That pins a view to constantly being at the top, I just want to bring it to the top.

travankor:
`follow-focus` with `raise_on_top` is probably the closest right now to what you are describing. But this would require a new plugin, I think.

JordanL2:
And also (more importantly actually) send it to the back, so all other views are on top.

If this feature doesn't exist I can add it, if we agree where it should go.

ammen99:
This feature doesn't really fit into the main repo, sounds something quite specific to your workflow. I'd suggest implementing a custom plugin or adding it to wayfire-plugins-extra, if @soreau thinks it is a good fit for it.

For reference: https://github.com/WayfireWM/wayfire/wiki/Project-goals#goals

JordanL2:
Really? It's very common window manager functionality. Example, with Linux Mint you can middle click a window's titlebar to send to the back.

ammen99:
> This feature doesn't really fit into the main repo, sounds something quite specific to your workflow. I'd suggest implementing a custom plugin or adding it to wayfire-plugins-extra, if @soreau thinks it is a good fit for it.
> 
> For reference: https://github.com/WayfireWM/wayfire/wiki/Project-goals#goals

Well, I haven't used or seen it much, but that might be me. Do you know any DEs/window managers which have this?

JordanL2:
Cinnamon and Plasma off the top of my head. Just talking about send-to-back, the bring-to-front can be done with a click (although that sends a click to the app, but I can live with that).

ammen99:
> the bring-to-front can be done with a click (although that sends a click to the app, but I can live with that).

https://github.com/WayfireWM/wayfire/blob/master/metadata/core.xml#L66

JordanL2:
That's handy thanks.

JordanL2:
MATE, XFCE and Unity can do this too.

ammen99:
Alright, so it is more common than I thought. `wm-actions` would indeed be the place to add this.

JordanL2:
[...]

PR #1270, opened 2021-08-24
WM-Actions Send to back functionality

JordanL2:
For feature request #1269 

I also noticed a missing binding removal in fini(), so I added it in a separate commit.

### DIFF
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -318,6 +318,7 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
         output->rem_binding(&on_toggle_maximize);
         output->rem_binding(&on_toggle_above);
         output->rem_binding(&on_toggle_fullscreen);
+        output->rem_binding(&on_toggle_sticky);
     }
 };
 
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -22,6 +22,8 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
         "wm-actions/toggle_fullscreen"};
     wf::option_wrapper_t<wf::activatorbinding_t> toggle_sticky{
         "wm-actions/toggle_sticky"};
+    wf::option_wrapper_t<wf::activatorbinding_t> send_to_back{
+        "wm-actions/send_to_back"};
 
     bool toggle_keep_above(wayfire_view view)
     {
@@ -50,8 +52,8 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
     }
 
     /**
-     * Find the selected toplevel view, or nullptr if the selected
-     * view is not toplevel.
+     * Find the selected toplevel view, or nullptr if the selected view is not
+     * toplevel.
      */
     wayfire_view choose_view(wf::activator_source_t source)
     {
@@ -88,8 +90,7 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
     };
 
     /**
-     * Ensures views marked as above are still above
-     * if their output changes.
+     * Ensures views marked as above are still above if their output changes.
      */
     wf::signal_connection_t on_view_output_changed
     {[=] (wf::signal_data_t *data)
@@ -115,8 +116,8 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
     };
 
     /**
-     * Ensures views marked as above are still above
-     * if they are minimized and unminimized.
+     * Ensures views marked as above are still above if they are minimized and
+     * unminimized.
      */
     wf::signal_connection_t on_view_minimized
     {[=] (wf::signal_data_t *data)
@@ -142,8 +143,8 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
     };
 
     /**
-     * Disables show desktop if the workspace is changed or any view is
-     * attached, mapped or unminimized.
+     * Disables show desktop if the workspace is changed or any view is attached,
+     * mapped or unminimized.
      */
     wf::signal_connection_t view_attached = [this] (wf::signal_data_t *data)
     {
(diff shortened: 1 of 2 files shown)


JordanL2:
Currently has a bug, when it focuses on the top-most view it can focus a view in a different workspace.

JordanL2:
Fixed, although the code is pretty ugly and there might be a better way of checking if two views are on the same workspace.

plugins/wm-actions/wm-actions.cpp
```diff
@@ -267,6 +268,35 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
...
+            auto views =
+                view->get_output()->workspace->get_views_on_workspace(ws,
+                    wf::LAYER_WORKSPACE);
+            wayfire_view bottom_view = views[views.size() - 1];
+            if (view != bottom_view)
+            {
+                // Move view to bottom by stacking it below the bottom-most view
+                view->get_output()->workspace->restack_below(view, bottom_view);
+
+                // Change focus to the top-most view
```
ammen99, line 285: Wayfire's stacking order isn't necessarily indicative of last-focused order. Instead, the next-to-focus view should be the view with the latest `last_focused_timestamp` (see [here](https://github.com/WayfireWM/wayfire/blob/master/src/output/output.cpp#L89)).

ammen99 (changes requested):
LGTM aside from the comment.

plugins/wm-actions/wm-actions.cpp
JordanL2: Done.

JordanL2:
I don't know why the code style check is failing, I've decrusted the only C++ file that's changed here.

JordanL2:
Uh. When I do the code style check command locally:
```
git ls-files | grep "hpp$\|cpp$" | xargs ./uncrustify/build/uncrustify -c uncrustify.ini --check
```
Every file fails except for `plugins/wm-actions/wm-actions.cpp`. I've not changed uncrustify.ini. There hasn't been an uncrustify release since May so it's not a mismatched version.

JordanL2:
Apparently my uncrustify thinks comment lines should be slightly wider. When I try uncrusting a random file this is the change it makes:
```
-/* Returns the intersection of the two boxes, if the boxes don't intersect,
- * the resulting geometry has undefined (x,y) and width == height == 0 */
+/* Returns the intersection of the two boxes, if the boxes don't intersect, the
+ * resulting geometry has undefined (x,y) and width == height == 0 */
```

JordanL2:
I don't know what's wrong, I'm using up-to-date uncrustify from Arch repo and uncrustify.ini is identical to the one on master. I'm uncrusting with:
```
uncrustify -c uncrustify.ini --no-backup plugins/wm-actions/wm-actions.cpp
```
I don't know if it's relevant but this is my LANG env var, in case it's charset related:
```
LANG=en_GB.UTF-8
```

JordanL2:
@ammen99 

Since comment lines are meant to be max 85 chars long:
```
# Try to wrap comments at N columns.
cmt_width                       = 85        # unsigned number
```
My instance of uncrustify actually seems to be working more correctly. Looking at the diff I posted above, the original line is 75 chars wide, the modified line is 79 chars wide. So I think the question is why the CI is limiting comment lines more than needed.

JordanL2:
Fixed my uncrustify issue by compiling uncrustify from source.

plugins/wm-actions/wm-actions.cpp
```diff
@@ -267,6 +268,38 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
...
+            auto views =
+                view->get_output()->workspace->get_views_on_workspace(ws,
+                    wf::LAYER_WORKSPACE);
+            wayfire_view bottom_view = views[views.size() - 1];
+            if (view != bottom_view)
+            {
+                // Move view to bottom by stacking it below the bottom-most view
+                view->get_output()->workspace->restack_below(view, bottom_view);
+
+                // Change focus to the last focused view on this workspace
```
ammen99, line 285: Sorry for the late suggestion, but this is mostly implemented already with this function https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/output.hpp#L197

There's a catch though - you should check that there is at least one other view on the workspace, otherwise, `refocus()` will set focus to NULL. But I think that a simple check like that is better than reinventing `refocus()`

JordanL2:
```
  285                 // Change focus to the last focused view on this workspace
  286                 if (views.size() > 1) {
  287                     view->get_output()->refocus(view, wf::LAYER_WORKSPACE);
  288                 }
```
This isn't quite working. If I have three views open: A, B, C (which appear in scale in that order, and I keep triggering the send_to_back, I get:
1. A goes to bottom, B is focused and is the top-most view
2. B goes to bottom, C is the top-most view, but A is focused
Before, I could keep triggering it and I could loop through all three windows in the same order forever, which is the ideal behaviour really. Certainly, it's not acceptable that the top-most view isn't the one focused.

JordanL2: Figured out the problem, don't quite understand it. This change in refocus makes it work:
```
@@ -92,8 +92,9 @@ void wf::output_impl_t::refocus(wayfire_view skip_view, uint32_t layers)
         bool visible_1 = view_on_current_ws(view1);
         bool visible_2 = view_on_current_ws(view2);

-        return std::make_tuple(visible_1, view1->last_focus_timestamp) <
-        std::make_tuple(visible_2, view2->last_focus_timestamp);
+        return (visible_1 && view1->last_focus_timestamp) < (visible_2 && view2->last_focus_timestamp);
```
So, it's the tuple comparison which doesn't seem to be working properly. This "fix" however treats all non-visible views the same, not caring about their last_focus_timestamp.

JordanL2: I've submitted a commit which doesn't use last_focus_timestamp at all and just uses the stacking order as it was before, this is working in my tests.

JordanL2:
I've added another commit, this one emits a `output-stack-order-changed` at the end of the method. This is to alert any relevant plugins that the stack order has changed. Case in point, this is needed for `follow-focus`.

plugins/wm-actions/wm-actions.cpp
```diff
@@ -267,6 +268,36 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
...
+                views =
+                    view->get_output()->workspace->get_views_on_workspace(ws,
+                        wf::LAYER_WORKSPACE);
+                view->get_output()->focus_view(views[0], false);
+            }
+
+            // Alert any relevant plugins that stack order has changed
```
ammen99, line 292: workspace-manager should emit this signal when restacking the view below, right? Why is this needed again here?

plugins/wm-actions/wm-actions.cpp
ammen99: Okay, I guess it makes sense that we don't follow last-focused order when we're changing the stacking order. I'm leaving this for you to decide, current version LGTM.

plugins/wm-actions/wm-actions.cpp
JordanL2: This was to make it work better with follow-focus so the view the cursor is pointing to gets the focus. We have to do this after the focus command of course. However, I'm not sure if we actually want to do the follow-focus change I submitted any more, after thinking about it more. So we possibly should take this signal out for now.

JordanL2: Removed the above change.

ammen99 (approved):
Thanks for the patience, LGTM now!

PR #1274, opened 2021-09-05
Option to show minimized views in scale

dkondor:
This can be helpful if scale is used as a primary method to switch between views. Of course, this is probably highly opinion-based, so I made this into a config option. Depends on #1906 

Issues / questions:
 - [x] Test the case when a view is minimized while scale is running ~~(should work, but not easy to test, I don't find the interactive scale option anymore)~~ simple test case [here](https://github.com/dkondor/wayfire/blob/scale_test/minimize_window.py)
 - [x] Should this be a global option? Yes
 - ~~[ ] Option for an other plugin starting scale to specify whether to include minimized views (e.g. the dbus plugin [here](https://github.com/damianatorrpm/wayfire-plugin_dbus_interface/blob/c705bf1e9dbe54b7ecc3608869db7c53e82bb38d/dbus_interface_backend.cpp#L1095))~~
 - [x] Should minimized view be differentiated in some way? E.g. more faded, different title overlay, etc.
 - [ ] Any view that is focused during scale is unminimzed. This is an issue when using the arrow keys which focuses the views under selection. I see that a related issue came up in #1698 
 - ~~[ ] This currently works by setting the state of the view unminimized for the duration of scale; this is the easiest option to make it show up without extra hacks. Still, it could be an issue whether it is "logically" minimized or not (e.g. if a taskbar is displaying minimized views separately). Also, setting the minimized state automatically adjusts focus ([here](https://github.com/WayfireWM/wayfire/blob/84ec9fce347c330ad04c30036d22d3367e915802/src/view/view.cpp#L397)), which is not really needed here.~~

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -68,6 +68,7 @@ struct view_scale_data
     };
 
     view_visibility_t visibility = view_visibility_t::VISIBLE;
+    bool was_minimized = false; /* flag to indicate if this view was originally minimized */
 };
 
 /**
@@ -108,7 +109,9 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
     wf::option_wrapper_t<int> spacing{"scale/spacing"};
     wf::option_wrapper_t<bool> middle_click_close{"scale/middle_click_close"};
     wf::option_wrapper_t<double> inactive_alpha{"scale/inactive_alpha"};
+    wf::option_wrapper_t<double> minimized_alpha{"scale/minimized_alpha"};
     wf::option_wrapper_t<bool> allow_scale_zoom{"scale/allow_zoom"};
+    wf::option_wrapper_t<bool> include_minimized{"scale/include_minimized"};
 
     /* maximum scale -- 1.0 means we will not "zoom in" on a view */
     const double max_scale_factor = 1.0;
@@ -198,6 +201,15 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         scale_data[view].transformer = tr;
         view->get_transformed_node()->add_transformer(tr, wf::TRANSFORMER_2D,
             "scale");
+        /* Handle potentially minimized views by making them visible,
+         * however, they start out as fully transparent. */
+        if (view->minimized)
+        {
+            tr->alpha = 0.0;
+            wf::scene::set_node_enabled(view->get_root_node(), true);
+            scale_data[view].was_minimized = true;
+        }
+
         /* Transformers are added only once when scale is activated so
          * this is a good place to connect the geometry-changed handler */
         view->connect(&view_geometry_changed);
@@ -231,6 +243,11 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
             for (auto& toplevel : e.first->enumerate_views(false))
             {
                 pop_transformer(toplevel);
+            }
+
+            if (e.second.was_minimized)
+            {
+                wf::scene::set_node_enabled(e.first->get_root_node(), false);
             }
 
             if (e.second.visibility == view_scale_data::view_visibility_t::HIDDEN)
@@ -379,7 +396,8 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
             }
 
             auto alpha = scale_data[v].transformer->alpha;
-            scale_data[v].fade_animation.animate(alpha, (double)inactive_alpha);
+            double target_alpha = (v->minimized) ? minimized_alpha : inactive_alpha;
+            scale_data[v].fade_animation.animate(alpha, target_alpha);
         }
     }
 
@@ -415,6 +433,11 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         if (!view)
         {
             return;
+        }
+
+        if (scale_data.at(view).was_minimized)
+        {
+            wf::scene::set_node_enabled(view->get_root_node(), false);
         }
 
         for (auto v : view->enumerate_views(false))
@@ -659,7 +682,8 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         view = find_view_in_grid(next_row, next_col);
         if (view && (current_focus_view != view))
         {
-            // view_focused handler will update the view state
+            /* view_focused handler will update the view state
+             * note: this will always unminimize a minimized view */
             output->focus_view(view, false);
         }
     }
@@ -707,7 +731,8 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
     /* Returns a list of views for all workspaces */
     std::vector<wayfire_toplevel_view> get_all_workspace_views()
     {
-        return output->wset()->get_views(wf::WSET_EXCLUDE_MINIMIZED | wf::WSET_MAPPED_ONLY);
+        return output->wset()->get_views(
+            (include_minimized ? 0 : wf::WSET_EXCLUDE_MINIMIZED) | wf::WSET_MAPPED_ONLY);
     }
 
     /* Returns a list of views for the current workspace */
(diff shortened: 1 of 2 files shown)


ammen99:
This needs a rebase, otherwise is a good idea.

I think it makes the most sense to have this as an option in scale itself.
As to the minimized state: now there is a separation between logically minimized (view->minimized), and to hide/show the view its root node is enabled or disabled, so you could just enable and then disable the root node to implement this.

dkondor:
Hi,

I gave this a try, since it seemed simple, but I cannot get it working. My attempt is in the [scale_minimized3](https://github.com/dkondor/wayfire/tree/scale_minimized3) branch.
The problem is that when trying to enable or disable showing minimized views (lines 210 and 250), if I use `get_transformed_node()` (as the code for hiding the filtered views does), then the views still do not show up (there is an empty slot in their place, but cannot be selected with the mouse).
However, if I use `get_root_node()` (as it is now in my code), everything seems to work well at first, but if I exit scale by selecting a non-minimized view and then start scale again, the minimized views do not show up anymore. If I exit scale by selecting a minimized view (which becomes unminimized as a consequence), I cannot minimize it anymore.

I wonder if I'm missing something how the scenegraph interacts with the minimized state? I tried to look into the implementation (in toplevel-view.cpp), but I don't see anything obvious, apart from the minimized / disappeared signals that get emitted.

soreau:
> Hi,
> 
> I gave this a try, since it seemed simple, but I cannot get it working. My attempt is in the [scale_minimized3](https://github.com/dkondor/wayfire/tree/scale_minimized3) branch. The problem is that when trying to enable or disable showing minimized views (lines 210 and 250), if I use `get_transformed_node()` (as the code for hiding the filtered views does), then the views still do not show up (there is an empty slot in their place, but cannot be selected with the mouse). However, if I use `get_root_node()` (as it is now in my code), everything seems to work well at first, but if I exit scale by selecting a non-minimized view and then start scale again, the minimized views do not show up anymore. If I exit scale by selecting a minimized view (which becomes unminimized as a consequence), I cannot minimize it anymore.
> 
> I wonder if I'm missing something how the scenegraph interacts with the minimized state? I tried to look into the implementation (in toplevel-view.cpp), but I don't see anything obvious, apart from the minimized / disappeared signals that get emitted.

Hi, try this:
```
wf::scene::set_node_enabled(view->get_root_node(), true);
wf::scene::set_node_enabled(view->get_transformed_node(), true);
```

when adding the transformer. Hope this helps.

dkondor:
Hi,

thank you, this fixed my first problem (view not appearing the 2nd time scale is used). Also, I did not realize before that `set_node_enabled()` uses a counter internally. This made me come up with a solution for my second problem (view becomes unminimizable), but it seems very hackish.

Tracking down why sometimes I don't disable the root node when needed, I found that the code for minimizing a view [here](https://github.com/WayfireWM/wayfire/blob/f848c7569e785b00edc5464c40a316577741c488/src/view/toplevel-view.cpp#L214) always sends a `view_disappeared_signal`, regardless whether a view is being minimized or restored. However, in scale ([here](https://github.com/WayfireWM/wayfire/blob/f848c7569e785b00edc5464c40a316577741c488/plugins/scale/scale.cpp#L1081)), this always results in the view in question being removed from `scale_data`. However, in this case, I would really need to distinguish between a view appearing or disappearing -- the [solution](https://github.com/dkondor/wayfire/blob/57f21ba89eaf867018e9f604f064cb1d54823ddc/plugins/scale/scale.cpp#L1111) I could come up with so far is a bad hack that will lead to hard-to-reproduce crashes. Is there a better way to distinguish between views appearing and disappearing?

ammen99:
> Tracking down why sometimes I don't disable the root node when needed, I found that the code for minimizing a view [here](https://github.com/WayfireWM/wayfire/blob/f848c7569e785b00edc5464c40a316577741c488/src/view/toplevel-view.cpp#L214) always sends a `view_disappeared_signal`, regardless whether a view is being minimized or restored.

Oh well, this is definitely an oversight, it should be emitted only when minimizing.

> Is there a better way to distinguish between views appearing and disappearing?

The disappear signal should never be called (current behavior is a bug) when a view 'appears', so I hope this answers your question?

dkondor:
Yes, it answers my question :) changing it solves the problem I'm having here as well. I'll open a separate PR for that. Thanks!

ammen99 (commented):
> Should minimized view be differentiated in some way? E.g. more faded, different title overlay, etc.

An alpha < 1.0 would probably make sense.

> Any view that is focused during scale is unminimzed.

Yeah, this is unfortunate. I think the best solution is probably to refactor scale to simply not change the focus at all, but keep track of the currently selected view manually, and focus the selected view only at the end.

plugins/scale/scale.cpp
```diff
@@ -198,6 +200,16 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         scale_data[view].transformer = tr;
         view->get_transformed_node()->add_transformer(tr, wf::TRANSFORMER_2D,
             "scale");
+        /* Handle potentially minimized views by making them visible,
+         * however, they start out as fully transparent. */
+        if (view->minimized)
+        {
+            tr->alpha = 0.0;
+            wf::scene::set_node_enabled(view->get_root_node(), true);
+            wf::scene::set_node_enabled(view->get_transformed_node(), true);
```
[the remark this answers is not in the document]
dkondor: Actually, no need for it -- I've added it before when experimenting

dkondor:
OK, I've addressed some of the remaining issues.

> > Should minimized view be differentiated in some way? E.g. more faded, different title overlay, etc.
> 
> An alpha < 1.0 would probably make sense.

I've added a configuration option for this; however, the difference with a "normal" unfocused view seems to be not that large to me subjectively, but this could be adjusted later.

> 
> > Any view that is focused during scale is unminimzed.
> 
> Yeah, this is unfortunate. I think the best solution is probably to refactor scale to simply not change the focus at all, but keep track of the currently selected view manually, and focus the selected view only at the end.

This is probably best addressed separately; I might have some time to work on this later this week.

ammen99:
> I've added a configuration option for this; however, the difference with a "normal" unfocused view seems to be not that large to me subjectively, but this could be adjusted later.

Personally, I would make such views much fainter (say alpha = 0.3), so an option sounds good :)

plugins/scale/scale.cpp
```diff
@@ -860,7 +885,9 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         {
             std::sort(views.begin(), views.end(), [=] (wayfire_toplevel_view a, wayfire_toplevel_view b)
             {
-                return wf::get_focus_timestamp(a) > wf::get_focus_timestamp(b);
+                /* note: avoid focusing minimized views if possible */
+                return (a->minimized < b->minimized) ||
```
ammen99, line 889: I would split this into an if - else for better readability.

plugins/scale/scale.cpp
```diff
@@ -1082,15 +1112,27 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
     {
         if (scale_data.count(get_top_parent(view)) != 0)
         {
-            remove_view(view);
-            if (scale_data.empty())
+            if (include_minimized && view->minimized && should_scale_view(view))
```
ammen99, line 1115: I think we should be really careful here, view_disappeared is not emitted only when the view is minimized. This path here could be triggered if for example a minimized view is closed (for example by killing the client).

I tested this locally and it doesn't actually cause any issues because `should_scale_view()` returns false for unmapped views - but I thought it might be worth adding a comment for future reference :)

ammen99 (approved):
Two small things, otherwise I tested in some complex scenarios and it always works, so good job and thanks :)

plugins/scale/scale.cpp
dkondor: I've come up with an alternative, since it's also a bit weird to compare bools this way.

plugins/scale/scale.cpp
dkondor: Yes, I've run into this issue while testing -- I was experimenting if there is a more efficient way to distinguish between these two cases than `should_scale_view()`, but decided that this is likely the most robust one.

PR #1279, opened 2021-09-08
merging get_snapshot() with take_snapshot()

marcusbritanicus:
Code refactoring to merge `void take_snapshot()` with `const wf::framebuffer_t get_snapshot()` int one function: `const wf::framebuffer_t take_snapshot()`. It's basically the `take_snapshot()` function which also returns the captured buffer.


ammen99 (changes requested):
Needs bumping of the API/ABI version in `plugin.hpp`. Otherwise LGTM

ammen99:
You should rebase on top of master and also make sure that uncrustify checks passes.

ammen99 (approved):
Thanks!

PR #1281, opened 2021-09-09
Scale focus bug

### LINKED ISSUE #1280: Scale - Mouse-wheel not scrolling correct window sometimes
Steps to reproduce:
1. Two windows open on desktop, A and B. Window A is selected. (Both are maximised in my testing.)
2. Activate scale, and hover mouse over window B, on the right hand side. (EDIT: You don't even have to move the mouse, if the cursor is already over window B when you open scale)
3. Deactivate scale, without selecting a different window (so A is still selected).
4. Without moving the mouse at all, scroll mouse wheel.

Expected: Window A scrolls
Actual: Window B scrolls.

I thought this would be related to the follow-focus plugin, but after disabling that I can still reproduce it.

This doesn't affect keyboard control - pressing the arrow keys scrolls window A as expected.

This also affects the switcher plugin.

JordanL2:
Both the Scale and Switcher plugins have a bug that can leave the wrong view focused by the pointer after using them. This means that scrolling the mouse wheel after selecting a view with these plugins can scroll the wrong view.

This fixes emits an `output-stack-order-changed` signal at the end of both of these plugins' animation when closing them, to cause `update_cursor_focus` in `src/core/seat/pointer.cpp` to recalculate which view should have focus.

Fixes #1280

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -1442,6 +1442,10 @@ class wayfire_scale : public wf::plugin_interface_t
         workspace_changed.disconnect();
         view_geometry_changed.disconnect();
         output->deactivate_plugin(grab_interface);
+
+        wf::stack_order_changed_signal data;
+        data.output = output;
+        wf::get_core().emit_signal("output-stack-order-changed", &data);
     }
 
     /* Utility hook setter */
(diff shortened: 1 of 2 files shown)


ammen99:
Nice catch!

Issue #1283, opened 2021-09-13
Wayfire wlr_virtual_pointer_v1_new_pointer_event ignores suggested seat and output


kode54:
**Describe the bug**
Wayfire's new virtual pointer event handler ignores the suggested seat and output parameters completely, which breaks single output pointer constraints by anyvnc.

https://github.com/any1/wayvnc/blob/master/src/main.c#L985-L987

**To Reproduce**
1. Start Wayfire with multiple outputs.
2. Start anyvnc on one of those outputs.
3. Cursor is spread across all outputs.

**Expected behavior**
Cursor is constrained to the suggested output, or even just seat.

**Wayfire version**
git: e83f1f6fb271911e74a404784a8699d45a525095

**Notes**
There are two messages for creating a new virtual pointer, one which can specify the seat it belongs to, and an extended function that can also specify the output it should belong to. AnyVNC uses the latter to constrain the client cursor to the requested output to be remoted into. Currently Wayfire ignores both of these hint variables and just treats all virtual pointers as a full surface space according to all active outputs.

**Related issues**
https://github.com/any1/wayvnc/issues/64

kode54:
It appears Sway does listen to these members. It maps the input to the given wlr seat if a seat is specified. Output handling appears to be rather simple to handle, too, it just uses `wlr_cursor_map_input_to_output` to handle that part.

mrsteve0924:
i've been using wayfire for past 6  months and just installed wayvnc and still see this same issue you are describing.  

i dont see any response here on this issue from wayfire.  have you just moved on to sway?  or maybe some other workaround you discovered?

kode54:
I've moved desktops multiple times, and back to Wayfire again. This appears to still be a problem.

quadcom:
As a newbie trying to figure his way through the Linux landscape and start getting his feet wet with RPIs, it's frustrating to see an issue like this be left in this state for so long.

I set up Bookworm and wrestled through getting a kiosk mode running, only to be stumped by why the mouse movements on a VNC connection are inverted. In this current state, it is ridiculously, if not impossible, to remote in and do anything with the Pi.

Researching solutions for this for a guy with my experience level is like hunting in a dark room with a tea light whilst getting battered in a hurricane. Everything is outdated, doesn't work or doesn't apply.

It's no wonder why the Linux community hasn't taken over the desktop space from MS. With all its quirks and PIA issues, it still works.

/end rant

soreau:
[...]

PR #1284, opened 2021-09-13
core: Implement new_pointer_event suggested_output

kode54:
This is a partial implementation, which is enough to get wayvnc
    cursor working on sessions with multiple outputs. It does not
    implement suggested_seat, but Wayfire does not appear to use
    that yet, anyway. Perhaps a simple check could be added to at
    least make sure that suggested_seat is either NULL or the one
    seat that Wayfire creates.

    Partial fix for #1283 , enough to get it basically working, but
    not exactly a full implementation of this functionality.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -263,7 +263,8 @@ void wf::compositor_core_impl_t::init()
         input->handle_new_input(&ptr->input_device);
         if (event->suggested_output)
         {
-            wlr_cursor_map_input_to_output(seat->cursor->cursor, &ptr->input_device, event->suggested_output);
+            wlr_cursor_map_input_to_output(seat->cursor->cursor, &ptr->input_device,
+                event->suggested_output);
         }
     });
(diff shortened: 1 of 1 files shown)


any1:
Maybe it would be good to add a `// TODO` for the seat, so that it's less likely to be forgotten if/when wayfire gets multi-seat support.

kode54:
Added the comment, it should hold over in case anyone else decides to try implementing it, in case it actually becomes relevant.

ammen99 (changes requested):
Unfortunately, this wouldn't work. Device mappings are done here: https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/input-manager.cpp#L57

As soon as that code is run, any other mapping you create (like in this PR) will be overwritten.

How I think this problem can be fixed: You need to set the mapping option to the correct value. Then, you should `lock()` the option so that the config file doesn't overwrite it - there is a wf-config API for such scenarios.

kode54:
Right, how do I do that? Why do I have to do that when you can just edit this PR?

PR #1287, opened 2021-09-17
workspace-impl: emit changeviewport on fixed views

Javyre:
The signal definition is also moved to core


src/api/wayfire/signal-definitions.hpp
```diff
@@ -595,6 +595,23 @@ struct view_geometry_changed_signal : public _view_signal
     wf::geometry_t old_geometry;
 };
 
+/**
+ * name: view-change-viewport
```
ammen99, line 599: viewport and workspace used to be used interchangeably, however, I think for new code it should only be 'workspace'. Since we're changing things I think now is a good time to rename this signal.

src/output/workspace-impl.cpp
```diff
@@ -783,6 +783,11 @@ class output_viewport_manager_t
 
         for (auto& v : fixed_views)
         {
+            wf::view_change_viewport_signal vdata;
+            vdata.view = v;
+            vdata.from = data.old_viewport;
```
ammen99, line 788: Fixed views are not necessarily on the current workspace. You can have fixed views everywhere (not that I know of any plugin which does that but still)

ammen99 (changes requested):

src/api/wayfire/signal-definitions.hpp
Javyre: Done.

src/output/workspace-impl.cpp
Javyre: Ok so to do this I used `get_view_main_workspace()` which meant that I have to save the old workspaces now since that function corrects for invalid workspace in a lossy way. Is this what you had in mind?

src/api/wayfire/signal-definitions.hpp
```diff
@@ -595,6 +595,23 @@ struct view_geometry_changed_signal : public _view_signal
...
+ */
+struct view_change_workspace_signal : public _view_signal
+{
+    wf::point_t from, to;
+
+    /**
+     * Indicates whether the old workspace is known.
+     * If false, then the `from` field should be ignored.
+     */
+    bool old_workspace_invalid = true;
```
Javyre, line 612: I realized this field name doesn't seem to make much sense. If the old workspace is invalid then shouldn't that mean that it should be ignored? Isn't this backwards? i.e.: this field should be called `old_workspace_valid`

ammen99: Ah, yeah, makes sense, nice catch! It should be `_valid` indeed.

src/output/workspace-impl.cpp
ammen99: Hmm, I haven't thought it in so much details but what you did makes sense, we need to somehow keep this mapping.

src/api/wayfire/signal-definitions.hpp
Javyre: Done.

ammen99 (approved):
LGTM, thanks!

PR #1289, opened 2021-09-21
plugin-loader: clarify dlclose() behaviour

Javyre:
This came up twice in conversation on IRC and it's quite verbose to
track it down from scratch every time.

I think it would be useful to have this in the code to easily find and
point at.


ammen99 (approved):
Thanks!

PR #1293, opened 2021-09-25
core: Set minimum workspaces to one

### LINKED ISSUE #1290: Accidentally setting workspace grid vheight or vwidth in [core] to zero crashes wayfire
Accidentally set vheight of workspace grid using the wcm configurator to zero. Wayfire immediately crashed kicking me back to the tty.

You can easily reproduce either by selecting 'core' in wcm and clicking Horizontal/Vertical virtual size minus sign until you reach zero OR editiing the ~/.config/wayfire.ini file and changing the value of vheight/vwidth to zero in the [core] stanza.

The configuration GUI should not allow a (meaningless) value of zero.

Wayfire version: 0.8.0-8 as provided by the debian/ubuntu PPA.

soreau:
Wayfire crashed if vwidth or vheight was set to <= 0. Set the minimum
vwidth and vheight to 1, so we always have at least one workspace.

Fixes #1290.

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -19,11 +19,13 @@
 			<_short>Horizontal virtual size</_short>
 			<_long>Sets the number of horizontal workspaces.  Currently, cannot be changed at runtime.</_long>
 			<default>3</default>
+			<min>1</min>
 		</option>
 		<option name="vheight" type="int">
 			<_short>Vertical virtual size</_short>
 			<_long>Sets the number of vertical workspaces.  Currently, cannot be changed at runtime.</_long>
 			<default>3</default>
+			<min>1</min>
 		</option>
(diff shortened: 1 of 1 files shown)


src/output/workspace-impl.cpp
```diff
@@ -483,7 +483,8 @@ class output_viewport_manager_t
         }
 
         auto old_grid = grid;
-        grid = {vwidth_opt, vheight_opt};
+        grid = {vwidth_opt > 0 ? vwidth_opt : 1,
```
ammen99, line 486: This clamping is not necessary, wf-config will sanitize the values once the option has a min/max in the .xml file

ammen99 (approved):
Thanks!

PR #1298, opened 2021-10-02
Track wlroots changes

JordanL2:
We needed some changes to track this wlroots change:
https://github.com/swaywm/wlroots/commit/0e342083440db978463dbf43795bfed8a5a11cd6

I also updated the wlroots submodule.


ammen99 (approved):
Thanks!

PR #1300, opened 2021-10-06
view: chase wlroots b72a217fcc (fix xdg maximize not working)

unrelentingtech:
https://github.com/swaywm/wlroots/pull/3199

I think that's the only place that needed changing… at least it seems that nobody saw any breakage other than maximize requests not working.

There are a couple `pending` accesses right in `initialize()` but I'm not sure how to even test them. Trying to e.g. launch vkquake preconfigured to start in fullscreen, it still does a fullscreen request later rather than initializing with fullscreen already pending.


ammen99:
> There are a couple `pending` accesses right in `initialize()` but I'm not sure how to even test them. Trying to e.g. launch vkquake preconfigured to start in fullscreen, it still does a fullscreen request later rather than initializing with fullscreen already pending.

I am not sure how to test this as well, but I assume it can happen - the xdg-shell protocol doesn't seem to forbid this, wlroots emits the `new_surface` event on the first commit, but the maximize request could have happened before that.

unrelentingtech:
Changed the `initialize` ones as well.

PR #1330, opened 2021-11-06
Rename output.layout field in example config to output.position

datMaffin:
When trying to set wayfire up, I noticed that the placeholder config has a "layout" field (that seemingly does nothing) instead of "position"


ammen99 (approved):
Thanks!

PR #1332, opened 2021-11-07
scale: allow small mouse movement between click and release

occivink:
Before, any movement between click and release would reset 
last_selected_view, resulting in BTN_LEFT not ending scale.
Now this only happens after the drag snap-off is triggered, making it
less likely to happen accidentally.

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -107,6 +107,7 @@ class wayfire_scale : public wf::plugin_interface_t
     bool all_workspaces;
     std::unique_ptr<wf::vswitch::control_bindings_t> workspace_bindings;
 
+    wf::point_t drag_origin;
     wf::shared_data::ref_ptr_t<wf::move_drag::core_drag_t> drag_helper;
 
   public:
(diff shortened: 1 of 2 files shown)


plugins/scale/scale.cpp
```diff
@@ -547,16 +552,16 @@ class wayfire_scale : public wf::plugin_interface_t
...
             wf::move_drag::drag_options_t opts;
             opts.join_views = true;
             opts.enable_snap_off    = true;
             opts.snap_off_threshold = 200;
 
             drag_helper->start_drag(last_selected_view, to, opts);
-            last_selected_view = nullptr;
-        } else if (drag_helper->view)
+            drag_started = true;
+        } else if (drag_started)
```
ammen99, line 564: I think we should still check drag_helper->view here, otherwise, LGTM.

occivink: Right, maybe we can entirely rely on it instead of the extra boolean

occivink:
Simplified to not require the extra boolean

ammen99:
Actually, on a second thought: does this really make sense? What if the snap-off threshold is big? Like, you can have drag threshold 300-400 pixels (I have it like this), and then, if I drag my mouse for whole 100 pixels, snap-off won't be triggered, but this would count as a click, which I find counterintuitive.

Maybe we should just manually check how much the pointer actually moved and make a decision based on that.

occivink:
Update to use a custom threshold, instead of the snap-off threshold. Two questions came to mind:
* the logic to determine the instead to the origin is a bit duplicated, should I add some kind of `distance_moved()` in core_drag_t?
 * should the threshold be configurable?

ammen99:
> * should the threshold be configurable?

I think it is fine to leave it hardcoded for now.

> * the logic to determine the distance dragged is a bit duplicated, should I add some kind of `distance_moved()` in core_drag_t?

If this will reduce code duplication (it sounds like it will), then yeah, sounds like a good idea.

ammen99 (approved):
LGTM, thanks!

PR #1333, opened 2021-11-07
workspace-manager: ignore promoted state for views not in WORKSPACE layer

### LINKED ISSUE #1076: Full-screened apps are lost after coming out of "Show desktop"
**Describe the bug**
Full-screened apps are lost (do not appear/vanish) after coming out of wm-actions/show_desktop. They do not crash/close, they're just not shown. They do not even get listed by switcher/fast-switcher.

**To Reproduce**
1. Enable plugin wm-actions
2. Open an app, and make it full-screen
3. Invoke "Show Desktop" - All windows are hidden to show the desktop
4. Invoke again to come out of "Show Desktop".
5. All windows except the one which was full-screen are shown.

**Expected behavior**
The full-screened window should reappear

**Wayfire version**
git

ammen99:
Promoted views make sense only for that layer. If a promoted view is moved to some of the other layers, it may become inaccessible because its promoted status is first cleared when it is moved back to the workspace layer.

We can simply ignore the promoted state as long as the view is not in the workspace layer. We also don't have to clear this state because it is unused. It will be recalculated when the view is brought back to the WORKSPACE layer.

Fixes #1076

### DIFF
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -330,8 +330,15 @@ class output_layer_manager_t
         rebuild_stack_order();
     }
 
+    enum class promoted_state_t
+    {
+        PROMOTED,
+        NOT_PROMOTED,
+        ANY,
+    };
+
     void push_views(std::vector<wayfire_view>& into, layer_t layer_e,
-        bool promoted)
+        promoted_state_t desired_promoted)
     {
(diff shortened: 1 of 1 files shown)


marcusbritanicus:
I got a chance to test this today. Works great.

PR #1337, opened 2021-11-09
decor & common: use pango to draw text

lilydjwg:
Use a serious library to draw all kinds of scripts.

The cairo text API is "a toy text API, for testing and demonstration
purposes. Any serious application should avoid them."

-- quoted from cairo's documentation


CrezyDud:
why leave cairo dependencies?

Javyre:
@CrezyDud wayfire uses Cairo for much more than just text rendering. Anyways it looks like Cairo buffers are the preferred output format for Pango ifaiu.

@lilydjwg any reason for only porting decor text rendering and not any other plugins?

lilydjwg:
Because that was the only one I saw. I'm starting using wayfire only for four days till now, and this is the second time I use pango :-) It's a lot of work to switch to Wayland.

Anyway scale is labeling windows without hollow boxes now.

@CrezyDud  Cairo is a vector drawing library that happens to be able to draw simple text. Pango only knows how to draw text. PangoCairo binds them together.

ammen99 (approved):
Unfortunately, I never used Pango and PangoCairo, but I'll trust you that this is the correct way to do it :)

Do you know of a way to test this (e.g. where cairo doesn't work but this does)?

lilydjwg:
Sure, in a web browser, open these links and see if the title renders correctly in decoration (need prefer server decorations) and scale.

* https://zh.wikipedia.org/wiki/Wikipedia:%E9%A6%96%E9%A1%B5
* https://ar.wikipedia.org/wiki/%D8%B7%D8%B1%D9%8A%D9%82_%D8%A7%D9%84%D8%AD%D8%B1%D9%8A%D8%B1
* https://emojipedia.org/transgender-flag/

ammen99:
Alright, I tested this, it works quite well, so thanks!

Unfortunately, this also breaks wayfire-plugins-extra because we do not export Wayfire's dependencies in the .pc file. I'll try to write a patch which fixes this solution, and then we can merge this PR.

lilydjwg:
That patch is here: https://github.com/lilydjwg/wayfire/commit/9d760f3378b97f63550f9928d4f4c2dc15d3968f. It works for wayfire-plugins-extra but I don't know if it's the way. If wayfire.pc includes pango and pangocairo, it probably should include others too.

PR #1342, opened 2021-11-12
Update wlroots path

marcusbritanicus:
Change the wlroots upstream path to gitlab.freedesktop.org/wlroots/wlroots.git


marcusbritanicus:
This commit will probably fail. And that will be because of breaking changes in wlroots.

src/core/output-layout.cpp
```diff
@@ -665,7 +665,14 @@ struct output_layout_output_t
         }
 
         wlr_dmabuf_attributes attributes;
-        if (!wlr_output_export_dmabuf(wo->handle, &attributes))
+        if (wo->handle->front_buffer == NULL)
+        {
+            LOGE("Got empty buffer on ", wo->handle);
```
ammen99, line 670: I am not sure why we're printing a pointer to the `wlr_output` here. It might make more sense to output `wo->to_string()` or `wo->handle->name`, this way, you get a human-readable identification of the output where the error happens.

src/core/output-layout.cpp
```diff
@@ -665,7 +665,14 @@ struct output_layout_output_t
...
+
+            return;
+        }
+
+        if (!wlr_buffer_get_dmabuf(wo->handle->front_buffer, &attributes))
         {
             LOGE("Failed reading mirrored output contents from ", wo->handle);
```
soreau, line 677: @ammen99 maybe because this was already wo->handle here.

ammen99: Then that place is mistaken as well ...

ammen99 (approved):
LGTM now, thanks!

PR #1343, opened 2021-11-15
manually compare wlr_layer_surface_v1_state

### LINKED ISSUE #1328: swaybg high CPU usage
When using swaybg the CPU usage (of swaybg) goes up massively (for me up to 85%). There is a [bug report for swaybg](https://github.com/swaywm/swaybg/issues/33) as well where it looks like it might be wayfires fault. I also think this is likely because it seems to have started with some update of wayfire.

I know that wf-shell has a background componet as well but I haven't tried if there are similar issues appearing or not.

**To Reproduce**
1. Run wayfire
2. Start swaybg
3. CPU usage goes immediately up

**Expected behavior**
CPU usage in the normal range.

**Screenshots or stacktrace**
See linked [issue](https://github.com/swaywm/swaybg/issues/33) by @demostanis.

**Wayfire version**
git (master)

lilydjwg:
configure_serial changes every time, so programs like slurp and swaylock
would enter a loop.

Fixes #1328

### DIFF
--- a/src/view/layer-shell.cpp
+++ b/src/view/layer-shell.cpp
@@ -467,6 +467,28 @@ void wayfire_layer_shell_view::unmap()
     wf_layer_shell_manager::get_instance().handle_unmap(this);
 }
 
+static bool same_state(const wlr_layer_surface_v1_state* a, const wlr_layer_surface_v1_state* b)
+{
+    if(a->committed == b->committed &&
+            a->anchor == b->anchor &&
+            a->exclusive_zone == b->exclusive_zone &&
+            a->margin.top == b->margin.top &&
+            a->margin.bottom == b->margin.bottom &&
+            a->margin.left == b->margin.left &&
+            a->margin.right == b->margin.right &&
+            a->keyboard_interactive == b->keyboard_interactive &&
+            a->desired_width == b->desired_width &&
+            a->desired_height == b->desired_height &&
+            a->layer == b->layer &&
+            a->actual_width == b->actual_width &&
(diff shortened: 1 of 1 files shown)


src/view/layer-shell.cpp
```diff
@@ -467,6 +467,28 @@ void wayfire_layer_shell_view::unmap()
...
+            a->exclusive_zone == b->exclusive_zone &&
+            a->margin.top == b->margin.top &&
+            a->margin.bottom == b->margin.bottom &&
+            a->margin.left == b->margin.left &&
+            a->margin.right == b->margin.right &&
+            a->keyboard_interactive == b->keyboard_interactive &&
+            a->desired_width == b->desired_width &&
+            a->desired_height == b->desired_height &&
+            a->layer == b->layer &&
+            a->actual_width == b->actual_width &&
```
ammen99, line 483: actual_width doesn't seem relevant, do we use it anywhere in this file? I think this is something wlroots uses for internal bookkeeping.

lilydjwg: I didn't check. All fields are exported so I compare all sensible ones to lock my screen quietly. I can remove it if it's not useful.

ammen99: On a second thought, we can also just look at the state->committed field. It will be nonzero iff anything of value changes, so we don't need to list all of this.

lilydjwg: Yes, that seems to work (tested in a vm).

ammen99 (approved):
Thanks!

PR #1345, opened 2021-11-15
WIP: ws-impl: promote top fullscreen view per sublayer

Javyre:
This adresses a bug where if a fullscreen view in sublayer A is
promoted, it stops being promoted as soon as there is a view in a higher
sublayer.

Now the top view of each sublayer is checked for fullscreen and promoted
accordingly.

### DIFF
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -1101,9 +1101,27 @@ class workspace_manager::impl
         });
         views.erase(it, views.end());
 
-        if (!views.empty() && views.front()->fullscreen)
+        if (!views.empty())
         {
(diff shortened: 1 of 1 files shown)


Javyre:
@ammen99 Ok so this solves the bug we were talking about last time on irc if i understood it correctly. However now that the SUBLAYER_FLOATING fullscreened view remains promoted, it goes above the non-fullscreen SUBLAYER_DOCKED_ABOVE view because view promotion doesn't consider SUBLAYER_DOCKED_ABOVE.

I'm not sure this is the desired behaviour.. I would've expected the docked above view to still be above the floating layer but just that the promoted view is above the panel...

PR #1353, opened 2021-11-18
resize: support for keeping ratio

lilydjwg:
In some cases it's desirable to keep the width / height ratio even when the application doesn't require it, e.g. resizing scrcpy out of ratio will leave unpleasant black space.

The patch adds another key binding to resize to keep ratio.

### DIFF
--- a/metadata/resize.xml
+++ b/metadata/resize.xml
@@ -9,5 +9,11 @@
 			<_long>When the specified button is held down, you can drag windows to resize them.</_long>
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
+
+		<option name="activate_preserve_aspect" type="button">
+			<_short>Activate resize while preserving aspect</_short>
+			<_long>When the specified button is held down, you can drag windows to resize them while perserving its orignal aspect.</_long>
+			<default>disabled</default>
+		</option>
 	</plugin>
 </wayfire>
--- a/metadata/resize.xml
+++ b/metadata/resize.xml
@@ -6,13 +6,13 @@
 		<category>Window Management</category>
 		<option name="activate" type="button">
 			<_short>Activate</_short>
-			<_long>When the specified button is held down, you can drag windows to resize them.</_long>
+			<_long>When the specified button is held down, you can drag a window to resize it.</_long>
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
 
 		<option name="activate_preserve_aspect" type="button">
 			<_short>Activate resize while preserving aspect</_short>
-			<_long>When the specified button is held down, you can drag windows to resize them while perserving its orignal aspect.</_long>
+			<_long>When the specified button is held down, you can drag a window to resize it while perserving its orignal aspect.</_long>
 			<default>disabled</default>
 		</option>
 	</plugin>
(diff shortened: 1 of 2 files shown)


metadata/resize.xml
```diff
@@ -9,5 +9,10 @@
 			<_long>When the specified button is held down, you can drag windows to resize them.</_long>
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
+		<option name="activate_preserve_aspect" type="button">
+			<_short>Activate resize while preserving aspect</_short>
+			<_long>When the specified button is held down, you can drag windows to resize them while perserving its orignal aspect.</_long>
+			<default>&lt;super&gt; &lt;alt&gt; BTN_RIGHT</default>
```
ammen99, line 15: By default, most of the bindings should be disabled to avoid conflicts with user's configs.

plugins/single_plugins/resize.cpp
```diff
@@ -104,6 +104,20 @@ class wayfire_resize : public wf::plugin_interface_t
         output->connect_signal("view-disappeared", &view_destroyed);
     }
 
+    bool activate(bool keep) {
```
ammen99, line 107: I'd like a longer name for the parameter name, `keep` does not immediately tell the reader what this parameter is all about.

ammen99 (changes requested):
Sorry for the long wait, I have a few comments which should be easy to fix, also please fix formatting.

metadata/resize.xml
lilydjwg: Updated.

plugins/single_plugins/resize.cpp
lilydjwg: I've changed it to `should_perserve_aspect`.

ammen99:
Could you rebase again?

lilydjwg:
I tried but it crashed. The custom crash handling is a huge obstacle to debug:

* no coredump, so no gdb
* no crash log (it meant to print something but I didn't see it)
* a lot of unrelated `strace` logs

lilydjwg:
Thanks, it was a silly typo.

metadata/resize.xml
```diff
@@ -9,5 +9,11 @@
 			<_long>When the specified button is held down, you can drag windows to resize them.</_long>
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
+
+		<option name="activate_preserve_aspect" type="button">
+			<_short>Activate resize while preserving aspect</_short>
+			<_long>When the specified button is held down, you can drag windows to resize them while perserving its orignal aspect.</_long>
```
ammen99, line 15:
```suggestion
			<_long>When the specified button is held down, you can drag a window to resize it while perserving its orignal aspect.</_long>
```

plugins/single_plugins/resize.cpp
```diff
@@ -93,14 +95,29 @@ class wayfire_resize : public wf::per_output_plugin_instance_t, public wf::point
         output->connect(&on_view_disappeared);
     }
 
+    bool activate(bool should_preserve_aspect)
```
ammen99, line 98: There is one more way to activate the plugin, see `on_resize_request` defined in the beginning of the class.

ammen99 (changes requested):
A few small things + please fix the code style, otherwise lgtm.

plugins/single_plugins/resize.cpp
lilydjwg: I've set `preserve_aspect` to `false` to keep the behavior.

ammen99 (approved):
LGTM, thanks!

PR #1354, opened 2021-11-18
use GL_NEAREST for integer scale

lilydjwg:
Currently wayfire scales windows with default GL_LINEAR method. When scaling up this method gives blurry edges. For integral scales GL_NEAREST will give shape lines and blocky text instead, which looks much nicer. See files attached in [this comment](https://github.com/swaywm/wlroots/issues/1770#issuecomment-515404622).

Sway has [already supported this](https://github.com/swaywm/sway/pull/4727). It defaults to GL_NEAREST and has an option to change.

This patch makes wayfire to use GL_NEAREST when scale is an integer but I don't know how to make it an option.

I have [a better solution for xwayland windows on HiDPI monitors](https://gitlab.freedesktop.org/lilydjwg/wlroots/-/tree/hidpi) but it depends on [an xwayland patch which has stalled for some days](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733). Anyway I made this before that so here's a pull request.

### DIFF
--- a/src/view/surface.cpp
+++ b/src/view/surface.cpp
@@ -438,6 +438,14 @@ void wf::wlr_surface_base_t::_simple_render(const wf::framebuffer_t& fb,
     OpenGL::render_begin(fb);
     OpenGL::render_texture(texture, fb, geometry, glm::vec4(1.f),
         OpenGL::RENDER_FLAG_CACHED);
+    // use GL_NEAREST for integer scale.
+    // GL_NEAREST makes scaled text blocky instead of blurry, which looks better
+    // but only for integer scale.
+    if (fb.scale - floor(fb.scale) < 0.001)
+    {
+        GL_CALL(glTexParameteri(texture.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST));
+    }
+
     for (const auto& rect : damage)
     {
         fb.logic_scissor(wlr_box_from_pixman_box(rect));


src/view/surface.cpp
```diff
@@ -438,6 +438,10 @@ void wf::wlr_surface_base_t::_simple_render(const wf::framebuffer_t& fb,
     OpenGL::render_begin(fb);
     OpenGL::render_texture(texture, fb, geometry, glm::vec4(1.f),
         OpenGL::RENDER_FLAG_CACHED);
+    if (fb.scale - floor(fb.scale) < 0.001)
```
ammen99, line 441: Can you please add a comment in the source code explaining why we do this?

ammen99 (changes requested):
LGTM, please add a comment and format the code.
If you rebase on top of master, you'll even get a diff that can be applied to fix formatting (it is printed in the github actions log)

src/view/surface.cpp
CrezyDud: he said it's because it only does nearest when it's an integer
which makes sense to me

ammen99: It may make sense now because we know what problem this PR attempts to solve. It may be much less clear if someone looks at this code in 2 years or so.

ammen99 (approved):
LGTM, will merge after the CI runs. Thanks!

src/view/surface.cpp
lilydjwg: Updated.

@CrezyDud s/he/she/

ammen99 (approved):

CrezyDud:
yoy

Issue #1351, opened 2021-11-17
-Duse_system_wlroots=enabled fails to build with 2021-12-01 snapshot


jbeich:
<details><summary>Old errors obsolete after 6720f6a6e902</summary>

Regressed by [wlroots@a04cfca](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/a04cfca4da42) + [wlroots@fdf3169](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/fdf3169b4185). From [error log](https://github.com/WayfireWM/wayfire/files/7565313/wayfire-0.7.2.170.log):

```c++
src/core/opengl.cpp:538:31: error: no member named 'inverted_y' in 'wlr_gles2_texture_attribs'
    this->invert_y = !attribs.inverted_y;
                      ~~~~~~~ ^
src/main.cpp:301:21: error: use of undeclared identifier 'wlr_backend_get_renderer'; did you mean 'wlr_backend_get_session'?
    core.renderer = wlr_backend_get_renderer(core.backend);
                    ^~~~~~~~~~~~~~~~~~~~~~~~
```
</details>

Regressed by [wlroots@1d3dd7f](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/1d3dd7fc08f2) + [wlroots@efeb834](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/efeb8346cfc6).
```c++
In file included from src/core/core.cpp:17:
src/api/wayfire/nonstd/wlroots-full.hpp:85:10: fatal error: 'wlr/backend/noop.h' file not found
#include <wlr/backend/noop.h>
         ^~~~~~~~~~~~~~~~~~~~
src/core/output-layout.cpp:662:25: error: no member named 'front_buffer' in 'wlr_output'
        if (wo->handle->front_buffer == NULL)
            ~~~~~~~~~~  ^
src/core/output-layout.cpp:669:48: error: no member named 'front_buffer' in 'wlr_output'
        if (!wlr_buffer_get_dmabuf(wo->handle->front_buffer, &attributes))
                                   ~~~~~~~~~~  ^
```

droc12345:
jbeich had a patch that works, https://github.com/WayfireWM/wayfire/pull/1355#issuecomment-981090104
I actually went through the code bit by bit as the compiler/linker complained and wound up with effectively the same thing.
[...]

PR #1355, opened 2021-11-18
Track wlroots again

unrelentingtech:
#1351 and more (https://github.com/swaywm/sway/commit/5865af75cf8029cc703cda36b68daafcb658c97b)

hopefully the comment is correct :D

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -295,10 +295,22 @@ int main(int argc, char *argv[])
     core.argv = argv;
 
     /** TODO: move this to core_impl constructor */
-    core.display  = display;
-    core.ev_loop  = wl_display_get_event_loop(core.display);
-    core.backend  = wlr_backend_autocreate(core.display);
-    core.renderer = wlr_backend_get_renderer(core.backend);
+    core.display = display;
+    core.ev_loop = wl_display_get_event_loop(core.display);
+    core.backend = wlr_backend_autocreate(core.display);
+
+    int drm_fd = wlr_backend_get_drm_fd(core.backend);
+    if (drm_fd < 0)
+    {
+        LOGE("Failed to get DRM file descriptor!");
+        wl_display_destroy_clients(core.display);
+        wl_display_destroy(core.display);
+        return EXIT_FAILURE;
+    }
+
+    core.renderer  = wlr_gles2_renderer_create_with_drm_fd(drm_fd);
+    core.allocator = wlr_allocator_autocreate(core.backend, core.renderer);
+    assert(core.allocator);
     core.egl = wlr_gles2_renderer_get_egl(core.renderer);
     assert(core.egl);
 
--- a/src/api/wayfire/core.hpp
+++ b/src/api/wayfire/core.hpp
@@ -85,6 +85,7 @@ class compositor_core_t : public wf::object_base_t
      */
     wlr_backend *backend;
     wlr_renderer *renderer;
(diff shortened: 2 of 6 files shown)


unrelentingtech:
huh uncrustify just.. crashed?? in CI

src/main.cpp
```diff
@@ -298,7 +298,9 @@ int main(int argc, char *argv[])
     core.display  = display;
     core.ev_loop  = wl_display_get_event_loop(core.display);
     core.backend  = wlr_backend_autocreate(core.display);
-    core.renderer = wlr_backend_get_renderer(core.backend);
+    core.renderer = wlr_renderer_autocreate(core.backend);
```
ammen99, line 301: Can we directly create a GLES renderer here? That's the only thing we support.

unrelentingtech: Looks like it is possible, `wlr_gles2_renderer_create_with_drm_fd` is public, I'll try that now

droc12345:
Needed to apply these to the latest wlroots git. Wouldn't compile without them.

Seems to be working fine, as I'm running it now.

jbeich:
After [wlroots@1d3dd7f](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/1d3dd7fc08f2) this is no longer enough:
```c++
In file included from src/core/core.cpp:17:
src/api/wayfire/nonstd/wlroots-full.hpp:85:10: fatal error: 'wlr/backend/noop.h' file not found
#include <wlr/backend/noop.h>
         ^~~~~~~~~~~~~~~~~~~~
```

ammen99:
To avoid many changes being in the same PR, I think we can merge this and then use another PR for dropping the noop output as well.

@unrelentingtech can you format with uncrustify? (or rebase on top of master, then you'll get a diff in the github actions log)

unrelentingtech:
Now instead of uncrustify crashing, the diff command crashes in CI o_0

ammen99:
> Now instead of uncrustify crashing, the diff command crashes in CI o_0

It doesn't crash, it returns 1 because the code is not formatted :)
If you look at the previous command, it shows the diff you can apply.

jbeich:
> To avoid many changes being in the same PR, I think we can merge this and then use another PR for dropping the noop output as well.

noop fix is probably very simple e.g., https://github.com/freebsd/freebsd-ports/blob/d0c7d052f319/x11-wm/wayfire/files/patch-wlroots-0.15

ammen99:
> noop fix is probably very simple e.g., https://github.com/freebsd/freebsd-ports/blob/d0c7d052f319/x11-wm/wayfire/files/patch-wlroots-0.15

This fix is however not correct. Headless backend has also other uses than what the noop output had. For example, it is perfectly fine to render on a headless backend, if it is used for wayvnc or similar.

unrelentingtech:
well, "other uses" aren't really part of the *fix*, they're extra possibilities for feature development

CrezyDud:
when this

ammen99:
> when this

I just squashed two of the commits and then we can merge the PR. Do note that there are other changes necessary to support wlroots master, so even after merging this PR there is still work to do.

ammen99:
Thanks @unrelentingtech :)

CrezyDud:
> > when this
> 
> I just squashed two of the commits and then we can merge the PR. Do note that there are other changes necessary to support wlroots master, so even after merging this PR there is still work to do.

oof

PR #1362, opened 2021-11-22
core: close dev_null before running command

ericonr:
Otherwise an extra file descriptor for /dev/null will leak into
applications launched by Wayfire. In my case, it was fd 58.


ammen99 (approved):
Nice catch!

PR #1364, opened 2021-11-24
output: minor wording simplification.

Javyre:
Randomly came accross this while making bindings for wflua. IIUC, this wording should be equivalent.


ammen99 (approved):
LGTM, thanks!

PR #1365, opened 2021-11-24
Remove signal_callback_t in favor of signal_connection_t

### LINKED ISSUE #834: Remove deprecated `signal_callback_t`
`signal_callback_t` is already deprecated, and up to now we have had to fix numerous bugs due to forgetting to disconnect it.

`signal_connection_t` is the new signal connection mechanism which uses destructors to automatically disconnect. It is already used in new code. So we should just go through all places where `signal_callback_t` is used, replace it with signal_connection_t` and then remove the former completely from core.

l3nkz:
This PR is an implementation for #834

Fixes #834

### DIFF
--- a/plugins/animate/animate.cpp
+++ b/plugins/animate/animate.cpp
@@ -357,7 +357,7 @@ class wayfire_animation : public wf::singleton_plugin_t<animation_global_cleanup
     }
 
     /* TODO: enhance - add more animations */
-    wf::signal_callback_t on_view_mapped =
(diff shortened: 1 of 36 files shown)


ammen99 (approved):
LGTM, thanks!

ammen99 (changes requested):
Ah, forgot one thing. We should bump the plugin API/ABI in `plugin.hpp`

ammen99 (approved):
Aand nevermind, fixed it.

PR #1371, opened 2021-11-29
Implement switch to last workspace feature for vswitch plugin

### LINKED ISSUE #1296: Switch to last workspace
Hey!

Please describe what you want Wayfire to do:

What I'm sometimes missing from Wayfire, is a way to get back to the last visited workspace with a key combination. (Openbox calls this GoToDesktop 'last').

For example, I'm at Workspace 2, switch to Workspace 5, then have a key combination that would send me back to Workspace 2. Pressing it again would send me to Workspace 5, etc.

Thanks!

l3nkz:
Extend the vswitch plugin with bindings that allow switching back to the
last active workspace. The functionality is achieved by always
remembering the switching direction at each switch, so that we can
easily reverse it. There in now not only a binding for switching to the *last*
workspace, but also two other bindings: 1) for taking the current view
with when switching and 2) for only sending the currently focused view
to the *last* workspace.

Fixes #1296

### DIFF
--- a/plugins/vswitch/wayfire/plugins/vswitch.hpp
+++ b/plugins/vswitch/wayfire/plugins/vswitch.hpp
@@ -334,6 +334,24 @@ class control_bindings_t
 #undef SETUP
         /* *INDENT-ON* */
 
+        // Setup the bindings for switching back to the last workspace for the output
+        /* *INDENT-OFF* */ // Uncrustify has problems with this macro
+#define SETUP_LAST(name, view, only) \
+        wf::option_wrapper_t<wf::activatorbinding_t> binding_##name { \
+            "vswitch/"#name}; \
+        activator_cbs.push_back(std::make_unique<activator_callback>()); \
+        *activator_cbs.back() = [=] (const wf::activator_data_t&) \
+        {\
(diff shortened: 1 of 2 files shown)


plugins/vswitch/wayfire/plugins/vswitch.hpp
```diff
@@ -404,6 +422,8 @@ class control_bindings_t
     binding_callback_t user_cb;
     std::vector<std::unique_ptr<wf::activator_callback>> activator_cbs;
 
+    wf::point_t last_dir;
```
ammen99, line 425:
```suggestion
    wf::point_t last_dir = {0, 0};
```

ammen99 (commented):
As a whole, LGTM, please fix the formatting issues.

plugins/vswitch/wayfire/plugins/vswitch.hpp
l3nkz: Good catch! Fixed

l3nkz:
I updated the PR with your comments and checked that uncrustify does not complain for my code changes.

ammen99 (approved):
Thanks!

Issue #1137, opened 2021-04-07
Relative input for tablets (e.g. wacom)


mfxbe:
An basic tablet support seems already to be included in wayfire (see #732). But the documentation yet is not showing any possibility to allow changing the input mode from absolute to relative. The latter is the default in many environments and even recommended by [libinput](https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#relative-motion-for-tablet-tools). So I think there should be at least a possibility to enable such behavior.

For reference something similar has already been included [in sway](https://github.com/swaywm/sway/issues/4139).

PR #1374, opened 2021-12-05
Tablet motion options

mfxbe:
Fixes the Issue about motion of graphical tablets opened from me a while ago: #1137 

I'm not an C++ developer and also do not really know how the wayfire internals work but I made this changes for myself and they seem to work. So maybe they can be merged with a couple of adjustments here and there.


mfxbe:
Close for now. Will open a new one soon :-)

Issue #1381, opened 2021-12-16
Behaviour: Fullscreen windows occupy the next space


IngwiePhoenix:
I come from macOS and what I love doing is to send a window into fullscreen mode especially on a laptop to get as much screen-realestate as I can out of them.

However, after configuring the fullscreen shortcut, I quickly noticed that it stood on the same space. So, I thought it might be a good idea to request this behaviour. When an application goes into fullscreen - either being made to by a shortcut or maybe even on its own -  it would be moved to the next space instead.

IngwiePhoenix:
Ah, yeah, workspace... Sorry, 12 years of Apple doesn't leave you that fast ;)

Awesome to hear! That would be really neat to have.

unrelentingtech:
@ammen99 it's way more than that.

macOS uses dynamic workspaces rather than a fixed grid. Fullscreen apps **act like workspaces themselves**.

I guess you can think of it as:

- fullscreening a surface creates a *special* workspace for it, to the right of the current one
  - and switches to it of course
- the special workspace is completely tied to that surface being full-screen: when it's closed/unfullscreened, that workspace is gone
  - if it's just unfullscreened, it goes back to the workspace it was fullscreened from
  - minimize is typically not allowed IIRC but I guess it's fine to treat it just like closing too

https://youtu.be/L1xsPlg96xc?t=139

---

You could technically "just" move&switch to the next workspace in our fixed grid but I think that's just a VERY weird thing to do. What if you're on the rightmost workspace?

unrelentingtech:
Actually since #1063 our workspaces aren't necessarily static, so it would be fine to create a plugin that would create and manage these transient workspaces. I might even try.

ammen99:
> Actually since #1063 our workspaces aren't necessarily static, so it would be fine to create a plugin that would create and manage these transient workspaces. I might even try.

It shouldn't be too hard ;)

In any case, I don't think this needs to be part of the core plugins. We can't accomodate every feature, so closing this, of course feel free to share a link to your plugin if you end up implementing this.

IngwiePhoenix:
Where do I find the plugin API? I might just give it a shot :)

I know the code is C++ - but does it neccessarily need to be in C++ or are there C bindings?

soreau:
> Where do I find the plugin API? I might just give it a shot :)
> 
> I know the code is C++ - but does it neccessarily need to be in C++ or are there C bindings?

[...]

PR #1384, opened 2021-12-16
expo: ask the workspace_implementation about moveability

unrelentingtech:
Something found while working on a #1381 related plugin. I wonder how nobody found this when working on tile :)

re: code style — I think skipping over `update_target_workspace(to.x, to.y);` with the `return` is fine, and the whole `if (view)` thing could've been ` if (!view) return;` to save an indent level, right?…


ammen99 (approved):
Seems simple enough, thanks!

PR #1387, opened 2021-12-17
seat: add touchpad hold gesture support

unrelentingtech:
Catching up to wlroots: https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/d069a783bcafe65c240f2369b68f5e8d45d0b9a2 (we depend on a newer revision I'm pretty sure)


ammen99 (approved):
Thanks!

PR #1388, opened 2021-12-18
Add ability for plugins to avoid touchpad gesture conflicts

valpackett:
(INCLUDES #1387)

Extend the `carried_out` pattern to input signals.

So… yeah, this is necessary for something really cool with touchpad gestures :)

UPD: by which I mean that I have [3-finger drag](https://github.com/DankBSD/wf-touchpad-gesture-drag) not conflicting with 3-finger vswipe! (and the same will extend to [globalgestures](https://github.com/DankBSD/wf-globalgestures))

### DIFF
--- a/src/api/wayfire/signal-definitions.hpp
+++ b/src/api/wayfire/signal-definitions.hpp
@@ -83,6 +83,7 @@ enum class input_event_processing_mode_t
  *   pointer_motion, pointer_motion_absolute, pointer_button, pointer_axis,
  *   pointer_swipe_begin, pointer_swipe_update, pointer_swipe_end,
  *   pointer_pinch_begin, pointer_pinch_update, pointer_pinch_end,
+ *   pointer_hold_begin, pointer_hold_end,
  *   keyboard_key,
  *   touch_down, touch_up, touch_motion,
  *   tablet_proximity, tablet_axis, tablet_button, tablet_tip
--- a/src/api/wayfire/signal-definitions.hpp
+++ b/src/api/wayfire/signal-definitions.hpp
@@ -118,6 +118,16 @@ struct input_event_signal : public wf::signal_data_t
      * pointer_button, keyboard_key, touch_down
      */
     input_event_processing_mode_t mode = input_event_processing_mode_t::FULL;
+
+    /**
+     * Whether the event has already been handled.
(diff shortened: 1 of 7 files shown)


FriederHannenheim:
Any update on this?

unrelentingtech:
@FriederHannenheim all PRs are on hold waiting for the big stabilize-api refactor to land.

src/api/wayfire/signal-definitions.hpp
```diff
@@ -114,9 +115,22 @@ struct input_event_signal : public wf::signal_data_t
...
+
+    /**
+     * Whether the event has already been handled.
+     *
+     * This is useful for e.g. preventing conflicts between different plugins
+     * handling touchpad gestures.
+     *
+     * Setting this flag also implicitly sets the mode to NO_CLIENT.
+     */
+    bool carried_out = false;
```
ammen99, line 133: As we already have an input processing mode, I was wondering whether it won't be simpler to add an additional processing mode `IGNORE` ?

unrelentingtech: maybe. But this is more consistent with other signals and possibly safer against unintentional modification in other places (since it's a new flag with only this meaning)?

ammen99: Looking at this again, I think you are right. Anyway, I think it is better to remove the `implicitly sets ...` bit, plugins should do that if they so desire, but there is no possibility for opt-out as it is currently.

ammen99: Looking at this a third time, I believe extending the input processing mode is better. I'll send a PR for that as I need it for something else (pointer buttons), and then if you still need this feature, feel free to send a new PR with the new gestures and handling for the gestures themselves. The relevant refactor is already over, so a potential PR with such changes could be merged now.

ammen99: FTR: #1804

valpackett: Hah, sorry, I haven't been following the project lately… sometime in the future I might come back…

ammen99:
Closing because of staleness. Feel free to resend in the future.

PR #1396, opened 2021-12-30
view_3D: use a sensible default for depth scale

### LINKED ISSUE #1359: Switcher crash when closing a window
**Describe the bug**
alt tab plugin bad and crashes when a window closes while alt tabbing
sometimes it goes to the wrong window when alt tabbing
FIX IT ALREADY

**To Reproduce**
1. Enable plugin X
2. Do Y
3. Z happens

**Expected behavior**

**Screenshots or stacktrace**

**Wayfire version**
0.5.0, git, package, something else?

dkondor:
Fixes #1359 

This is a workaround that avoids a segfault if output is null by using a default value for "size". I'm assuming that a view without an output will not be rendered, so the actual value does not really matters.

**Notice: Wayfire's development is temporarily happening in the `stabilize-api` branch. If you open a pull request for the master branch, it will likely not be merged before the stabilize-api branch is itself merged into master.**

### DIFF
--- a/src/view/view-3d.cpp
+++ b/src/view/view-3d.cpp
@@ -198,9 +198,15 @@ wf::view_3D::view_3D(wayfire_view view, uint32_t z_order_) : z_order(z_order_)
 /* TODO: cache total_transform, because it is often unnecessarily recomputed */
 glm::mat4 wf::view_3D::calculate_total_transform()
 {
-    auto og = view->get_output()->get_relative_geometry();
-    glm::mat4 depth_scale =
-        glm::scale(glm::mat4(1.0), {1, 1, 2.0 / std::min(og.width, og.height)});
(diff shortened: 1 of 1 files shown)


ammen99:
Looking at this again, I have the feeling we must be doing something wrong, but I don't know what :/ This seems more like a hack to get it working, not a real solution (the current state of the code isn't any better)

dkondor:
My intuition is that the correct value for the depth scale should depend on the relation between the distance the user's eyes are from the screen and the physical size of the display. But since the output is 2D content, there is likely a broader range of values that look reasonably well, so maybe any dependence on screen size could be avoided?

ammen99:
I submitted a slightly better fix in #1883, thanks for this PR nonetheless :)

PR #1400, opened 2022-01-06
Fix window position error

ChaojiangLuo:
The window is positioned by client, but it will be placed to center by
the place plugin. Set is_positioned to place plugin to avoid this issue.

Signed-off-by: Chaojiang Luo <luochaojiang@uniontech.com>
Change-Id: I925ced36e9750667afc3c196db98e0e050b3c0c7

**Notice: Wayfire's development is temporarily happening in the `stabilize-api` branch. If you open a pull request for the master branch, it will likely not be merged before the stabilize-api branch is itself merged into master.**

### DIFF
--- a/src/api/wayfire/view.hpp
+++ b/src/api/wayfire/view.hpp
@@ -229,6 +229,8 @@ class view_interface_t : public surface_interface_t
      * If the view is tiled to all edges, it is considered maximized. */
     uint32_t tiled_edges = 0;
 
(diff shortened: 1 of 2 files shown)


ammen99:
Hi, I know a long time has passed, but is this still relevant? In what cases is the position set by the client? Xwayland already has a code for tracking this, and with xdg-shell, it is not possible for clients to position themselves.

PR #1409, opened 2022-01-14
Avoid using [] which implicitly inserts.

Javyre:
`[]` should be avoided where possible as it causes things that look like
simple lookups to actually perform insertions.

`[]` replaced with explicit operators where possible.

`[] = ...` assignments aren't too confusing but `try_emplace()` is much
clearer.

C++ magic looks very pretty but can be very dangerous.


ammen99:
Closing as stale .. Thanks for the PR nonetheless ;)

PR #1410, opened 2022-01-16
add keepabove window rule

TRPB:
I've been trying to control the Z-order of various UI elements. In particular, I've been trying to get Waybar to appear over `wf-dock`. 

This allows setting up a window rule as such:

```
rule_1 = on created if app_id is "waybar" then keepabove
```

This isn't quite the same as the shortcut key as it sets `wf::LAYER_TOP` to force the window to the top. The keypress which targets workspace windows sets `wf::WORKSPACE` and `SUBLAYER_DOCKED_ABOVE`. 

This might need a different name? E.g. 

```
rule_1 = on created if app_id is "waybar" then toppanel # uipanel  depanel or whatever
```

### DIFF
--- a/plugins/window-rules/view-action-interface.cpp
+++ b/plugins/window-rules/view-action-interface.cpp
@@ -77,6 +77,11 @@ bool view_action_interface_t::execute(const std::string & name,
         _unminimize();
 
         return false;
+    } else if (name == "keepabove")
+    {
+        _keepabove();
+
+        return false;
     } else if (name == "snap")
     {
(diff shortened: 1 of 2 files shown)


TRPB:
just a thought, would it be better going forward to support something like:

```
on created if app_id is "....." then layer top above
on created if app_id is "....." then layer workspace above
on created if app_id is "....." then layer top below
on created if app_id is "....." then layer background above
```

it would add more flexibility for customisation, thinks like desktop widgets and the like.

ammen99:
Just realized this feature is available on master since quite some time, nonetheless thanks for sending this PR!

PR #1417, opened 2022-01-24
view: use focus_request for self-requests from xwayland and gtk-shell

valpackett:
The code in xwayland and gtk-shell actually would not carry the request out!
Let's just use focus_request (which now has an argument for the self flag) rather than duplicating the carrying-out code.

### DIFF
--- a/plugins/protocols/gtk-shell.cpp
+++ b/plugins/protocols/gtk-shell.cpp
@@ -9,6 +9,7 @@
 #include <wayfire/nonstd/wlroots-full.hpp>
 #include <wayfire/plugin.hpp>
 #include <wayfire/toplevel-view.hpp>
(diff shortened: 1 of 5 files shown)


src/api/wayfire/view.hpp
```diff
@@ -243,7 +243,7 @@ class view_interface_t : public surface_interface_t
     /** Request that an interactive move starts for this view */
     virtual void move_request();
     /** Request that the view is focused on its output */
-    virtual void focus_request();
+    virtual void focus_request(bool self_request = false);
```
ammen99, line 246: Might be a good idea to bump the API/ABI version, 'just in case' :)

ammen99 (approved):
Rebased the changes to fit master, thanks :)

PR #1425, opened 2022-01-31
place: Add "nearmouse" mode for placing under cursor

yobert:
Add a new mode for placing new windows, where it will attempt to be placed centered under the mouse cursor. If it would run off the edge of the screen it is clamped. I think this mode is a lot friendlier than the default of placing in the center. Let me know what you think!

I saw your note about the stabilize-api branch--- but I couldn't get that one to compile. It's okay if this takes a while to merge though! No hurry.

I really love this project you've made. Thank you for your hard work and I hope I can help with it.

### DIFF
--- a/metadata/place.xml
+++ b/metadata/place.xml
@@ -24,6 +24,10 @@
 				<value>random</value>
 				<_name>Random</_name>
 			</desc>
+			<desc>
+			  <value>nearmouse</value>
+			  <_name>Near Mouse</_name>
+			</desc>
 		</option>
 	</plugin>
 </wayfire>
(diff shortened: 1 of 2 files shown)


yobert:
One note-- I don't fully understand all the coordinate systems used in Wayfire. It's possible I'm doing it wrong when there are multiple outputs.

Nuc1eoN:
Sounds like it could be a nice experience! :)

One think MS Windows does to prevent windows stacking directly on top of each other, is spawning the next window just a few pixel shifted, diagonally to the bottom right. Something similar might be considered for the "standard" behavior, since stacking windows exactly on top of each other is not very user friendly.

Nuc1eoN:
image

This is how Windows does it, interesting, huh?

yobert:
Yeah! makes sense. This feature idea as it is comes from Fluxbox, my favorite window manager from Ye Olde Xorg. I like to move my mouse to one of my corners when I launch a terminal and have the new window come up in that corner, snug against the edge.

yobert:
Closing for now just to remove cruft. I'll try again in the future with some fancy features.

PR #1430, opened 2022-02-12
Updates for latest master of wlroots

yobert:
- wlr_output_layout_get_box (now takes box as param)

- wlr_xdg_toplevel_set_fullscreen_event has been replaced by fields in
  struct wlr_xdg_toplevel_requested

I understand if you want to close this and not follow wlroots' master quite so closely. This fixes a crash for me on nvidia proprietary (wlr_renderer_init_wl_display: Assertion `argb8888 && xrgb8888" failed). I think the good commit comes from https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/59b9518f072527ac59593e51df7f5d5331a34f0e so I was hoping we could get at least to that new.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -1423,8 +1423,9 @@ class output_layout_t::impl
                 wlr_output_layout_add_auto(output_layout, handle);
 
                 /* Get the correct position */
-                auto box = wlr_output_layout_get_box(output_layout, handle);
-                assert(box);
+                wlr_box box;
+                wlr_output_layout_get_box(output_layout, handle, &box);
+                assert(!wlr_box_empty(&box));
                 lo->apply_state(state);
             }
         }
@@ -1473,11 +1474,12 @@ class output_layout_t::impl
             auto head = wlr_output_configuration_head_v1_create(
                 wlr_configuration, output.first);
 
-            auto box = wlr_output_layout_get_box(output_layout, output.first);
-            if (box)
+            wlr_box box;
+            wlr_output_layout_get_box(output_layout, output.first, &box);
+            if (!wlr_box_empty(&box))
             {
-                head->state.x = box->x;
-                head->state.y = box->y;
+                head->state.x = box.x;
+                head->state.y = box.y;
             }
         }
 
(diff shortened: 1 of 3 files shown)


droc12345:
src/core/output-layout.cpp - line 1428 (new)
 assert(!wlr_box_empty(&box));

This caused my system to be in a weird state, so I added a debug line to see what was going on
and the box is "considered" empty (at this point ie. starting up) with 0, 0, 0, 0, 
but the assert never triggered, it just stayed in a half-gui/half-console mode where I had to ssh in to kill wayfire.

When I commented out the assert, then it came up and ran fine.

yobert:
Yes interesting--- I'm getting that now too on a different video card. (Intel Tiger Lake Iris XE). It's interesting because it should have been failing the assert before, if I understand the wlroots upstream change. I'll keep investigating.

yobert:
Commenting out the assert always seems like a great fix, but I'd like to wait for someone who understands it better maybe to chime in :laughing:

droc12345:
Not sure what to say on it, according to the include file wlr_box_empty is a bool that is supposed to trigger with either width or height being zero. 
---
Checks whether a box is empty or not.
A wlr_box is considered empty if its width and/or height is zero or negative.
---
and assert should trigger off on wlr_box_empty returning zero 

But I didn't get an assert reaction, I didn't get anything, might be something from the display area, video card (rx 550) or display LG 4k monitor having some weird interaction with the wlr* function call.  *scratches head*

droc12345:
As as follow on to the "ramblings" from the last post

I don't think assert is needed, in the old way, box was a pointer and the assert was checking to make sure the box really pointed to a valid struct. 

With the new way, box is a struct (that exists) so you know that assert/wlr_box_empty will never see just a pointer, without a struct. In other words it's a meaningless check (in this case).

caniko:
Will this make Wayfire available to my single nvidia GPU PC? I got a 1080ti

ammen99:
> Will this PR make Wayfire available to my single nvidia GPU PC? I got a 1080ti

The real problem is GBM support in the nvidia proprietary driver. Pretty sure even wayfire 0.4.0 would work if you have the newest nvidia drivers which do support GBM.

yobert:
Closing for now--- wlroots has many more updates. I'll have a new PR up soon with fixes.

PR #1474, opened 2022-05-09
Update wobbly license

### LINKED ISSUE #1470: Unclear (non-free?) license for plugins/wobbly/wobbly.h file
Hi,

I was reviewing the code base of Wayfire project, which is a prerequisite of packaging this software in Debian/Ubuntu. While reviewing the code licenses, I noticed the following declaration for `plugins/wobbly/wobbly.h` file:

https://github.com/WayfireWM/wayfire/blob/94f15048fe91fde0bc65126306ab2dc13648c148/plugins/wobbly/wobbly.h#L1-L6

This certainly makes the source code no longer free (as defined by open source software).

Luckily I believe the original author is still around and should be @soreau . I have no idea where did this source code come from; it clearly was introduced in https://github.com/WayfireWM/wayfire/commit/83653e530b1796995ff165f317dc50d0f50716d7 , but the files seem to have been copied from elsewhere.

@soreau , if you were the author and the header file was indeed released under "All Rights Reserved", is it possible to re-license it to make it compatible with Wayfire (e.g., using MIT license as in `plugins/wobbly/wobbly.c`)? If not, is there any possibility to migrate away from such code base?

Thank you very much in advance!

soreau:
Fixes #1470

### DIFF
--- a/plugins/wobbly/wobbly.h
+++ b/plugins/wobbly/wobbly.h
@@ -1,9 +1,26 @@
-/**************************************************************************
+/*
+ * The MIT License (MIT)
  *
- * Copyright 2014 Scott Moreau <oreaus@gmail.com>
- * All Rights Reserved.
+ * Copyright (c) 2014 Scott Moreau
  *
- **************************************************************************/
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
LGTM (obviously), thanks! The CI failure is fixed in master, so I'll just merge.

ammen99:
Oh wait, this is for stabilize-api .. still fine :)

PR #1496, opened 2022-06-08
input: allow directly specifying a keymap file

occivink:
A new `xkb_keymap_file` option is added, with which a complete keymap
can be specified.
If this option is set, the keymap is loaded from the corresponding
file and the other xkb options  (rules, mnodel, layout, variant
and options) are ignored.

### DIFF
--- a/src/core/seat/keyboard.cpp
+++ b/src/core/seat/keyboard.cpp
@@ -62,6 +62,7 @@ wf::keyboard_t::keyboard_t(wlr_input_device *dev) :
     layout.load_option("input/xkb_layout");
     options.load_option("input/xkb_options");
     rules.load_option("input/xkb_rules");
+    keymap_file.load_option("input/xkb_keymap_file");
 
     repeat_rate.load_option("input/kb_repeat_rate");
     repeat_delay.load_option("input/kb_repeat_delay");
@@ -73,6 +74,7 @@ wf::keyboard_t::keyboard_t(wlr_input_device *dev) :
     layout.set_callback([=] () { this->dirty_options = true; });
     options.set_callback([=] () { this->dirty_options = true; });
     rules.set_callback([=] () { this->dirty_options = true; });
+    keymap_file.set_callback([=] () { this->dirty_options = true; });
     repeat_rate.set_callback([=] () { this->dirty_options = true; });
     repeat_delay.set_callback([=] () { this->dirty_options = true; });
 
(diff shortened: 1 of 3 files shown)


src/core/seat/keyboard.cpp
```diff
@@ -106,29 +108,54 @@ void wf::keyboard_t::reload_input_options()
...
+        if (file)
+        {
+            keymap = xkb_keymap_new_from_file(ctx, file,
+                XKB_KEYMAP_FORMAT_TEXT_V1,
+                XKB_KEYMAP_COMPILE_NO_FLAGS);
+            std::fclose(file);
+            if (!keymap)
+            {
+                LOGE("Could not create keymap from file \"", path, "\"");
+            }
```
soreau, line 126: Opportunity to branch for success message here?

occivink: Why not, but is this really useful? Other options don't log anything either when they're successfully set.

soreau: I suppose it would be a way to know if the other options are being ignored at least.

src/core/seat/keyboard.cpp
```diff
@@ -106,29 +108,54 @@ void wf::keyboard_t::reload_input_options()
...
-        XKB_KEYMAP_COMPILE_NO_FLAGS);
+    struct xkb_keymap *keymap = nullptr;
 
-    if (!keymap)
+    if (!this->keymap_file.value().empty())
+    {
+        std::string path = this->keymap_file;
+        FILE *file = std::fopen(path.c_str(), "r");
+        if (file)
+        {
```
ammen99, line 118: I think it would make sense to split the two branches into separate helper functions.

occivink:
Unless there is any interest in it, I'll probably drop the PR. I've solved the issue by storing my configuration as RMLVO in `~/.config/xkb` and using that in my `wayfire.ini`, this feels more in line with the intended configuration of xkb than adding a keymap file option. If anybody is curious, my (relatively minimal) xkb configuration is published in my [dotfiles](https://github.com/occivink/config/tree/master/.config/xkb)
Thanks for the feedback nonetheless

ammen99:
> Unless there is any interest in it, I'll probably drop the PR. I've solved the issue by storing my configuration as RMLVO in `~/.config/xkb` and using that in my `wayfire.ini`, this feels more in line with the intended configuration of xkb than adding a keymap file option. If anybody is curious, my (relatively minimal) xkb configuration is published in my [dotfiles](https://github.com/occivink/config/tree/master/.config/xkb) Thanks for the feedback nonetheless

I didn't know that XKB allows you to add this to your home directory. With such functionality, I also think adding this to Wayfire doesn't make sense. Thank you for the PR nonetheless :)

Issue #1320, opened 2021-10-15
Remove/refactor grab APIs


ammen99:
The plugin grab APIs are a bit weird right now, and they duplicate functionality provided by signals for input handling. It might make more sense to remove the grab API and instead use the raw events + make sure that plugins can block all handling without listening to all possible input events.

The input capability should however still be required for doing this, in order to ensure that a single plugin holds such an exclusive grab at a time.

This can be easily done by moving grabs to be scene nodes above the other nodes. We need to be careful with grabs like move, which need to get a release event without the corresponding press (needs to be special cased in the focus hand-off)

PR #1501, opened 2022-06-14
Scenegraph

ammen99:
Here is another part of the upcoming refactor in Wayfire. It is still very WIP, so take any of the current APIs with a grain of salt.

Goals:

- Remove special cases in rendering and input handling as much as possible. To that purpose, the changes add a new scene graph to Wayfire's API. Eventually, this scenegraph should contain almost everything related to input and output in Wayfire. Examples of special cases intended to be made non-special by this code:
  - Drag and drop icons. They are currently hardcoded as special views visible on multiple outputs.
  - Custom renderers and input grabs (to be done later in #1320)
  - ~~Bindings in general~~
  - ~~Focused layers~~ Unsure how this will work out.
  - ~~Shortcuts inhibition (once this lands)~~ Seems quite complex for now, we may want to add some special cases still.
 
  All of this is already possible, but is hidden in core behind special-cases and special code paths for handling these things. The scenegraph should contain this state in a way which plugins can understand and manipulate.
- The scenegraph should also lift the restriction of views being visible on a single output only (although it will remain the default). Views, however, will remain with a single associated 'main output'. Plugins which desire to allow views visible on multiple outputs will need to ensure that workspace operations, etc. work correctly.

- Decouple the management of input and output elements (scenegraph nodes) from management of other window-manager specific state (geometry, maximized/fullscreen/minimized, workspaces, etc).

These changes will also help with the upcoming transactions API, workspace sets and the refactor of the view and surface interfaces.

Plan for this PR (mostly notes to myself):

- [x] Build the scenegraph from workspace-manager
  - [x] Basic views
  - [x] Dialogs
  - [x] Fullscreen promoted views 
- [x] Migrate basic input handling to the scenegraph
- [x] Reimplement basic rendering with the scenegraph
- [x] Remove sublayers, restacking, etc. functionality from the workspace-manager, as this is available via the scenegraph API
- [x] Move DnD icons to the scenegraph
- [x] Measure performance and see whether optimizations are necessary
- [x] Restore restack_below/above functionality for wm-actions
- [ ] Consider exposing wf::seat_t and piping it to nodes for input events


soreau:
In firefox, popups always appear flickering at the upper left of the output, until they stop flickering. The input region seems to be correct but the popup is somewhat unresponsive, it will stick on one item for example.. just open firefox and click on the hamburger icon to see.

soreau:
Blur performance is so bad that it is nearly unusable on modest settings that work well on current master.

ammen99:
Another bug: switcher + blur causes visual glitches after exiting switcher.

soreau:
I can crash wayfire with [this bt](http://ix.io/4bW0) by enabling blur, disabling it and repeating once more. When I disable it the second time, wayfire crashes.

ammen99:
A bug reported by @soreau: http://ix.io/4c0l
Probably happens when blur is active and move snap preview is about to be shown.

soreau:
Pointer constraints are not working for fullscreen applications that use it.

soreau:
All plugins need porting, some are not ported yet (such as idle/screensaver).

soreau:
The input region of wf-panel menu widget is off, moved up about the height of the panel.

soreau:
Wayfire decorations appear but are not clickable.

soreau:
wf-recorder cannot complete video on ctrl+c in scenegraph.

PR #1512, opened 2022-06-27
workspace-impl: Fix get_view_main_workspace()

soreau:
Offset by current workspace to get the correct result.


ammen99 (approved):
Thanks!

PR #1529, opened 2022-07-27
simplify run() to use daemon(3)

epsilon-0:
Signed-off-by: Aisha Tammy <aisha@bsd.ac>

### DIFF
--- a/plugins/wobbly/wobbly.h
+++ b/plugins/wobbly/wobbly.h
@@ -1,9 +1,26 @@
-/**************************************************************************
+/*
+ * The MIT License (MIT)
  *
- * Copyright 2014 Scott Moreau <oreaus@gmail.com>
- * All Rights Reserved.
+ * Copyright (c) 2014 Scott Moreau
  *
- **************************************************************************/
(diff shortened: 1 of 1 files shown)


ammen99:
This makes it not possible to track the PID of the started process. This doesn't work with dbus-activated apps, true, but it is useful when you know what client you are starting. Can we somehow still pipe the PID to the compositor with `daemon(3)`?

epsilon-0:
I don't see the return value from `run(...)` being used anywhere right now, so the information is lost in any case. It would be easier to log the `command` being run instead of the PID, as the command itself might do some internal daemonize stuff and change PIDs.

ammen99:
> I don't see the return value from `run(...)` being used anywhere right now, so the information is lost in any case. It would be easier to log the `command` being run instead of the PID, as the command itself might do some internal daemonize stuff and change PIDs.

I use it in the scenegraph branch for telling the tests (which call run() via IPC) the PID of the started process. I know it doesn't work for everything, but it does work when you know in advance that the processes you are starting don't do advanced stuff.

epsilon-0:
moved to https://github.com/WayfireWM/wayfire/pull/1596

Issue #217, opened 2019-05-05
Introduce sets of 2D workspaces


ammen99:
I just had an idea that we can have multiple sets of 2D workspaces, which could then be moved between outputs.

This can even be made a plugin, so in theory very little (or none) core changes would be needed.

ammen99:
On a second thought, doesn't look too useful, and doesn't fit with the main plugins anyway.

ammen99:
On a third thought, this makes sense. This can be a solution to saving and restoring output state on unplug and replug, moving whole workspace sets between outputs (kinda what [this plugin](https://github.com/ammen99/wayfire-plugins/blob/master/src/primary-monitor-switch.cpp) attempts to do)

It is also the way to provide independent workspaces without special-casing everywhere (#977) - instead of using workspaces, a plugin who wants to provide independent workspaces can just use different workspace sets with a single workspace.

The implementation might be a bit tricky, and will potentially require splitting the current `workspace_manager` into `workspace_set` and a manager.

Javyre:
@ammen99 How would you see plugins dealing with workspace sets moving from one output to another? Would it be on a view-by-view basis that it should accept the incoming views on the output still? This would break tiling layouts on the workspace sets for example. Or should the plugin instances be by workspace set now instead of by output?

ammen99:
Plugins will be free to store their data on the workspace set or per output, depending on how they want to operate.

PR #1535, opened 2022-08-02
don't remove DPMS off outputs

lilydjwg:
After I lock my screen and [turn off my monitors with DPMS to save power](https://github.com/lilydjwg/dpms-off), the output will first be turned on, then off, then on again. This makes the display flickers, the screen locker fails to receive all my keystrokes, and windows move back and forth which necessitate a resize and sometimes a redraw.

This is because that wayfire removes outputs when power saving and adds them back when resuming. By only disabling the outputs but not removing them from layout this flickering can be avoid.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -1375,7 +1375,8 @@ class output_layout_t::impl
             auto& state  = entry.second;
             auto& lo     = this->outputs[handle];
 
-            if (!(state.source & OUTPUT_IMAGE_SOURCE_SELF))
+            if (!(state.source & OUTPUT_IMAGE_SOURCE_SELF ||
+                  state.source & OUTPUT_IMAGE_SOURCE_DPMS))
             {
                 /* First shut down the output, move its views, etc. while it
                  * is still in the output layout and its global is active.
@@ -1384,6 +1385,17 @@ class output_layout_t::impl
                  * wl_surface.leave events for the to be destroyed output */
                 lo->apply_state(state);
                 wlr_output_layout_remove(output_layout, handle);
+            }
+        }
+
+        /* Disable DPMS state outputs */
+        for (auto& entry : config)
+        {
+            auto& handle = entry.first;
+            auto& state  = entry.second;
+            if (state.source & OUTPUT_IMAGE_SOURCE_DPMS)
+            {
+                wlr_output_enable(handle, false);
             }
         }
 
@@ -1399,6 +1411,7 @@ class output_layout_t::impl
                 !entry.second.position.is_automatic_position())
             {
                 ++count_enabled;
+                wlr_output_enable(handle, true);
                 wlr_output_layout_add(output_layout, handle,
                     state.position.get_x(), state.position.get_y());
                 lo->apply_state(state);
@@ -1420,6 +1433,7 @@ class output_layout_t::impl
                 entry.second.position.is_automatic_position())
             {
                 ++count_enabled;
+                wlr_output_enable(handle, true);
                 wlr_output_layout_add_auto(output_layout, handle);
 
                 /* Get the correct position */
@@ -1438,6 +1452,7 @@ class output_layout_t::impl
 
             if (state.source == OUTPUT_IMAGE_SOURCE_MIRROR)
             {
+                wlr_output_enable(handle, true);
                 lo->apply_state(state);
                 wlr_output_layout_remove(output_layout, handle);
             }


src/core/output-layout.cpp
```diff
@@ -1375,7 +1375,8 @@ class output_layout_t::impl
             auto& state  = entry.second;
             auto& lo     = this->outputs[handle];
 
-            if (!(state.source & OUTPUT_IMAGE_SOURCE_SELF))
+            if (!(state.source & OUTPUT_IMAGE_SOURCE_SELF ||
```
ammen99, line 1378: This is not necessary if you look at the enum value definitions, I admit this is confusing. `DPMS` is nothing else but `SELF | OFF`, so that a dpms output is NOT removed here, even with the current master version.

src/core/output-layout.cpp
```diff
@@ -1387,6 +1388,17 @@ class output_layout_t::impl
...
         }
 
+        /* Disable DPMS state outputs */
+        for (auto& entry : config)
+        {
+            auto& handle = entry.first;
+            auto& state  = entry.second;
+            if (state.source & OUTPUT_IMAGE_SOURCE_DPMS)
+            {
+                wlr_output_enable(handle, false);
```
ammen99, line 1398: `wlr_output` state is double-buffered, so you need a `wlr_output_commit` after this. In general though, I do not see why you need this.

ammen99 (commented):
In general, I don't like moving the enable/disable logic where you have put it, the `apply_state` of each output is where the stuff should be happening. Also, have you tested your patch, does DPMS actually work? I have the feeling the problem you are having is likely the problem where the hardware output disconnects itself (or maybe it is the driver, unsure) when DPMS is active. Wayfire then takes this to mean 'DPMS output was physically removed` and destroys the output. For comparison, DPMS works (with master branch) without the output being reset on my hardware.

The problem you are likely trying to solve (correct me if the above is wrong) has been reported multiple times, and one proposed solution is the preserve-output plugin which attempts to restore the layout of views when something like this happens.

A better solution will become possible in 0.8.0 when workspace sets (#217) are implemented, as then you can just restore the workspace set on the output after it is readded when DPMS ends.

src/core/output-layout.cpp
lilydjwg: Oh I missed that. But weird it worked, not very well though. I've got a crash just now. I'll check the code more carefully.

lilydjwg:
The code is wrong and crashes. Sorry for the noise.

ammen99:
> 

No worries, this is a difficult problem, which is why it has not been resolved already :)

lilydjwg:
So you mean with the workspace set stuff, the windows can be avoided to be moved back and forth in this case? That sounds a good solution. I'm thinking about delaying handling the output remove event.

Another issue this causes for me is that waybar will be removed and re-added, causing a few more zombie processes (https://github.com/Alexays/Waybar/issues/837) and some HTTP requests to be sent out unnecessarily (which may fail). The CPU indicator also wrongly indicates a high load for a while.

ammen99:
The workspace set stuff will fix the issue for regular windows. Waybar is a layer-shell surface, so it will have to be recreated, I suggest fixing the bug in Waybar instead.

> I'm thinking about delaying handling the output remove event.

layer-shell surfaces are hardcoded to be only on one `wlr_output`. As long as wlroots destroys the `wlr_output`, they have to be closed, thus this solution requires cooperation from wlroots. The compositor will have to keep the `wl_output` (the wayland protocol object, not the `wlr_output`) artificially until DPMS ends, then rescan connectors and potentially map the new `wlr_output` to the old `wl_output`. AFAIK Sway should also suffer from the same problem, at least with Waybar, you might want to ask in #sway-devel to see whether the other wlroots developers find this worth pursuing.

lilydjwg:
> The workspace set stuff will fix the issue for regular windows. Waybar is a layer-shell surface, so it will have to be recreated, I suggest fixing the bug in Waybar instead.

I see. Unfortunately waybar is also written in C++ which I'm less experienced with. I'm even thinking about writing my own bar program but there are a lot of things to fix in Wayland...

> AFAIK Sway should also suffer from the same problem, at least with Waybar, you might want to ask in #sway-devel to see whether the other wlroots developers find this worth pursuing.

I'll try with sway.

szbergeron:
Is there any reason that requires you to set DPMS::Off instead of Standby/Suspend?

PR #1543, opened 2022-08-13
vswitch: Fix vswitch with multiple views

soreau:
Before, vswitch could only switch workspaces with a single view at a time.
This makes it possible to switch with multiple views.


ammen99 (commented):
Seems like this would need quite an update to work with the new API if you want to work on this .. Otherwise, we should close the PR.

Do you happen to have a test case where this is actually needed?

PR #1544, opened 2022-08-13
Add release binding feature to core command plugin

veRYANgry:
Creates another type of command for the command plugin called release. On release of the binding the command will be called, instead of on press. This allows for things like Push to talk (unmute then mute) using a single button.

Tested by running this for weeks using it to trigger push to talk on Arch Linux.

### DIFF
--- a/metadata/command.xml
+++ b/metadata/command.xml
@@ -45,5 +45,18 @@
 				<_long>Sets the triggering activator binding.</_long>
 			</entry>
 		</option>
+		<option name="release_bindings" type="dynamic-list" type-hint="dict">
+			<_short>Realease bindings</_short>
+			<_long>Sets the bindings which activate on release</_long>
+			<entry prefix="command_" type="string" name="command">
+				<_short>Command</_short>
+				<_long>Sets the command to execute.</_long>
+				<hint>file</hint>
+			</entry>
+			<entry prefix="release_binding_" type="activator" name="binding">
+				<_short>Binding</_short>
+				<_long>Sets the triggering activator binding.</_long>
+			</entry>
+		</option>
 	</plugin>
 </wayfire>
--- a/plugins/single_plugins/command.cpp
+++ b/plugins/single_plugins/command.cpp
@@ -24,7 +24,7 @@ static int repeat_once_handler(void *callback)
 
 /* Provides a way to bind specific commands to activator bindings.
  *
- * It supports 2 modes:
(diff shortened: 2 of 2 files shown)


ammen99 (approved):
The idea looks good to me. I'll let the CI run but I think you'll need to fix the code style. The CI's output contains a diff to apply, so that should be easy.

ammen99:
Oh, and can you rebase on top of master? I think the CI may not run successfully otherwise.

veRYANgry:
Rebased! Please approve the workflow. Thanks for taking the time to look at it ^_^

ammen99:
@veRYANgry The code style is not entirely correct, take a look at the CI output (second-to-last action), it prints the necessary changes.

veRYANgry:
I'm not seeing a usable diff see below. Also, it seems like Uncrustify is not working on my computer and that is probably why there is formatting issues. I'll look into it more later and see if I can get it working. 
image

marcusbritanicus:
@veRYANgry You'll have to use [this fork](https://github.com/ammen99/uncrustify) of uncrustify. For the usable diff, look at the "tick-mark" above failed item.

ammen99:
> I'm not seeing a usable diff see below. Also, it seems like Uncrustify is not working on my computer and that is probably why there is formatting issues. I'll look into it more later and see if I can get it working. image

Take a look at the **second-to-last** command, the one above this. It is a simple `git diff`

veRYANgry:
> @veRYANgry You'll have to use [this fork](https://github.com/ammen99/uncrustify) of uncrustify. For the usable diff, look at the "tick-mark" above failed item.

Thanks! This will help in the future

veRYANgry:
I think that should be correct! Thanks for the support

metadata/command.xml
```diff
@@ -45,5 +45,18 @@
 				<_long>Sets the triggering activator binding.</_long>
 			</entry>
 		</option>
+		<option name="release_bindings" type="dynamic-list">
```
ammen99, line 48: Since this PR was created, we also had some changes to the dynamic-list options. They should have a `type-hint` property and entries should have `names`. Would you mind adding them here? (The other command options have been updated, so you can use them as a reference)

veRYANgry: I have fixed the xml attributes. Thanks for all the help!

ammen99:
Thanks for the PR!

Issue #1415, opened 2022-01-22
Crash when connecting wacom tablet


mfxbe:
**Describe the bug**
While working on  #1414 I noticed that sometimes wayfire will crash when I connect/disconnect my wacom tablet. Happens in stabilize-api as well as in master. Doesn't happen every time but quit often.
It seems to me that it happens more often if any kind of window is open but I migh be wrong here.

**To Reproduce**
(1. Open any kind of window)
2. Connect/Disconnect tablet a couple of times (USB)
3. Wayfire crash

**Expected behavior**
Wayfire doesn't crash and tablet  gets recognized.

**Screenshots or stacktrace**
See attached log.
[testlog.log](https://github.com/WayfireWM/wayfire/files/7918470/testlog.log)

**Wayfire version**
git master, git stabilize-api

lbcat06:
happens to me too

ethan2-0:
The following patch on top of wayfire 0.7.2 fixes this for me:

```
diff --git a/src/core/seat/tablet.cpp b/src/core/seat/tablet.cpp
index 274a5c75..8fec4fc1 100644
--- a/src/core/seat/tablet.cpp
+++ b/src/core/seat/tablet.cpp
@@ -479,7 +479,7 @@ void wf::tablet_pad_t::update_focus(wlr_surface *focus_surface)
     {
         wlr_tablet_v2_tablet_pad_notify_enter(pad_v2,
             attached_to->tablet_v2, focus_surface);
-    } else
+    } else if(old_focus)
     {
         wlr_tablet_v2_tablet_pad_notify_leave(pad_v2, old_focus);
     }
```

If there's interest I could submit a PR. I haven't tested for 0.7.3. It doesn't seem to break anything, but I haven't tested thoroughly.

wloxyz:
@ethan2-0 I'd appreciate a PR if you're willing to take the time, I keep running into this on my system.

mfxbe:
Looks like its fixed thanks to @ethan2-0

PR #1555, opened 2022-08-19
core: fix crash when connecting graphics tablet

ethan2-0:
Fixes issue #1415, which is a crash when plugging in a graphics tablet, by adding a null check. It appears to be necessary to have a window open and XWayland running to trigger the bug.

I developed this against `v0.7.2` but rebased to `master` for this pull request. I can switch to another branch as well - I'm a bit new to this whole open-source thing, so I'm not sure what's best to do.


ammen99 (approved):
LGTM, thanks!

PR #1572, opened 2022-09-19
Winrules updates

soreau:


plugins/window-rules/view-action-interface.cpp
```diff
@@ -206,6 +237,40 @@ void view_action_interface_t::_unminimize()
     _view->set_minimized(false);
 }
 
+void view_action_interface_t::_make_sticky()
+{
+    _view->set_sticky(1);
+}
+
+void view_action_interface_t::_always_on_top()
```
ammen99, line 245: We are really duplicating code here. Doesn't wm-actions have a signal to toggle always-on-top for a view?

plugins/window-rules/view-action-interface.cpp
```diff
@@ -342,6 +407,30 @@ void view_action_interface_t::_set_geometry(int x, int y, int w, int h)
...
+    }
+
+    auto og = output->get_relative_geometry();
+
+    x = std::clamp(x, 0, 100);
+    y = std::clamp(y, 0, 100);
+    w = std::clamp(w, 0, 100);
+    h = std::clamp(h, 0, 100);
+
+    x = (og.width / 100) * x;
```
ammen99, line 425: Why not use double/float/(multiply first and then divide) for more precision here?

plugins/window-rules/view-action-interface.cpp
soreau: The code is not really duplicated. In this case I thought it would make more sense that `set always_on_top` meant enable it, and not toggle it (and the function has been modified to reflect that). Consider if you tied it to maximize. Every *other* time you maximize the view, it would become always on top and the other times it would be normal floating. So yes, I could use the wm_actions_toggle_above signal, but I think it wouldn't do what the rule says.

soreau: > I see. Then it would make sense to either extend the wm-actions signal or add a new one. Duplicating functionality is not a good idea in any case.

Sure. I will find a better way.

plugins/window-rules/view-action-interface.cpp
soreau: Good idea, will update.

plugins/wm-actions/wm-actions.cpp
```diff
@@ -203,7 +207,8 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
     {
         auto view = choose_view(ev.source);
 
-        return toggle_keep_above(view);
+        in_always_above_layer = !in_always_above_layer;
```
ammen99, line 210: in_always_above_layer is a per-output variable, but we want to track this state per-view, right? There can be multiple always-on-top views (so we cannot use a single boolean variable to keep track of their state).

There is in fact no real need for the `in_always_above_layer` at all (I think), we store custom_data on each view in the above layer, so the state of a view can be checked via `view->has_data(...)`

ammen99 (changes requested):

ammen99 (approved):
LGTM, thanks!

PR #1582, opened 2022-10-02
Remove get_promoted_views() from header

soreau:
The function was removed, which broke idle. Fix idle by counting fullscreen views manually instead of using get_promoted_views() and remove the function from the header.


ammen99:
Should we inhibit idle even if a fullscreen window is not the focused one? Note that the old logic (promoted views) counted only focused fullscreen views on the current workspace.

soreau:
> Should we inhibit idle even if a fullscreen window is not the focused one? Note that the old logic (promoted views) counted only focused fullscreen views on the current workspace.
> 
> Also the signals emitted from core follow that logic.

Good point, I'll add a focused condition.

ammen99 (approved):
Thanks!

PR #1585, opened 2022-10-12
build: Remove WF_SRC_DIR macro

### LINKED ISSUE #1534: Consider removing WF_SRC_DIR
Currently `config.h.in` defines the macro `WF_SRC_DIR`, which hardcodes the source code directory path during build:

https://github.com/WayfireWM/wayfire/blob/ce3a37c7ee230e917be194a56bbec822bdb78aec/config.h.in#L6

https://github.com/WayfireWM/wayfire/blob/ce3a37c7ee230e917be194a56bbec822bdb78aec/meson.build#L77

However, it seems that `WF_SRC_DIR` is not used anywhere in the wayfire project. Looking at project history, it was introduced for debugging at https://github.com/WayfireWM/wayfire/commit/eb735a005eb04d481d4fd61055a405f5f5c660c3

```cpp
    static int srclen = sizeof(WF_SRC_DIR);
    if (strstr(filepath, WF_SRC_DIR) == filepath) {
        filepath += srclen;
```
...but its actual use was later removed. As seen in https://github.com/WayfireWM/wayfire/search?q=WF_SRC_DIR , no real code is using this macro.

------------

Now, keeping `WF_SRC_DIR` macro is causing different files (to be more concrete, different `config.h`) being generated when the build src dir is different, which breaks the [reproducible build initiative](https://reproducible-builds.org/) that aims to make software robustly generate bit-to-bit identical binaries during build.

You may find downstream distribution bug report at https://bugs.debian.org/1016486 . An analysis on current wayfire's reproducibility in Debian can be found at https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/wayfire.html .

If `WF_SRC_DIR` is not to be used for future Wayfire, I am wondering if it can be removed from upstream code base. If it will be kept, please let me know so that we can patch out this line from `config.h` with confidence. Thanks!

soreau:
Fixes #1534.


ammen99 (approved):

ammen99 (changes requested):
Uhm, actually, we should remove this from the config.h.in file, too.

ammen99 (approved):

ammen99:
Thanks!

PR #1596, opened 2022-10-21
simplify run() to use daemon(3)

epsilon-0:
current method is a 'hack' and using daemon(3)
reduces code burden by reusing standard libraries.

Signed-off-by: Aisha Tammy <aisha@bsd.ac>


src/core/core.cpp
```diff
@@ -766,59 +764,61 @@ wayfire_view wf::compositor_core_impl_t::find_view(const std::string& id)
...
+    int daemonized, status, sp[2];
+    struct pollfd pfd;
+    int TIMEOUT = 500;
+
+    if (socketpair(AF_UNIX, SOCK_STREAM, PF_UNSPEC, sp) == -1)
+    {
+      return -1;
+    }
 
-    /* The following is a "hack" for disowning the child processes,
```
ammen99, line 775: I think the comment should stay. We still do a double fork (even though daemonize does it for us), and I think that the purpose of this may not be clear to all readers.

src/core/core.cpp
```diff
@@ -766,59 +764,61 @@ wayfire_view wf::compositor_core_impl_t::find_view(const std::string& id)
...
-    {
-        close(pipe_fd[WRITE_END]);
-
-        int status;
-        waitpid(pid, &status, 0);
-
-        pid_t child_pid;
-        read(pipe_fd[READ_END], &child_pid, sizeof(child_pid));
+#endif
+        execl("/bin/sh", "/bin/sh", "-c", command.c_str(), NULL);
```
ammen99, line 803: I see you have removed the `_exit()` call. Wouldn't that cause issues if `execl` fails?

ammen99 (changes requested):
Also the formatting is off, see the CI logs for a patch you can apply to fix the code style.

ammen99:
Ping?

src/core/core.cpp
epsilon-0: I still do have the `_exit` but it's moved above to line 787.

src/core/core.cpp
epsilon-0: readded

src/core/core.cpp
ammen99: Yeah but the exit should happen after a failed exec

src/api/wayfire/core.hpp
```diff
@@ -274,7 +274,7 @@ class compositor_core_t : public wf::object_base_t, public signal::provider_t
      * This also sets some environment variables for the new process, including
      * correct WAYLAND_DISPLAY and DISPLAY.
      *
-     * @return The PID of the started client, or -1 on failure.
+     * @return The PID of the started client, or 0 on failure.
```
ammen99: Can we avoid breaking changes like this? I think we can easily continue returning -1 on failure.

ammen99:
Thanks for the PR but I have decided not to accept it. I think this PR makes much more than the title suggests - simply using daemon instead of fork might be a good idea, using anything more complicated than pipes is not.

PR #1607, opened 2022-11-03
Add an option "rules" to window-rules with type "dynamic-list"

marcusbritanicus:
Currently, window-rules does not have any option. Therefore, it's difficult for the parser to auto-magically load the stored rules. Furthermore, since it does not have any option defined, wcm also does not list the rules, or allow it's modification. This MR adds the "rules" option of type "dynamic-list" to window-rules.xml file.

### DIFF
--- a/metadata/window-rules.xml
+++ b/metadata/window-rules.xml
@@ -4,5 +4,13 @@
 		<_short>Window Rules</_short>
 		<_long>A plugin to apply specific commands to windows by using various criteria.</_long>
 		<category>Window Management</category>
+		<option name="rules" type="dynamic-list">
+			<_short>Window rules</_short>
+			<_long>Apply specific commands to windows by using various criteria</_long>
+			<entry prefix="rule_" type="string">
+				<_short>Rule</_short>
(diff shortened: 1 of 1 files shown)


lilydjwg:
This caused the following error on startup:

```
Malformed input:

^ Rule parser error. Expected 'on' keyword.
```

and a lot of `Window-rules: Error while executing rule on created signal` messages during running.

marcusbritanicus:
@lilydjwg Yes, I too noticed this lately. The window-rules plugin needs some modifications. If you look at [window-rules.cpp#L198](https://github.com/WayfireWM/wayfire/blob/master/plugins/window-rules/window-rules.cpp#L198), you'll notice that we're simply reading all entries under `window-rules` section as normal options. It should be modified to read only one option: `rules` which is a compound option. This "lazy" code is also the reason why window rules don't work with hjson-config-backend.

I am, at the moment, rather busy to make the modification. Please feel free to fix the code and open a PR. I am sure @ammen99 and @soreau will help you, should you need any.

PR #1612, opened 2022-11-10
Take plugin name as --config-backend argument

marcusbritanicus:
As an alternative to providing the full path, provide plugin name as the option's argument:
```
wayfire --config-backend hjson-config-backend
```

The current behaviour (which is providing the full path) is still valid.
```
wayfire --config-backend /usr/lib/wayfire/libdefault-config-backend.so
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -162,28 +162,8 @@ static wf::config_backend_t *load_backend(const std::string& backend)
         if (backend.compare(0, 1, "/") != 0)
         {
             // Not a full path so try to get the full path.
-            std::vector<std::string> plugin_prefixes;
-            if (char *plugin_path = getenv("WAYFIRE_PLUGIN_PATH"))
-            {
-                std::stringstream ss(plugin_path);
-                std::string entry;
-                while (std::getline(ss, entry, ':'))
-                {
-                    plugin_prefixes.push_back(entry);
-                }
-            }
-
-            plugin_prefixes.push_back(PLUGIN_PATH);
-
-            for (std::filesystem::path plugin_prefix : plugin_prefixes)
-            {
-                auto plugin_path = plugin_prefix / ("lib" + backend + ".so");
-                if (std::filesystem::exists(plugin_path))
-                {
-                    config_plugin = plugin_path;
-                    break;
-                }
-            }
+            std::vector<std::string> plugin_prefixes = wf::get_plugin_paths();
+            config_plugin = wf::get_plugin_path_for_name(plugin_prefixes, backend);
         }
     }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -158,13 +158,9 @@ static wf::config_backend_t *load_backend(const std::string& backend)
 
     if (backend.size())
     {
-        /** Get the full path if the given name is not a full path */
-        if (backend.compare(0, 1, "/") != 0)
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -153,7 +154,40 @@ static std::optional<std::string> choose_socket(wl_display *display)
 
 static wf::config_backend_t *load_backend(const std::string& backend)
 {
-    auto [_, init_ptr] = wf::get_new_instance_handle(backend);
+    std::string config_plugin(backend);
+
+    if (backend.size())
```
ammen99, line 159: Is this not duplicating code from the plugin loader for normal plugins? Can we not refactor the code to avoid the duplication?

marcusbritanicus: Added two functions to `output/plugin-loader.[cpp|hpp ]` to avoid code duplication.

src/main.cpp
```diff
@@ -153,7 +154,20 @@ static std::optional<std::string> choose_socket(wl_display *display)
...
+    {
+        /** Get the full path if the given name is not a full path */
+        if (backend.compare(0, 1, "/") != 0)
```
ammen99, line 162: This is also already done by the plugin-loader, I'd say we should move this in the get_plugin_path_for_name() function. Maybe also mention in the docstring that it accepts full paths.

marcusbritanicus: Done

src/main.cpp
```diff
@@ -153,7 +154,16 @@ static std::optional<std::string> choose_socket(wl_display *display)
 
 static wf::config_backend_t *load_backend(const std::string& backend)
 {
-    auto [_, init_ptr] = wf::get_new_instance_handle(backend);
+    std::string config_plugin(backend);
+
+    if (backend.size())
+    {
+        // Not a full path so try to get the full path.
```
ammen99, line 161: The comment is outdated.

src/main.cpp
```diff
@@ -153,7 +154,16 @@ static std::optional<std::string> choose_socket(wl_display *display)
 
 static wf::config_backend_t *load_backend(const std::string& backend)
 {
-    auto [_, init_ptr] = wf::get_new_instance_handle(backend);
+    std::string config_plugin(backend);
```
ammen99, line 157: Why do we need the extra variable here?

ammen99 (changes requested):
Two small issues, LGTM otherwise.

src/output/plugin-loader.cpp
```diff
@@ -271,3 +246,41 @@ void plugin_manager::load_static_plugins()
...
+            plugin_prefixes.push_back(entry);
+        }
+    }
+
+    plugin_prefixes.push_back(PLUGIN_PATH);
+
+    return plugin_prefixes;
+}
+
+std::string wf::get_plugin_path_for_name(std::vector<std::string> plugin_paths,
```
ammen99, line 268: I think it would be best to return std::optional<std::string> to indicate that this function may indeed fail.

marcusbritanicus: Done

src/main.cpp
marcusbritanicus: Removed.

ammen99 (approved):
LGTM now, thanks!

PR #1613, opened 2022-11-10
Enhancements to XML files for Hjson support

marcusbritanicus:
These changes are completely backward compatible with the existing ini backend.

Following additions have been made:
1. To each `option` node of `dynamic-list` type, `type-hint` attribute is added.
    Options which are a simple list should specify `type-hint='plain'`. Example: autostart/autostart, window-rules/rules
    Options which need one or more (key, value) pair, should specify `type-hint='dict'`. Example: command/bindings, vswitch/workspace_bindings
    Options which take a list of strings should specify `type-hint='tuple'`. Potentially, session management.
2. To each 'entry' node of compound option, 'name' attribute is added.
    This attribute is specifcally needed to store dict type options.

These attributes are needed save the config in hjson format such that it's intuitive to the user.

### DIFF
--- a/metadata/window-rules.xml
+++ b/metadata/window-rules.xml
@@ -4,10 +4,10 @@
 		<_short>Window Rules</_short>
 		<_long>A plugin to apply specific commands to windows by using various criteria.</_long>
 		<category>Window Management</category>
-		<option name="rules" type="dynamic-list">
+		<option name="rules" type="dynamic-list" type-hint="plain">
 			<_short>Window rules</_short>
 			<_long>Apply specific commands to windows by using various criteria</_long>
-			<entry prefix="rule_" type="string">
+			<entry prefix="" type="string">
 				<_short>Rule</_short>
 				<_long>Defines the rule to be applied</_long>
 			</entry>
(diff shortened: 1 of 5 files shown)


metadata/window-rules.xml
```diff
@@ -4,10 +4,10 @@
 		<_short>Window Rules</_short>
 		<_long>A plugin to apply specific commands to windows by using various criteria.</_long>
 		<category>Window Management</category>
-		<option name="rules" type="dynamic-list">
+		<option name="rules" type="dynamic-list" type-hint="plain">
 			<_short>Window rules</_short>
 			<_long>Apply specific commands to windows by using various criteria</_long>
-			<entry prefix="rule_" type="string">
```
ammen99, line 10: Reason for this change?

marcusbritanicus: Two reasons:
1. It does not make sense to have the prefix: single entry.
2. The current code does not care for the prefix.

PS: The current code does not treat `rules` as a compound option at all - I just looked at the [code](https://github.com/WayfireWM/wayfire/blob/master/plugins/window-rules/window-rules.cpp#L199), but as a bunch of normal options.

ammen99: Makes sense.

ammen99 (approved):
LGTM, thanks!

PR #1618, opened 2022-11-16
Fix typo in spelling of 'focusable'. Closes #1617

### LINKED ISSUE #1617: Incosistent spelling of "focusable"
The word 'focusable' is misspelled in several places in the code and docs. In particular, the configuration docs say to use "is focuseable" (incorrect spelling) but then the code which reads the configuration looks for "is focusable" (correct spelling). It looks like it needs to be corrected in about 7 places in the code, and then on the wiki. I'm happy to fork and submit a pull request, but I thought I'd just check in quick before I did.

emakman:


ammen99:
As this is a breaking change, the API version [here](https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/plugin.hpp#L170) should be bumped as well (it is just the current date).

ammen99 (approved):
Thanks!

PR #1619, opened 2022-11-16
Add maximize_regions option for outputs

BwackNinja:
I have a 5120x1440 monitor, which is easier to use as 3 separate regions. I can do that in X with `xrandr --setmonitor` commands.

Here I'm reproducing that functionality with a new option. It also affects the behavior of the grid plugin. Fullscreen still uses the output's full size.

Example configuration:

```
[output:DP-3]
maximize_regions = 1280x1440+0+0,2560x1440+1280+0,1280x1440+3840+0
```

This could still use more validation of the set of maximize regions are valid within the output, but is working for me.

### DIFF
--- a/metadata/output.xml
+++ b/metadata/output.xml
@@ -14,5 +14,8 @@
     <option name="transform" type="string">
       <default>normal</default>
     </option>
+    <option name="maximize_regions" type="string">
+      <default></default>
+    </option>
   </object>
 </wayfire>
--- a/plugins/grid/grid.cpp
+++ b/plugins/grid/grid.cpp
@@ -179,7 +179,7 @@ class wayfire_grid : public wf::plugin_interface_t
 
         view->get_data_safe<wf_grid_slot_data>()->slot = slot;
         ensure_grid_view(view)->adjust_target_geometry(
-            get_slot_dimensions(slot) + delta,
+            get_slot_dimensions(slot, view->get_wm_geometry()) + delta,
             get_tiled_edges_for_slot(slot));
     }
 
@@ -188,9 +188,9 @@ class wayfire_grid : public wf::plugin_interface_t
      * 4 5 6
      * 1 2 3
      * */
-    wf::geometry_t get_slot_dimensions(int n)
+    wf::geometry_t get_slot_dimensions(int n, wf::geometry_t g)
     {
-        auto area = output->workspace->get_workarea();
+        auto area = output->workspace->get_maximize_region(g);
         int w2    = area.width / 2;
         int h2    = area.height / 2;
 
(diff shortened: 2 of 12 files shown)


ammen99:
Hi, this looks like a very specific feature, so I do not think it belongs in core or in the default set of plugins. The current [goals](https://github.com/WayfireWM/wayfire/wiki/Project-goals#goals) are to have the basic functionality in this main repo, and more specific features in external plugins.

For this particular feature, I think there are no core changes really needed - you can just fork the grid plugin and put all the logic there.

Nonetheless, thanks for the PR :)

BwackNinja:
I can agree with that, but given the required changes to the move plugin as well, that's forking two plugins and putting information in two places -- unless the coupling between move and grid is changed. They're otherwise synchronized only by information about outputs. I'm only familiar with the codebase from implementing this, how would you suggest that I approach moving this functionality into a separate plugin?

ammen99:
As far as I can see, the only change required for move is that you pass the coordinates of the pointer? In this case, the grid plugin can just retrieve this position from core. (core::get_cursor_position()). And it can also check whether move is active with `output->is_plugin_active("move")`.

BwackNinja:
That's half of it. In the move plugin, `calc_slot` also independently determines where the window will snap based on the geometry of the output. If it instead queried that information from the grid plugin, then two plugins wouldn't need to be forked in order to implement this functionality.

ammen99:
I see. I"d say we should refactor the default implementation so that actually move queries grid whether snap should be enabled for a given location (and if yes, what rectangle to show as a preview). After all, move should not care about slots and their geometry at all - that part of the code belongs in grid. With these changes, it should be possible to fork just grid to add multiple maximizing regions, right?

BwackNinja:
I think that would be enough. Then I'd just have to move the configuration of the maximizing regions.

soreau:
Using the window-rules plugin, you could adjust the size and position when a client surface is opened or maximized, for example:
```
[window-rules]
rule1 = on created if app_id is "app.id.here" then set geometry_ppt 33 4 33 96
```
`geometry_ppt` interprets the numbers as percentages, but there is also `geometry` that treats them as absolute screen geometry. You would probably have to tweak some rules for your exact setup. The wiki has more information about the window-rules syntax. Hopefully this helps.

BwackNinja:
What I'm looking for is more complicated, to have regions on the screen where a window might maximize to depending on where it is on the screen -- like virtual monitors. Unless I missed it, the window-rules plugin doesn't have criteria based on where the window is. This pull request implemented that, and I'm in the middle of extracting the grid pieces from the move plugin in order to have a self-contained alternative grid plugin with my desired functionality.

PR #1622, opened 2022-11-19
Update to build against wlroots 0.16.0

soreau:


src/view/xdg-shell.cpp
```diff
@@ -263,9 +264,8 @@ void wayfire_xdg_view::initialize()
     });
     on_request_fullscreen.set_callback([&] (void *data)
     {
-        auto ev = static_cast<wlr_xdg_toplevel_set_fullscreen_event*>(data);
-        auto wo = wf::get_core().output_layout->find_output(ev->output);
-        fullscreen_request(wo, ev->fullscreen);
+        wlr_xdg_toplevel_requested *req = &xdg_toplevel->requested;
+        fullscreen_request((wf::output_t*)req->fullscreen_output, req->fullscreen);
```
ammen99, line 268: This seems very wrong. Casting wlr_output to a wf::output_t is not possible, because wf::output_t is not a subclass of wlr_output. The find_output method should be used as it was before.

ammen99 (changes requested):

src/view/xdg-shell.cpp
soreau: This code works because `wlr_output *handle` is the first member in class `wf::output_t`. However I will change it to find_output() as you suggest.

ammen99 (approved):
Thanks!

PR #1624, opened 2022-11-21
Update to build against wlroots 0.16.0

soreau:


plugins/ipc/stipc.cpp
```diff
@@ -89,10 +93,54 @@ static std::string layer_to_string(uint32_t layer)
...
+    wlr_touch_init(virtual_touch, &touch_impl, "wlr_virtual_touch_v1");
+
+    return &virtual_touch->base;
+}
+
 class headless_input_backend_t
 {
   public:
     wlr_backend *backend;
+    wlr_virtual_pointer_v1 virtual_pointer;
```
ammen99, line 141: What is the problem with creating a wlr_pointer here, instead of `wlr_virtual_pointer_v1`? The virtual_pointer is the object created by the virtual-pointer protocol. Here, we create generic input devices.

plugins/ipc/stipc.cpp
```diff
@@ -103,9 +151,14 @@ class headless_input_backend_t
...
         wlr_multi_backend_add(core.backend, backend);
 
-        pointer  = wlr_headless_add_input_device(backend, WLR_INPUT_DEVICE_POINTER);
-        keyboard = wlr_headless_add_input_device(backend, WLR_INPUT_DEVICE_KEYBOARD);
-        touch    = wlr_headless_add_input_device(backend, WLR_INPUT_DEVICE_TOUCH);
+        pointer  = virtual_pointer_device_create(&virtual_pointer);
+        keyboard = virtual_keyboard_device_create(&virtual_keyboard);
+        touch    = virtual_touch_device_create(&virtual_touch);
+
+        wl_signal_emit_mutable(&backend->events.new_input, pointer);
```
ammen99, line 158: Why _mutable here? We should use `wlr_signal_emit_safe` here.

plugins/ipc/stipc.cpp
```diff
@@ -89,10 +93,54 @@ static std::string layer_to_string(uint32_t layer)
...
+}
+
 class headless_input_backend_t
 {
   public:
     wlr_backend *backend;
+    wlr_virtual_pointer_v1 virtual_pointer;
+    wlr_virtual_keyboard_v1 virtual_keyboard;
+    wlr_touch virtual_touch;
     wlr_input_device *pointer;
```
ammen99, line 144: Since we create the structures ourselves, I don't think we have to store pointers to the `wlr_input_device`s. Instead, we can just store wlr_pointer/wlr_touch/wlr_keyboard, I think this will also make the code simpler, as we don't have to cast from device to the concrete type.

plugins/ipc/stipc.cpp
soreau: This is no such function in wlroots 0.16.0. The only option is `wl_signal_emit_mutable`.

plugins/ipc/stipc.cpp
soreau: Yes, I agree.

ammen99 (approved):
Thanks!

PR #1625, opened 2022-11-21
input: Fix input position with more than one wayland backend output

soreau:


src/core/seat/input-manager.cpp
```diff
@@ -72,7 +72,18 @@ void wf::input_manager_t::refresh_device_mappings()
         auto mapped_output = section->get_option("output")->get_value_str();
         if (mapped_output.empty())
         {
-            mapped_output = nonull(dev->name);
+            if (dev->type == WLR_INPUT_DEVICE_POINTER)
```
ammen99, line 75: Is there a mapping for tablet devices as well?

soreau: Not that I know of, `output_name` was moved to pointer and touch devices only. See wlroots a92e5f8d.

ammen99:
Thanks!

PR #1626, opened 2022-11-22
wf-utils: bump subproject

soreau:
Fixes matching app ids that have more than one dot in the string.
Previously, the lexer was treating a string with any number of dots as a double.


soreau:
This needs [wf-utils #4](https://github.com/WayfireWM/wf-utils/pull/4) to pass clang CI.

PR #1629, opened 2022-11-24
BugFix: fix inverted image when writing to file

marcusbritanicus:
Patch by @soreau to fix the inverted image while taking view snapshot


ammen99 (approved):
Thanks!

PR #1646, opened 2022-12-03
ShowDesktop: Minimize current workspace views

marcusbritanicus:
@soreau 's patch to minimize current workspace views in wm-actions plugin's show desktop.

### DIFF
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -10,7 +10,7 @@
 class wayfire_wm_actions_t : public wf::plugin_interface_t
 {
     wf::scene::floating_inner_ptr always_above;
-    bool showdesktop_active = false;
+    std::vector<std::vector<bool>> showdesktop_active;
 
     wf::option_wrapper_t<wf::activatorbinding_t> toggle_showdesktop{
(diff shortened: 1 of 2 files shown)


plugins/wm-actions/wm-actions.cpp
```diff
@@ -344,6 +340,21 @@ class wayfire_wm_actions_t : public wf::plugin_interface_t
             output->node_for_layer(wf::scene::layer::WORKSPACE),
             always_above);
 
+        auto wsize = output->workspace->get_workspace_grid_size();
+        showdesktop_active.resize(wsize.width);
+        for (int x = 0; x < wsize.width; x++)
+        {
+            showdesktop_active[x].resize(wsize.height);
```
soreau, line 347:
```suggestion
            showdesktop_active[x].resize(wsize.height);
            for (int y = 0; y < wsize.height; y++)
            {
                showdesktop_active[x][y] = false;
            }
```

and then omit the for loop below (on line 350)

marcusbritanicus:
Major rework is required to make this PR work with the current Wayfire-git. I will close this for the moment. 

It would be much better to write a separate plugin for show desktop.

PR #1654, opened 2022-12-08
Implement idle plugin disable_initially setting

### LINKED ISSUE #1653: Have option for idle plugin to be initially disabled
Hi, I would like to have the idle plugin, triggering DPMS initially disabled. And then activated through the toggle hotkey.
I would like to try and accomplish this and create a PR. But beforehand I would like to discuss approach as well as desirability.   

Case:
I only want idle behavior for a laptop on battery. I am now sometimes annoyed at having wayfire trigger DPMS trigger and having to remember disabling it. When I am on battery, for example in meeting rooms, I was used to pressing buttons to enable screensaver, to prolong battery life. This works great as is, only with the added annoyance of having to manually disable the IDLE plugin on boot.

My questions:
- is the proposed strategy (below) acceptable
- would this indeed be something that, once done [w/c]ould be merged?

Context:
the idle plugin toggle mechanism seems to work through an `std::optional` type parameterized with `wf::idle_inhibitor_t`. Currently called `hotkey_inhibitor`. (Perhaps after added option as initial source of inhibition, `plugin_inhibitor` would be more apt, but meh. The plugin binding `idle/toggle` is bound to a `toggle` function, that, with some guard, checks the `std::optional` and either `emplace()`s (constructing new `wf::idle_inhibitor_t` object, or `reset()`s, calling the destructor of that object, lowering the static `wf::idle_inhibitor_t::inhibitors` and updating through `wlr__idle_set_enabled`.

Strategy:
create new pluging option: `wf::option_wrapper_t<bool> disable_initially{"idle/disable_initially"};`
put in metadata and declare in `..../single_plugins/idle.cpp`.  
in `init()` put condition based on the option, and call `toggle` method once, if true.

bklaase:
Patch to add an option to disable the idle plugin initially, only to be activated by the toggle shortcut.
As advocated in: #1653 

Fixes #1653

### DIFF
--- a/metadata/idle.xml
+++ b/metadata/idle.xml
@@ -28,6 +28,12 @@
 			<_long>Disables idle on fullscreen.</_long>
 			<default>true</default>
 		</option>
+		<!-- Disable initially -->
+		<option name="disable_initially" type="bool">
+			<_short>Disable idle timeout initially</_short>
+			<_long>Disables idle timeout innitially as if the toggle was called once after loading the plugin.</_long>
+			<default>false</default>
+		</option>
 		<!-- Screensaver animation -->
 		<option name="cube_zoom_speed" type="int">
 			<_short>Cube zoom speed</_short>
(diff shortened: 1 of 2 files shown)


ammen99 (approved):
Seems small enough to be merged until a better mechanism exists (as discussed in IRC). Thanks :)

PR #1664, opened 2022-12-16
decor: Fix size reported when view is fullscreen

soreau:


plugins/decor/deco-subsurface.cpp
```diff
@@ -90,7 +90,14 @@ class simple_decoration_surface : public wf::surface_interface_t,
 
     virtual wf::dimensions_t get_size() const final
     {
-        return size;
+        if (view->fullscreen)
+        {
+            auto vg = view->get_wm_geometry();
+            return {vg.width, vg.height};
```
ammen99, line 96:
```suggestion
            return wf::dimensions(view->get_wm_geometry());
```

ammen99 (approved):

Issue #1309, opened 2021-10-11
Split `view_interface_t`


ammen99:
The view_interface_t actually combines three unrelated concepts:

1. Desktop state (position, size, tiled edges, fullscreen, etc.)
2. Content (surface_interface_t)
3. 'Entity' displayed on a layer on an output with transformers, this would the actual 'view'.

There is no reason to put all of these into a single mega structure like `view_interface_t` does. It should be split, with each of the three components having a clearly defined API. Benefits aside from cleaner code:

- It might be possible for plugins to replace xdg-shell/xwayland/whatever handling by replacing component 1. This would be especially beneficial if someone is up to extending the Xwayland support.
- Compositor views would now need to implement the second component and leave the rest untouched.
- This would also be a step towards allowing views to be on multiple outputs (it'd be easier to create a view which shares desktop state and drawable content with other views).

PR #1669, opened 2022-12-30
Remove surface_interface_t

### LINKED ISSUE #1314: Unify input handling
Currently, we have ugly `if (dynamic_cast<...>(...) != null)` checks everywhere in the input code, because it differentiates between compositor surfaces and wlroots surfaces. Instead, the `surface_interface_t` should be extended so that the branching happens there, not in the input code (basically `compositor_surface_t` will be absorbed into `surface_interface_t`)

https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/pointer.cpp#L123
https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/pointer.cpp#L159

and many other places in the touch and tablet inputs as well.

Also pointer constraints:

https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/pointer.cpp#L168

ammen99:
This series of commits completely removes the surface_interface_t class from the codebase. The surface and subsurface functionality has been superseded by the scenegraph and custom nodes.

This PR is the first part of #1309 and also fixes #1314.


ssfdust:
Hello, I have tested with the PR, it fixes the input-method popup problem.

But when I hover on firefox window in expo mode, the firefox window appears to be jittering. It seems that only it only influences firefox window.


https://user-images.githubusercontent.com/3088610/210216940-546c0310-20c8-429a-b462-fa0c995486fd.mp4

ammen99:
> Hello, I have tested with the PR, it fixes the input-method popup problem.
> 
> But when I hover on firefox window in expo mode, the firefox window appears to be jittering. It seems that it only influences firefox window.
> 
>  jittering-ff.mp4

Do you have blur enabled, and if yes, does disabling it alleviate the problem?

ssfdust:
After disabling blur plugin, the problem disappears.

ammen99:
> Hello, I have tested with the PR, it fixes the input-method popup problem.
> 
> But when I hover on firefox window in expo mode, the firefox window appears to be jittering. It seems that it only influences firefox window.
> 
>  jittering-ff.mp4

I somehow managed to reproduce this once, but cannot reproduce it anymore .. Do you have any idea whether there is any specific action which might cause this bug?

ammen99:
@ssfdust Actually nevermind, I managed to find a way to reproduce reliably. Does http://ix.io/4kgE fix the issue for you as well?

ssfdust:
Yes, the issue was fixed. Thank you very much.

ammen99:
> Yes, the issue was fixed. Thank you very much.

Thanks for testing!

PR #1670, opened 2023-01-01
Backport graphics tablet fix to 0.7.x

ethan2-0:
Straightforward backport of #1555 to the `0.7.x` branch. I'd appreciate if you could make a new release including this change, so that this change can hopefully make it into Debian bookworm.


ammen99:
Thanks! I'll see about the release, I don't think we have enough changes for a new release.

PR #1673, opened 2023-01-02
Fix 'lazy' code of window rules

marcusbritanicus:
This fixes the [problem](https://github.com/WayfireWM/wayfire/pull/1607#issuecomment-1338902195) reported by @lilydjwg. Besides, it also makes window-rules work with hjson config backend.

### DIFF
--- a/plugins/window-rules/window-rules.cpp
+++ b/plugins/window-rules/window-rules.cpp
@@ -14,6 +14,7 @@
 #include <wayfire/rule/lambda_rule.hpp>
 #include <wayfire/rule/rule.hpp>
 #include <wayfire/util/log.hpp>
+#include <wayfire/option-wrapper.hpp>
 
 #include "lambda-rules-registration.hpp"
 #include "view-action-interface.hpp"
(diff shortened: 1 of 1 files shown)


plugins/window-rules/window-rules.cpp
```diff
@@ -194,11 +195,13 @@ void wayfire_window_rules_t::setup_rules_from_config()
...
-    // Build rule list.
-    auto section = wf::get_core().config.get_section("window-rules");
-    for (auto opt : section->get_registered_options())
+    wf::option_wrapper_t<wf::config::compound_list_t<std::string>> rule_list_option{"window-rules/rules"};
+    auto rule_list = rule_list_option.value();
+
+    for (const auto& [name, rule_str] : rule_list)
     {
-        _lexer.reset(opt->get_value_str());
+        LOGW("Registering ", rule_str);
```
ammen99, line 203: Why is this a warning?

marcusbritanicus: Oh.. I had added it during testing... What's the function for debug? Would it be LOGD?

ammen99 (approved):

ammen99:
Thanks!

Issue #1510, opened 2022-06-27
Make Wayfire's default bindings optional


ammen99:
With #1501, bindings will stop being a special case in Wayfire's input handling system and can be implemented as a plugin. Therefore, it would make sense to move bindings to a separate plugin which provides an API for registering bindings on an output. This API should make it possible for an alternative implementation of a plugin for handling bindings, which can then for example support multi-stroke bindings, or bindings based on the xkb state instead of raw keycodes, etc. transparently to existing plugins.

As this will be a very minor breaking change for plugins, it can be left for 0.9 to avoid delaying 0.8 infinitely.

ammen99:
Seems that instead of working with activator callbacks, it would be even better to allow plugin IPC actions.

PR #1674, opened 2023-01-04
global plugins

### LINKED ISSUE #1320: Remove/refactor grab APIs

ammen99:
Currently, plugins necessarily have one instance per output. While this makes sense for many plugins, it is an artificial restriction which is sometimes bypassed with the singleton plugin class. Nonetheless, it would make far more sense to have a single instance of each plugin, and then have the plugins create per-output objects if they need to do so. This makes a lot of sense especially with plugins like IPC, idle or oswitch. Not to mention that things like bindings are currently artifically constrained to a single output, but they should actually be global state, just like every other input state.

This PR aims to do the following:

- [x] Remove the old grab interfaces, which are per-output. Instead, plugins can use the input-grab helper.
- [x] Make plugins have one global instance instead of an instance per-output. There should also be a helper for plugins which prefer the old behavior (to begin with, that include basically all plugins).
- [x] Revamp bindings. They should work globally and not just per-output.
- [x] #1510

Fixes #1320


ammen99:
In this PR I have refactored the existing IPC used for tests in a more general-purpose IPC framework. A demo plugin utilizing it is the demo-ipc.cpp. Unfortunately, there is nothing too exciting for users as of yet. Anyway, how it works:

1. There is an ipc method repository which contains callbacks for various IPC calls which can be made. Multiple plugins can register callbacks here. The idea is that eventually most plugins would provide their actions (currently activator callbacks) via this IPC method repository.
2. The ipc plugin itself creates the ipc socket and manages incoming connections.

Currently, there is a demo ipc plugin which is a proof-of-concept of how this works together. It provides just 4 IPC calls - watch (register the IPC client so that it can get view-mapped events over IPC), get view info, get output info, set view geometry. The IPC client in itself, together with the message parsing logic (in python, but could be any other language) can be found below. What it does is center each view when it becomes mapped, essentially place's center mode (but it also sets the view size to half the output size, even though that can obviously be changed by everybody).

```py
import socket
import json as js

def get_msg_template():
    message = {}
    message["data"] = {}
    return message

class WayfireSocket:
    def __init__(self, socket_name: str):

    def read_exact(self, n):
        response = bytes()
        while n > 0:
            if not read_this_time:
            n -= len(read_this_time)
            response += read_this_time

        return response

    def read_message(self):

    def send_json(self, msg):
        self.client.send(header)
        self.client.send(data)
        return self.read_message()

    def watch(self):
        message = get_msg_template()
        message["method"] = "demo-ipc/watch"

    def get_view_info(self, id):
        message = get_msg_template()
        message["method"] = "demo-ipc/view-info"
        message["data"]["id"] = id

    def get_output_info(self, id):
        message = get_msg_template()
        message["method"] = "demo-ipc/output-info"
        message["data"]["id"] = id

    def set_view_geometry(self, id, x, y, w, h):
        message = get_msg_template()
        message["method"] = "demo-ipc/view-set-geometry"
        message["data"]["id"] = id
        message["data"]["geometry"] = {}
        message["data"]["geometry"]["x"] = x
        message["data"]["geometry"]["y"] = y
        message["data"]["geometry"]["width"] = w
        message["data"]["geometry"]["height"] = h

inbound_socket = WayfireSocket('/tmp/wayfire.socket')
outbound_socket = WayfireSocket('/tmp/wayfire.socket')

inbound_socket.watch()
while True:
    msg = inbound_socket.read_message()

    output = msg["view"]["output"]
    output_info = outbound_socket.get_output_info(output)
    print(output_info)

    geometry = {}

    w = output_info["info"]["geometry"]["width"]
    h = output_info["info"]["geometry"]["height"]
    print(outbound_socket.set_view_geometry(msg["view"]["id"], w // 4, h // 4, w // 2, h // 2))
```

The long-term plans are to add many more commands in a dedicated IPC plugin, the current one is just a demo to ensure that what I imagine is feasible.

PS. The ipc plugin by default sets the Wayfire socket to `/tmp/wayfire-<wayland display name>.socket`, however, that can be overridden with the `_WAYFIRE_SOCKET` env var before starting Wayfire. The python test client currently hardcodes the expected `/tmp/wayfire.socket` socket path.

ammen99:
Porting guide for plugins:

- The grab interfaces have been removed. Input grabs can be done with the input grab helper from `wayfire/plugins/common/input-grab.hpp`. See bedbeb78629efed804f0382a5a2f263b32d6f17c for more details.
- Plugins now have single instance (as opposed to a per-output instance) by default. To get the old behavior, plugins should now derive from `per_output_plugin_instance_t` defined in `wayfire/per-output-plugin.hpp` and then use `DECLARE_WAYFIRE_PLUGIN(wf::per_output_plugin_t<your_plugin_class_here>)`. See fe3a36fd780ea716cce8571f50da543ef1ff2023 for more details.
- The `grab_interface` field was completely removed from the plugin interface. Plugins that need to use activation (e.  `output->activate_plugin()`) can instead create a `wf::plugin_activation_data_t` member and use that struct to activate themselves. For example https://github.com/WayfireWM/wayfire/pull/1674/commits/fe3a36fd780ea716cce8571f50da543ef1ff2023#diff-ec11d3b1cb8b17c685647070cfdccd7f3003b59b99b3f56f75e2f7aabfe500bfR122-R127 and the subsequent rename 11a5e45e8203f2c41e0d82658d74792c6445b0c1

PR #1678, opened 2023-01-06
Drop outdated signals API + Separate foreign toplevel

### LINKED ISSUE #1577: Move all signals to the new API
The new API seems to work quite well, see #1495 . Currently however, the majority of signals are emitted via the old and less-safe API. At some point, we should remove the old API completely.

ammen99:
Goals for this PR:

- [x] Fix #1577 
- [x] Separate foreign toplevel impl into its own plugin
- [x] Separate gtk-shell into its own plugin

After this, the view impl should be ready to tackle transactions and #1309


ammen99:
Breaking change for users: in order to have wf-panel, waybar, etc. working you likely need to add the `foreign-toplevel` and `gtk-shell` plugins to your config file.

ssfdust:
Hello, I have tested with the PR. when I switch window to my kitty terminal with switcher, which runs in xwayland, the window behaves strangely. In fact, I don't hit any key at that time.


https://user-images.githubusercontent.com/3088610/211153687-5391eb18-f855-4921-be06-fc120424e981.mp4

ammen99:
> Hello, I have tested with the PR. when I switch window to my kitty terminal with switcher, which runs in xwayland, the window behaves strangely. In fact, I don't hit any key at that time.

Strangely enough I can reproduce the issue even on master without any plugins involved - just starting terminator and then xterm. I'll look into this, I am wondering why it wasn't noticed before.

ssfdust:
Confirm the bug has been fixed in the latest update, thank you.😇

ammen99:
@ssfdust Thanks for testing, it is really helpful :)

PR #1681, opened 2023-01-08
fast-switcher: add alpha transition animations

occivink:
One advantage of this change is that it avoids the harsh flickering if you quickly press and release the fast-switcher shortcut.

### DIFF
--- a/plugins/single_plugins/fast-switcher.cpp
+++ b/plugins/single_plugins/fast-switcher.cpp
@@ -10,11 +10,19 @@
 #include <wayfire/workspace-set.hpp>
 #include <wayfire/util/log.hpp>
 #include <wayfire/seat.hpp>
+#include <wayfire/render-manager.hpp>
+#include <wayfire/util/duration.hpp>
 
 /*
  * This plugin provides abilities to switch between views.
  * It works similarly to the alt-esc binding in Windows or GNOME
  */
+
+class animation_data_t : public wf::custom_data_t
+{
+  public:
+    wf::animation::simple_animation_t anim;
+};
 
 class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf::keyboard_interaction_t
 {
@@ -22,11 +30,13 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
     wf::option_wrapper_t<wf::keybinding_t> activate_key_backward{
         "fast-switcher/activate_backward"};
     wf::option_wrapper_t<double> inactive_alpha{"fast-switcher/inactive_alpha"};
+    wf::option_wrapper_t<int> animation_duration{"fast-switcher/animation_duration"};
     std::vector<wayfire_view> views; // all views on current viewport
     size_t current_view_index = 0;
     // the modifiers which were used to activate switcher
     uint32_t activating_modifiers = 0;
     bool active = false;
+    bool animations_ongoing = false;
     std::unique_ptr<wf::input_grab_t> input_grab;
 
     wf::plugin_activation_data_t grab_interface = {
@@ -86,6 +96,9 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
             return;
         }
 
+        ev->view->erase_data<animation_data_t>();
+        ev->view->get_transformed_node()->rem_transformer(transformer_name);
+
         views.erase(views.begin() + i);
 
         if (views.empty())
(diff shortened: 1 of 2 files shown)


occivink:
Changed the smoothing function to sigmoid, which further reduces the flickering upon quick press-and-release.

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -8,23 +8,56 @@
...
     wf::option_wrapper_t<wf::keybinding_t> activate_key_backward{
         "fast-switcher/activate_backward"};
     wf::option_wrapper_t<double> inactive_alpha{"fast-switcher/inactive_alpha"};
+    wf::option_wrapper_t<int> animation_duration{"fast-switcher/animation_duration"};
     std::vector<wayfire_view> views; // all views on current viewport
     size_t current_view_index = 0;
     // the modifiers which were used to activate switcher
     uint32_t activating_modifiers = 0;
     bool active = false;
+    bool animations_ongoing = false;
```
soreau, line 60: This variable seems to serve no real purpose.

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -8,23 +8,56 @@
...
 #include <wayfire/util/log.hpp>
+#include <wayfire/render-manager.hpp>
+#include <wayfire/util/duration.hpp>
 
 /*
  * This plugin provides abilities to switch between views.
  * It works similarly to the alt-esc binding in Windows or GNOME
  */
 
+class animated_alpha_transformer : public wf::scene::view_2d_transformer_t
```
soreau, line 19: Do you actually have to subclass this for some reason or can you just use a wf::scene::view_2d_transformer_t as it is, [like in animate](https://github.com/WayfireWM/wayfire/blob/master/plugins/animate/basic_animations.hpp#L33)?

soreau (commented):
The general idea seems good. I haven't tested it but left some comments.

plugins/single_plugins/fast-switcher.cpp
occivink: That's not quite correct: I imagine you mean that it's redundant with `active` , but it's actually not. That's because when exiting the switcher (for example, after pressing `esc` or activating a view), `active` is set to false, but the animations are still being performed for X milliseconds, so we need to keep the hooks on. This variable indicates whether the blur effect is on or not. 

In addition, if the user re-activates the switcher during that duration, we should not re-add the hooks a second time, since they're already active.

plugins/single_plugins/fast-switcher.cpp
occivink: Possibly not, I might have done this by getting inspiration from another plugin. Thanks for the tip, I will look.

occivink: So in the end I need a way to store an instance of `wf::animation::simple_animation_t` per-view. In the case of basic_animations there is already a subclass of the per-view `animation_base` that stores it, but there is no such equivalent in the fast-switcher plugin, so I don't see a simple way of getting rid of this subclass.

plugins/single_plugins/fast-switcher.cpp
soreau: > That's not quite correct: I imagine you mean that it's redundant with `active` , but it's actually not. 

I do not mean that it is redundant with active. I mean that you could remove this variable in every place that it is used and the code would behave the same. Let me explain. At first, when the keybinding is activated, you check if `!animations_ongoing` and set it to true, when hooking up the pre and post callbacks. Then in post callback, you assert `animations_ongoing` and then unset it when unhooking pre and post. The act of adding and removing the pre/post effects is enough and the variable can be omitted.

occivink: >I think what you are trying to avoid is adding multiple effects of the same type while animation is in flight? If so, perhaps this makes sense.

Yes exactly, that's what I mentioned in the second part of my comment.

plugins/single_plugins/fast-switcher.cpp
ammen99: Like soreau, I don't think inheritance makes sense here. Better simply store a custom_data_t on the view with the animation running (and don't inherit from view_2d_transformer)

occivink:
Thank you for the review, I've tried to address both of your comments.

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -86,6 +96,9 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
             return;
         }
 
+        ev->view->erase_data<animation_data_t>();
```
ammen99, line 99: I'd put these two in a separate function to avoid code duplication.
In this case, the duplicated lines are just 2, but they logically belong together (fast-switcher does not care about the view anymore), and having a single function would probably make future adjustments easier as well.

plugins/single_plugins/fast-switcher.cpp
```diff
@@ -105,12 +118,77 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
...
+            {
+                continue;
+            }
+
+            transformer->alpha = static_cast<double>(anim_data->anim);
+            view->damage();
+        }
+    };
+
+    wf::effect_hook_t post_hook = [this] ()
```
ammen99, line 144: I am confused as to why we need a pre and a post hook, is it not enough to use a pre-hook only?

ammen99:
Ping :)

occivink:
Sorry, I've not been able to resolve the newest merge conflicts (at least, such that the proposed feature is still functional), and I've lost the motivation to work on this feature, so I'll be closing this PR.

PR #1697, opened 2023-01-19
scale: let views with the same geometry appear in the same order

occivink:
The problem that this addresses can be reproduced with the following steps:
1. Have the `place` (with `mode = center`) and `scale` plugins
2. Open multiple windows with the same dimensions (e.g. terminals)
3. Start 'scale' mode
4. Focus a different window
5. Start 'scale' mode again

The position of the windows in the two scale activations should be different, which can make it difficult to keep track of which window is which.

This is because the sorting of the views is purely geometry based, and when the geometry is identical the order is essentially unspecified. 
While there is no ordering that makes more sense that the other, ensuring that the order is always the same (which we do by just comparing pointers) makes things more predictable.

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -846,7 +846,15 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
         std::vector<wayfire_view>& views)
     {
         std::vector<std::vector<wayfire_view>> view_grid;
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
LGTM!

ammen99:
Thanks :)

PR #1698, opened 2023-01-19
scale: highlight view when hovering with the mouse

occivink:


occivink:
As discussed on IRC, it is not ideal that the focus is changed constantly as the user moves the cursor around (as this messes with the 'recent' ordering). Instead, only the highlight should change on hover, and the focus should be set when exiting scale. 
Note that the same issue applies to changing the active window using the arrow keys.

ammen99:
This PR has gotten quite stale and also doesn't really contain much of what is needed to 'properly' fix the issue, so closing.

Issue #1702, opened 2023-01-22
Moving windows in expo is broken if the output position is not 0,0


timgott:
**Describe the bug**
If an output is not at position (0,0) and the window is grabbed with the mouse, it jumps to a different position. Probably some offset is not computed correctly.

**To Reproduce**

1. Enable plugin expo
2. In the section `[output]`, set `position = 100, 100`
3. Toggle expo overview, grab a window with the mouse
4. The window jumps by approximately (100,100)

Minimal config to reproduce (expo opens with Ctrl+Tab):

[expotest.ini.txt](https://github.com/WayfireWM/wayfire/files/10474416/expotest.ini.txt)

**Expected behavior**
Window stays at the same position relative to the mouse.

**Wayfire version**
git commit `3711f658e539ce4f50a1e5f5219ae7922f0e8fe4` (latest commit on master)

Issue #1701, opened 2023-01-22
"Scale" title overlay briefly appears on all windows after opening a new window


occivink:
**Describe the bug**
The title overlay appears when it should not be visible.

**To Reproduce**
1. Enable plugin 'scale'
2. Have some windows open
3. Open a new window
4. The title overlay (as per the `title_overlay` option of the scale plugin) appears for a few seconds (seems to be variable?) on all visible windows

**Expected behavior**
Since the scale mode is not activated, the overlay should not appear.

**Screenshots or stacktrace**
As you can see in this screenshot, the "foot" title overlay appears on all windows even though the scale mode is not active (the windows are not laid out in a grid)
![wayfire-scale-overlay](https://user-images.githubusercontent.com/5473453/213911896-db9a65e0-2b0f-4d7a-9dc0-4ce9c9a765e0.png)
Log: [scale-title.log](https://github.com/WayfireWM/wayfire/files/10474205/scale-title.log)

**Wayfire version**
commit 3711f658e539ce4f50a1e5f5219ae7922f0e8fe4

Issue #1700, opened 2023-01-22
Crash when closing last window using scale


occivink:
**Describe the bug**
Wayfire crashes when using the scale plugin and closing the last remaining window

**To Reproduce**
1. Enable plugin 'scale'
2. Have a single window open on the current workspace
3. Toggle 'scale' mode
4. Close the window using the middle-click shortcut

**Expected behavior**
The window is simply closed (and scale ends)

**Screenshots or stacktrace**
log with sanitizer: [crash.log](https://github.com/WayfireWM/wayfire/files/10474159/crash.log)

**Wayfire version**
commit 3711f658

stefonarch:
Confirming.

Issue #1686, opened 2023-01-17
Crash when using vswitch while scale is active


occivink:
**Describe the bug**
Wayfire crashes when combining scale and vswitch

**To Reproduce**
1. Enable plugins `vswitch` and `scale`
2. Toggle scale mode 
3. Use one of the vswitch shortcuts
4. wayfire crashes

**Screenshots or stacktrace**
[crash.log](https://github.com/WayfireWM/wayfire/files/10433699/crash.log)

**Wayfire version**
git master, commit c8f22f72

ssfdust:
Hello, i can't reproduce it on the master branch, could you please upgrade to the lastest git version?

occivink:
I can't reproduce this issue with these exact steps anymore on master. But I've experienced another crash that seemed related in terms of actions performed, though I don't have reproduction steps atm, so I'll keep this open until I have more info

ammen99:
Should we close this?

occivink:
Yes, I can't reproduce it even with some 'stress-testing'.

PR #1705, opened 2023-01-23
bugfixes

### LINKED ISSUE #1695: Wayfire crashes when dragging a window from one output to another
Wayfire crashes when dragging a window from one output to another with [this bt](http://ix.io/4ly3).

Wayfire c566170b.

ammen99:
This PR aims to fix #1695, #1700, #1701 and #1702. Let me know if some of the bugs have slipped through, ideally if this is the case let me know how to adjust the tests: https://github.com/ammen99/wayfire-tests

### DIFF
--- a/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
+++ b/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
@@ -499,8 +499,6 @@ class core_drag_t : public signal::provider_t
             grab_origin = grab_position;
             view_held_in_place = true;
         }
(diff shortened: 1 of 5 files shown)


occivink:
Can confirm that #1700 is fixed, but #1701 is not. 
Possibly #1686 is fixed as well? Though I can't reproduce it anymore, so I can't say for sure, it could have been already fixed.

AR-CADE:
#1700 seems to be fixed, i have not tested the others mentioned bugs

Issue #1637, opened 2022-11-27
"always on top" binding of wm-actions plugin is not functional


occivink:
**Describe the bug**
When always-on-top is toggled on a window, the window does not always appear above other windows

**To Reproduce**
1. Enable plugin wm-actions
2. Assign a binding to `toggle_always_on_top`
3. Use said binding on a window
4. Other windows can still appear on top of the previous window, when they have focus

**Expected behavior**
The window on which the binding was used should always appear on top of the first window

**Wayfire version**
Self-built from git (using non-system wlroots and wf-config), commit c3b1edf1

UltraBlackLinux:
I can confirm this. Furthermore, pressing the key combo moves the window to the very back every time another one is focused

stefonarch:
Same hear now, I remember it worked with release version.

soreau:
Does [this patch](http://ix.io/4hAt) help?

occivink:
It seems to, although if a window is being moved (with the `move` plugin) it will still appear above the one supposed to be on top. I don't remember if this used to be the case previously

stefonarch:
Looks like the patch doesn't apply anymore now:

```
patching file plugins/wm-actions/wm-actions.cpp
Hunk #1 FAILED at 108.
```

soreau:
For me, it applies cleanly to wayfire master with `curl http://ix.io/4hAt | git apply`

stefonarch:
My fault, it works fine:

* both shortcut and window rule `always_on_top` are working
* moving windows from another workspaces only  for a short moment cover the always_on_top window

Has the wiki to be updated for the window rules? I found them only [here](https://github.com/WayfireWM/wayfire/issues/1567#issuecomment-1251463878) after searching.

soreau:
> Has the wiki to be updated for the window rules? I found them only [here](https://github.com/WayfireWM/wayfire/issues/1567#issuecomment-1251463878) after searching.

If not, it's open for anyone to update.

stefonarch:
I see a weird behavior: when an application with rule `always_on_top` is activated (lxqt-panel or lxqt-runner) also waybar and yatbwf (panel) will be shown always, for example with an youtube video on fullscreen. In this case pressing the shortcut for lxqt-runner appear also those 2  bars instead of the runner only.

stefonarch:
> For me, it applies cleanly to wayfire master with `curl http://ix.io/4hAt | git apply`

[...]

PR #1713, opened 2023-02-08
Fix for #1637

marcusbritanicus:
This was the patch @soreau cooked up to fix the issue with `always_on_top`. It works :tm:, but has a few issues - See the last two comments in #1637


ammen99 (approved):
LGTM, thanks!

PR #1715, opened 2023-02-09
Split wf shell into plugin

marcusbritanicus:
Move the wayfire-shell code into it's own plugin.


ammen99 (approved):
LGTM, I assume you have tested this?

marcusbritanicus:
Yes. I use hotspots are a part of DesQ. They work well. I also had rigged up a small test for wayfire_surface, which also runs.

ammen99:
Breaking change for users: add wayfire-shell to your plugin list.

PR #1736, opened 2023-02-24
use native binary for wayland-scanner

epsilon-0:
This is similar to how [swaylock](https://github.com/swaywm/swaylock/blob/ac3b49b6571ceda3f8db11a98bfe320106996280/meson.build#L34) finds the wayland-scanner binary. This gives better support for cross compilation.


ammen99 (approved):
Thanks!

Issue #1738, opened 2023-02-25
Memory leak in cube plugin (in git)


tsujan:
With the latest git source (https://github.com/WayfireWM/wayfire/commit/a1f575701ab3ba41006a38c0a01f39d31fdf2dbb), whenever I rotate the desktop cube (`rotate_left` or `rotate_right`) or just activate it (`activate`), the memory usage (shown by Conky) increases.

This does not happen with the latest release, Wayfire 0.7.5.

tsujan:
In case you need it, this the config:
```
[cube]
activate = <alt> <ctrl> BTN_LEFT
rotate_left = <alt> <ctrl> KEY_LEFT
rotate_right = <alt> <ctrl> KEY_RIGHT
zoom = 0.2
background_mode = skydome
skydome_texture = /home/pedram/Pictures/Compiz/morning-clear.jpg
skydome_mirror = false
```

tsujan:
@stefonarch found out that the problem is in skydome and its texture. Without it, there's no memory leak; with it, I only see a memory leak, but he also sees a laggy session.

tsujan:
I tested again with git Wayfire and https://github.com/WayfireWM/wayfire/pull/1744. The problem still exists. Moreover, it also happens without `skydome`. It's easy to increase the memory usage by 1GiB, even without `skydome`.

ammen99:
Thanks for the detailed report, it really helped :) The issue itself was very simple to fix (fix will land soon with #1845)

ammen99:
Weird, I am sure I tested and failed to reproduce after the patch.

tsujan:
It may depend on settings. Mine:
```
[cube]
activate = <alt> <ctrl> BTN_LEFT
rotate_left = <alt> <ctrl> KEY_LEFT
rotate_right = <alt> <ctrl> KEY_RIGHT
zoom = 0.2
background_mode = skydome
skydome_texture = /home/pedram/Pictures/Compiz/morning-clear.jpg
skydome_mirror = false
```

ammen99:
I just tested this once again with latest master and with your config, but I cannot get the issue to happen. Both normal RAM and VRAM usage stay constant, no matter how many times I start cube.

tsujan:
> I just tested this once again...

Thanks! I'd close the report if I were you. I may not find time to test it soon but will add a comment here otherwise.

ammen99:
> > 
> 
[...]

PR #1744, opened 2023-02-28
Fix cube

soreau:


tsujan:
I don't know if this is for https://github.com/WayfireWM/wayfire/issues/1738, but it doesn't fix the memory leak. I can add 1GiB to the memory usage in less than 1 minute just by rotating the cube + skydome. The leak is huge.

ammen99 (approved):
Thanks!

PR #1746, opened 2023-03-01
foreign-toplevel: Cleanup remaining leftovers from moving to a plugin

### LINKED ISSUE #1739: Waybar's taskbar doesn't work with git Wayfire
With the latest git source (https://github.com/WayfireWM/wayfire/commit/a1f575701ab3ba41006a38c0a01f39d31fdf2dbb), Waybar's taskbar doesn't work anymore.

Waybar's taskbar works well with the latest release, Wayfire 0.7.5.

soreau:
Wayfire was advertising the protocol twice which breaks apps like waybar.

Fixes #1739.


ammen99 (approved):
Good catch!

PR #1751, opened 2023-03-05
wlr-surface-node: Don't send frame events to surfaces with a buffer

soreau:
After removing a view from workspace-manager, wayfire was sending frame done events immediately on commit, if it was not visible on any output. In addition, we can check if the surface has a buffer to tell if it's mapped or not.


ammen99 (approved):
LGTM, thanks!

Issue #1756, opened 2023-03-12
The state of the mouse cursor doesn't change when entering scale/expo view.


ErynGalen:
**Describe the bug**
The cursor state doesn't change when entering scale/expo view.

**To Reproduce**
1. Enable plugin expo or scale
2. Place your cursor over some text so that the cursor changes its shape
Or place the cursor over a video in VLC for example, and wait for it to disappear
3. Enter scale or expo view with a keybinding without moving the mouse.
4. Move the mouse around: the cursor keeps the shape it had before entering scale/expo view. If it was hidden (after watching a video for example), it doesn't reappear.
5. In scale, if you interact with a window (for example drag it around), the cursor comes back to its default shape.

**Expected behavior**
When entering scale or expo, the cursor should immediately come back to its default shape.

**Wayfire version**
`0.8.0-90c370fa (Mar 11 2023, branch 'master')`

ErynGalen:
Alright, so...
After some testing, I recompiled the master branch with no changes, and it worked...
I don't really know what was the problem

ErynGalen:
I've come with a fix, which is adding a `wf::get_core().set_cursor("default");` to the respective `activate()` function, is that the correct way of doing it?
Should I may a PR for that?

PR #1769, opened 2023-04-01
Restore the cursor shape to default when activating scale or expo

ErynGalen:
Without that, when switching to expo/scale view, the cursor keeps its shape (or remains hidden if it was hidden before switching)
#1756

### DIFF
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -1316,6 +1316,9 @@ class wayfire_scale : public wf::per_output_plugin_instance_t,
             return false;
         }
 
+        // hack because otherwise the cursor keeps its previous shape
+        wf::get_core().set_cursor("default");
+
         initial_workspace  = output->workspace->get_current_workspace();
         initial_focus_view = output->get_active_view();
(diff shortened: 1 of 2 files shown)


ammen99 (commented):
I suppose the question is, should grabbing the input automatically set the cursor? I suspect other plugins will have a similar problem.

ErynGalen:
Yes indeed, it happens as well with others, but generally I was fine with it.

Also I've no idea how the plugins handle input grabbing or anything like that, so I only made what I knew would work.

There's probably a better way to do it though.

ammen99:
I think it should be safe (and better) to set the cursor in the input-grab helper, namely [here](https://github.com/WayfireWM/wayfire/blob/master/plugins/common/wayfire/plugins/common/input-grab.hpp#L134).

ErynGalen:
You're probably right, I'm not familiar at all with the code base, I can't really help. 
Someone else will be more able to solve the problem than me.

ammen99:
Resubmited in #1875 with the correct fix.

Issue #1771, opened 2023-04-06
Fullscreen window state not emitted (0.7.5)


simplejack-src:
**Describe the bug**
Fullscreen window state (`zwlr_foreign_toplevel_`) not emitted. Noticed when debugging [xfce4-panel](https://gitlab.xfce.org/xfce/xfce4-panel/-/issues/728) external plugins. Normally, xfce4-panel runs external plugin surfaces at a layer-shell level of `OVERLAY` and when detecting an active fullscreen window demotes these plugin surfaces to `TOP`. However, (on 0.7.5 NOT master), this window state event is not being seen. 

**To Reproduce**
1. Run xfce4-panel with an external plugin.
2. Fullscreen another surface.
3. xfce4-panel external plugins overlay fullscreen window (as the fullscreen state event was not received to demote these surfaces layer)

**Expected behavior**
This event should be emitted when the other surface is fullsreened. 

**Wayfire version**
0.7.5

Issue is not present on Wayfire(git-master), so I'm looking for a fix on the 0.7.x branch.

marcusbritanicus:
[This](https://github.com/WayfireWM/wayfire/pull/1557) is the PR that fixed this issue on the git master. You should be able to make a patch similar to this.

simplejack-src:
That (offhand) looks pretty simple to patch into. Whats the release process for Wayfire look like? i.e. Is it worth it for me to look to put an MR up for this on the 0.7.x branch?

simplejack-src:
Patch cleanly applied and very much fixed this issue (thanks @marcusbritanicus!). PR opened to have it applied to the `0.7.x` branch (#1772).

PR #1557, opened 2022-08-22
Implement wlr_foregin_toplevel_handle's fullscreen management

marcusbritanicus:
**Notice: Wayfire's development is temporarily happening in the `stabilize-api` branch. If you open a pull request for the master branch, it will likely not be merged before the stabilize-api branch is itself merged into master.**

PR #1772, opened 2023-04-07
core: Backport 'Implement wlr_foregin_toplevel_handle's fullscreen' fix

simplejack-src:
(#1557) fix fix for (#1771).


ammen99:
Why the second commit?

ammen99:
I removed the second commit because I do not think that it is the proper fix. The other commit can be merged in 0.7.x.

Issue #1725, opened 2023-02-17
simple-tile: all input ignored after moving or resizing a window


kbtz:
While using the simple tile plugin, all keyboard and mouse input seems stop after releasing the `super` key for moving or resizing a window. I've installed [keyd](https://github.com/rvaiya/keyd) recently but the issue is reproducible even with its service stopped (ie no virtual kb).

**Wayfire version**
I've compiled targeting master and b8b8486613b297b8fe02ce1a0439e18577cfbbea (jan/07) and the issue persists, while the version compiled against 8d7b4cf36d6ad245580fd473c8faa3c0fab15561 (jan/04) works fine.

**Stacktrace**
No errors reported in the log however [input-method-relay](https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/input-method-relay.cpp) keeps emitting messages like these even on the jan/04 version so I'm not sure if related:
```log
Disabling text input, but input method is gone
Enabling text input, but input method is gone
Committing text input, but input method is gone
Inactive text input tried to commit
Inactive text input tried to commit
Committing text input, but input method is gone
Committing text input, but input method is gone
```

ammen99:
Has this been fixed on latest master? There seems to be a PR which supposedly fixed it, and I cannot reproduce.

PR #1774, opened 2023-04-12
simple-tile: Fix no response to input after move/resize a tiled window.

jpy794:
After resizing a tiled window with the key binding `button_resize`, wayfire does not response to any input. To avoid this, We need to ungrab the input when deactivating the plugin.
I'm not sure if this is related to #1725, which seems to have a similar behavior.

In addition, the plugin never receives the  `WLR_BUTTON_RELEASED` event corresponding to the `WLR_BUTTON_PRESSED` event in the key binding. We have to mark the key binding as unhandled so that there's a record of the `WLR_BUTTON_PRESSED` event, then grab input with `retain_pressed_state` set to true.
I've also seen similar issue in `wrot`. I guess it could be fixed like this.

### DIFF
--- a/plugins/tile/tile-plugin.cpp
+++ b/plugins/tile/tile-plugin.cpp
@@ -219,23 +219,22 @@ class tile_plugin_t : public wf::per_output_plugin_instance_t, public wf::pointe
     }
 
     template<class Controller>
-    bool start_controller()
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Thanks!

PR #1775, opened 2023-04-14
DRAFT scale: include all views visible on workspace

mntmn:
Intuitively, "scale" should include all views that are visible in (overlap with) the current workspace, not only those that are mostly inside of it.


ammen99:
@soreau Thoughts?

By the way, if we change the behavior as you propose, the whole function can be replaced with a single call:

```cpp
output->wset()->get_views(wf::WSET_EXCLUDE_MINIMIZED | wf::WSET_MAPPED_ONLY | wf::WSET_CURRENT_WORKSPACE);
```

NamorNiradnug:
I think the behavior should be configurable to allow users to choose between old and new behavior.

PR #1778, opened 2023-04-21
Fix broken scale

### LINKED ISSUE #1750: window becomes weird when DPI scale set 2.0
https://user-images.githubusercontent.com/53224655/222875349-417b75c6-8eda-482b-a8ce-df463c516da4.mp4

**Wayfire config**
test with a new user and use default wayfire.ini config

**Wayfire version**
0.8.0-4dd373d2

soreau:
This missing line was causing only the top left quadrant of surfaces to be rendered with scale = 2, for example. Fixes #1750.


ammen99 (approved):
Thanks!

PR #1793, opened 2023-05-20
allow wayfire to search for plugins in the users home directories

epsilon-0:
~~There's A LOT to be done to make this a nicer feature~~

- ~~make the code not suck~~
- ~~add hashing values to the URLs~~
- ~~add icon loading capabilities for WCM~~

~~But hey, its a hack and a slash and a whole lotta all that.~~


ammen99 (approved):
LGTM, thanks!

PR #1797, opened 2023-05-21
Workspace Sets (#217)

ammen99:
As a prerequisite for finishing the transactions API, it seemed best to start with the workspace management reorganizations, as that also affects how views are handled.

Workspace sets are intended as a new feature which can provide solutions to multiple interesting problems in Wayfire, namely:

- Workspace sets consist of a normal workspace grid (where workspaces are adjacent to each other), however, they are not tied to a particular output.
- Each workspace set is completely independent of other workspace sets. In a way, having a workspace set with just 1 workspace is the same as having independent workspaces ala Sway-style.
- Workspace sets should also be transferrable between outputs. This should nicely solve issues which plugins like `preserve-output` tries to work around: when an output is temporarily disconnected, a plugin can just save its workspace set, and restore it when the output is reconnected.
- An output may have more than one workspace sets, even though only one of them is active. In this way, it acts like a third dimension in the workspace grid, but is way more flexible.

Of course, complete support for workspace set will also require plugins which can handle and work with them. As time goes, I will implement a simple plugin for switching workspace sets.

Rough roadmap:

- [ ] Refactor workspace-manager to not handle any information regarding layers, only workspaces.
- [ ] Make sure that workspace-manager only tracks toplevel views which are affected by workspace changes. Layer-shell and similar can add themselves to the scenegraph directly.
- [ ] Implement workspace-sets
- [ ] Implement demo plugin


ammen99:
Summary of the big changes for people updating external plugins:

- `output->workspace->get_workarea()` is now `output->workarea->get_workarea()` (and you need `#include <wayfire/workarea.hpp>` for it)
- `output->workspace` is now `output->wset()`
- Workspace sets have most of the functions that `workspace_manager` (aka `output->workspace`) had. The main differences are:
  - Workspace sets contain only toplevel views. Panels, backgrounds, etc. are not part of the list returned by `wset->get_views()`. If for some reason you need all views, you can use `collect_views_from_output` or `collect_views_from_scenegraph`: https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/view-helpers.hpp#L29
    In practice, plugins should either use `wf::get_core().get_all_views()` (example: cleaning up state when unloading the plugin), or `workspace_set_t::get_views()`. The latter function combines `output->workspace->get_views_in_layer()` and `output->workspace->get_views_on_workspace()`. It supports several usages, most of which should be self-explanatory:
```cpp
wset.get_views(wf::WSET_MAPPED_ONLY) // filters out unmapped views beforehand
wset.get_views(wf::WSET_EXCLUDE_MINIMIZED)
wset.get_views(wf::WSET_CURRENT_WORKSPACE)
wset.get_views(wf::WSET_SORT_STACKING) // When set, the views are sorted (otherwise, they are unsorted!)
wset.get_views(<any flags, bitwise OR of the above>)
wset.get_views(<any flags, bitwise OR of the above>, <workspace to get views on>)
```

  - The current workspace set on an output may change. The `wsets` plugin is an example of this, it manages a collection of workspace sets which may be moved between outputs like Sway workspaces. Each workspace set has its own grid of workspaces though.

- Because workspace sets have a ton of interactions with many plugins, current status is that they are optionally supported. That is, many plugins don't care about them at all, so they might not work very well together. For example, clicking on a view from another workspace set in wf-panel won't work (the foreign-toplevel plugin doesn't represent workspace-sets currently). The plan is to gradually add the necessary features/cases to other plugins as people need them. Some plugins (esp. external plugins) might just opt out to ignore the possibility of multiple workspace sets and treat output->wset() as constant, which is fine and shouldn't break anything (except of course the wset plugin).

PR #1810, opened 2023-06-17
Install also metadata for wayfire-shell plugin

dkondor:
This was left out from #1715 ?
At least when I did a recompile, the wayfire-shell.xml file did not get installed.


ammen99 (approved):
Thanks!

PR #1812, opened 2023-06-19
Switcher: Rotate and switch views from the active one

Teselka:
hI, i added ability to change rotation of the thumbnail images 

<details>
    <summary>Thumbnail rotated by 0 degrees</summary>

Screenshot 2023-06-19 19-24-21

</details>

also added ability to switch views from the current active window

### DIFF
--- a/metadata/switcher.xml
+++ b/metadata/switcher.xml
@@ -15,6 +15,12 @@
 			<_long>Switches to the previous view with the specified key.</_long>
 			<default>&lt;alt&gt; &lt;shift&gt; KEY_TAB</default>
 		</option>
+		<!-- Behavior -->
+		<option name="start_from_active" type="bool">
+			<_short>Start from active</_short>
+			<_long>Starts to switch views from the active window.</_long>
+			<default>false</default>
+		</option>
 		<!-- Effects -->
 		<option name="speed" type="int">
 			<_short>Duration</_short>
--- a/metadata/switcher.xml
+++ b/metadata/switcher.xml
@@ -15,12 +15,6 @@
 			<_long>Switches to the previous view with the specified key.</_long>
 			<default>&lt;alt&gt; &lt;shift&gt; KEY_TAB</default>
 		</option>
-		<!-- Behavior -->
-		<option name="start_from_active" type="bool">
-			<_short>Start from active</_short>
-			<_long>Starts to switch views from the active window.</_long>
-			<default>false</default>
-		</option>
 		<!-- Effects -->
 		<option name="speed" type="int">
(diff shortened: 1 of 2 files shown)


ammen99:
In general LGTM. I don't think we need the `start_from_active` option, the other switcher (fast-switcher) always orders by focus timestamp.

Does anyone have thoughts on this? @marcusbritanicus @soreau  ?

marcusbritanicus:
@ammen99 @Teselka I agree `start_from_active` isn't needed.

If I remember correctly, one behaviour that has always bugged me is that the switcher plugin does not order the views in the same manner as fast-switcher.

To be specific, from the active view (say, V1), if `alt-tab` is pressed, it goes to the next view (say, V2). Pressing `alt-tab` again should make the previous view active, but currently, it goes to third view in the list, V3.

Of course, if you want to go to V3, from V1, bypassing V2, the press `alt` and then `tab` twice.

Teselka:
> To be specific, from the active view (say, V1), if `alt-tab` is pressed, it goes to the next view (say, V2). Pressing `alt-tab` again should make the previous view active, but currently, it goes to third view in the list, V3.

Sorry, maybe i didn't understand the idea but isn't expected behavior? Why it should make V1 active when you just switched to V2?

> I agree `start_from_active` isn't needed.
I think that it's really needed because it's really annoying when you have for example 15 opened views and you need to select 10 view that you just used (maybe this option must be renamed to something like `order_by_activity`?). Also it's extremly handful to see previews of the views when switching them because they can be over another view (with fast-switcher you can see them transparent but what if you have 2-3 fullscreen windows of the same color?).

ammen99:
@Teselka I agree sorting by last focus timestamp is the better behavior, what I am suggesting is that it is not optional (i.e. the option is not needed), it is simply always sorted by timestamp.

Teselka:
I just fixed code style issue so uncrustify checks must pass now. Also i think that it's ready to merge

marcusbritanicus:
@Teselka @ammen99 I can confirm that the views are ordered by lat focus timestamp.

However, I have one tiny nitpick: When I press `Alt+Tab`, the animation shows and the currently focused view is the one that's at the center. To focus the next view, `Tab` needs to be pressed once more (something like, `Alt+Tab+Tab`). While it's not much of a burden to do this, it would be better, if the next view is shown directly.

I have a request: It would be great if you implemented: Would it be possible to add a time delay before the animation begins? If the user presses `Alt+Tab` and releases both within a short duration (say 500 ms, adjustable), then animation is not shown. Otherwise, show the animation.

ammen99:
Thanks!

PR #1865, opened 2023-08-25
ipc-rules: add focus method for ipc

ssfdust:
* Add window-rules/focus-view method to window-rules plugins
* Add focus method to wayfire_socket.py

### DIFF
--- a/ipc-scripts/wayfire_socket.py
+++ b/ipc-scripts/wayfire_socket.py
@@ -65,6 +65,11 @@ class WayfireSocket:
         message["data"]["view_id"] = view_id
         return self.send_json(message)
 
+    def set_focus(self, view_id: int):
+        message = get_msg_template("window-rules/focus-view")
+        message["data"]["id"] = view_id
+        return self.send_json(message)
+
     def set_always_on_top(self, view_id: int, always_on_top: bool):
         message = get_msg_template("wm-actions/set-always-on-top")
         message["data"]["view_id"] = view_id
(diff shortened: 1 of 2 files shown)


ssfdust:
A simple example that uses wofi to switch between windows with this patch
```Python
#!/bin/env python

import socket
import os
import json as js
import subprocess

def get_msg_template(method: str):
    message = {}
    message["method"] = method
    message["data"] = {}
    return message


class WayfireSocket:
    def __init__(self, socket_name):

    def read_exact(self, n):
        response = bytes()
        while n > 0:
            if not read_this_time:
            n -= len(read_this_time)
            response += read_this_time

        return response

    def read_message(self):

    def send_json(self, msg):
        self.client.send(header)
        self.client.send(data)
        return self.read_message()

    def get_views(self):
        message = get_msg_template("stipc/list_views")

    def set_focus(self, view_id: int):
        message = get_msg_template("window-rules/focus-view")
        message["data"]["id"] = view_id


def main():
    sock = WayfireSocket(addr)

    windows_map = {}
    for view in sock.get_views():
        if view['layer'] == 'workspace':
            index = view['app-id'] + ' - ' + view['title']
            windows_map[index] = view['id']

    indexs = "\n".join(windows_map.keys()).encode()
    proc = subprocess.Popen(["wofi", "--dmenu"], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
    window_index = proc.communicate(input=indexs)[0].decode().strip()
    window_id = windows_map[window_index]

    print(sock.set_focus(window_id))

if __name__ == '__main__':
    main()
```

ammen99 (approved):
Thanks!

ammen99:
> A simple example that uses wofi to switch between windows with this patch


Cool, seems like an interesting idea!

ammen99:
We can merge once the code style is fixed

PR #1882, opened 2023-09-01
xwayland + other fixes

### LINKED ISSUE #1859: Several problems with Steam
**Describe the bug**

The steam desktop application is quite problematic under Wayfire. One problem is that clicking on 'Friends & Chat' in the left lower corner should open the chat window, but it doesn't do anything. Several Wayfire builds ago, it used to work, but then the window would often appear underneath the main window, instead of on top, as is expected.

Another problem (but unfortunately one that's harder to pin down) is that the application's UI often stops responding properly, with popup menus possibly being the culprit. Just try navigating the application for a while - it should be apparent what I mean.

**To Reproduce**

Click on 'Friends & Chat' to open the chat window

**Expected behavior**

The Chat window should appear, but doesn't (it seems Steam does not register the mouse click properly?)

**Screenshots or stacktrace**

**Wayfire version**

0.8.0-f1d524df (Aug  3 2023, branch 'master'), wayfire-git

### LINKED ISSUE #1862: system crashes using follow-focus
after updating to latest master and enabling follow-focus plugin i get constant crashes.  i disabled follow-focus and no more crashes.

### LINKED ISSUE #1881: Steam is not visible in multi-monitor setup when it tries to open itself on another output
Wayfire doesn't properly constrain steam to the visible part of the current output.

ammen99:
- stipc: add get_xwayland_display
- xwayland: rewrite mapping functionality
- xwayland-unmanaged-view: improve tracking of position

@EoceneMiacid @soreau Can you test again with steam (and any other Xwayland clients which are important for you)?

Fixes #1859 
Fixes #1881 
Fixes #1862

### DIFF
--- a/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
+++ b/plugins/common/wayfire/plugins/common/move-drag-interface.hpp
@@ -609,12 +609,12 @@ class core_drag_t : public signal::provider_t
         wf::get_core().default_wm->set_view_grabbed(view, false);
         view = nullptr;
         all_views.clear();
-        current_output     = nullptr;
-        view_held_in_place = false;
+        current_output = nullptr;
         wf::get_core().set_cursor("default");
 
         // Lastly, let the plugins handle what happens on drag end.
         emit(&data);
+        view_held_in_place = false;
         on_view_unmap.disconnect();
     }
 
--- a/plugins/decor/decoration.cpp
+++ b/plugins/decor/decoration.cpp
@@ -1,5 +1,6 @@
 #include <wayfire/per-output-plugin.hpp>
 #include <wayfire/view.hpp>
+#include <wayfire/workarea.hpp>
 #include <wayfire/matcher.hpp>
 #include <wayfire/workspace-set.hpp>
 #include <wayfire/output.hpp>
@@ -110,6 +111,10 @@ class wayfire_decoration : public wf::plugin_interface_t
         if (!pending.fullscreen && !pending.tiled_edges)
         {
             pending.geometry = wf::expand_geometry_by_margins(pending.geometry, pending.margins);
+            if (view->get_output())
+            {
+                pending.geometry = wf::clamp(pending.geometry, view->get_output()->workarea->get_workarea());
+            }
         }
     }
 
--- a/plugins/ipc/stipc.cpp
+++ b/plugins/ipc/stipc.cpp
@@ -345,6 +345,7 @@ class stipc_plugin_t : public wf::plugin_interface_t
         method_repository->register_method("stipc/tablet/pad_button", do_pad_button);
         method_repository->register_method("stipc/delay_next_tx", delay_next_tx);
         method_repository->register_method("stipc/get_xwayland_pid", get_xwayland_pid);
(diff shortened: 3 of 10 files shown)


EoceneMiacid:
Hi, just compiled the xwayland-fixes branch (0.8.0-d34b291b (Sep  1 2023, branch 'HEAD'))

The menu bug in Steam on a multimonitor setup appears to be fixed. Regardless of the monitor Steam is on, the menu appears  to work. The 'friends & chat' problem (click to open, click again to hide, click to re-open now shows the window below the main window) is still there.

EoceneMiacid:
It happens with other windows as well, and is consistently reproducible.

ammen99:
This is weird, I cannot reproduce any of the two issues .. Can you attach logs with `-d -d xwayland -t txn -d txni -d views` for the steam disappearing?

ammen99:
It seems to me the problem is caused by the very big snap off threshold (it is still a bug in Wayfire though)

EoceneMiacid:
I see... i made the snapoff that big because of your suggestion, since it makes for a nice effect when combined with wobbly windows.

ammen99:
> I see... i made the snapoff that big because of your suggestion, since it makes for a nice effect when combined with wobbly windows.

I would say the correct fix would be that the window is not moved to the other output at all when the snap off threshold is that big (of course as long as it is not exceeded). I'll work on this.

ammen99:
@EoceneMiacid The disappearing when dragging maximized window should be fixed now.

EoceneMiacid:
Can confirm, nice.

This also appears to have fixed a tiny micro-bug where trying to drag a maximized window with alt held down (so it can be dragged from anywhere instead of just the titlebar) now behaves consistently.

Before, it would stretch normally the first time, but from the wrong offsets (if that's the correct terminology here) the second time, then normally again, wrong, etc.

ammen99:
@EoceneMiacid I tested Steam on Weston. The chat & friends button works the same as in Wayfire. On first click, the window shows up. On subsequent clicks, nothing happens, however, the main Steam window is brought to the front. If steam is maximized, the chat window ends up on the back so you might think it was closed, but it never gets closed - so everything works as expected.

I suppose you'd want to set the chat window to always-on-top (see for ex. wm-actions, you can also write a short IPC script to do that automatically based on the chat window's app-id/title), but there is no bug in Wayfire related to this.

EoceneMiacid:
Thanks, I appreciate the effort you took to verify this.

It's hard to nail down this kind of problem at the moment because there are so many variables and there is still a lot of transitioning to Wayland going on. From now on I will test in Weston first before filing bugs.

I still experience a crash every now and then when exiting a fullscreen XWayland application, by the way.

ammen99:
> I still experience a crash every now and then when exiting a fullscreen XWayland application, by the way.

Weird, please open a new issue and attach address sanitizer stacktrace to it.

EoceneMiacid:
I'm working on narrowing down the cause of the crash.

I just recompiled the latest master from git, but to my surprise the large-snapoff-makes-windows-disappear-when-dragged-across-displays bug is back.

EoceneMiacid:
I don't see how.

Anyway it doesn't occur anymore now, so I'm not sure what happened.

The crash appears to be gone as well.

PR #1884, opened 2023-09-03
wobbly: Fix wobbly after second grab while wobbling

### LINKED ISSUE #1690: Grabbing a wobbly window causes it to quit wobbling
Grabbing a window that is still wobbling causes it to stop wobbling even though it is still grabbed.

To repro, grab a window and wobble it, then let go and grab again, before the wobbly animation completes. Wobbly stops happening and the window grab behaves as if wobbly is not enabled.

I expect to be able to grab the window as many times in any state and it will continue wobbling as long as a grab is held.

Latest git c566170b.

soreau:
Fixes #1690.


ammen99 (approved):
Nice catch!

PR #1887, opened 2023-09-05
[Switcher Plugin]: Animate to show next view on trigger

marcusbritanicus:
With this PR, the "next view" is shown as the "center view" as soon as the plugin is activated. This removes the need to press `Alt-Tab-Tab` to focus the "next view".


ammen99 (approved):
Thanks!

Issue #1745, opened 2023-03-01
fast-switcher: alternative animation


Vladimir-csp:
Please add something that would animate the window to switch to, not all other windows except that. For example, a simple color highlight.

When there are two windows on screen, instinct tells you've switched to the one that flickers, but in case of fast switcher's alpha fade it's the other way around.

Also, please make it cycle through minimized windows (at the far end of cycle)

timgott:
I think it would be nice if fast-switcher would directly focus the window and not just when the key it is released. Then the normal window decorations would change color and show which one you switched to.

Is there a reason why it currently does not do this?

occivink:
It's not exactly what you're asking for, but I submitted a PR for alpha animations, maybe you can try it out?

>I think it would be nice if fast-switcher would directly focus the window and not just when the key it is released. Then the normal window decorations would change color and show which one you switched to.

The problem of doing this is that it adds a lot of noise to the history of focused windows, which in turns affects the order in later invocations of the fast-switcher.

timgott:
> The problem of doing this is that it adds a lot of noise to the history of focused windows, which in turns affects the order in later invocations of the fast-switcher.

That sounds like an implementation detail, it should be possible to avoid that and only commit the window to the focus history after releasing the key.

Your PR is nice but I'm not sure whether it solves this usability problem. The only way to know which window is selected is to find the one that does not become transparent, which is sometimes hard to tell. Is it easier with the animation?

There should be a positive hint for the selected window, not just negative hints for the non-selected windows.

timgott:
Setting `view->set_activated` works and does not change the order. I tried it, it's much better than before if you use a strong focus effect like I do.

<details>
<summary>Patch</summary>

``` diff
diff --git a/plugins/single_plugins/fast-switcher.cpp b/plugins/single_plugins/fast-switcher.cpp
index 9bef90b9..a919bf2e 100644
--- a/plugins/single_plugins/fast-switcher.cpp
+++ b/plugins/single_plugins/fast-switcher.cpp
@@ -67,6 +67,7 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
 
         if (reorder_only)
         {
+            views[i]->set_activated(true);
             output->workspace->bring_to_front(views[i]);
         } else
         {
@@ -193,6 +194,7 @@ class wayfire_fast_switcher : public wf::per_output_plugin_instance_t, public wf
     {
 #define index current_view_index
[...]

PR #1897, opened 2023-09-12
fast-switcher: Set activated on selected view

timgott:
Set selected view to activated to get focus decorations. This makes the currently selected view much easier to see (described in #1745).

Also I just had to remove that #define.


ammen99 (approved):
I see you even included a bug fix in `view_chosen()`, nice catch!

Thanks for the PR :)

PR #1898, opened 2023-09-15
Draft: Implement xdg-activation v1 protocol plugin

kode54:
This is a draft, in case it is desired to implement the XWayland equivalent of this protocol into the same plugin.


plugins/protocols/xdg-activation.cpp
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if (!view->get_output())
+        {
+            LOGI("View has no output");
+            return;
+        }
+
+        LOGI("Activating view");
+        view->get_output()->focus_view(view, true);
```
ammen99, line 60: Focus request is more appropriate here (it will also switch to the target workspace, etc.):

https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/window-manager.hpp#L45

With it, you also shouldn't need the check above (about the output)

plugins/protocols/xdg-activation.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  private:
+    static void xdg_activation_handle_request_activate(struct wl_listener *listener, void *data)
+    {
+        auto event = static_cast<const struct wlr_xdg_activation_v1_request_activate_event*>(data);
+
+        wayfire_view view = wf::wl_surface_to_wayfire_view(event->surface->resource);
+        if (!view)
+        {
+            LOGI("Could not get view");
```
ammen99, line 43: Sounds like an error to me :)

plugins/protocols/xdg-activation.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <memory>
+#include <wayfire/plugin.hpp>
+#include <wayfire/view.hpp>
+#include <wayfire/nonstd/wlroots-full.hpp>
+#include <wayfire/window-manager.hpp>
+#include "config.h"
+
+extern "C"
+{
+#include <wlr/types/wlr_xdg_activation_v1.h>
```
ammen99, line 13: This should go to `wayfire/nonstd/wlroots-full.hpp`

ammen99 (commented):
After looking at the sway pr, I see that the launcher contexts are optional (we don't have that concept in wayfire yet so it can be left aside).

Also, leaving a few comments, even though this is a draft :)

plugins/protocols/xdg-activation.cpp
kode54: I've changed this, and it still does what it's supposed to.

plugins/protocols/xdg-activation.cpp
kode54: Changed this to log an error.

plugins/protocols/xdg-activation.cpp
kode54: Added this include to `wayfire/nonstd/wlroots-full.hpp`

ammen99:
I think something went wrong, because I do not see any changes since last time ..

kode54:
Oh, that's ducking stupid. I somehow managed to amend commit without staging the new files.

Issue #1811, opened 2023-06-17
Is there an alternative to call_plugin()?


dkondor:
output_t::call_plugin() was removed in #1674 (as plugins are not created for each output anymore). Is there a replacement for its functionality? I would like to be able to programmatically activate other plugins (as if the user pressed the associated keybinding), and this was a very convenient feature, since it only required the name of the action (e.g. "scale/toggle").
Thanks!

ammen99:
The idea was to enable calling plugins via IPC, replacing the call_method in this way. It would also allow additional parameters to be passed. I have yet to decide on a good API, I think it would make sense to have this ready before 0.8.

ammen99:
Forgot to mention: IPC methods can of course be also called from other plugins, and for many plugins we'd want a simple wrapper which combines a regular binding + ipc call.

dkondor:
Thank you, this makes sense!

ammen99:
In #1864 I have added support for these ipc-activators to the scale plugin since that's what you mentioned here. Other plugins can be handled similarly, PRs welcome.

Here's an example of how to call scale (haven't tested but should work):

```cpp
nlohmann::json data;
data["output_id"] = output->get_id();

shared_data::ref_ptr_t<ipc::method_repository_t> repo;
repo->call_method("scale/toggle", data);
```

PR #1899, opened 2023-09-18
expo: add support for ipc triggering

dkondor:
Following up on #1811 I've implemented the IPC based activator for the Expo plugin (based on the example with scale). I have tested activating it from another plugin and via IPC. So far, I haven't included the workspace_bindings, since I'm not sure if there is a good use-case for them (they are only active when expo is activated).


ammen99 (approved):
Thanks!

PR #1900, opened 2023-09-19
Fix wf-config version when using wf-config from system

ssfdust:
* correct version from 0.16 which is the version of wlroots to 0.8


ammen99 (approved):
Thanks!

PR #1903, opened 2023-09-22
Fix "No free output buffer slot" messages

soreau:
The problem here was that moving the cursor during repaints was sometimes causing all free buffer slots in wlroots to be used because we were calling wlr_cursor_set_surface() more than once per frame. Call this on idle so that we only do it once per frame and avoid using all the free buffer slots.

### DIFF
--- a/src/core/seat/cursor.cpp
+++ b/src/core/seat/cursor.cpp
@@ -30,10 +30,49 @@ wf::cursor_t::cursor_t(wf::seat_t *seat)
 
     wf::get_core().connect(&config_reloaded);
 
-    request_set_cursor.set_callback([&] (void *data)
+    request_set_cursor.set_callback([=] (void *data)
     {
         auto ev = static_cast<wlr_seat_pointer_request_set_cursor_event*>(data);
-        set_cursor(ev, true);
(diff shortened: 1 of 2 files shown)


src/core/seat/cursor.cpp
```diff
@@ -33,7 +33,10 @@ wf::cursor_t::cursor_t(wf::seat_t *seat)
     request_set_cursor.set_callback([&] (void *data)
     {
         auto ev = static_cast<wlr_seat_pointer_request_set_cursor_event*>(data);
-        set_cursor(ev, true);
+        idle_set_cursor.run_once([=] ()
+        {
+            set_cursor(ev, true);
```
ammen99, line 38: This is definitely use-after-free, the event may be freed until idle.

ammen99: Also, we have to make sure that the cursor surface doesn't get destroyed until idle

ammen99 (changes requested):

ammen99:
As discussed on IRC, this is kind of hard to fix, and the benefits are not very clear ..

PR #1904, opened 2023-09-22
Send configure to views when output configuration is changed

soreau:
This effectively fixes wf-background updating when resizing the wayland backend window.


ammen99:
Closing in favor of #1905, thanks for the investigation though :)

PR #1906, opened 2023-09-23
Do not send a view-disappeared signal when a minimized view is restored

dkondor:
See the discussion in #1274 
I did some quick testing and a search for where this signal is used, but let me know if I could do some more tests for regressions.

### DIFF
--- a/src/view/toplevel-view.cpp
+++ b/src/view/toplevel-view.cpp
@@ -211,9 +211,12 @@ void wf::toplevel_view_interface_t::set_minimized(bool minim)
     if (get_output())
     {
         get_output()->emit(&data);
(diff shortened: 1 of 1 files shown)


ammen99:
> I did some quick testing and a search for where this signal is used, but let me know if I could do some more tests for regressions.

Well, I have a few tests here: https://github.com/ammen99/wayfire-tests, however, they are quite brittle because they use small timeouts for most tasks .. I usually install a slightly outdated Wayfire version and compare it with the version I want to test.

Anyway, I ran all the tests and there are no regressions, so this should be quite good :) Thanks!

ammen99 (approved):

PR #1907, opened 2023-09-23
Visibility & surface frame done improvements

ammen99:
When we have a pending transaction, we want to make sure we send a frame done to the client, so that it can resize. However, if it is already visible on an output, we should not send additional frame done events. Instead, we should wait until the 'normal' frame done events are sent as part of the rendering loop.

### DIFF
--- a/src/view/wlr-surface-node.cpp
+++ b/src/view/wlr-surface-node.cpp
@@ -110,7 +110,7 @@ wf::scene::wlr_surface_node_t::wlr_surface_node_t(wlr_surface *surface, bool aut
     {
         if (!wlr_surface_has_buffer(this->surface) && this->visibility.empty())
         {
-            send_frame_done();
+            send_frame_done(false);
         }
 
         if (this->autocommit)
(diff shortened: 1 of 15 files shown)


src/view/wlr-surface-node.cpp
```diff
@@ -193,13 +193,19 @@ wf::touch_interaction_t& wf::scene::wlr_surface_node_t::touch_interaction()
...
-void wf::scene::wlr_surface_node_t::send_frame_done()
+void wf::scene::wlr_surface_node_t::send_frame_done(bool delay_until_vblank)
 {
-    if (surface)
+    if (surface && (!delay_until_vblank || visibility.empty()))
     {
         timespec now;
         clock_gettime(CLOCK_MONOTONIC, &now);
         wlr_surface_send_frame_done(surface, &now);
+    } else if (surface)
```
soreau, line 203: since you're checking `surface != NULL` in both cases, maybe just wrap these two conditions in `if (surface)`.

soreau:
Tested by @soreau

PR #1909, opened 2023-09-24
Change workspace_bindings to wsets_bindings in wsets

marcusbritanicus:


ammen99 (approved):
LGTM, thanks!

This is a breaking change in the config file, but I don't think anyone except me uses wsets so far ..

PR #1911, opened 2023-09-27
DRAFT: IPC method to start scale with an app-id filter

dkondor:
This is a long-time wish of mine that would allow a panel / launcher to start the scale plugin, filtering for only views that belong to a specific type of application (based on app-id). This would allow recreating the functionality that was available when using Compiz + Cairo-Dock together, where when clicking on an icon that represents a group of windows, scale would open including only that group.

This PR adds an IPC method that starts scale and also applies an app-id filter to it. For now, it can be invoked e.g. with the following Python script (using the functionality from [wayfire_socket.py](https://github.com/WayfireWM/wayfire/blob/master/ipc-scripts/wayfire_socket.py)):
```
import wayfire_socket
ws = wayfire_socket.WayfireSocket('/tmp/wayfire-wayland-2.socket')
msg = wayfire_socket.get_msg_template('scale/activate_appid')
msg['data']['app_id'] = 'lxterminal'
msg['data']['all_workspaces'] = True
ws.send_json(msg)
```

Would this functionality belong in the main repository? If yes, should this be part of the scale plugin or an IPC plugin? (I think that a separate plugin for this is overkill)


ammen99:
Anyway, thanks for the PR :)

PR #1913, opened 2023-09-28
api: make input focus global state

### LINKED ISSUE #1892: Moving view activates view on other display
**Describe the bug**
If there are windows open on two displays, and you move a window by dragging, the decoration of the last focused window on the other display changes to focused state. It happens once you release the drag.

Keyboard focus does not change, only the visual state of the decoration, with client-side and with server-side deco.

**To Reproduce**
1. Open windows on two different displays
2. Move windows by dragging them
3. Decoration on other window will show focus once you release, you might have two focused-looking windows at the same time

**Expected behavior**
Moving a window should have no side effects

**Screenshots or stacktrace**
/

**Wayfire version**
git `17d4eee`

ammen99:
The keyboard input focus has always been a global state (to be precise, per-seat, but Wayfire has only one seat anyway). However, previously the focused view has been managed per-output, which makes little sense and does not reflect reality, nor the desired behavior. This commit moves all keyboard-focus functionality out of `wf::output_t` and into `wf::seat_t`.

Fixes #1892

### DIFF
--- a/plugins/common/wayfire/plugins/common/input-grab.hpp
+++ b/plugins/common/wayfire/plugins/common/input-grab.hpp
@@ -8,6 +8,7 @@
 #include <wayfire/debug.hpp>
 #include <memory>
 #include <string>
+#include <wayfire/seat.hpp>
 namespace wf
 {
 namespace scene
@@ -146,8 +147,7 @@ class input_grab_t
         children.insert(idx, grab_node);
         root->set_children_list(children);
         wf::get_core().transfer_grab(grab_node);
-        scene::update(root, scene::update_flag::CHILDREN_LIST);
-        output->refocus();
+        scene::update(root, scene::update_flag::CHILDREN_LIST | scene::update_flag::REFOCUS);
 
         // Set cursor to default.
         wf::get_core().set_cursor("default");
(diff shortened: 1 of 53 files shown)


ammen99:
Breaking changes for plugins: focus is no longer per-output, hence:

1. `wf::get_core().get_active_output()` and `wf::get_core().focus_output()` have been moved to `wf::seat_t` (so for example use `wf::get_core().seat->focus_output()` to focus an output).
2. `output->refocus()` becomes `wf::get_core().seat->refocus()`.
3. `output->get_active_view()` can be replaced with `wf::get_active_view_for_output(output)` or `wf::get_core().seat->get_active_view()`. The latter returns the globally focused view, the former checks whether the output is focused, if not, it returns null, otherwise the same as the seat's `get_active_view()`.
4. `output->focus_view()` is not a very well defined function, because it does a ton of things. There are multiple alternatives available, depending on what one wants to do:
    - `wf::get_core().seat->focus_view()`: this is a very basic function and just focuses the view, no unminimizing, raising the view, etc. This is a rather low-level function.
    - `wf::get_core().default_wm->focus_raise_view()`: this is pretty much calling `output->focus_view(view, true)`: focus the view, raise to the top, unminimize, etc.
    - `wf::get_core().default_wm->focus_request()`: this is the same as `focus_raise_view()`, but it emits a focus_request signal, which gives plugins a chance to deny the focus request. This is used for example when a new view is mapped, so that focus stealing prevention can work.
    - There is no output-specific view-focused signal, instead, plugins can use the global keyboard focus changed in core to figure out what is going on with focus. Also, the pre-focus signal has been removed in favor of the focus-request as explained in the previous point.

Issue #1918, opened 2023-09-30
Need to set WLR_USE_UNSTABLE when compiling external plugins?


dkondor:
Since Wayfire headers include wlroots headers, WLR_USE_UNSTABLE needs to be set before including them. This can be a bit strange if compiling a plugin that does not use wlroots directly (e.g. a simple example [here](https://github.com/dkondor/wayfire-scale-ipc/blob/main/meson.build): removing `-DWLR_USE_UNSTABLE` from `cpp_args` makes the build fail).

It might be nice to add this to build options automatically, e.g. have `pkg-config --cflags wayfire` provide it.

marcusbritanicus:
I guess it is for the user to know and understand that the current wlroots is unstable... A license agreement, sort of. 

> By using `-DWLR_USE_UNSTABLE` you understand that wlroots is unstable and the developers of wlroots are not to be held accountable for sudden and api-breaking changes that may come in the future.

And by adding `-DWLR_USE_UNSTABLE` to cflags, wayfire will be deciding for the user, which is not right.

ammen99:
> e.g. a simple example [here](https://github.com/dkondor/wayfire-scale-ipc/blob/main/meson.build)

As a side note, the repo seems to return 404, maybe it is a private repo?

> And by adding `-DWLR_USE_UNSTABLE` to cflags, wayfire will be deciding for the user, which is not right.

I agree with the general idea. Are there any Wayfire headers (except wlroots-full.hpp) which require `-DWLR_USE_UNSTABLE`?

dkondor:
> I guess it is for the user to know and understand that the current wlroots is unstable... A license agreement, sort of.
> 
> > By using `-DWLR_USE_UNSTABLE` you understand that wlroots is unstable and the developers of wlroots are not to be held accountable for sudden and api-breaking changes that may come in the future.
> 
> And by adding `-DWLR_USE_UNSTABLE` to cflags, wayfire will be deciding for the user, which is not right.

Yes, I guess it is a trade-off between making life easier for plugin writers who don't care about wlroots at all (only use Wayfire features) vs. harder for users who explicitly care about wlroots. However, currently, most of the wlroots API is unstable, so any plugin that uses wlroots will likely define this. And including the Wayfire headers will include the unstable wlroots symbols anyway.

dkondor:
> > e.g. a simple example [here](https://github.com/dkondor/wayfire-scale-ipc/blob/main/meson.build)
> 
[...]

PR #1923, opened 2023-10-01
Always define WLR_USE_UNSTABLE

dkondor:
Suggested improvement for #1918 

This is the simplest possible solution that will leave WLR_USE_UNSTABLE defined -- if this is not desired, I could change this to conditionally undefine it after the wlroots includes (if it was not originally defined).


ammen99 (approved):
Yeah, seems that we already force an include to `wlr_input_device.h` which is unstable - and the user cannot choose to use Wayfire's headers without `WLR_USE_UNSTABLE` - so this makes sense. Thanks!

PR #1924, opened 2023-10-01
Add metadata files for the IPC and IPC rules plugin

### LINKED ISSUE #1917: No metadata for IPC plugin?
I've noticed that the IPC plugin does not have an xml file in metadata, which means that it does not show up in WCM and can only be enabled by editing the config file. It could be nice to have this (unless there are plans to change how the IPC plugin works or if it is not yet intended for general usage).

Alternatively, I've noticed that there are a lot of "protocol-like" plugins (IPC and all the other protocols) that do not have any configuration options and clutter WCM. Maybe grouping them together on a "protocols" or "utilities" page can be helpful?

dkondor:
Fixes #1917


ammen99 (approved):
Thanks :)

PR #1926, opened 2023-10-01
wm-actions: allow show desktop to be called from IPC (and other plugins)

dkondor:
Adding some more functionality in line with #1811


dkondor:
I have no idea why the build is failing, it works fine for me, and the error comes from ipc-activator.hpp which I did not touch

ammen99:
You probably need to add `plugins_common_inc` or whatever the meson var was called for the `wayfire/plugins/common` include directory - it is needed for the ipc activators.

ammen99 (approved):
Thanks!

PR #1941, opened 2023-10-07
resize: allow keyboard interaction

dkondor:
This is something I had in a half-finished state for a long time.
This adds the possibility of a keybinding to start resize using the arrow keys. Enter or a click ends it. I've also added an IPC interface.

### DIFF
--- a/metadata/resize.xml
+++ b/metadata/resize.xml
@@ -15,5 +15,16 @@
 			<_long>When the specified button is held down, you can drag a window to resize it while perserving its orignal aspect.</_long>
 			<default>disabled</default>
 		</option>
+		
+		<option name="activate_key" type="key">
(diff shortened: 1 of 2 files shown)


ammen99:
I don't really think this belongs upstream, see https://github.com/WayfireWM/wayfire/pull/1319 as well - this can be mostly achieved with a simple ipc script + the command plugin (it also supports repeatable bindings)

ammen99:
The notable difference that I see is that this here starts a grab and then allows keybindings which are just arrows - is this important for you?

dkondor:
> The notable difference that I see is that this here starts a grab and then allows keybindings which are just arrows - is this important for you?

Not that much -- this was always low priority for me. I'll look into the other possibilities you linked.

PR #1944, opened 2023-10-07
build: Tell meson to ignore the return value of git commands

soreau:
We grab the commit hash and branch name for use in the log, but this does not work when building from a tarball. Tell meson to ignore the return value by setting 'check: false' for git commands.


ammen99:
Thanks :)

Issue #1949, opened 2023-10-08
Some windows maximize to underneath the panel


soreau:
Some windows like wcm and geany are problematic to maximize. They have the wrong position and end up to the left and under the panel.

soreau:
Fixed by #1950.

PR #1950, opened 2023-10-08
xdg-toplevel: update wm-offset before emitting signal

ammen99:


soreau:
This also fixes #1949.

PR #1963, opened 2023-10-19
ipc-rules-demo.py: Fix error setting 'workarea'.

flexibeast:


ammen99 (commented):
This seems wrong, we do a query for the output, not for the view. Have you tested this change?

flexibeast:
@ammen99: Yes; the reason i made this change is that the script didn't work for me until i made it.

ammen99:
Ohh I see what is going wrong, this is not the correct fix though. I'll try to fix this and will open a new PR (with the actual fix)

flexibeast:
Ah, fair enough - thanks!

PR #1964, opened 2023-10-19
ipc-scripts: fix ipc-rules-demo

ammen99:


ammen99:
@flexibeast: this should be the correct fix, let me know whether it works.

flexibeast:
@ammen99: Yep, that works here. :-) Thanks!

PR #1974, opened 2023-10-25
support reposition popups

lilydjwg:
This implements [xdg-shell v3](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3514). GTK 4 can use the reposition functionality to move its popups, which avoids flickering by the fallback way to hide and show popups.

This is especially useful since Wayfire doesn't fully implement the wayland input protocol and input methods resort to client side rendering.

(This patch is adapted from an earlier patch about which I don't remember many details, but it works.)

### DIFF
--- a/src/view/xdg-shell.hpp
+++ b/src/view/xdg-shell.hpp
@@ -16,7 +16,7 @@ class wayfire_xdg_popup : public wf::view_interface_t
 {
   protected:
     wf::wl_listener_wrapper on_destroy, on_new_popup,
-        on_map, on_unmap, on_ping_timeout, on_repostion;
+        on_map, on_unmap, on_ping_timeout, on_reposition;
 
     wf::signal::connection_t<wf::view_geometry_changed_signal> parent_geometry_changed;
     wf::signal::connection_t<wf::view_title_changed_signal> parent_title_changed;
--- a/src/view/xdg-shell.hpp
+++ b/src/view/xdg-shell.hpp
@@ -16,7 +16,7 @@ class wayfire_xdg_popup : public wf::view_interface_t
 {
   protected:
     wf::wl_listener_wrapper on_destroy, on_new_popup,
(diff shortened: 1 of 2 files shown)


src/view/xdg-shell.hpp
```diff
@@ -16,7 +16,7 @@ class wayfire_xdg_popup : public wf::view_interface_t
 {
   protected:
     wf::wl_listener_wrapper on_destroy, on_new_popup,
-        on_map, on_unmap, on_ping_timeout;
+        on_map, on_unmap, on_ping_timeout, on_repostion;
```
ammen99, line 19: typo: reposItion

ammen99 (approved):
Looks simple enough. Which clients use this functionality so that I can also test this?

ammen99:
Oh I almost forgot .. there is a destroy() function where all of the wl_listeners need to be disconnected. Otherwise, you might get crashes in certain situations (destroying a popup with animation for example)

lilydjwg:
I'm using gnome-font-viewer and fcitx5 to test this.

* install the `fcitx5-im` package group on Arch Linux and an input method, e.g. `fcitx5-chinese-addons`
* add a Chinese input method in fcitx5-configtool
* GTK_IM_MODULE=fcitx gnome-font-viewer
* click the search icon to have an input box. activate fcitx5 (click tray icon if you don't know the key)
* move the window to the right corner so that the space on the right isn't enough for fcitx5 to place its window
* type something. you can see the fcitx5 window moves a bit as its size changes

Without this patch, the window will disappear and then re-appear frequently.

I don't know other applications affected actually. fcitx5 is the only reason I want this (but there aren't a lot of GTK 4 applications currently; Qt6 seems to still not having the support for reposition).

ammen99:
Ok, this sounds a little bit difficult to test, I will trust that you have tested this ;) Thanks!

Issue #1326, opened 2021-10-26
Change keyboard layout with shortcut


JustinDroege:
Maybe there is already a option for it, but is there any way to change the keyboard layout with a shortcut?

This would be a nice feature when this is not already in the current state

AlexJakeGreen:
What is a "shortcut" in this case?

This is an example how I configured keyboard layouts switch using CapsLock key
```
[input]
*** skipped ***
xkb_layout = pl,ru
xkb_model = pc104
xkb_options = grp:caps_toggle,grp_led:caps
xkb_rules = evdev
xkb_variant = ,ruu
```

Also I use a plugin to "remember" keyboard layout per window https://github.com/AlexJakeGreen/wayfire-kbdd-plugin

Voklen:
This can be achieved by setting `xkb_options` the same way the layout and variants are set. So although I don't think it's necessary to add any new features for this, I think it could be beneficial to make it clear how to do this from the default `wayfire.ini` since this is an essential feature if you have multiple layout. Something like (of course commented out)
```ini
[input]
xkb_layout = us,fr
xkb_variant = dvorak,bepo
xkb_options = grp:win_space_toggle
```
and then in the wiki link to https://man.archlinux.org/man/xkeyboard-config.7#Switching_to_another_layout

ammen99:
> This can be achieved by setting `xkb_options` the same way the layout and variants are set

Yeah absolutely. Feel free to send a PR for the README :)

In the meantime, I am closing this issue, since there isn't much else to be done about it.

PR #1981, opened 2023-10-26
Add `xkb_options` to `wayfire.ini`

Voklen:
This makes it clearer how to set the key to switch the keyboard layout, I think this should be included in the config because to use multiple layouts you need to set a way to switch between them. This was inspired by the discussion at #1326


ammen99 (approved):
Thanks!

PR #1986, opened 2023-10-29
Small fix to setting focus of on_demand layer shell surfaces

### LINKED ISSUE #1276: layer surface 'on_demand' keyboard focus does not receive focus until the user clicks on it.
**Describe the bug**
If we have a client that uses layer shell protocol, and set the keyboard focus to `on_demand`, it does not get the keyboard focus until the user clicks on it.

**To Reproduce**
1. Open a client that uses layer shell protocol and requires `on_demand` keyboard focus.
2. You'll see that it does not have keyboard focus, until you click it with a mouse pointer.

**Expected behavior**
Give keyboard focus to the layer surface as soon as it is mapped.

**Wayfire version**
git

**PS:** This was tested using a Qt app. It is possible that this bug is due to Qt's bad implementation of things.

mfxbe:
Fixes #1276

### DIFF
--- a/src/view/layer-shell/layer-shell.cpp
+++ b/src/view/layer-shell/layer-shell.cpp
@@ -508,7 +508,7 @@ void wayfire_layer_shell_view::map()
     wf_layer_shell_manager::get_instance().handle_map(this);
 
     auto& state = lsurface->current;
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Thanks :)

PR #1993, opened 2023-11-01
core: add an option to accept or discard command output

### LINKED ISSUE #1973: Command output is lost
It's very helpful to see what goes wrong when commands spawned from wayfire fails in some way, but wayfire currently discards their output:

https://github.com/WayfireWM/wayfire/blob/ada3a6ba2176134dc084cf373a38cd3521c1996d/src/core/core.cpp#L394-L399

I'm using lightdm and all output from the compositor is saved in `~/.xsession-errors`.

lilydjwg:
fixes #1973.


ammen99 (approved):
LGTM, thanks!

PR #2006, opened 2023-11-06
Config backend fixups

soreau:


src/default-config-backend.cpp
```diff
@@ -53,7 +53,7 @@ static int handle_config_updated(int fd, uint32_t mask, void *data)
 
     const auto last_slash = config_file.find_last_of('/');
```
ammen99, line 54: I guess we can use the built-in std::filesystem functions here to get the base file name? https://en.cppreference.com/w/cpp/filesystem/path/filename

src/default-config-backend.cpp
```diff
@@ -66,6 +66,13 @@ static int handle_config_updated(int fd, uint32_t mask, void *data)
         // - Config file was created inside parent directory
         should_reload |=
             (event->wd == wd_cfg_file) || (cfg_file_basename == event->name);
+
+        if ((event->name == cfg_file_basename) && (event->wd == wd_cfg_dir))
+        {
+            inotify_rm_watch(fd, wd_cfg_file);
+            wd_cfg_file =
+                inotify_add_watch(fd, (config_dir + "/" + cfg_file_basename).c_str(), IN_CLOSE_WRITE);
```
ammen99, line 74: I wonder, do we need to concatenate the two files here, isn't this basically `config_file.c_str()` here?

ammen99 (approved):
I tested this and it seems to work. I left two small suggestions, otherwise we can merge.

src/default-config-backend.cpp
soreau: It would be, except in the case where the config file is specified as a relative path. I wanted to make sure the file path we're watching is complete and not rely on relative paths.

ammen99 (approved):
Thanks :)

PR #2028, opened 2023-11-26
ipc-rules: change output from name to ID.

lcolitti:
This is more useful because the other IPCs take IDs and not names.

For backwards compatibility, also include the output name in an "output-name" member.


ammen99:
The idea is good, but why not just add a new output-id field, this way this is not a breaking change?

ammen99 (approved):
Thanks!

lcolitti:
Done. Though technically I anyone using this will have to change their code already, since until a couple of weeks ago list_views was in the stipc plugin and now it's in ipc-rules.

ammen99:
stipc is for tests only, I do not count it in the public API :)

PR #2029, opened 2023-11-26
Add an example script to move all windows to another workspace

lcolitti:


ammen99 (approved):
Thanks!

I'm wondering whether it would make sense to maybe open a separate repo with all of the ipc scripts and examples, where people can also share whatever they use?

PR #2031, opened 2023-11-26
ipc-rules: Add list-outputs function

soreau:
This can be used by scripts like trailfocus.py, where views need to be divided by outputs. The list can be cobbled together from list_views() and view['output'], but that is a hack, at best.


ammen99 (commented):
I think it might make sense to separate the code from the call for querying a single output to a function and then give the same information with the same structure in both list-outputs and output-info, otherwise lgtm

ammen99 (approved):
Thanks!

PR #2038, opened 2023-11-29
plugin-loader: Only load plugins once

### LINKED ISSUE #2036: Plugin loaded twice causes crash
**Describe the bug**
If you put resize twice into the plugins list, Wayfire crashes when you resize a window. I somehow got in this situation accidentally and it took me too long to figure out what happened.

**To Reproduce**
```
plugins = command autostart resize resize
```

Open a window and try to resize it. Wayfire crashes.

**Expected behavior**
Ignore duplicates, print warning

**Wayfire version**
git

soreau:
Duplicate plugins are now ignored.

Fixes #2036.

### DIFF
--- a/src/core/plugin-loader.cpp
+++ b/src/core/plugin-loader.cpp
@@ -1,15 +1,11 @@
 #include <sstream>
 #include <algorithm>
-#include <set>
 #include <memory>
 #include <filesystem>
 #include <dlfcn.h>
 
 #include "plugin-loader.hpp"
-#include "wayfire/output-layout.hpp"
-#include "wayfire/output.hpp"
 #include "../core/wm.hpp"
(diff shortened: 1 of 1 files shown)


ammen99:
We already have a check for this https://github.com/WayfireWM/wayfire/blob/9947528db40c94cc95e30379b3754bf4b6c106a5/src/core/plugin-loader.cpp#L203

The only question is why it doesn't work

ammen99:
Oh, I see. We ought to make sure the plugin isn't duplicated in the pending plugins as well.

soreau:
If I am understanding correctly, this PR would make it so the latter non-working check is not needed.

ammen99:
Thanks @soreau for the initial work on this :)

PR #2045, opened 2023-12-02
wm-actions: Make show desktop per-workspace

soreau:
This was per-workspace until 92a44dd3dfe43c7621a98e6a3e45521e1e0c8461.


ammen99:
> I am wanting to figure out how to make `bool showdesktop_active` a per-workspace variable.

The best way would be to store custom data on the workspace set, and in the custom data have a 2d bool vector.

PR #2049, opened 2023-12-06
Shortcuts inhibit core xml

marcusbritanicus:
I think it would be beneficial if shortcuts-inhibit was enabled by default. I've added the plugin to core.xml and the sample wayfire.ini


ammen99:
I guess this makes sense, since most other compositors enable it unconditionally ...

PR #2056, opened 2023-12-07
Call shutdown on SIGINT/SIGTERM

### LINKED ISSUE #729: Handle SIGINT/SIGTERM
Wayfire should handle these signals and shutdown gracefully.

soreau:
Fixes #729.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -124,6 +124,16 @@ static void signal_handler(int signal)
         error = "Fatal error(SIGABRT)";
         break;
 
+      case SIGINT:
+        LOGI("Got SIGINT, shutting down");
+        wf::get_core().shutdown();
+        return;
+
+      case SIGTERM:
+        LOGI("Got SIGTERM, shutting down");
+        wf::get_core().shutdown();
+        return;
+
       default:
         error = "Unknown";
     }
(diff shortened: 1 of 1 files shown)


src/core/opengl.cpp
```diff
@@ -705,7 +705,7 @@ void program_t::free_resources()
 {
     for (int i = 0; i < wf::TEXTURE_TYPE_ALL; i++)
     {
-        if (this->priv->id[i])
+        if (this->priv && this->priv->id[i])
```
ammen99, line 708: This sounds like a workaround for a race condition which happens somewhere else. free_resources() should never be called if this->priv is NULL - that would mean that we call the function after the program object has been freed.

plugins/protocols/wayfire-shell.cpp
```diff
@@ -448,6 +448,11 @@ class wayfire_shell_protocol_impl : public wf::plugin_interface_t
 
     void fini() override
     {
+        if (wf::get_core().get_current_state() == wf::compositor_state_t::SHUTDOWN)
```
ammen99, line 451: Why is this necessary? Is this something all protocol plugins need to do, for example gtk-shell as well?

ammen99 (changes requested):

ammen99:
I merged this manually in my shutdown fixes branch, thanks for the PR!

PR #2057, opened 2023-12-07
decor: Don't put minimize buttons on child views

### LINKED ISSUE #705: Disallow minimize for child views
For some reason, the foobar2000 about window cannot be minimized, and when it is minimized with the animate plugin running, it animates a minimize, but then immediately reappears at its original location. Without the animate plugin, it just can't be minimized.

https://f.losno.co/v/wayfire_minimize_foobar2000_about.mp4

soreau:
Fixes #705.

### DIFF
--- a/plugins/decor/deco-button.hpp
+++ b/plugins/decor/deco-button.hpp
@@ -19,9 +19,9 @@ class decoration_theme_t;
 
 enum button_type_t
 {
-    BUTTON_CLOSE,
-    BUTTON_TOGGLE_MAXIMIZE,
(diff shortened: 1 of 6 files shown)


soreau:
Tested with VLC.

ammen99 (approved):
LGTM. I wonder whether we shouldn't also add a check or at least log an error in case a view with a parent is minimized?

soreau:
> LGTM. I wonder whether we shouldn't also add a check or at least log an error in case a view with a parent is minimized?

At what time point or source location?

PR #2058, opened 2023-12-08
output: Implement adaptive sync option

### LINKED ISSUE #1009: Add support for variable refresh rate / adaptive sync
Since wlroots added support for this 6 months ago, now may be a good time to add an output option to actually enable or disable it.

See: wlr_output_enable_adaptive_sync

soreau:
Enable by setting [output:<NAME>] vrr = true, or using a client like wlr-randr.

Fixes #1009.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -600,7 +600,7 @@ struct output_layout_output_t
             } else
             {
                 LOGE("Failed to change adaptive sync on output: ", handle->name);
-                wlr_output_enable_adaptive_sync(handle, adaptive_sync_enabled);
+                wlr_output_rollback(handle);
             }
         }
     }
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -588,31 +588,19 @@ struct output_layout_output_t
 
         wlr_output_commit(handle);
 
-        if ((handle->adaptive_sync_status == WLR_OUTPUT_ADAPTIVE_SYNC_DISABLED) && current_state.vrr)
+        const bool adaptive_sync_enabled = (handle->adaptive_sync_status == WLR_OUTPUT_ADAPTIVE_SYNC_ENABLED);
+
+        if (adaptive_sync_enabled != current_state.vrr)
         {
             wlr_output_enable_adaptive_sync(handle, current_state.vrr);
             if (wlr_output_test(handle))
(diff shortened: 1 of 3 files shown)


src/core/output-layout.cpp
```diff
@@ -581,6 +586,29 @@ struct output_layout_output_t
...
+            {
+                wlr_output_enable_adaptive_sync(handle, current_state.vrr);
+                wlr_output_commit(handle);
+                if (handle->adaptive_sync_status == WLR_OUTPUT_ADAPTIVE_SYNC_DISABLED)
+                {
+                    LOGE("Failed to enable adaptive sync on output: ", handle->name);
+                    wlr_output_enable_adaptive_sync(handle, false);
+                } else
+                {
+                    LOGI("Enabled adaptive sync on output: ", handle->name);
```
ammen99, line 601: Sounds like a good candidate for LOGD, not LOGI. Optionally, maybe even add a new logging category to debugging.hpp (output-related)?

src/core/output-layout.cpp
```diff
@@ -581,6 +586,29 @@ struct output_layout_output_t
             wlr_output_set_custom_mode(handle, mode.width, mode.height, mode.refresh);
         }
 
+        if (!is_nested_compositor)
```
ammen99, line 589: Is this something wlroots doesn't support or why put this check?

ammen99 (commented):
As a whole LGTM, two small nitpicks.

src/core/output-layout.cpp
soreau: When I use `if  (1)` here, nested wayland backend does not start. it fails with:
```
[types/output/swapchain.c:109] Swapchain for output 'WL-1' failed test
```

src/core/output-layout.cpp
soreau: Sure, we can make it LOGD but I'm not sure it warrants a new debug category.

By the way, is there a way to enable all debug categories at once with a single option?

src/core/output-layout.cpp
soreau: I think I fixed it without the special case.

ammen99 (approved):
Just pushed a small commit to reduce duplicated logic, otherwise LGTM. Thanks!

PR #2060, opened 2023-12-08
Fix memory leak when activating cube plugin

### LINKED ISSUE #2059: Cube leaks memory every time it's activated
**Describe the bug**
When using cube, the wayfire process leaks several MB of shared memory every time the cube is activated.

**To Reproduce**
1. Enable cube
2. Run "watch -n0.1 ps u -p $(pidof wayfire)" and "watch -n0.1 free -m"
3. Activate cube
4. Observe memory usage going up.

**Expected behavior**
No memory leak.

**Wayfire version**
0.8.0 and git

lcolitti:
Currently it looks like the framebuffers are allocated on activate but never released. This causes a substantial memory leak. Fix this by adding a destructor to cube_render_instance_t that releases them.

Fixes #2059

### DIFF
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -81,6 +81,17 @@ class wayfire_cube : public wf::per_output_plugin_instance_t, public wf::pointer
                 }
             }
 
+            ~cube_render_instance_t()
+            {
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Nice catch!

PR #2061, opened 2023-12-09
ipc-scripts: fix wrong command name

lcolitti:
This fixes the list_views method and the move-to-workspace.py example script.


ammen99 (approved):
Thanks ;)

Issue #2043, opened 2023-12-01
The 'maximized is true' criteria is not working.


xiaohuirong:
**Describe the bug**
I set `ignore_views = app_id is "xfce4-terminal" | maximized is true` for decoration plugin, but the `maximized is true` criteria is not working and it will render the `app_id is "xfce4-terminal"` ineffective.

**To Reproduce**
1. enable ssd and decoration plugin.
2. set `ignore_views = app_id is "xfce4-terminal" | maximized is true` for decoration plugin.
3. The xfce4-terminal's ssd still exists, and it won't be hidden when the window is maximized.

**Expected behavior**
The SSD can be hidden when the window is maximized.

**Wayfire version**
wayfire 0.8.0.r52.g8f7787e1-1 built from latest 8f7787e1 and wayfire 0.7.5

ammen99:
The criteria here will be applied only once, when the window is mapped. The intention has never been to allow this to hide the decoration when the window is maximized and restore decorations when it is not ...

xiaohuirong:
Is it possible to support this feature? It seems that apart from automatically hiding decorations, there doesn't seem to be any other place where the `maximized is true` criteria can be used.

ammen99:
The maximized criteria was originally intended for window-rules, but I agree it is not particularly interesting. The decoration plugin could, of course, handle this, but I am not sure what the best way to expose the functionality is - (ab)using the existing option seems less than ideal.

Maybe we want configurable decoration size for different states, so that you can set the size to 0 for maximized windows?

PS. For blur it won't work either, they all have a one-off effect: only when the view is mapped. The maximized criteria was meant for window-rules primarily.

xiaohuirong:
Dynamic setting of window height also requires the decoration to change along with the window's state. Therefore, the key to solving this issue is to ensure that these one-off effects can refresh along with the window state changes. This design should be more reasonable; otherwise, the usability of criteria that change with window state is very limited.

timgott:
Also affects winshadows. "floating is true" does not work either. This used to work, so I consider this a bug. Probably the view_matcher is broken, but I did not investigate yet.

See https://github.com/timgott/wayfire-shadows/issues/19

timgott:
The criteria parser does not even accept "maximized is true" for me:

```
[src/core/matcher.cpp:27] Failed to parse condition maximized is true from option ignore_views
[src/core/matcher.cpp:28] Reason for the failure: Condition parser error. Unexpected symbol.
[...]

PR #2062, opened 2023-12-11
Listen to tiled signal in deco plugin and emit view_tiled signal on core

timgott:
Fixes https://github.com/timgott/wayfire-shadows/issues/19 and partially https://github.com/WayfireWM/wayfire/issues/2043

For the decoration plugin that should be very low overhead even if the maximized criterion is not used, so why not.

But please, can all view signals be consistently emitted on view, output and core at once.

### DIFF
--- a/plugins/decor/decoration.cpp
+++ b/plugins/decor/decoration.cpp
@@ -58,11 +58,19 @@ class wayfire_decoration : public wf::plugin_interface_t
         update_view_decoration(ev->view);
     };
 
+    // allows criteria containing maximized or floating check
+    wf::signal::connection_t<wf::view_tiled_signal> on_view_tiled =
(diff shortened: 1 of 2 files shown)


soreau:
> But please, can all view signals be consistently emitted on view, output and core at once.

:+1:

I've run into problems like this before, builds fine, plugin source hasn't changed but wayfire did, so it doesn't work.

timgott:
So I had only tested with "floating is false", this fixes #2043 only partially, see there.

ammen99:
> But please, can all view signals be consistently emitted on view, output and core at once.

I see no reason not to do it, it may even make sense to have a helper function so that we always emit on all three at once. PRs welcome :)

ammen99 (approved):
Thanks!

PR #2072, opened 2023-12-19
add keybindings: prev_output, prev_output_with_win

### LINKED ISSUE #2071: Allow keybindings to move to the previous output
The plugin oswitch only supports `next_output` and `next_output_with_win`. This is convenient enough with a 2 monitors setup. 
However if you have 3 or more monitors, it is not that convenient as you have to iterate through all monitors to get to the target one, if you only use keyboard.
Thus there's a demand for `prev_output` and `prev_output_with_win` in this case.

A7R7:
provide an api `wf::output_t *get_prev_output(wf::output_t *output)` in core/output-layout.hpp
and use this api to implement `prev_output` and `prev_output_with_win`
fixes #2071

### DIFF
--- a/metadata/oswitch.xml
+++ b/metadata/oswitch.xml
@@ -14,5 +14,15 @@
 			<_long>Moves focus to the next output with the focused window with the specified activator.</_long>
 			<default>&lt;super&gt; &lt;shift&gt; KEY_O</default>
 		</option>
+		<option name="prev_output" type="activator">
+			<_short>Previous output</_short>
+			<_long>Moves focus to the previous output with the specified activator.</_long>
+			<default></default>
+		</option>
+		<option name="prev_output_with_win" type="activator">
+			<_short>Previous output with window</_short>
+			<_long>Moves focus to the previous output with the focused window with the specified activator.</_long>
+			<default></default>
+		</option>
 	</plugin>
 </wayfire>
--- a/plugins/single_plugins/oswitch.cpp
+++ b/plugins/single_plugins/oswitch.cpp
@@ -9,44 +9,74 @@
 
 class wayfire_oswitch : public wf::plugin_interface_t
 {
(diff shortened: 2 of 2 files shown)


A7R7:
Haven't noticed there's already #1570. I may take up his work then

ammen99 (changes requested):
As mentioned in the other PR (#1570) there are two changes needed:

- The get_prev_output() shouldn't be added to core at all, oswitch can do that on its own.
- It would be nice to refactor the code so that the functions for next and previous output reuse the same code. Currently, they do the exact same thing but with the only difference of previous/next output. I imagine it should be easy to refactor them to use a common function `switch_to_output(target_output)` and `switch_to_output_with_window(target_output)`

A7R7:
I did some refactor to the original functions, most importantly add `get_output_relative` to get the target monitor instead of changing the core api. The 4 activator_callbacks now only use `switch_to_output(target_output)` and `switch_to_output_with_window(target_output)` to finish their jobs.

I haven't implement `swap_output` and `swap_output_prev` in #1570, because I'm not sure if it is better to swap only the top view or swap the whole active workspace on the 2 outputs, or if it is good to add both functionalities (which I think will be a little mess).

ammen99:
> I haven't implement `swap_output` and `swap_output_prev` in #1570, because I'm not sure if it is better to swap only the top view or swap the whole active workspace on the 2 outputs, or if it is good to add both functionalities (which I think will be a little mess).

That is fine, since you are doing the work, you can only add whatever you find useful :)

ammen99 (approved):
LGTM now, thanks!

ammen99:
Please fix code style as well (there is a diff in the CI logs)

A7R7:
I've fixed the code style with your fork of  uncrustify, hopefully it can pass!

PR #2092, opened 2024-01-06
idle: Fix stuck black screen after timeout with cube disabled

### LINKED ISSUE #2085: Occasional black screen when switching to Wayfire's TTY
**Describe the bug**
Black screen with a cursor. The cursor changes occasionally when moving, which suggests the clients are still alive and affect the cursor when I hover them, but I cannot see any surfaces anymore. I have no idea how to resurrect Wayfire from that state, so I just kill it.

**To Reproduce**
1. Launch X.Org on TTY7.
2. Launch Wayfire on TTY3.
3. Launch a few clients in Wayfire.
4. Switch to TTY7 for a few hours.
5. Switch back to TTY3.

**Expected behavior**
Wayfire survives running in background for a long time well. I suspect it's related to an upgrade, as I used an old build before and forgot it on other TTY like that for many days, and it survived well. Not sure which version exactly that was, but it used wlroots 0.11, AFAIR. So a regression happened somewhere in between.

**Screenshots or stacktrace**

**Wayfire version**
a1d2607aa5d63c895b83994957d3e250ddc9cf2a
wlroots: fe53ec693789afb44c899cad8c2df70c8f9f9023

soreau:
Fixes #2085.

### DIFF
--- a/plugins/single_plugins/idle.cpp
+++ b/plugins/single_plugins/idle.cpp
@@ -242,6 +242,12 @@ class wayfire_idle_plugin : public wf::per_output_plugin_instance_t
             return;
         }
 
+        if (!timeout_screensaver.is_connected() && output_inhibited)
+        {
+            uninhibit_output();
+            return;
+        }
+
         if (!timeout_screensaver.is_connected() && (state == CUBE_SCREENSAVER_RUNNING))
         {
             stop_screensaver();


plugins/single_plugins/idle.cpp
```diff
@@ -242,7 +242,8 @@ class wayfire_idle_plugin : public wf::per_output_plugin_instance_t
             return;
         }
 
-        if (!timeout_screensaver.is_connected() && (state == CUBE_SCREENSAVER_RUNNING))
+        if (!timeout_screensaver.is_connected() &&
```
ammen99, line 245: Should we maybe only call uninhibit_output if the screensaver is not running?

soreau: I pushed a change, can you have another look?

benkant:
Confirming #2092 fixes #2085 thank you 👍

ammen99 (approved):
Thanks!

PR #2099, opened 2024-01-17
Add subsurfaces to the back that are meant to be below the parent surface

### LINKED ISSUE #1206: Handle subsurfaces below and above main surface
We already have support for this, but we are not reading the Z-order provided by wlroots (see #1205). Basically, we need to just add the subsurfaces with the appropriate mode - above or below.

### LINKED ISSUE #1615: Black screen on waydroid
**Describe the bug**

When running `waydroid show-full-ui` the resulting window is black, when dragging the screen to the top of window to auto resize, it briefly shows the content of the screen, touch and keyboard is being passed through to the android container perfectly fine

**To Reproduce**
1. Install waydroid
2. run waydroid with `waydroid show-full-ui`

**Expected behavior**
Waydroid should open a fullscreen android container

**Screenshots or stacktrace**
N/A also running WAYLAND_DEBUG with waydroid produces no output at this time

**Wayfire version**
wayfire-git 0.7.4.r341.g866cdf28-1

**Additional Info**
Originally though to be a wlroots issue, turns out it isn't has been successfully tested on both sway and cage installed side by side with wayfire. 

the screen manages to flicker when snapping and unsnapping from edges

soreau:
Most applications add their subsurfaces above the parent, so we were blindly adding them to the front. This patch checks if the subsurface should be below, and adds it to the back in that case, or else adds it to the front.

Fixes #1206.
Fixes #1615.

### DIFF
--- a/src/view/surface.cpp
+++ b/src/view/surface.cpp
@@ -45,6 +45,15 @@ wf::wlr_surface_controller_t::wlr_surface_controller_t(wlr_surface *surface,
         // Allocate memory, it will be auto-freed when the wlr objects are destroyed
         auto sub_controller = new wlr_subsurface_controller_t(sub);
         create_controller(sub->surface, sub_controller->get_subsurface_root());
+        wlr_subsurface *s;
+        wl_list_for_each(s, &surface->current.subsurfaces_below, current.link)
+        {
+            if (sub == s)
+            {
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Looks good, thanks :)

PR #2101, opened 2024-01-18
Modify blur defaults to avoid visual artifacts

soreau:
The degrade setting was too high, causing problems especially when moving a view around the screen.


ammen99 (approved):
LGTM

PR #2110, opened 2024-01-22
Fix `wayfire` AUR link

trigg:
The URL link for `wayfire` Arch/AUR package collides with the URL link for `https://wayfire.org/` earlier in the document


ammen99 (approved):
Thanks :)

PR #2112, opened 2024-01-22
output: Add bit depth property

### LINKED ISSUE #2103: 10-bit color
wlroots should have it, sway has implemented it

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3333
https://github.com/mstoeckl/sway/commit/aecc6e07c3dcea2f3ca08ad8621193ed447b2ed7

soreau:
Currently supports 8 and 10 bpc.

Fixes #2103.

### DIFF
--- a/metadata/output.xml
+++ b/metadata/output.xml
@@ -17,5 +17,10 @@
     <option name="vrr" type="bool">
       <default>false</default>
     </option>
+    <option name="depth" type="int">
+      <default>8</default>
+      <min>8</min>
+      <max>10</max>
+    </option>
   </object>
 </wayfire>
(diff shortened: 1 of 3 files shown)


ammen99 (approved):
LGTM. I pushed two small changes, mostly was wondering why we need the invalid format at the end of the list? When we have a std::vector, we can simply iterate over the  list and don't need the last invalid format, right @soreau? (asking to make sure I didn't break something I don't understand)

soreau:
> LGTM. I pushed two small changes, mostly was wondering why we need the invalid format at the end of the list? When we have a std::vector, we can simply iterate over the list and don't need the last invalid format, right @soreau? (asking to make sure I didn't break something I don't understand)

Yes, your changes LGTM too. 👍

PR #2119, opened 2024-02-01
layer-shell: Refocus the keyboard focus view after arranging layers

### LINKED ISSUE #2047: Click on layer shell windows renderes active application as inactive
**Describe the bug**
In older Wayfire versions (0.7.4) you could click anywhere on layer-shell-windows like panels and the currently open application would still be displayed as currently focused window. That changed (probably with the reimplementation of the layer-shell-protocol as plugin) with the newer versions of Wayfire. Now the „visible“-focus gets lost when clicking on a layer-shell-window.     
I write „visible“-focus because only the style of the windows is affected (I guess there are better terms to describe it but I do not know them.) things like keyboard focus still stay with the application at least as long as the keyboard mode is set that way.

I think that this might be a bug because the old behavior seemed much more logical but if this was on purpose I would also propose a change back.

I noticed that with my own panel implementation but it happens with wf-panel as well. Also the problem does not occur in Sway.

**To Reproduce**
1. Have a theme/application that renders differently when not in focus (for example Firefox)
2. Click on a toplevel that uses layer shell (e.g. wf-panel)
3. The window will loose its focus/is rendered as inactive

**Expected behavior**
The open application is still rendered as active.

**Screenshots or stacktrace**

https://github.com/WayfireWM/wayfire/assets/10778257/d80b81e0-7642-41a5-adf8-f9d4657468ab

**Wayfire version**
git

### LINKED ISSUE #2118: wf-panel does not lose keyboard focus when closing menu
**Describe the bug**
I found it nice to have the new feature to open / close the menu with a keyboard shortcut (from #2097), but I found that when closing the menu with the keyboard, the panel retains the keyboard focus. To me, the natural behavior would be that the last focused app gets back the focus. I feel that this is actually the intended behavior, since wf-panel sends a `zwlr_layer_surface_v1#40.set_keyboard_interactivity(0)` request when the menu is closed. Also [here](https://github.com/WayfireWM/wayfire/blob/9c9630d1a03293a6bca4b7496124eec6ccddd3e4/src/view/layer-shell/layer-shell.cpp#L546), it is indicated that a refocus might happen as a consequence of changing keyboard-interactivity, but the `arrange_layers()` function does not seem to handle focus changes.

**To Reproduce**
1. Run wf-panel and open the menu (with at least one open view).
2. Close the menu with keyboard interaction (pressing ESC, or the shortcut set for `wayfire-shell/toggle_menu`).
3. The panel keeps keyboard focus.

**Expected behavior**
The last focused view gets back the keyboard focus.

**Wayfire version**
latest git (9c9630d1a03293a6bca4b7496124eec6ccddd3e4)

soreau:
This effectively makes it so the view that had focus before interacting with wf-shell apps, is focused after interactivity is complete.

Fixes #2118.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,7 +23,7 @@ jobs:
     steps:
     - run: sed -i 's/SigLevel    = Required DatabaseOptional/SigLevel    = Optional TrustAll/' /etc/pacman.conf
     - run: pacman --noconfirm --noprogressbar -Syyu
-    - run: pacman --noconfirm --noprogressbar -Sy git clang lld libc++ pkgconf meson ninja wayland wayland-protocols libinput libxkbcommon pixman glm libdrm libglvnd cairo pango systemd scdoc base-devel seatd hwdata libdisplay-info
+    - run: pacman --noconfirm --noprogressbar -Sy git clang lld libc++ pkgconf cmake meson ninja wayland wayland-protocols libinput libxkbcommon pixman glm libdrm libglvnd cairo pango systemd scdoc base-devel seatd hwdata libdisplay-info
 
       # Build Wayfire
     - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


benkant:
:rocket:

ammen99 (approved):
Thanks, LGTM!

PR #2123, opened 2024-02-03
Don't resize clients when output is resized

marcusbritanicus:
soreau's fix to prevent clients from resizing proportional to output size

### DIFF
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -233,7 +233,7 @@ struct workspace_set_t::impl
 
             if (view->get_output() && view->toplevel()->current().fullscreen)
             {
-                view->set_geometry( new_geometry );
+                view->set_geometry(new_geometry);
             } else if (view->toplevel()->current().tiled_edges)
             {
                 // Do nothing. This is taken care of, by the grid plugin.
--- a/src/output/workspace-impl.cpp
+++ b/src/output/workspace-impl.cpp
@@ -230,13 +230,21 @@ struct workspace_set_t::impl
             auto wm  = view->get_geometry();
             float px = 1. * wm.x / old_w;
             float py = 1. * wm.y / old_h;
-            float pw = 1. * wm.width / old_w;
-            float ph = 1. * wm.height / old_h;
 
-            view->set_geometry({
-                    int(px * new_geometry.width), int(py * new_geometry.height),
-                    int(pw * new_geometry.width), int(ph * new_geometry.height)
-                });
+            if (view->get_output() && view->toplevel()->current().fullscreen)
+            {
+                view->set_geometry( new_geometry );
+            } else if (view->toplevel()->current().tiled_edges)
+            {
+                // Do nothing. This is taken care of, by the grid plugin.
+                // If the user does not have grid enabled, we ignore it anyways.
+            } else
+            {
+                view->set_geometry({
+                        int(px * new_geometry.width), int(py * new_geometry.height),
+                        wm.width, wm.height
+                    });
+            }
         }
 
         workspace_geometry = new_geometry;


src/output/workspace-impl.cpp
```diff
@@ -230,13 +230,23 @@ struct workspace_set_t::impl
...
-            float ph = 1. * wm.height / old_h;
 
-            view->set_geometry({
-                    int(px * new_geometry.width), int(py * new_geometry.height),
-                    int(pw * new_geometry.width), int(ph * new_geometry.height)
-                });
+            auto edges = view->toplevel()->current().tiled_edges;
+            if (view->get_output() && view->toplevel()->current().fullscreen)
+            {
+                wf::get_core().default_wm->fullscreen_request(view, view->get_output(), false);
```
ammen99, line 237: This feels a bit hacky, I would directly adjust the view geometry for fullscreen views.

src/output/workspace-impl.cpp
```diff
@@ -230,13 +230,23 @@ struct workspace_set_t::impl
...
+                wf::get_core().default_wm->fullscreen_request(view, view->get_output(), true);
+            } else if (edges)
+            {
+                wf::get_core().default_wm->tile_request(view, 0);
```
ammen99, line 241: Grid already handles this without untiling and re-tiling

https://github.com/WayfireWM/wayfire/blob/7ce8332fe63583933d1a38494178be7d5bafad94/plugins/grid/grid.cpp#L145

In general I feel like un-tiling and re-tiling is a very hacky thing to do, esp. since for example tiled state is set by the simple-tile plugin. You ought to not just untile views like that ...

ammen99: So I think that in general we can just not do anything for this case. Alternatively, we'd need a new weird signal, `adjust_view_on_workspace_set_resize_signal` which then simple-tile/grid can register to, say 'I have adjusted the view' (with a carried_out flag), and if nobody takes care of those views, then core can do that - set fullscreen/tiled geometry or only adjust the position.

ammen99 (changes requested):

src/output/workspace-impl.cpp
marcusbritanicus: You mean to say, resize the view directly?
```
view->set_geometry(...)
```

src/output/workspace-impl.cpp
marcusbritanicus: I guess, we can take the code from grid. Let me see how I can incorporate it.

ammen99: No, the code should remain in grid! Core doesn't know about all the available slots.

src/output/workspace-impl.cpp
ammen99: Yes

src/output/workspace-impl.cpp
marcusbritanicus: I see. So, how do we "retile" tiled (maximized) views, then?

marcusbritanicus: Nevermind.. I see that they remain tiled, automagically!! :)

ammen99: Not magically, the grid plugin does that :)

The only problem is if the grid plugin is not enabled, but I think this is very very rare in actual Wayfire setups, so I don't consider it a problem that we don't re-maximize if grid is not enabled :)

marcusbritanicus: Okay... We now do nothing when the view is maximized!

(I=When I said automagically, I meant that we don't need to add any code here..! At that time I had not yet realised that grid plugin listens to workarea_changed_signal. :smiley: But, thank for the explanation.)

ammen99 (approved):
LGTM, please fix code style and I will merge ;)

marcusbritanicus:
> LGTM, please fix code style and I will merge ;)

Yeah, I noticed. Fixed it.. :)

PR #2131, opened 2024-02-10
ipc-rules: Add get-view-pid to get the active view pid

soreau:


killown:
could we have the pid info in "window-rules/get-focused-view" instead ?

so for example we could do things like [i["pid"] for i in list_view()]  to get pid from all views

ammen99:
I think we don't need an additional method  we should report this as a property of the view in the other methods like get-focused-view, list-views etc

plugins/single_plugins/ipc-rules.cpp
```diff
@@ -494,6 +495,31 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
...
+    pid_t get_view_pid(wayfire_view view)
+    {
+        pid_t pid = -1;
+        if (!view)
+        {
+            return pid;
+        }
+
+        wlr_surface *wlr_surface = view->get_wlr_surface();
+        int is_xwayland_surface  = 0;
```
ammen99, line 508: sounds like the type should be bool :)

but also why is this outside the if-def and do we need a variable for this at all?

ammen99 (approved):
LGTM, thanks!

I'll fix the tests and merge once I verify that everything works correctly :)

Issue #1912, opened 2023-09-27
A better support for out-of-tree plugins


ammen99:
As the number of external plugins slowly rises, it has become clear we need something like https://github.com/timgott/wfplug, but I think for a better UX it should be built-in into Wayfire itself, or at least an official tool. A vague plan that I am making up on the spot, feedback and suggestions are very welcome:

- Wayfire should start looking at a user-writable location by default (say `~/.local/lib/wayfire` or whatever the standards say is a good place)
- This folder should probably vary based on the installation prefix: this way we can have multiple installed Wayfire versions
- We'd need a bit more sophisticated script/tool which should support things like:
  - Install/Update/Remove a new plugin from an URL for a specific wayfire installation
  - Rebuild all plugins for a specific wayfire installation
  - It would be also cool to enable/disable plugins from this tool, but it gets a bit hard: we do not know the exact config backend used by Wayfire ... So we are not quite certain which config file(s) to edit (or gsettings, or ...).

In any case, this will happen for Wayfire 0.9.0 at earliest.

cc @timgott

timgott:
> Wayfire should start looking at a user-writable location by default

It would be easier to manage if in that user plugins folder each plugin has its own installation folder, instead of one binary and one metadata folder for all plugins.

> Rebuild all plugins for a specific wayfire installation

This means you need some integration with a build system (e.g. wfplug only works with meson). Maybe you could also make it easier for single cpp file plugins.

> Install/Update/Remove

I had these in wfplug but removed it because it was just a wrapper around git and makes it more complicated if you e.g. need an older commit for compatibility.

Btw initially wfplug was meant to work similar to vim-plug for vim. But now I use wfplug more like a python virtualenv to test plugins (i.e. winshadows). That's also why it can generate a test config with only a specific plugin enabled. It has to be separate from my main install, otherwise it is too easy to crash the main session.

For the actual system I switched to NixOS by now and I have a simple Nix expression that simply compiles all subdirectories of a plugins folder and adds them to wayfire's path variables. This way my plugins also rebuild automatically for a new wayfire version. I could also have multiple wayfire versions but I don't need that (you might be overestimating the number of people that have multiple installations of wayfire at the same time).

soreau:
> > Wayfire should start looking at a user-writable location by default
> 
> It would be easier to manage if in that user plugins folder each plugin has its own installation folder, instead of one binary and one metadata folder for all plugins.

[...]

PR #2134, opened 2024-02-11
Support scaling only the windows of the focused app

lcolitti:
This adds support to the scale plugin to show only the windows of a given app (or of a given app on all workspaces). This is similar to the compiz feature that scales only the windows of a window group.

### DIFF
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -16,6 +16,16 @@
 				<_long>Toggles scale showing windows from all workspaces.</_long>
 				<default></default>
 			</option>
+			<option name="toggle_app" type="activator">
+				<_short>Toggle for app</_short>
+				<_long>Toggles scale for the focused app's windows.</_long>
+				<default></default>
+			</option>
+			<option name="toggle_app_all" type="activator">
+				<_short>Toggle for app on all workspaces</_short>
+				<_long>Toggles scale for the focused app's windows on all workspaces..</_long>
+				<default></default>
+			</option>
 			<option name="duration" type="int">
 				<_short>Animation Transition Time</_short>
 				<_long>Time it takes for views to transition. Units are in milliseconds.</_long>
--- a/plugins/scale/scale.cpp
+++ b/plugins/scale/scale.cpp
@@ -71,6 +71,12 @@ struct view_scale_data
     bool was_minimized = false; /* flag to indicate if this view was originally minimized */
 };
 
+struct scale_view_list
+{
+    std::vector<wayfire_toplevel_view> views;
+    std::vector<wayfire_toplevel_view> filtered_views;
+};
+
 /**
  * Scale has the following hard coded bindings are as follows:
  * KEY_ENTER:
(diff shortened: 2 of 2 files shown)


ammen99:
Duplicate of https://github.com/WayfireWM/wayfire/pull/1911

Unfortunately, my stance on this hasn't changed.

Thanks for your PR nonetheless :)

ammen99:
Here is the functionality as a standalone plugin, I think you should easily be able to get this working with an ipc script bound to a hotkey.https://github.com/dkondor/wayfire-scale-ipc

lcolitti:
I get that this is possible to do with an out-of-tree plugin (and thanks for the link), but why require users to do that? The fact that several people have asked for this suggests it's a useful feature. [Mac OS has it](https://support.apple.com/lt-lt/guide/mac-help/mh35798/mac), and compiz does as well (or did before [apparently it broke](https://askubuntu.com/questions/39735/how-do-i-use-initiate-window-picker-for-window-group-in-compiz)). Why not include it? It doesn't seem to be a lot of work to maintain.

FWIW what I am trying to do here is not write code for a specific dock, it's just to improve the user experience of using wayfire and wf-panel. My use case is: I always have dozens of windows open on lots of workspaces, because I use multi-window setups in all apps. I want to pick a window visually from the app I'm already using. Say I'm working on scale.cpp and want to switch to the .hpp file. I know it's another text editor window, but I don't know where it is because its obscured by other windows from other apps (e.g., browsers open at C++ documentation). If I use scale without filters I probably won't be able to pick the window among the dozens of windows I have open - I can see which windows are text editors, but I can't see the one I want because the text in them is too small to read. With this keybinding I can find the window I want.

Sure, it may not be something that most users need. But again - why not support it? Is there a downside?

ammen99:
> Sure, it may not be something that most users need. But again - why not support it? Is there a downside?

The downside is that this repo gets bigger and bigger. I understand the motivation for this particular feature, but this is about the general principle - there are tons of small features which would make sense, but if you put them together, Wayfire's main repo will become too large. This means that basically all features which can be easily delegated to a third-party plugin and I don't deem crucial for the project (or are necessary as a demo of a functionality) are left as an exercise for the reader :)

I myself am using a few external plugins: https://github.com/ammen99/wayfire-plugins

Note that my vision for the project from the very start has been users using multiple plugins from different places. So I don't see this as a downside. I agree that we ought to make installing plugins easier - see #1912.

ammen99:
Also note that, if contributors/Wayfire plugin authors feel like this is a suboptimal way of doing things, nothing prevents people from creating one mega-reposjtory with all kinds of plugins written by various plugins. Think of wayfire-plugins-extra for example :)

soreau:
> This adds support to the scale plugin to show only the windows of a given app (or of a given app on all workspaces). This is similar to the compiz feature that scales only the windows of a window group.

It seems like you can get the same job done with scale title filter plugin, in addition to scale. This way, you can type and match.

PR #2148, opened 2024-02-18
output-layout: Don't initialize output position with values unconditi…

### LINKED ISSUE #2147: Enabling/disabling output via DPMS will duplicate mouse cursors when using multi monitors.
The title describes the issue that affects the output layout and results in duplicated mouse cursors. This means you will get two identical mouse cursors, one for each screen instead of just one cursor for all screens, also duplicating the cursor movement.

wlr-randr --output DP-1 --off ; wlr-randr --output DP-1 --on

To quick fix the output layout issue, you need to power off/on the output using the power button.

soreau:
…onally

When initializing output::position, passing x and y arguments means that the position will not be marked automatic, even if it's set to auto or the default value. This spells trouble later, when enabling an output that was previously disabled that is not the default 0,0 output. This commit checks the position is set to automatic before blindly supplying x and y values to the position.

Fixes #2147.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -1047,7 +1047,8 @@ class output_layout_t::impl
                 state.mode.refresh = head->state.custom_mode.refresh;
             }
 
-            state.position  = {head->state.x, head->state.y};
+            state.position = {head->state.x, head->state.y};
+
             state.scale     = head->state.scale;
             state.transform = head->state.transform;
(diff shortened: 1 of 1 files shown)


ammen99:
As discussed on IRC, this is not really fixable in Wayfire (at least in this way), because the client itself sends a request to make the outputs overlapping. An easy 'fix' is to specify the position when calling `wlr-randr --on`

Thanks for looking into the issue though :)

PR #2156, opened 2024-02-20
ipc: support workspace sets

CharlieQLe:
This is a draft PR for an implementation of querying and configuring workspace sets via IPC. This would be a way to interface with workspaces via bars such as waybar or window rule scripts specific to workspaces. Leaving it as a draft for now since I am unsure if this should live in `wsets` or `ipc-rules`. For now, it lives in `ipc-rules`. This is a very quick, very untested implementation as I do not have time at the moment to build and test the plugin.

This would likely require signals for seeing if a workspace set is created or destroyed. I'm not too sure if they already exist, but this would require more signals to be defined I imagine, if they aren't defined already.

Views should also know which workspace set and viewport they exist on, but I don't know if that would complicate anything in the codebase, or if that is considered out of scope for this PR. On paper, listing views and specifying the workspace set id and/or viewport position _should_ list all the views that exist on that workspace set (and viewport if specified). This would make it less simple for scripts to know which workspace set they exist on however- an example would be seeing if a view's title change is on the second workspace. One would have to list all the views that exist on the second workspace in order to check if the view is indeed on it whereas ideally, one could simply check properties such as `workspace-set-id` and `workspace-point`. Again, that might be out of scope for this specific PR as it doesn't necessarily pertain to interfacing with workspace sets.

### DIFF
--- a/plugins/single_plugins/ipc-rules.cpp
+++ b/plugins/single_plugins/ipc-rules.cpp
@@ -135,8 +135,10 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         method_repository->register_method("window-rules/events/watch", on_client_watch);
         method_repository->register_method("window-rules/list-views", list_views);
         method_repository->register_method("window-rules/list-outputs", list_outputs);
+        method_repository->register_method("window-rules/list-workspace-sets", list_workspace_sets);
         method_repository->register_method("window-rules/view-info", get_view_info);
         method_repository->register_method("window-rules/output-info", get_output_info);
+        method_repository->register_method("window-rules/workspace-set-info", get_workspace_set_info);
         method_repository->register_method("window-rules/configure-view", configure_view);
         method_repository->register_method("window-rules/focus-view", focus_view);
         method_repository->register_method("window-rules/get-focused-view", get_focused_view);
@@ -158,8 +160,10 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         method_repository->unregister_method("window-rules/events/watch");
         method_repository->unregister_method("window-rules/list-views");
         method_repository->unregister_method("window-rules/list-outputs");
+        method_repository->unregister_method("window-rules/list-workspace-sets");
         method_repository->unregister_method("window-rules/view-info");
         method_repository->unregister_method("window-rules/output-info");
+        method_repository->unregister_method("window-rules/workspace-set-info");
         method_repository->unregister_method("window-rules/configure-view");
         method_repository->unregister_method("window-rules/focus-view");
         method_repository->unregister_method("window-rules/get-focused-view");
@@ -171,6 +175,8 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         output->connect(&_tiled);
         output->connect(&_minimized);
         output->connect(&_fullscreened);
+        output->connect(&on_workspace_set_changed);
+        output->connect(&on_workspace_changed);
     }
 
     void handle_output_removed(wf::output_t *output) override
@@ -192,14 +198,44 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         return response;
     };
 
-    wf::ipc::method_callback list_views = [=] (nlohmann::json)
+    wf::ipc::method_callback list_views = [=] (nlohmann::json data)
     {
         auto response = nlohmann::json::array();
+        WFJSON_OPTIONAL_FIELD(data, "workspace-set-id", number_integer)
+        WFJSON_OPTIONAL_FIELD(data, "workspace-set-point", object)
 
-        for (auto& view : wf::get_core().get_all_views())
+        if (data.contains("workspace-set-id"))
         {
-            nlohmann::json v = view_to_json(view);
-            response.push_back(v);
+            auto ws = wf::ipc::find_workspace_set_by_id(data["workspace-set-id"]);
+            if (!ws)
+            {
+                return wf::ipc::json_error("workspace set not found");
+            }
+            if (data.contains("workspace-set-point"))
+            {
+                auto point = wf::ipc::point_from_json(data["workspace-set-point"]);
+                for (auto& view : ws->get_views(0, point))
+                {
+                    nlohmann::json v = view_to_json(view);
+                    response.push_back(v);
+                }
+            }
+            else
+            {
+                for (auto& view : ws->get_views(0))
+                {
+                    nlohmann::json v = view_to_json(view);
+                    response.push_back(v);
+                }
+            }
+        }
+        else
+        {
+            for (auto& view : wf::get_core().get_all_views())
+            {
+                nlohmann::json v = view_to_json(view);
+                response.push_back(v);
+            }
         }
 
         return response;
@@ -333,6 +369,44 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         return wf::ipc::json_ok();
     };
 
+    nlohmann::json workspace_set_to_json(wf::workspace_set_t* ws)
+    {
+        nlohmann::json response;
+        response["id"]   = ws->get_id();
+        response["name"] = ws->to_string();
+        
+        auto output = ws->get_attached_output();
+        response["output-id"] = output ? output->get_id() : -1;
+        response["output-name"] = output ? output->to_string() : "";
+
+        response["current-workspace"] = wf::ipc::point_to_json(ws->get_current_workspace());
+        response["grid-size"] = wf::ipc::dimensions_to_json(ws->get_workspace_grid_size());
+        return response;
+    };
+
+    wf::ipc::method_callback list_workspace_sets = [=] (nlohmann::json)
+    {
+        auto response = nlohmann::json::array();
+        for (auto& workspace_set : wf::workspace_set_t::get_all())
+        {
+            response.push_back(workspace_set_to_json(workspace_set.get()));
+        }
+        return response;
+    };
+
+    wf::ipc::method_callback get_workspace_set_info = [=] (nlohmann::json data)
+    {
+        WFJSON_EXPECT_FIELD(data, "id", number_integer);
+        auto ws = wf::ipc::find_workspace_set_by_id(data["id"]);
+        if (!ws)
+        {
+            return wf::ipc::json_error("workspace set not found");
+        }
+
+        auto response = workspace_set_to_json(ws);
+        return response;
+    };
+
   private:
     wf::shared_data::ref_ptr_t<wf::ipc::method_repository_t> method_repository;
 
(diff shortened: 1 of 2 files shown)


ammen99:
Hi and thanks for starting the work on this. A few things:

- I don't think we ought to filter views in the plugin. Instead, we should return the full list of views, on all workspaces/workspace-sets. We should indicate the wset id for each view (if the view is on a wset id) - this should be as simple as adding a new field to the `view_to_json` function. Then the ipc client can do the filtering on its own.

- A view is not really 'on' a particular viewport/workspace/however you call it, Views can be visible on multiple of those. Again, I think it is better that we don't have a bigger API than we need. If ipc scripts need to query something in particular, they can easily go through all views and check whether they fit/overlap a particular workspace rectangle.

- About the signal for output-workspace-changed: this signal should be `wset-workspace-changed`, because workspaces change within a workspace set, not within an output.

- About workspace-set-info and output-info commands: with your patch they have a different syntax for the workspace information, output-info has output[workspace][x/y/width/height], I'd prefer workspace-set-info to report in the same format.

CharlieQLe:
@ammen99 Updated `ipc-rules` to reflect your comments on this.

> About workspace-set-info and output-info commands: with your patch they have a different syntax for the workspace information, output-info has output[workspace][x/y/width/height], I'd prefer workspace-set-info to report in the same format.

Does the output data need to contain any workspace information except for the workspace set's id? It could easily just provide the id, and scripts can query for the workspace set info itself afterwards with that. Right now, the output data still supplies the workspace data to clients.

ammen99:
> @ammen99 Updated `ipc-rules` to reflect your comments on this.
> 
> > About workspace-set-info and output-info commands: with your patch they have a different syntax for the workspace information, output-info has output[workspace][x/y/width/height], I'd prefer workspace-set-info to report in the same format.
> 
> Does the output data need to contain any workspace information except for the workspace set's id? It could easily just provide the id, and scripts can query for the workspace set info itself afterwards with that. Right now, the output data still supplies the workspace data to clients.

Yes, that's true but I'd like to avoid breaking changes - and I guess that this is information which would be commonly requested so I think overall it is better to spare the additional IPC call needed otherwise.

ammen99:
Took a brief look over the PR, looks much better now :) I'll do a full review on the weekend & merge if everything is fine.

CharlieQLe:
> Took a brief look over the PR, looks much better now :) I'll do a full review on the weekend & merge if everything is fine.

Sounds good. I just added `configure_output` and `configure_wset` as well, allowing scripts to manipulate workspace sets.

ammen99:
> Sounds good. I just added `configure_output` and `configure_wset` as well, allowing scripts to manipulate workspace sets.

I think this features should belong to the wsets plugin though, since it can also create/destroy wsets and already has all the logic for moving wsets around (and also I think it keeps an internal mapping of wsets). It can also show a popup when the wset changes (which I guess should happen also for ipc-triggered switches).

CharlieQLe:
> > Sounds good. I just added `configure_output` and `configure_wset` as well, allowing scripts to manipulate workspace sets.
> 
> I think this features should belong to the wsets plugin though, since it can also create/destroy wsets and already has all the logic for moving wsets around (and also I think it keeps an internal mapping of wsets). It can also show a popup when the wset changes (which I guess should happen also for ipc-triggered switches).

Good point! I'll revert the commit for now and look into adding this feature before the merge, if it is in scope for this PR.

ammen99:
> > 
> 
> Good point! I'll revert the commit for now and look into adding this feature before the merge, if it is in scope for this PR.

I don't personally mind merging either two separate PRs or one bigger PR, feel free to organize it as you wish.

plugins/single_plugins/ipc-rules.cpp
```diff
@@ -333,6 +340,46 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
...
     };
 
+    nlohmann::json wset_to_json(wf::workspace_set_t *wset)
+    {
+        nlohmann::json response;
+        response["id"]   = wset->get_id();
+        response["name"] = wset->to_string();
+
+        auto output = wset->get_attached_output();
+        response["output-id"]   = output ? output->get_id() : -1;
```
ammen99, line 350: `get_id()` returns a uint, so we need to cast it explicitly to `int` to avoid overflow for the -1. There are also other places where the exact same bug happens, I will fix them as well at some point, but we shouldn't introduce even more bugs ;)

plugins/single_plugins/ipc-rules.cpp
```diff
@@ -509,7 +577,9 @@ class ipc_rules_t : public wf::plugin_interface_t, public wf::per_output_tracker
         description["fullscreen"]  = toplevel ? toplevel->pending_fullscreen() : false;
         description["minimized"]   = toplevel ? toplevel->minimized : false;
         description["activated"]   = toplevel ? toplevel->activated : false;
-        description["focusable"]   = view->is_focusable();
+        description["sticky"]    = toplevel ? toplevel->sticky : false;
+        description["wset-id"]   = toplevel && toplevel->get_wset() ? toplevel->get_wset()->get_id() : -1;
```
ammen99, line 581: wsets also have their internal indices that we can access by get_wset()->get_index(). The id()s are global IDs for all 'objects' - so views, outputs, wsets all have IDs in the same ID space. But wset->get_index() is special for wsets, and they are usually the indices displayed in the overlay of wsets.

In addition, the get_id() can change for the 'same' workspace set as seen by wsets: for example, if we create wset 3, it becomes empty and we delete it, then at some point go there again, we will get a new object with the same index.

So I think that maybe for IPC clients the index should be used instead of id (not just here, but in all calls), what do you think?

plugins/single_plugins/wsets.cpp
```diff
@@ -113,6 +121,48 @@ class wayfire_wsets_plugin_t : public wf::plugin_interface_t
...
+        }
+
+        auto response = wf::ipc::json_ok();
+        wset->attach_to_output(o);
+        return response;
+    };
+
+    wf::ipc::method_callback set_output_wset = [=] (nlohmann::json data)
+    {
+        WFJSON_EXPECT_FIELD(data, "id", number_integer);
```
ammen99, line 147: should be called `output-id`

plugins/single_plugins/wsets.cpp
```diff
@@ -113,6 +121,48 @@ class wayfire_wsets_plugin_t : public wf::plugin_interface_t
...
+        }
+
+        auto wset = wf::ipc::find_workspace_set_by_id(data["wset"]);
+        if (!wset)
+        {
+            return wf::ipc::json_error("workspace set not found");
+        }
+
+        auto response = wf::ipc::json_ok();
+        o->set_workspace_set(wset->shared_from_this());
```
ammen99, line 162: This should call `select_workspace()` if we use index instead of id - the function might need to be refactored a bit so that it gets the output as an additional parameter, which defaults to the currently focused output (but can also be set to an arbitrary output)

plugins/single_plugins/wsets.cpp
```diff
@@ -113,6 +121,48 @@ class wayfire_wsets_plugin_t : public wf::plugin_interface_t
...
+        }
+
+        auto o = wf::ipc::find_output_by_id(data["output"]);
+        if (!o)
+        {
+            return wf::ipc::json_error("output not found");
+        }
+
+        auto response = wf::ipc::json_ok();
+        wset->attach_to_output(o);
```
ammen99, line 141: This has the potential to crash Wayfire, if we move one output's active workspace set to another output: https://github.com/WayfireWM/wayfire/blob/master/src/output/workspace-impl.cpp#L336

I am in general wondering about the usefulness of this particular IPC call. If we do not immediately make the wset visible on another output, what is the use of this operation? Do you have a particular use-case?

ammen99 (changes requested):

plugins/single_plugins/ipc-rules.cpp
CharlieQLe: Sounds good to me!

plugins/single_plugins/wsets.cpp
ammen99: There is no need to pre-create workspace sets, the wsets plugin automatically creates a wset when it is needed. Otherwise you're wasting resources :)

ammen99 (approved):
I pushed one small commit with a few smaller fixes and also made it so that when selecting a workspace set on an output which doesn't exist, the workspace set with the given index will be created. This matches the behavior of the bindings that the `wset` plugin provides - otherwise we'd need a dedicated command for creating a wset which seems clumsy.

Otherwise LGTM, thanks!

bluebyt:
I would like to switch between workspace in a bar.
Can I have an example of a script to do that?

ammen99:
@bluebyt Workspace or workspace set? Here is an example for workspaces using the vswitch plugin:

```py
from wayfire_socket import WayfireSocket, get_msg_template
import os

sock = WayfireSocket(addr)

msg = get_msg_template('vswitch/set-workspace')
msg['data']['x'] = x
msg['data']['y'] = y
msg['data']['output-id'] = output-id
sock.send_json(msg)
```

For reference: 
https://github.com/WayfireWM/wayfire/blob/4f65d16420aad0dae6f26018088cfb839e93a622/plugins/vswitch/vswitch.cpp#L242

https://github.com/WayfireWM/wayfire/blob/4f65d16420aad0dae6f26018088cfb839e93a622/plugins/vswitch/vswitch.cpp#L251-L255

bluebyt:
@ammen99 Something like this, all workspace are displayed in the bar with numbers or icons and with the mouse you can navigate between them by clicking or scrolling on the number in the bar.


I don't known what is workspace set?

I tried the script on the terminal, but I get multiple error, my system should be all set, I can run other script like inactive-alpha.py without any error.

./workspace.py: line 1: from: command not found
import: unable to grab mouse '': Resource temporarily unavailable @ error/xwindow.c/XSelectWindow/9354.
import: unable to read X window image '': Success @ error/xwindow.c/XImportImage/4961.
import: unable to read X window image '': Resource temporarily unavailable @ error/xwindow.c/XImportImage/5068.
import:  `os' @ error/import.c/ImportImageCommand/1289.
./workspace.py: line 4: syntax error near unexpected token `('
./workspace.py: line 4: `addr = os.getenv('WAYFIRE_SOCKET')'

ammen99:
@bluebyt In case you're not looking at the matrix chat .. `maybe you need #!/bin/env python3 at the start?`

bluebyt:
After added the line "#!/bin/env python3" I have less error.
Traceback (most recent call last):
  File "/home/bluebyt/.config/ipc-scripts/./workspace.py", line 9, in <module>
    msg['data']['x'] = x
                       ^
NameError: name 'x' is not defined

ammen99:
Yes, replace x/y and the output-id appropriately xD

bluebyt:
What is output-id? I tried this HDMI-A-1, but I get an error.

ammen99:
You can see a list of available output by making a query `window-rules/list-outputs`, no arguments needed. Then you'll get in return a json object with a list of all outputs with their names and IDs

killown:
also you can try: pip install wayfire

from wayfire import sock
focused_view_id = sock.get_focused_view()["id"]
sock.set_workspace({"x":1, "y":1}, focused_view_id)

bluebyt:
Thanks @ammen99 and @killown I will play around with this!

PR #2162, opened 2024-02-25
Implement the session lock protocol

### LINKED ISSUE #1358: mako notifications cannot be made above fullscreen views but below lockscreens
**Describe the bug**
See title.

**To Reproduce**
1. Watch a YouTube video in fullscreen
2. Scroll down to read comments
3. Select a strange word and call via a key binding a dictionary lookup script that sends a notification as the result
4. The notification isn't visible unless I exit fullscreen or spawn a normal window above

Also, I want to see new messages even in fullscreen so I can react. Or watch videos while compiling things and sending myself a notification when done. These kinds of usage should be supported.

**Expected behavior**
The notification should be visible.

If I set `layer=overlay` for mako, the notification can appear above the fullscreen window but it also appears on lockscreen (created by swaylock).

**Wayfire version**
git

### LINKED ISSUE #1494: Support ext-session-lock
This will improve screenlocking security.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/staging/ext-session-lock

https://github.com/swaywm/swaylock/commit/3e84316ce294e329c23a8c2b2c89cd6c0cfbd182

(Refs from https://github.com/labwc/labwc/issues/215)

lcolitti:
Implement the session lock protocol

Currently implemented:

- Locking and unlocking.
- Display above all desktop layers, including OVERLAY (but below
  DWIDGET).
- Waiting for the client to display locked surfaces before
  declaring the screen to be locked. If the client does not
  display all surfaces, locks anyway after a timeout.
- Multiple outputs, including outputs being added/removed while
  locking or locked.
- Leaving the screen locked if the client crashes (but allows a
  new client to come in and lock the screen.

TODO:
- Deal with outputs being resized.
- Properly display text if the screenlocker crashes. Currently it
  just displays two a big "explosion" emoji.
- Use a view instead of a plain wlr_surface_node?

Tested with swaylock.

Fixes #1494
Fixes #1358


plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+                auto node = new std::shared_ptr<wf::scene::wlr_surface_node_t>(
+                    new lock_surface_node_t(lock_surface->surface, true));
+
+                lock_surface->data = node;
+                auto target_layer  = wf::scene::layer::TOP;
+                auto wo = wf::get_core().output_layout->find_output(lock_surface->output);
+                auto layer_node = wo->node_for_layer(target_layer);
+
+                (*node)->set_enabled(true);
+                wf::scene::add_front(layer_node, *node);
```
soreau, line 101: You might need to set this to `add_back` and/or change the `target_layer` above, to get swaylock to appear.

src/api/wayfire/core.hpp
```diff
@@ -144,6 +144,8 @@ class compositor_core_t : public wf::object_base_t, public signal::provider_t
         wlr_xdg_foreign_registry *foreign_registry;
         wlr_xdg_foreign_v1 *foreign_v1;
         wlr_xdg_foreign_v2 *foreign_v2;
+
+        wlr_session_lock_manager_v1 *session_lock;
```
ammen99, line 148: Do we need to define this in core, since the functionality is implemented as a plugin?

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+        class lock_surface_keyboard_interaction : public wf::keyboard_interaction_t
+        {
+          public:
+            lock_surface_keyboard_interaction(wlr_surface *surface) : surface(surface)
+            {}
+
+            void handle_keyboard_enter(wf::seat_t *seat)
+            {
+                // TODO: use a keyboard grab instead?
```
ammen99, line 43: I think this is fine - wlroots grabs receive events if Wayfire decides to forward requests to wlroots - with HIGH focus importance and putting the lockscreen above everything else, Wayfire should always feed the focus to the lockscreen first.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+            wlr_surface *surface;
+        };
+
+        wf::keyboard_interaction_t& keyboard_interaction()
+        {
+            return *interaction;
+        }
+
+      private:
+        wf::keyboard_interaction_t *interaction;
```
ammen99, line 68: I think this should be a unique_ptr

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+            plugin = parent;
+            lock   = data;
+
+            new_surface.set_callback([this] (void *data)
+            {
+                wlr_session_lock_surface_v1 *lock_surface = (wlr_session_lock_surface_v1*)data;
+                uint32_t serial = wlr_session_lock_surface_v1_configure(lock_surface,
+                    lock_surface->output->width,
+                    lock_surface->output->height);
+                LOGI("surface_configure output=", lock_surface->output->name, " ",
```
ammen99, line 85: Once we're ready to merge, this should be LOGD or maybe even LOGC(VIEWS, ...)

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+                auto node = new std::shared_ptr<wf::scene::wlr_surface_node_t>(
+                    new lock_surface_node_t(lock_surface->surface, true));
+
+                lock_surface->data = node;
+                auto target_layer  = wf::scene::layer::TOP;
+                auto wo = wf::get_core().output_layout->find_output(lock_surface->output);
+                auto layer_node = wo->node_for_layer(target_layer);
+
+                (*node)->set_enabled(true);
```
ammen99, line 100: Is this necessary? I thought nodes are created enabled by default ..

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+            {
+                cur_lock.reset(new wayfire_session_lock(this, wlr_lock));
+                // TODO: blank the screen, or set a timeout for the lock app to commit a surface.
+                wlr_session_lock_v1_send_locked(wlr_lock);
+                LOGI("new_lock");
+            } else
+            {
+                LOGI("new_lock: already locked");
+                wlr_session_lock_v1_destroy(wlr_lock);
+            }
```
ammen99, line 164: The old locking implementation also inhibits the output so that no plugins can be activated (I think this implementation doesn't do that?): https://github.com/WayfireWM/wayfire/blob/2e0926f8f5a1c3834c51e97673bfbf074f4506ac/src/output/output-impl.hpp#L67-L79

We'd probably want to move these functions to the public API so that lockscreens can be implemented from plugins as well. Also, please change inhibit/uninhibit to `set_inhibited(bool)` in the process :)

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+                {
+                    auto node = (std::shared_ptr<wf::scene::wlr_surface_node_t>*)lock_surface->data;
+                    if (node != nullptr)
+                    {
+                        wf::scene::remove_child(*node);
+                    }
+
+                    lock_surface_destroy.disconnect();
+                });
+                lock_surface_destroy.connect(&lock_surface->events.destroy);
```
ammen99, line 114: You will need one listener per created surface (because listeners cannot connect to multiple object's signals unlike Wayfire's signals). So lock_surface_destroy() probably will have to be moved to the node class.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+                LOGI("unlock");
+            });
+            unlock.connect(&lock->events.unlock);
+
+            destroy.set_callback([this] (void *data)
+            {
+                new_surface.disconnect();
+                unlock.disconnect();
+                destroy.disconnect();
+                LOGI("destroy");
```
ammen99, line 129: I think that the point of ext-screenlock was that we should keep the screen locked if the client crashes and we haven't unlocked.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,198 @@
...
+        auto display = wf::get_core().display;
+        manager = wlr_session_lock_manager_v1_create(display);
+
+        new_lock.set_callback([this] (void *data)
+        {
+            wlr_session_lock_v1 *wlr_lock = (wlr_session_lock_v1*)data;
+            if (cur_lock.get() == nullptr)
+            {
+                cur_lock.reset(new wayfire_session_lock(this, wlr_lock));
+                // TODO: blank the screen, or set a timeout for the lock app to commit a surface.
```
ammen99, line 157: Maybe same as in the 'crashed' case, I would create a simple cairo surface saying 'Waiting for screenlock' or 'Screenlocker crashed?' 

Here is a small example of how to show a cairo-rendered text in a node: https://github.com/WayfireWM/wayfire/blob/2e0926f8f5a1c3834c51e97673bfbf074f4506ac/plugins/single_plugins/wsets.cpp#L25-L78

If you feel like it, you could also make this into a helper class in `plugins/common/...` - a node implementation which can take cairo options and text, can be set the geometry, etc. But if you feel like this is too much, feel free to also copy-paste the code and adapt it.

ammen99 (commented):
I have a few requests/suggestions on the implementation, but the general approach seems good!

ammen99:
> * The swaylock indicator isn't displayed. Perhaps this is because swaylock uses a subsurface to draw it, and wlr-surface-node doesn't support child surfaces?

Yeah, you'd need to create a subsurface controller: https://github.com/WayfireWM/wayfire/blob/master/src/view/surface-impl.hpp#L15

Probably should be moved to the unstable APIs.

ammen99:
> * Deal with outputs being resized

I think it might actually make sense to actually create two nodes: one node is compositor-generated, saying 'Waiting for screenlocker ...' and above it, we should add the screenlock.

This covers basically all cases: before the screenlock surface is there, we hide the contents of the desktop. If the screenlock crashes, the second node below will still cover / lock the screen. And, if we have to resize the output, the second node will cover the screen while the screenlocker is resizing.

plugins/protocols/session-lock.cpp
lcolitti: Removed.

src/api/wayfire/core.hpp
lcolitti: Left over from a previous attempt :-). Removed.

plugins/protocols/session-lock.cpp
lcolitti: Changed to add_back. Not sure what the difference between front and back would be here. I guess it doesn't matter because the only thing in the lock layer is this surface?

lcolitti:
> > * The swaylock indicator isn't displayed. Perhaps this is because swaylock uses a subsurface to draw it, and wlr-surface-node doesn't support child surfaces?
> 
> Yeah, you'd need to create a subsurface controller: https://github.com/WayfireWM/wayfire/blob/master/src/view/surface-impl.hpp#L15
> 
> Probably should be moved to the unstable APIs.

Done in [this commit.](https://github.com/WayfireWM/wayfire/pull/2162/commits/b8028e620bbe6a7d7e21f8b72db356a8ffae7b84). I also renamed them, but I can keep the existing names if you prefer.

plugins/protocols/session-lock.cpp
lcolitti: Thanks for the suggestion. Done, I think.

plugins/protocols/session-lock.cpp
lcolitti: Done in [this commit](https://github.com/WayfireWM/wayfire/pull/2162/commits/714cd91a88758cca1237fc2388ef0e31e90fe4d4). Let me know if you'd like me to split out these refactoring commits to another PR so we can merge them first.

lcolitti:
> I think it might actually make sense to actually create two nodes: one node is compositor-generated, saying 'Waiting for screenlocker ...' and above it, we should add the screenlock.
> 
> This covers basically all cases: before the screenlock surface is there, we hide the contents of the desktop. If the screenlock crashes, the second node below will still cover / lock the screen. And, if we have to resize the output, the second node will cover the screen while the screenlocker is resizing.

I'm not sure if this is the best approach. For one thing, if we create the surface immediately, it will be visible/jarring. Ideally we just display the lockscreen surface like we do today without this plugin. Personally my lockscreen is the same as my background so when the lockscreen activates it just looks as if the windows smoothly fade out. I'd hate to make that less slick :smile:  I guess we could add this extra surface after a timeout? Not sure how to do that yet but there's probably a way.

Also, what would happen when an output is added? With the code as is, the plugin doesn't do anything. The screenlocker will probably react to this by creating a new surface, and when it does, the new output will be locked. But while the client creates the surface, I think there is a race where it's possible to display or interact with the new output, even if the others are locked. But can we close that race completely even if we create the surface in the compositor? There could still be a window between when the output is created and the surface is added.

ammen99:
> But can we close that race completely even if we create the surface in the compositor? There could still be a window between when the output is created and the surface is added.

We most certainly can. The compositor is in full control of drawing every frame on every output. If you create a window/node on a new output in the added handler for that event, it will be present on the very first output frame :)

> I'd hate to make that less slick 😄 I guess we could add this extra surface after a timeout? Not sure how to do that yet but there's probably a way.

Ok, I guess this is a good point. Maybe the screenlocker wants to implement fade-in or a slide effect, so we shouldn't create the surface immediately. Maybe say 5 seconds after the lockscreen has kicked in? I'm not sure whether we want this to be configurable, so maybe just hardcode it and we can always add an option later.

plugins/protocols/session-lock.cpp
lcolitti: BTW, any reason not to use wf::input_manager_t::set_exclusive_focus here? Seems like it would be simpler. If the client crashes I assume that focus will go nowhere, but it sounds like that's what we want.

ammen99: set_exclusive_focus was meant for the older implementation of screenlockers, I think we can even remove this function nowadays. It is better to use the scenegraph API for this ;)

plugins/protocols/session-lock.cpp
ammen99: > Let me know if you'd like me to split out these refactoring commits to another PR so we can merge them first.

We can do it as you wish. For my personal PRs I usually use git rebase and squash related commits, but if you find multiple PRs simpler, we can do that too.

lcolitti:
Ok, this looks closer to done now.

I think it deals correctly with outputs being added and removed, and (as required by the protocol) it now waits for the screenlocker to display a frame before it sends back the lock event. It uses @ammen99 's suggestion to use two nodes to ensure that if the screenlocker crashes, the screen remains locked. The UX of that is a bit primitive (a big, pixelated explosion emoji), but that shouldn't be too hard to fix by modifying simple-text-node.

It still also uses a surface instead of a view, but maybe that's OK for now.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+    UNLOCKED,
+    DESTROYED,
+    ZOMBIE,
+} state;
+
+class wf_session_lock_plugin : public wf::plugin_interface_t
+{
+    class wayfire_session_lock;
+
+    class lock_surface_node_t : public wf::scene::wlr_surface_node_t
```
ammen99, line 31: Note that all nodes are shared_ptrs - this means that in general, it is possible that a node outlives the session-lock plugin to begin with, it can also outlive the output it is on, etc.

So, you should try to add code so that the object can 'work' (at least not crash) even as the underlying surface, lock, output, etc. are destroyed. I have tried to add some suggestions on how to accomplish this.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+            output(output),
+            interaction(std::make_unique<lock_surface_keyboard_interaction>(lock_surface->surface,
+                lock_surface->output))
+        {
+            lock_surface_destroy.set_callback([this] (void *data)
+            {
+                wf::wlr_surface_controller_t::try_free_controller(this->lock_surface->surface);
+                wf::scene::remove_child(shared_from_this());
+                lock_surface_destroy.disconnect();
+                this->lock->surface_destroyed(this->output);
```
ammen99, line 49: Re lifetimes: I would personally set the destroy callback in the wayfire_session_lock class (this way, you don't have to pass the `lock` parameter.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+                const char *name = this->output->handle ? this->output->handle->name : "(deleted)";
+                LOGD("lock_surface on ", name, " destroyed");
```
ammen99, line 51: You can/should reset the keyboard interaction to something else (something like this: https://github.com/WayfireWM/wayfire/blob/master/src/view/wlr-surface-node.cpp#L102) so that we don't get a crash if someone holds on to the node after the surface/output are destroyed.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+            auto layer_node = output->node_for_layer(wf::scene::layer::LOCK);
+            wf::scene::add_front(layer_node, shared_from_this());
+            wf::wlr_surface_controller_t::create_controller(lock_surface->surface, layer_node);
+            focus();
+        }
+
+        void focus()
+        {
+            wf::get_core().seat->set_active_node(shared_from_this());
+            interaction->handle_keyboard_enter(wf::get_core().seat.get());
```
ammen99, line 67: Why is this necessary? A simple `refocus()` call should be enough to focus the node. Manually calling `enter()` is usually not what you want, this is what core uses when it focuses your node.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+        {
+          public:
+            lock_surface_keyboard_interaction(wlr_surface *surface, wlr_output *output) : surface(surface),
+                output(output)
+            {}
+
+            void handle_keyboard_enter(wf::seat_t *seat)
+            {
+                wlr_seat_keyboard_enter(seat->seat, surface, nullptr, 0, nullptr);
+                LOGI("Focused ", output->name);
```
ammen99, line 90: This should be a LOGD at most (though it probably should be removed altogether)

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+        }
+
+        void focus()
+        {
+            wf::get_core().seat->set_active_node(shared_from_this());
+            interaction->handle_keyboard_enter(wf::get_core().seat.get());
+        }
+
+        wf::keyboard_focus_node_t keyboard_refocus(wf::output_t *output) override
+        {
```
ammen99, line 71: Shouldn't this focus only if our own output is focused?

plugins/protocols/session-lock.cpp
ammen99: And also re lifetimes, you should set lock_surface to NULL here, and make sure it is not used in the other functions if it is NULL already (where it makes sense)

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        wf::keyboard_interaction_t& keyboard_interaction()
+        {
+            return *interaction;
+        }
+
+      private:
+        wayfire_session_lock *lock;
+        wlr_session_lock_surface_v1 *lock_surface;
+        wf::output_t *output;
```
ammen99, line 117: Re lifetimes; I see that the output is a necessary thing to keep a reference to, which means you ought to also have a handler for when it is destroyed.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+    };
+
+    class wayfire_session_lock
+    {
+      public:
+        wayfire_session_lock(wf_session_lock_plugin *plugin, wlr_session_lock_v1 *lock) :
+            plugin(plugin), lock(lock)
+        {
+            auto& ol = wf::get_core().output_layout;
+            output_added.set_callback([this] (wf::output_added_signal *ev)
```
ammen99, line 171: Instead of doing this manually, check out https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/per-output-plugin.hpp#L30

You can add it as a base class and then it will connect to the signals, etc. for you.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+            {
+                handle_output_added(output);
+            }
+
+            new_surface.set_callback([this] (void *data)
+            {
+                wlr_session_lock_surface_v1 *lock_surface = (wlr_session_lock_surface_v1*)data;
+                wlr_output *wo = lock_surface->output;
+                wlr_session_lock_surface_v1_configure(lock_surface,
+                    wo->width / wo->scale,
```
ammen99, line 193: I don't think this takes output transform into account. You can however find the `wf::output_t` associated with the wlr output and get its geometry - the geometry will be scaled and rotated correctly.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+                    wo->height / wo->scale);
+                LOGD("surface_configure output=", wo->name, " ", wo->width, "x", wo->height);
```
ammen99, line 195: I think a lot of the log messages in this file can be put into a debugging category (see `debug.hpp`) - either VIEWS or LSHELL (I know this protocol technically is not layer-shell but they are closely related, at least in my mind :P)

plugins/protocols/session-lock.cpp
ammen99: Looking at the code below, I assume you did this per hand simply because the mixin does not support non-plugin template types (yet). I think it should be very easy (and a good idea) to fix the mixin class - with an `if constexpr (std::is_base_v<per_output_plugin_instance_t, ConcretePluginType>) {}` around code blocks which are specific for per-output-plugin instances, so that the helper can be used generically.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+            break;
+
+          default:
+            break;
+        }
+    }
+
+    void fini() override
+    {
+        // TODO: unlock everything?
```
ammen99, line 418: I don't think we actually need to do anything. Plugins that cannot be unloaded like this one will be destroyed only at compositor shutdown. At that point the wayland display is being destroyed, so wlroots will auto-destroy the protocol and it will emit destruction signals for the various lock surfaces, etc. (I hope it does so at least).

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,434 @@
...
+            if (state == LOCKED)
+            {
+                lock_output(output, output_states[output]);
+            }
+        }
+
+        void handle_output_removed(wf::output_t *output)
+        {
+            output_states.erase(output);
+            auto next_output_state = output_states[wf::get_core().seat->get_active_output()];
```
ammen99, line 276: Again, is this needed? Core should auto-refocus correctly if you filter by the output in the refocus_keyboard() method.

ammen99 (changes requested):

ammen99:
I went through your code, the overall design & implementation looks good. Lifetimes are hard to get right, especially when you have a shared_ptr, that needs some care.

plugins/protocols/session-lock.cpp
lcolitti: Attempted to reset the keyboard interaction.

I don't think lock_surface can be used after it's destroyed. The only other method that uses it is attach_to_layer. But attach_to_layer will only be called if the output_state->surface is non-null. But when the lock_surface is destroyed, the callback for the destroy event calls this->lock->surface_destroyed, which nulls out the output_state->surface, so attach_to_layer will not be called.

plugins/common/wayfire/plugins/common/simple-text-node.hpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    void set_text_params(wf::cairo_text_t::params params)
+    {
+        this->params = params;
+    }
+
+    void set_text(std::string text)
+    {
+        cr_text.render_text(text, params);
```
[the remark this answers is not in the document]
lcolitti: Done, I think.

plugins/protocols/session-lock.cpp
lcolitti: I did use the mixin for a while, to create the per-output objects. The reason I stopped using it is that the mixin expects the per-output instances to have zero-arg constructors. But here, the instances need to have a backpointer to a plugin-global object, for example to call notify_lock_state. Not sure if this is easy to do in the mixin - my command of C++ syntax isn't great.

plugins/protocols/session-lock.cpp
ammen99: Ok, it might still be useful to set it to NULL so that if in the future we change anything, we directly get a null pointer crash instead of a dangling pointer.

ammen99:
@lcolitti Sorry I mistakenly merged this PR without even reviewing your latest changes .. Would you mind opening a new pull request or reopening this one? I'm not sure how to make GitHub do that ...

lcolitti:
Sorry I didn't update this PR for a while. I've been travelling and can't test with my multi-monitor setup.

I don't think there's a way to reopen this after the merge. Opened #2237 instead. @ammen99 can you take a look? The code there addresses most of your review comments, including hopefully not crashing on shutdown. I've been running this code on my machine for a while and it seems to work well.

PR #2163, opened 2024-02-25
ipc-rules: add output-added and output-removed signals

CharlieQLe:
Haven't tested this yet, but this would be nice for laptop users plugging into different monitors.


ammen99 (approved):
LGTM, thanks :)

PR #2170, opened 2024-03-02
xdg-popup: include popup's own position in the calculation

### LINKED ISSUE #2141: Tooltip flicker
**Describe the bug**
When hovering over widgets in wf-panel the tooltip will flicker intensely when mouse moves over tooltip box

I first noticed it with the Teams chromium web app.  It is not happening with Chromium though.  
Other place it happens consistently is wf-panel with the window-list widget and system tray.
I couldn't reproduce in firefox, thunar and thunderbird

**Wayfire version**
this started when i updated to commit 7388778. Previously I was on d6abead.

ammen99:
Fixes #2141


mrsteve0924:
that fixed it.  thanks!

PR #2186, opened 2024-03-07
im-v1: send null preedit_string if it's empty

lilydjwg:
This makes GTK to emit preedit-end signals so that vte paints its cursor. The preedit_string from input-method-v1 can't be null.


ammen99 (approved):

lilydjwg:
PS: GNOME does this. KDE ignores it and a following `done` event resets the preedit thing.

There is a related issue https://github.com/fcitx/fcitx5/issues/514 but for input-method-v2.

PR #2192, opened 2024-03-09
change overlay to workspace as default in scale plugin

### LINKED ISSUE #1528: command_screenshot + toggle + hotspot: may freeze UI
When I initiate the screenshot commandand in turn get the selection rectangle tool, but then before selecting anything, accidentally move the mouse into the hotspot top-left area, then wayfire shows the window/workspace overview but I cannot exit from there anymore. 

It seems that waiting for the screen selection (by mouse) plus waiting for a mouse click to select the window from the overview somehow block each other.

This is my configuration:

`toggle_all = <super> KEY_W | hotspot left-top 10x10 200`
`binding_screenshot = <super> KEY_S`
`command_screenshot = grim -g "$(slurp)" -t jpeg /tmp/screenshot.jpg
`

You need to in stall grim + slurp for that.

### LINKED ISSUE #1861: Scale plugin, interact with panel
If possible I would like to be able to interact with the panel, in my case waybar, while looking at the scale view.
This is most relevant when I'm playing a game in full-screen since "tabbing out" with switcher does not put the panel on top.

My current solution is switching to another workspace and do things there, but it's sometimes a little awkward.

Thanks in advance for you consideration.

killown:
unless there is a reason to use overlay as default, workspace makes more sense to turn any visible panels interactions

Fixes #1861
Fixes #1528


ammen99:
I think this also 'accidentally' fixes #1528 :)

ammen99 (approved):
Thanks :)

PR #2194, opened 2024-03-10
Set space between the edges

killown:
There is an option to set space between views which is useful to set in the edges too

the default space for edges look like this:

scale1

scale2

scale4

scale6


ammen99:
I think this should be a separate option though, not reuse the `scale/spacing` option because maybe you want bigger spacing on the edges but to have the windows more tightly packed together.

ammen99 (approved):
LGTM, thanks!

PR #2196, opened 2024-03-11
Add views to signals

marcusbritanicus:
Add wayfire_toplevel_view arg to two signals:
- view_activated_state_signal
- view_parent_changed_signal


ammen99 (changes requested):
I think we ought to bump the API/ABI version in plugin.hpp. Otherwise LGTM.

marcusbritanicus:
Ah!! I knew I had missed something... :) Give me a few minutes.

ammen99 (approved):
Thanks :)

Issue #2205, opened 2024-03-12
Decoration plugin title bar height size issue


killown:
you are allowing heights with no limits like 2000px, if you set that, you can't disable anymore inside a wayfire session, because the bar is so huge, the only way is leaving wayfire and disabling the plugin from wayfire.ini

PR #2208, opened 2024-03-12
set max height for decoration plugin

killown:
fix for #2205


killown:
cannot even update the branch anymore

PR #2211, opened 2024-03-13
core: do not resize view in move_view_to_output

ammen99:
We already changed the behavior similarly in #2123, so it should be fine to change it here too.

### DIFF
--- a/plugins/protocols/wayfire-shell.cpp
+++ b/plugins/protocols/wayfire-shell.cpp
@@ -312,6 +312,18 @@ class wfs_output
     void create_hotspot(uint32_t hotspot, uint32_t threshold, uint32_t timeout,
         uint32_t id)
     {
+        if (!this->output)
+        {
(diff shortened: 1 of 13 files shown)


mark-herbert42:
Works perfectly - no more window growth when disconnecting  external display.

mark-herbert42:
I've opened a bug in wf-shell regarding wf-background crash - it seems to belong there, not a wf core functionality. The way it crashes very much look that it still tries to paint next wallpaper image on disconnected monitor.  Wf-background is detecting the new monitor well and paint the active wallpaper on it immediately. When monitor is disconnected and time comes to paint a new wallpaper - then new wallpaper appersr on  the main monitor and then it crahes almost imeediately, sometimes I can see the new wallper well somertimes it crahes in the middle of animation. So  seems that wf-wallpaper does not catch the output discinnect event properly, but clearly works well when monitor is connected.

I've compiled wf-shell from current git.

PR #2215, opened 2024-03-13
conf: fix reloading on IN_CREATE

### LINKED ISSUE #2210: inotify watch on config file should include IN_CREATE
**Describe the bug**

Currently, Wayfire will fail to automatically reload the config if that config is replaced by a new file (e.g., if it's a symlink). This causes difficulty specifically with NixOS and home-manager, where, to keep the config reproducible, it's symlinked in from the nix store. This is likely caused by inotify only watching the config for IN_CLOSE_WRITE events when a more liberal selection would be more appropriate.

**To Reproduce**

Write a config file, and symlink it to `~/.config/wayfire.ini` or similar. Start wayfire. Write a new config file, and symlink (`-f`) that new file into `~/.config/wayfire.ini`. Observe that wayfire does not reload the config.

**Expected behavior**

Wayfire should reload the config regardless of how it's updated.

**Wayfire version**

0.8.0. A cursory glance at git makes me think that this is still an issue.

---

Thanks so much! (By the way, my apologies if I've made a mistake reporting this; it's late here!)

atalii:
See the linked issue for context.

This is a strange bug, and I honestly don't know why this fix works (or, rather, why the original code was buggy). I wrote some smaller test programs, and `operator==` generally *does* seem to work over a flexible array member. I may be missing something obvious, or there may be some UB going on when `->name` is inspected in the buffer returned from `read(2)`. In any case, this fix appears to work, though I'd like some external confirmation or refutation of that given how shaky it seems. Thanks!

Commit message below:

```
If the config file is a symlink, wayfire attempts to watch the parent directory so that, if replaced via a subsequent `ln -sf`, it'll still be notified. When notified, however, wayfire doesn't immediately reload the config file. Rather, as this event may be triggered by any file creation in the directory, we first make sure that the config file really is the one being created anew.

The filename from inotify is stored as a flexible array member in the event struct, though, and this was somehow causing problems when compared with the known std::string basename.

Fixes #2210.
```

### DIFF
--- a/src/default-config-backend.cpp
+++ b/src/default-config-backend.cpp
@@ -7,6 +7,7 @@
 #include <wayfire/plugin.hpp>
 #include <wayfire/core.hpp>
 
+#include <cstring>
 #include <sys/inotify.h>
 #include <filesystem>
 #include <unistd.h>
@@ -20,7 +21,7 @@ static int wd_cfg_dir, wd_cfg_file;
 
 static void add_watch(int fd)
 {
-    wd_cfg_dir  = inotify_add_watch(fd, config_dir.c_str(), IN_CREATE);
+    wd_cfg_dir  = inotify_add_watch(fd, config_dir.c_str(), IN_CREATE | IN_MOVED_TO);
     wd_cfg_file = inotify_add_watch(fd, config_file.c_str(), IN_CLOSE_WRITE);
 }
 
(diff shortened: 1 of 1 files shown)


src/default-config-backend.cpp
```diff
@@ -58,18 +59,22 @@ static int handle_config_updated(int fd, uint32_t mask, void *data)
...
-        // - Config file itself was modified
-        // - Config file was created inside parent directory
-        should_reload |=
-            (event->wd == wd_cfg_file) || (cfg_file_basename == event->name);
+        // 1. The config file itself was modified, or...
+        should_reload |= event->wd == wd_cfg_file;
+        // 2. The config file was moved/created inside the parent directory.
+        if (event->len > 0) {
+            // This is UB unless event->len > 0.
+            auto name_matches = strcmp(cfg_file_basename.c_str(), event->name);
```
ammen99, line 69: Wait, strcmp returns 0 iff the config file and the name match, isn't that right? So, the variable is actually `name_does_not_match`

atalii: ....and that's what was going on. `==` never succeeded whenever it was supposed to. I'll revise the patch to make it actually correct. Thanks!

atalii:
The issue was a little less subtle than I thought; the new commit should fix it. Thanks :)

ammen99 (approved):
LGTM, thanks :)

ammen99:
Please also fix the code style (https://github.com/WayfireWM/wayfire/actions/runs/8317772853/job/22774734664?pr=2215#step:8:1)

atalii:
> Please also fix the code style (https://github.com/WayfireWM/wayfire/actions/runs/8317772853/job/22774734664?pr=2215#step:8:1)

Oops, thanks :)

Issue #1492, opened 2022-06-01
[tile] Customizeable Tiling of New Nodes


Nitepone:
**Currently**

Wayfire seems to assign a sliver of the screen to fill (1/(N+)) percent of the workspace (Where N is the number of nodes). And resizes over the existing windows relatively to make space for itself.

The direction of where the new window is put seems to be a hidden attribute of the root node from my testing. I have been able to put the workspace into a state where closing all windows, then opening new windows causes **all** new children to open from the same side of the screen (right or bottom).

**Suggestion**

Provide the user better options as to where new nodes spawn, and reduce invisible behaviors.

The paradigm I would like to see is one such that the new node spawned is allocated half the space of the currently (or last) active node. Specifically splitting the node against it's longest dimensions to keep the resulting children as close to square as possible.
This is the same paradigm used in bspwm.

Nitepone:
I understand the very limited scope of simple tile. #771 is far more specific than this feature.

This ticket only pertains to a standard behavior of "how new nodes are spawned", and not meant to cover predefined layouts.

Current behavior of how nodes are spawned is very confusing and not desirable in my _opinion_. It has some really weird behaviors.

e.g
1. spawn 2 windows
2. stack the windows {horizontally, vertically}
3. close **all** windows
4. spawn any N number windows
5. note that all windows now spawn {horizontally, vertically}

Scrapping the customization aspect, I would still suggest making it so that windows spawn in a more deterministic fashion. With of course a preference to my own suggestion ;)

Finally, adding better node spawning to simple-tile is something I would be willing to contribute to this project. If there is interest of merging such a change, I'd like to make something that could be PR'd instead of maintaining my own fork with a single change.

ammen99:
> Finally, adding better node spawning to simple-tile is something I would be willing to contribute to this project. If there is interest of merging such a change, I'd like to make something that could be PR'd instead of maintaining my own fork with a single change.

Just a heads up (if you still use Wayfire), with the upcoming tiling IPC I really do not think we need anything more complicated than the current rule upstream: https://github.com/WayfireWM/wayfire/pull/2219

PR #2219, opened 2024-03-15
simple tile ipc

### LINKED ISSUE #1401: [simple-tiling] request: Drag between monitors
Firstly, I just discovered this project and it's exactly what I've been looking for. Minimal but good looking and animated, wayland and inbuilt tiling! Great work!

One slight frustration I have is that with the tiling plugin enabled I can't find a way to move a window from one monitor to another, the window can only ever exist on the monitor I opened it on.

`<super>` + drag works (although it's slightly intermittent, but that's a different issue) 

It doesn't matter if the target monitor has any windows open or not, trying to drag a window with `<super>` + drag doesn't let me drag the window between monitors.

I have a minor UI suggestion as well, when dragging the window, it would be nice if the target window was resized so I could see what the target window will be like before I move a window over it.

### LINKED ISSUE #2021: dwindle layout
add a dwindle tiling layout like hyprland, idk there's only one tiling layout and it's not that good

### LINKED ISSUE #2109: scale + scale-title-filter plugins: extra title text flickering
**Describe the bug**
When activating the scale plugin with scale-title-filter, there is title text flickering at the center. After filtering it may appear elsewhere. It can even be cropped from the middle. Moving mouse cursor to a window stops it.

**To Reproduce**
1. Enable plugin scale & scale-title-filter
2. Activate it
3. See flickering text

**Expected behavior**
It shouldn't be there.

**Screenshots or stacktrace**

https://github.com/WayfireWM/wayfire/assets/440661/78d4c715-20fc-42fb-b5ae-98c0600b9ff8

**Wayfire version**
git.

ammen99:
- main: add a commented-out assert handler
- simple-tile: add ipc supports
- add master-stack-tile-layout.py example

Fixes #2021, #771, #1492
Fixes #1401 
Fixes #2109


killown:
sometimes that happen if you move too much, low priority fix needed, the simple tile is pretty much advanced now
nice

killown:
Hi there!
about the simple tile bug:
to reproduce the issue you must have at least two monitors and firefox with fractional scaling (I am not sure if fractional scaling is causing this)


1 - open firefox in a empty workspace in the output from the right
2 - open a terminal which will be set in  the right side
3 - drag the right side view with super key + BTN_RIGHT and move as fast you can to the monitor on the lef

result: The terminal view will cover firefox view

PR #2221, opened 2024-03-16
ipc: add get view alpha

killown:
working in wayfire module:

In [3]: sock.get_view_alpha(10)
Out[3]: {'alpha': 0.7899999618530273, 'result': 'ok'}


plugins/single_plugins/alpha.cpp
```diff
@@ -53,6 +53,7 @@ class wayfire_alpha : public wf::plugin_interface_t
         min_value.set_callback(min_value_changed);
         wf::get_core().bindings->add_axis(modifier, &axis_cb);
         ipc_repo->register_method("wf/alpha/set-view-alpha", ipc_set_view_alpha);
+        ipc_repo->register_method("wf/alpha/get-view-alpha", ipc_get_view_alpha);
```
ammen99, line 56: needs also `unregister_method` in the fini() function to ensure proper unloading of the plugin

plugins/single_plugins/alpha.cpp
```diff
@@ -73,6 +74,22 @@ class wayfire_alpha : public wf::plugin_interface_t
...
 
+    wf::ipc::method_callback ipc_get_view_alpha = [=] (nlohmann::json data) -> nlohmann::json
+    {
+        WFJSON_EXPECT_FIELD(data, "view-id", number_unsigned);
+
+        auto view = wf::ipc::find_view_by_id(data["view-id"]);
+        if (view && view->is_mapped())
+        {
+            auto response = wf::ipc::json_ok();
+            response["alpha"] = ensure_transformer(view)->alpha;
```
ammen99, line 85: ensure_transformer will add a new transformer to the view even if the alpha==1.0. It would be better to check whether a transformer exists (if not, alpha == 1.0) and if it does, then get its alpha.

killown: done

plugins/single_plugins/alpha.cpp
killown: done

ammen99 (approved):
LGTM, thanks :)

PR #2225, opened 2024-03-18
Copy config.h approach from wlroots-full.hpp to other public api headers

parona-source:
* Otherwise wayfire could use the config.h of an already installed wayfire installation and lead to confusing linking issues if the options differed (like installed wayfire has xwayland disabled and the new build has it enabled).

### DIFF
--- a/src/api/wayfire/debug.hpp
+++ b/src/api/wayfire/debug.hpp
@@ -1,8 +1,12 @@
 #ifndef DEBUG_HPP
 #define DEBUG_HPP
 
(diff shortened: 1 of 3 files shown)


ammen99 (commented):
That is indeed a good catch! Could you also fix `debug.hpp` to use the same (WF_USE_CONFIG_H instead of the other macro)?

parona-source:
Pushed, didnt notice debug.hpp as my grep was too naive for that.

ammen99 (approved):
LGTM, thanks!

PR #2226, opened 2024-03-18
Explicitly check wlroots features and fail if required ones are missing

parona-source:
DRM, GLES2, libinput and session are all expected to be present unconditionally in wlroots-full.hpp which is a part of the wayfire api

https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/nonstd/wlroots-full.hpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,15 +57,21 @@ These are the dependencies needed for building Wayfire.
 These are the dependencies needed for building wlroots, and should be installed before building it.
 They are relevant for cases when the system doesn't have a version of wlroots installed.
 
-#### DRM Backend (optional, but you most likely want this)
+#### DRM Backend (required)
 
 - [libdisplay-info-dev](https://gitlab.freedesktop.org/emersion/libdisplay-info)
 - [hwdata-dev](https://github.com/vcrhonek/hwdata)
 
-#### Session Provider (optional, recommended)
+#### GLES2 renderer (required)
+- [libglvnd](https://gitlab.freedesktop.org/glvnd/libglvnd)
+- [mesa](https://gitlab.freedesktop.org/mesa/mesa) (with libEGL and gbm support)
 
-- [systemd](https://systemd.io/) **or**
-- [elogind](https://github.com/elogind/elogind) **or**
+#### Libinput Backend (required)
+- [libinput](https://gitlab.freedesktop.org/libinput/libinput)
+
+#### Session Provider (required)
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -57,15 +57,20 @@ These are the dependencies needed for building Wayfire.
...
 
-#### DRM Backend (optional, but you most likely want this)
+#### DRM Backend (required)
 
 - [libdisplay-info-dev](https://gitlab.freedesktop.org/emersion/libdisplay-info)
 - [hwdata-dev](https://github.com/vcrhonek/hwdata)
 
-#### Session Provider (optional, recommended)
+#### GLES2 renderer (required)
+- [libglvnd](https://gitlab.freedesktop.org/glvnd/libglvnd)
```
ammen99, line 66: Is libglvnd an actual dependency of wlroots? I don't see it listed on their README. Also by looking at the meson.build files, the gles2 renderer needs egl, gbm and glesv2

ammen99 (changes requested):
LGTM except the dependencies for the GLES renderer in the README.

README.md
parona-source: The most direct dependency is libglvnd as it provided the glesv2 pkgconfig file that is used by wlroots. 

https://gitlab.freedesktop.org/glvnd/libglvnd/-/blob/master/src/GLESv2/glesv2.pc.in?ref_type=heads

parona-source: but egl and gbm (so mesa) should also be there.

ammen99: I think we should be even more explicit about these things, because on other distros (like mine, Fedora) the mesa packages is split in multiple parts. For example GBM is not mesa-devel but mesa-libgbm-devel. Maybe just specify that we need the gbm and egl components of it.

parona-source:
```suggestion
- [libglvnd](https://gitlab.freedesktop.org/glvnd/libglvnd)
- [mesa](https://gitlab.freedesktop.org/mesa/mesa) (with libEGL and gbm support)
```

Atleast on Fedora I notice that the minimum required explicit packages would be [libglvnd-devel](https://koji.fedoraproject.org/koji/rpminfo?rpmID=37559750) and [mesa-libgbm-devel](https://koji.fedoraproject.org/koji/rpminfo?rpmID=38042568)

ammen99: looks great :)

ammen99 (approved):
Thanks!

PR #2237, opened 2024-03-22
Implement the session lock protocol, take 2

### LINKED ISSUE #1358: mako notifications cannot be made above fullscreen views but below lockscreens

### LINKED ISSUE #1494: Support ext-session-lock

lcolitti:
Re-applies #2162 (which was reverted) plus addresses more review comments.

Fixes #1494
Fixes #1358


plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+      private:
+        wlr_session_lock_surface_v1 *lock_surface;
+        wf::output_t *output;
+        std::unique_ptr<wf::keyboard_interaction_t> interaction;
+    };
+
+    class backup_node : public simple_text_node_t
+    {
+      public:
```
ammen99, line 115: This node doesn't have keyboard or pointer interaction set - but it should have it, otherwise, it is not really 'locking' the desktop.

Also on my machine there are a few thin semi-transparent strips along some of the emoji's sides, I am not sure why that happens - I'm mentioning this because through them, I could see that I am clicking on the terminal behind the crashed swaylock :)

ammen99 (commented):
I've tested this and it seems to work. There is one smaller issue regarding the backup node, otherwise LGTM.

Also please update the API/ABI version in `plugin.hpp` - it should be updated every time we change something in the API headers.

plugins/protocols/session-lock.cpp
```diff
@@ -0,0 +1,435 @@
...
+            {
+                wlr_seat_keyboard_notify_key(seat->seat, event.time_msec, event.keycode, event.state);
+            }
+
+          private:
+            wlr_surface *surface;
+            wlr_output *output;
+        };
+
+        wf::keyboard_interaction_t& keyboard_interaction()
```
ammen99, line 102: needs an `override` to prevent compiler warning

plugins/protocols/session-lock.cpp
lcolitti: Urg, yes. If I make the backup nodes transparent, it's pretty clear that they're not actually preventing keyboard or mouse input. Fixed in latest patch series (I think) by overriding find_node at. Not sure if this is correct or whether it will fix touch as well (I don't have a way to test touch).

ammen99: > Urg, yes. If I make the backup nodes transparent, it's pretty clear that they're not actually preventing keyboard or mouse input. Fixed in latest patch series (I think) by overriding find_node_at. Not sure if this is correct or whether it will fix touch as well (I don't have a way to test touch).

Yes, this is enough to fix touch, tablet and pointer. For the keyboard you'd also have to implement the refocus_keyboard() function very similarly to the lock surface node type.

ammen99: Ah, you already have this, nice! I'll test on multi-monitor as well to make sure the logic doesn't break somewhere and merge.

ammen99 (approved):
Looks great, I'll do one last round of testing and merge if nothing comes up. Thanks a lot!

ammen99:
@lcolitti Unfortunately I found one bug which seems quite important to fix, namely start Wayfire with two outputs, start swaylock, disconnect one of them => system is not responsive. Weirdly enough this does not happen if I use nested Wayfire outputs. So far I haven't been able to see what causes this though .. And it does not seem to happen when I use nested wayland backend outputs.

PR #2247, opened 2024-03-23
fix 1601

### LINKED ISSUE #1601: Setting keyboard layout to colemak does not remap keybindings
**Describe the bug**
After setting the xkb layout as follows:
```
xkb_layout = us
xkb_model = 
xkb_options = 
xkb_rules = evdev
xkb_variant = colemak
```

A key `KEY_*` will not be bound to the key in the Colemak layout, but in the QWERTY layout.

**To Reproduce**
Using the configuration above, bind any key (e.g. KEY_T). Using colemak, the binding will be placed on KEY_F (which corresponds to the same button in the QWERTY layout).

**Expected behavior**
I would expect all the keys to be bound with respect to the colemak layout, as opposed to the qwerty layout.

**Wayfire version**
0.7.4

ammen99:
This PR is an attempt at fixing #1601.

The solution is a bit more generic than that, though: we allow generic activators in the config file. Then, a plugin needs to parse them and activate them instead of core.

Fixes #1601

### DIFF
--- a/plugins/single_plugins/meson.build
+++ b/plugins/single_plugins/meson.build
@@ -2,7 +2,7 @@ plugins = [
   'move', 'resize', 'command', 'autostart', 'vswipe', 'wrot', 'expo',
   'switcher', 'fast-switcher', 'oswitch', 'place', 'invert',
   'fisheye', 'zoom', 'alpha', 'idle', 'extra-gestures', 'preserve-output',
-  'wsets', 'ipc-rules'
+  'wsets', 'ipc-rules', 'xkb-bindings'
 ]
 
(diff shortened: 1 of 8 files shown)


ammen99:
Example config:

```
[core]
plugins = xkb-bindings expo ...

[expo]
toggle = Mod4+e | <super> KEY_J | ...
```

In other words, the xkb-bindings plugin extends the available activator types with a new syntax which resembles Sway's and other compositor's a bit more closely. It also uses xkb to resolve keycodes to keysyms first, i.e the `e` in the example is not equivalent to `KEY_E`, but depends on how it is mapped exactly by the current layout. If you for example switch to a cyrillic layout (or anything else) then the binding stops working, because then `KEY_E` is mapped to a cyrillic letter which has a different keysym.

I'm open to feedback from the people who requested this feature, let me know whether this is what you had in mind.

PR #2250, opened 2024-03-23
transformer optimizations

### LINKED ISSUE #2184: Alpha plugin huge Vram usage
As time passes, the VRAM usage of Wayfire increases significantly, going from 320MB to 1GB and sometimes reaching 3GB depending on the amount of views. Disabling/enabling the plugin will reset the VRAM usage by around 320MB, as this action removes the alpha effect from every view.

I am wondering if you could optimize the plugin?

ammen99:
This PR aims to add more optimizations to Wayfire transformers and plugins, mostly aimed at reducing their VRAM usage and also some generic optimizations.

Fixes #2184

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,7 @@ jobs:
   test_musl_gcc:
     name: "Test with GCC/musl/libstdc++/BFD on Alpine Linux"
     runs-on: ubuntu-latest
-    container: alpine:edge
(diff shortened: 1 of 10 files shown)


ammen99:
This is a small breaking change for plugins which provide view transformers.

Transformers should now have `transformer_base_node_t` as a base class instead of simply` floating_inner_node_t`. The rest should work without any changes.

PR #2262, opened 2024-03-25
ipc: dynamically connect signals to core/outputs

CharlieQLe:
Connect signals to core and outputs if any clients exist, and disconnect them when all clients are disconnected.

This should ensure that no unnecessary creation of JSON data is done when no clients are available to receive such data.


killown:
hello charliQLe, I love your commits, always helping ipc, but I am wondering why not:

```
void send_event_to_subscribers(const nlohmann::json& data) {
    if (data.find("event_name") == data.end()) {
        return;
    }
    
    std::string event_name = data["event_name"];
    
    for (auto& [client, events] : clients) {
        if (events.empty() || events.count(event_name)) {
            client->send_json(data);
        }
    }
}
```

ammen99:
I was thinking of doing something a bit more finer-grained (i.e connect only some signals which are needed and not all) because for example you might have a few window-rules listening for view-mapped but nothing else. I will try to implement this, if I fail I will let you know and merge this PR :)

PR #2265, opened 2024-03-25
scale: auto-close scale when a new view opens

CharlieQLe:
I'll add a config option for this behavior before marking this PR as ready, but this seems to work well at the moment.

### DIFF
--- a/metadata/scale.xml
+++ b/metadata/scale.xml
@@ -37,6 +37,11 @@
 				<_long>Whether to show views that are minimized when starting scale.</_long>
 				<default>false</default>
 			</option>
+			<option name="close_on_new_view" type="bool">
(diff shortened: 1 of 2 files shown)


soreau:
> I'll add a config option for this behavior before marking this PR as ready, but this seems to work well at the moment.

Please make sure the behavior is the same as before this PR with the default option value. I don't think it's safe to assume this is expected/desired behavior for most users.

CharlieQLe:
Added an option to enable this behavior (it is disabled by default). Tested it with the option unset, set to true, and set to false and everything seems to work properly for all of them (new views only close scale when the option is set to true). It should be ready for a PR now!

ammen99 (approved):
LGTM, thanks :)

PR #2267, opened 2024-03-26
ipc-rules: add mechanism to connect signals only when a subscriber is really there

ammen99:
Alternative to #2262

@CharlieQLe @killown Please take a look and test.


ammen99:
> Doesn't seem to break anything, my custom bar runs perfectly fine and I have a Python script that prints out all the events as they come. Not too sure if output events are sent properly though, mostly because hotplugging one of my monitors seems to freeze everything. Not sure if this is a problem with Wayfire, Wlroots, Wayland, Linux, etc. or just something weird with my setup since this started happening back when I used Hyprland.

Thanks for testing! About the freeze, I would say a wlroots or kernel bug, since it happens in Hyprland too. I would recommend looking at Wayfire's logs (stdout) or dmesg, maybe you will see something related.

PR #2270, opened 2024-03-26
Update to wlroots 0.17.2

### LINKED ISSUE #2266: Update wlroots submodule to 0.17.2
This should be a drop in requiring zero changes, at least I'd hope it would be. This will make it possible for me to use the -git package again with labwc, since I need 0.17.2 for that to work cleanly.

(There's an issue with 0.17.1, now fixed by 0.17.2, that causes rapid fire mode changes on startup, such as when using kanshi, to cause a race condition, that causes the cursor plane surface to end up pointing at a NULL and crash the drivers on commit.)

marcusbritanicus:
This fixes #2266


ammen99 (approved):
Thanks :)

PR #2294, opened 2024-03-29
core: transfer views to same workspace as on destroyed output

lcolitti:
Currently, when an output is destroyed, views are transferred to the new output in roughly the same position as they were on the old output, but they are all transferred to the first workspace.

This is annoying: if they were not on the first workspace, then the user must have moved them, and after they're transferred, the user must will have to move them agein.

Instead, in the (common?) case that the old and the new output have the same number of workspaces, move each view to the same workspace it was on before.


soreau:
Is this not able to be fixed by only modifying the preserve-output plugin? Seems like that might be a better place for this to live.

lcolitti:
> Is this not able to be fixed by only modifying the preserve-output plugin? Seems like that might be a better place for this to live.

Perhaps, but why require the user to run the preserve output plugin to get the correct behaviour? Core obviously needs to do the right thing here in case the user is not running the plugin. And in fact the code does try to do the right thing, by putting the views in roughly the same place as they were on the previous output, scaled by the ratio of output sizes. This commit just makes it put the views on the right desktop as well. :smile:

ammen99 (changes requested):
This isn't ideal because move_view_to_output is used in other places as well, for example the oswitch plugin, and this will break the plugin.

I think the idea in general makes sense when destroying an output. It might make sense to add more flags to the function (instead of a bool). Maybe even add two overloads (to avoid breaking changes), one bool, one with flags, have one call the other :)

lcolitti:
> I think the idea in general makes sense when destroying an output. It might make sense to add more flags to the function (instead of a bool). Maybe even add two overloads (to avoid breaking changes), one bool, one with flags, have one call the other :)

Done

src/core/core.cpp
```diff
@@ -536,6 +540,11 @@ void wf::move_view_to_output(wayfire_toplevel_view v, wf::output_t *new_output,
         {
             auto new_g = wf::clamp(view_g, new_output->workarea->get_workarea());
             v->set_geometry(new_g);
+            if (same_workspace &&
```
ammen99, line 543: Wouldn't it also make sense to have similar treatment for fullscreen and tiled windows? You can pass the requested workspace to fullscreen/tile request too.

src/core/output-layout.cpp
```diff
@@ -546,7 +547,13 @@ struct output_layout_output_t
 
         /* It doesn't make sense to transfer to another output if we're
          * going to shut down the compositor */
-        transfer_views(wo, shutdown ? nullptr : get_core().seat->get_active_output());
+        wf::output_t *new_output = shutdown ? nullptr : get_core().seat->get_active_output();
+        transfer_views(wo, new_output);
+        if (new_output &&
```
ammen99, line 552: Ehh I'm not quite sure about this particular change. Is this really what users expect?

src/core/output-layout.cpp
```diff
@@ -1119,7 +1126,9 @@ class output_layout_t::impl
 
         if (!noop_output)
         {
-            auto handle = wlr_headless_add_output(noop_backend, 1280, 720);
+            // NOOP output should be at least as large as actual screen sizes. Otherwise, when
+            // when windows are temporarily mapped to it, they will be moved/cropped to match it.
+            auto handle = wlr_headless_add_output(noop_backend, 3840, 2160);
```
ammen99, line 1131: The noop output can be configured from the config file (IIRC), so this is only the default size if nothing else is set. I would rather not make it that big by default because if the user has some underpowered GPU mayybe this buffer would be too big?

lcolitti:
@ammen99 could you take a look? Or should I open another PR, since this one is marked stale?

src/core/output-layout.cpp
```diff
@@ -1626,8 +1626,22 @@ class output_layout_t::impl
...
-             * enabled again */
-            ensure_noop_output();
+             * enabled again.
+             *
+             * To ensure that when outputs are enabled again, windows are
+             * placed where they were before, the noop output must be large
+             * enough to hold the windows of the largest output that is
+             * being removed.
+             */
+            wf::dimensions_t max_size = {0, 0};
```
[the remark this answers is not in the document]
lcolitti: Yes, setting it in the config file does work. Personally I don't think this is a great solution - it will be very hard for users to understand why their windows are sometimes positioned correctly if their screen is <= 1280x720, but not if it's bigger - but at least it's possible to configure things to make it work.

I've removed this commit from the patch series for now.

Why are you concerned that the output might be too large? It's (roughly) only as large as the largest output that was already configured, so that memory was already being used, and it (roughly) only sticks around for as long as there are no other outputs - which (likely) means that nothing else is using that memory. Or is it because it could be larger, for example if you remove one output that is 128x3840 and another output that is 2160x128, the noop output becomes 3840x2160, which uses much more memory?

ammen99 (approved):
LGTM, thanks!

src/core/output-layout.cpp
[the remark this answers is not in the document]
lcolitti: This is a different use case from preserve-output: that saves window positions on *specific outputs*, and restores them if those exact outputs reappear. This PR was for switching outputs - for example, "user has windows on several workspaces on laptop screen, connects external monitor, and each window stays on the same workspaces where it was before".

Do you think this feature should be added to preserve-output? The current behaviour is enough for me because I know how to configure it :smile:, so I'm happy now that this PR is merged. I'm just thinking whether this should happen by default, or if there should be an easier way to configure it if other think this behaviour is better.

PR #2304, opened 2024-04-02
Session lock plugin fixes

lcolitti:
- Ensure that if an output is added when the screenlocker has crashed, the new output is locked.
- Support resizing outputs when locked.

### DIFF
--- a/plugins/protocols/session-lock.cpp
+++ b/plugins/protocols/session-lock.cpp
@@ -78,6 +78,12 @@ class lock_surface_node : public lock_base_node<wf::scene::wlr_surface_node_t>
         interaction(std::make_unique<lock_surface_keyboard_interaction>(lock_surface->surface))
     {}
 
+    void configure(wf::dimensions_t size)
+    {
+        wlr_session_lock_surface_v1_configure(lock_surface, size.width, size.height);
+        LOGC(LSHELL, "surface_configure on ", lock_surface->output->name, " ", size);
+    }
+
     void display()
     {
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
LGTM, we can merge after the CI passes.

lcolitti:
Yeah, somehow uncrustify never likes my formatting 🙂. I've added a git
hook that runs it before push so hopefully that won't happen again. Updated.

I considered adding the hook to the repo as well, so other folks can use
it, but the version I have on Debian disagrees with the GitHub version on
how to format one line in one file, and I didn't want to write a hook sync,
compile and run locally.

That said, if you think this could be useful, I can send a PR for it. My
hook is very simple and just ignores the file that doesn't match, and maybe
that's good enough for now.

ammen99 (approved):
Thanks!

lcolitti:
> We use a custom fork of uncrustify for formatting because the official one was kinda broken when I last used it. So I think it is not very ideal to make a git commit hook yet. I have considered moving to clang-format at some point, maybe if that happens a git hook would also be a good addition ..

Ack. FWIW the debian uncrustify only disagrees about this in the whole project:

-        if constexpr (std::is_base_of_v<wf::signal::provider_t, ObjectType>)
+        if constexpr (std::is_base_of_v<wf::signal::provider_t, ObjectType> )

ammen99:
> Ack. FWIW the debian uncrustify only disagrees about this in the whole project:

Interesting, maybe they have fixed the bugs I was experiencing, last time I tested was ~2 years ago or so.

PR #2329, opened 2024-04-11
Remove comment that says wf-dock started with autostart_wf_shell

### LINKED ISSUE #2327: wf-dock doesn't start with autostart_wf_shell (wayconfig.ini comments confirms it should)
**Describe the bug**
wf-dock does not start when autostart_wf_shell=true. wayfire.ini mentions:

```
# These are started by the autostart_wf_shell option above.
#
# background = wf-background
# panel = wf-panel
# dock = wf-dock
```

but wf-dock isn't actually started.

[autostart.cpp](https://github.com/WayfireWM/wayfire/blob/74ebf688ff506282a07f5abe424da9694948c830/plugins/single_plugins/autostart.cpp#L43C1-L51C10) also shows that panel and background are started, but not dock.

**To Reproduce**
1. Enable:
```
[autostart]
autostart_wf_shell = true
```
2. Start wayfire

**Expected behavior**
Dock should start 

**Wayfire version**
wayfire 0.8.1-4, Arch AUR

tyressk:
Quick PR to remove comment from waybar.ini that says wf-dock is started by autostart_wf_shell ([issue](https://github.com/WayfireWM/wayfire/issues/2327))

Fixes #2327

### DIFF
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -130,7 +130,6 @@ autostart_wf_shell = true
 #
 # background = wf-background
 # panel = wf-panel
-# dock = wf-dock
 
 # Output configuration
 # https://wayland.emersion.fr/kanshi/
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -130,6 +130,11 @@ autostart_wf_shell = true
 #
 # background = wf-background
 # panel = wf-panel
+#
+# You may also use wf-dock,
+# which is included in wf-shell but is not enabled by default.
+#
+# dock = wf-dock
 
 # Output configuration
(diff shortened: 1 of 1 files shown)


ammen99 (commented):
Maybe we can nonetheless leave a comment about wf-dock (otherwise I guess most people won't even notice its existence if they are looking for a basic dock)? But explicitly saying that it's not enabled by default.

tyressk:
@ammen99 I think that's a good idea. I added a comment in

ammen99 (approved):
Looks great now, thanks!

PR #2330, opened 2024-04-12
xwayland: add lazy start config option

dep4:
Add lazy start config option for XWayland which is built into wlroots, similar to sway.


ammen99 (approved):
LGTM, thanks!

ammen99:
Damn why does it change the commit author when I update your commit ..

PR #2331, opened 2024-04-13
send preferred_buffer_scale event after entering an output

### LINKED ISSUE #2253: Set fractional and preferred surface scale
wayland and wlroots now have an option to do that: `wlr_surface_set_preferred_buffer_scale` and `wlr_fractional_scale_v1_notify_scale`.

lilydjwg:
Qt 6.7.0+ requires this to correctly scale when fractional scale is not enabled.

A simple PyQt6 program to demostrate:

```py
#!/usr/bin/python3

import sys

from PyQt6 import QtWidgets, QtCore

def main():
  QtWidgets.QApplication.setHighDpiScaleFactorRoundingPolicy(
    QtCore.Qt.HighDpiScaleFactorRoundingPolicy.RoundPreferFloor)

  app = QtWidgets.QApplication(sys.argv)
  test = QtWidgets.QLabel("测试 Qt")
  test.show()
  sys.exit(app.exec())

if __name__ == '__main__':
  main()
```

Setting `QT_SCALE_FACTOR_ROUNDING_POLICY=RoundPreferFloor` for Qt6 apps should also reproduce it but not tested. This happens with Telegram (with "presice High DPI scaling" disabled) + Qt6 6.7.0.

Fixes #2253

### DIFF
--- a/src/view/wlr-surface-node.cpp
+++ b/src/view/wlr-surface-node.cpp
@@ -485,7 +485,6 @@ void wf::scene::wlr_surface_node_t::update_pending_outputs()
             if (surface)
             {
                 wlr_surface_send_enter(surface, wo->handle);
-                wlr_fractional_scale_v1_notify_scale(surface, wo->handle->scale);
             }
         } else if (delta < 0)
         {
@@ -506,6 +505,18 @@ void wf::scene::wlr_surface_node_t::update_pending_outputs()
                 visibility.erase(wo);
             }
         }
+    }
+
+    if (surface && (visibility.size() > 0))
+    {
+        float max_scale = 1;
+        for (auto x : visibility)
+        {
+            max_scale = std::max(max_scale, x.first->handle->scale);
+        }
+
+        wlr_fractional_scale_v1_notify_scale(surface, max_scale);
+        wlr_surface_set_preferred_buffer_scale(surface, max_scale);
     }
 
     pending_visibility_delta.clear();


ammen99 (commented):
In general maybe we should iterate over all the visible outputs and select the highest scale?

lilydjwg:
We may select the highest one among the outputs the surface in on.

src/view/wlr-surface-node.cpp
```diff
@@ -508,5 +510,16 @@ void wf::scene::wlr_surface_node_t::update_pending_outputs()
         }
     }
 
+    if (entered)
```
ammen99, line 513: We need to update not only when a new output is entered, but also when we send a leave event. It could be that the output with highest scale was removed and the rest remain visible (this could easily happen if you use the move plugin to drag a view from output with high to low scale factor).

src/view/wlr-surface-node.cpp
```diff
@@ -486,6 +487,7 @@ void wf::scene::wlr_surface_node_t::update_pending_outputs()
             {
                 wlr_surface_send_enter(surface, wo->handle);
                 wlr_fractional_scale_v1_notify_scale(surface, wo->handle->scale);
```
ammen99, line 489: Maybe we should move this below as well, so that we only notify the highest fractional scale?

ammen99 (changes requested):

ammen99 (approved):
LGTM now, thanks!

PR #2370, opened 2024-05-27
Provide "natural scrolling" option for mice.

mgriego:
Many mice support natural/reversed scrolling through libinput these days. This enables mouse natural scrolling throug a separate config option specifically for mice.

### DIFF
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -111,6 +111,11 @@
 				<default>1.0</default>
 				<min>0.0</min>
 			</option>
+			<option name="mouse_natural_scroll" type="bool">
+				<_short>Mouse natural scroll</_short>
+				<_long>Enables or disables mouse natural (inverted) scrolling.</_long>
+				<default>false</default>
+			</option>
 		</group>
 		<!-- Touchpad -->
 		<group>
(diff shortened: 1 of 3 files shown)


ammen99:
Thanks for the PR, I am however thinking whether it wouldn't be much simpler to simply allow negative mouse scroll speed, as we already have an option for that?

mgriego:
I would argue that using the natural scrolling option is more inline with what other graphical systems offer as an option and inline with what is expected for libinput.

marcusbritanicus:
Purely from a user point of view, especially for those migrating from other places (kde/gnome/.../mac/win), it would be nice if we have that check box. But I do not have any opinion on how it's implemented, though.

I do have a small nitpick in the xml file: Instead of `<_short>Mouse natural scroll</_short>`, it would be preferable to use `<_short>Natural scroll</_short>`, just as in touch pad configuration. In wcm, this option will appear in the "Mouse" tab, so it will not cause any confusion to the users.

mgriego:
@marcusbritanicus The scroll speed and cursor speed options are both prefixed with `Mouse`... ie: https://github.com/WayfireWM/wayfire/blob/master/metadata/input.xml#L226 . To me, this feels similar: it's a shared option that is scoped to mouse vs trackpad.

mgriego:
@ammen99 Any other concerns here?

ammen99 (approved):
LGTM, thanks!

PR #2385, opened 2024-06-19
Close animation duration config option

NamorNiradnug:
Opening and closing animations currently use the same `duration` option.


ammen99:
I am planning a redesign of how the animation options are handled (including making it easier to set per-view animation duration & animation type) so I am not sure I want to do a breaking change right now. I will leave the PR open until I finalize my thoughts on this matter, but I do not plan to merge before 0.9.

PR #2400, opened 2024-07-21
Set default decoration mode when it is adjusted at runtime

wb9688:
This calls `wlr_server_decoration_manager_set_default_mode` again when the setting `core/preferred_decoration_mode` in `wayfire.ini` is updated during runtime.

I figured out why it didn't work before and tried to implement that, but couldn't do that with valid C++ syntax due to my lack of C++ knowledge, so @soreau made the patch with proper C++ syntax, which is why I added him as co-author to the commit.

I tested this PR and it works completely as expected. Only with this PR, you will notice a bug in GTK, which I submitted a MR for at https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7475

### DIFF
--- a/src/view/xdg-shell/xdg-toplevel-view.cpp
+++ b/src/view/xdg-shell/xdg-toplevel-view.cpp
@@ -462,7 +462,7 @@ struct wf_xdg_decoration_t
 static wf::wl_listener_wrapper on_org_kde_decoration_created;
 static void init_legacy_decoration()
 {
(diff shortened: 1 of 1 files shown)


wb9688:
By the way it should be possible to make existing windows switch decorations as well, by sending them all a `mode` event, which a client can deny by sending a `request_mode` request back and the compositor then has to acknowledge that using a `mode` event, but I really doubt that there are many clients properly implementing that, so I haven't bothered for now.

ammen99 (approved):
LGTM, thanks!

PR #2401, opened 2024-07-21
Set initial decoration mode properly

wb9688:
This sets the initial decoration mode properly for clients that announce the `wl_surface` with the decoration protocol before the xdg-surface view, such as GTK 4, so that they are properly decorated.

### DIFF
--- a/src/view/xdg-shell/xdg-toplevel-view.cpp
+++ b/src/view/xdg-shell/xdg-toplevel-view.cpp
@@ -304,7 +304,7 @@ void wf::xdg_toplevel_view_t::map()
 {
     if (xdg_toplevel && uses_csd.count(xdg_toplevel->base->surface))
     {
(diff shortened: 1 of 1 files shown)


ammen99:
I think the real problem is that the following happens:

- start_map_tx is called
- decoration plugin sees that we'll map a view but it hasn't got its decorated status updated yet.
- we update decoration status in the actual map()

So the proper fix would be to have the old code (which just sets the decoration status without calling the function which also emits the decoration updated signal) in start_map_tx. Would you mind trying this out to see whether it works? Because with this patch, the flow of events is like this (and I think it is suboptimal):


- start_map_tx
- decoration plugin processes the view with the wrong state
- we start map()
- a signal is emitted and the decoration plugin now has to change the view's state in the next transaction

ammen99:
Closing as this seems pretty stale and also the wrong approach as outlined in my previous comment. Feel free to submit another PR with the updated approach.

PR #2404, opened 2024-07-24
move: Check if grid is enabled before activating edges on drag

### LINKED ISSUE #2403: [move] Do not show snap previews for halves and quarters of the screen when grid is disabled
**Describe the bug**
`move` without `grid` causes snapping to maximize instead of just not snapping at all.

**To Reproduce**
1. Enable plugin `move`, but not `grid`
2. Drag a window to the left
3. It shows the blue thingy to suggest snapping and when you drop it, the window maximizes

**Expected behavior**
When `grid` is not enabled, the blue thingy to suggest snapping should not appear and the window should definitely not be maximized.

**Wayfire version**
44e1fa9

soreau:
Before, dragging to an edge with move enabled and grid disabled would result in the indicator being drawn for the slot but always maximized. Fix this by emitting a signal that grid or another plugin can tell move the request is handled, so move can try gridding. In this implementation, we just hard code top to maximize if grid is disabled.

Fixes #2403.

### DIFF
--- a/plugins/grid/grid.cpp
+++ b/plugins/grid/grid.cpp
@@ -102,7 +102,15 @@ class wayfire_grid : public wf::plugin_interface_t, public wf::per_output_tracke
                 return false;
             });
         }
+
+        wf::get_core().connect(&grid_request_signal_cb);
     }
+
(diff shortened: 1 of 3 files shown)


ammen99 (approved):
LGTM, thanks!

PR #2408, opened 2024-07-25
animate: Add squeezimize animation

soreau:


soreau:
[squeezimize](https://www.youtube.com/watch?v=1QYC5Q0ihXM)

soreau:
It is noteworthy, that minimize animations in general require participation from the panel client, to tell the compositor where the minimize target is, within the panel. This way, the compositor knows where to animate the minimizing surface. Specifically, the panel client must call zwlr_foreign_toplevel_handle_v1_set_rectangle() to set the geometry for the surface within the parent panel window.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+  public:
+    using duration_t::duration_t;
+    timed_transition_t squeeze{*this};
+};
+class squeezimize_transformer : public wf::scene::view_2d_transformer_t
+{
+  public:
+    wayfire_view view;
+    OpenGL::program_t program;
+    wf::output_t *output;
```
ammen99, line 134: I think we can skip storing the output altogether, since it looks like we can compute `upward` at the beginning of the animation, right?

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+        program.compile(squeeze_vert_source, squeeze_frag_source);
+        OpenGL::render_end();
+    }
+
+    wf::geometry_t get_bounding_box() override
+    {
+        return this->animation_geometry;
+    }
+
+    wf::effect_hook_t pre_hook = [=] ()
```
ammen99, line 288: Is there a reason why this prehook is needed? The animation plugin is written in such a way that every animation gets this hook automatically set and managed by the main plugin: https://github.com/WayfireWM/wayfire/blob/0daf5c9a1f52383192d53bb20ad0d770581413f9/plugins/animate/animate.cpp#L86

In particular, this `animation_hook` automatically damages the view and calls the `step` method, so I think there should be no need at all for a pre-hook like this.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+                minimize_target.width),
+                (minimize_target.x + minimize_target.width) - bbox.x),
+                (bbox.x + bbox.width) - minimize_target.x);
+        animation_geometry.height =
+            std::max(std::max(std::max(bbox.height,
+                minimize_target.height),
+                (minimize_target.y + minimize_target.height) - bbox.y),
+                (bbox.y + bbox.height) - minimize_target.y);
+        OpenGL::render_begin();
+        program.compile(squeeze_vert_source, squeeze_frag_source);
```
ammen99, line 279: I wonder whether this is an expensive operation. Maybe we can store the program as custom data on core, so that we don't recompile all of the time?

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+        auto hint = toplevel->get_minimize_hint();
+        auto tmgr = view->get_transformed_node();
+        auto node = std::make_shared<wf::squeezimize::squeezimize_transformer>(view, hint, bbox);
+        tmgr->add_transformer(node, wf::TRANSFORMER_HIGHLEVEL + 1, sqeezimize_transformer_name);
+        node->init_animation(type & HIDING_ANIMATION);
+    }
+
+    void pop_transformer(wayfire_view view)
+    {
+        if (view->get_transformed_node()->get_transformer(sqeezimize_transformer_name))
```
ammen99, line 344: This check is not necessary, because rem_transformer is safe to call even if the transformer is not there.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+        {
+            view->get_transformed_node()->rem_transformer(sqeezimize_transformer_name);
+        }
+    }
+
+    bool step() override
+    {
+        if (!view)
```
ammen99, line 352: Is it possible that view becomes null? It seems to me that this shouldn't be necessary.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+        {
+            return false;
+        }
+
+        auto tmgr = view->get_transformed_node();
+        if (!tmgr)
```
ammen99, line 358: transformed_node is always valid (i.e non-nullptr) as long as the view is valid.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+            self->program.use(wf::TEXTURE_TYPE_RGBA);
+            self->program.uniformMatrix4f("matrix", target.get_orthographic_projection());
+            self->program.attrib_pointer("position", 2, 0, vertex_data_pos);
+            self->program.attrib_pointer("uv_in", 2, 0, vertex_data_uv);
+            self->program.uniform1i("upward", upward);
+            self->program.uniform1f("progress", progress);
+            self->program.uniform4f("src_box", src_box_pos);
+            self->program.uniform4f("target_box", target_box_pos);
+            self->program.set_active_texture(src_tex);
+            GL_CALL(glDrawArrays(GL_TRIANGLE_FAN, 0, 4));
```
ammen99, line 250: Seems like this is lacking scissoring to the damaged region. It probably is not noticeable when the animated view is on top, but it may be causing issues if for example the animated view ends up behind another opaque view which is blocking part of it.

ammen99: Simplest way to do it would be just iterating over the damage rects and calling glDrawArrays multiple times (core does this).

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+    class simple_node_render_instance_t : public wf::scene::transformer_render_instance_t<transformer_base_node_t>
+    {
+        wf::signal::connection_t<node_damage_signal> on_node_damaged =
+            [=] (node_damage_signal *ev)
+        {
+            push_to_parent(ev->region);
+        };
+
+        squeezimize_transformer *self;
+        wayfire_view view;
```
ammen99, line 148: Seems like `view` is not used anywhere at all.

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+{
+  public:
+    wayfire_view view;
+    OpenGL::program_t program;
+    wf::output_t *output;
+    wf::geometry_t minimize_target;
+    wf::geometry_t animation_geometry;
+    squeezimize_animation_t progression{squeezimize_duration};
+
+    class simple_node_render_instance_t : public wf::scene::transformer_render_instance_t<transformer_base_node_t>
```
ammen99, line 139: You can specify the actual transformer type here `wf::scene::transformer_render_instance_t<squeezimize_transformer>` as a base class. Then you can remove `squeezimize_transformer *self` that you have a few lines below, because the base class will already store this pointer for you: https://github.com/WayfireWM/wayfire/blob/0daf5c9a1f52383192d53bb20ad0d770581413f9/src/api/wayfire/view-transform.hpp#L148

plugins/animate/squeezimize.hpp
```diff
@@ -0,0 +1,388 @@
...
+            self->program.set_active_texture(src_tex);
+            GL_CALL(glDrawArrays(GL_TRIANGLE_FAN, 0, 4));
+            OpenGL::render_end();
+        }
+    };
+
+    squeezimize_transformer(wayfire_view view,
+        wf::geometry_t minimize_target, wf::geometry_t bbox) : wf::scene::view_2d_transformer_t(view)
+    {
+        this->view = view;
```
ammen99, line 258: I believe with the other suggested changes, this->view will not be needed.

ammen99 (changes requested):
Haven't looked into the details of how the animation works, I trust your videos which show how cool it is ;) So I have focused my comments on suggestions about the general Wayfire animation/transformer API. 

Most probably there will be similar comments in your other animation PRs, so I will not review them until we're finished with this first one.

plugins/animate/squeezimize.hpp
soreau: The output is for damaging in the pre_hook.

plugins/animate/squeezimize.hpp
soreau: It is needed because the animation geometry is larger than the view; it renders outside of the view's area. Thus, it must be damaged. You can see with the showrepaint plugin, and obviously, the animation renders outside of the view damage area.

plugins/animate/squeezimize.hpp
soreau: No, I actually checked this (by asking in #dri-devel). As long as shader cache is enabled, it shouldn't recompile, because it stores a hash of the shader source and returns the same program if it's in the cache.

plugins/animate/squeezimize.hpp
soreau: Well, that's something that could be changed throughout wayfire, but noted.

plugins/animate/squeezimize.hpp
soreau: Probably not necessary.

plugins/animate/squeezimize.hpp
soreau: Ok.

plugins/animate/squeezimize.hpp
soreau: Yes, sounds like it should work, so long as the pre_hook is in place to damage the full region.

plugins/animate/squeezimize.hpp
soreau: Looks like you are correct.

plugins/animate/squeezimize.hpp
soreau: Interesting, thanks for the tip.

plugins/animate/squeezimize.hpp
soreau: Indeed.

soreau: Actually it is needed, for in gen_render_instances() when creating the simple_node_render_instance_t object instance.

plugins/animate/squeezimize.hpp
soreau: Turns out this isn't needed because it's handled in transform_damage_region(). Thanks for pointing this out.

plugins/animate/squeezimize.hpp
soreau: On second take, I managed to not need to store it.

metadata/animate.xml
```diff
@@ -116,5 +133,10 @@
 			<_long>Sets the color of the fire effects, alpha is ignored</_long>
 			<default>#b22303ff</default>
 		</option>
+		<option name="squeezimize_duration" type="animation">
+			<_short>Squeezimize duration</_short>
+			<_long>Sets the duration of the squeezimize animation in milliseconds.</_long>
+			<default>1000ms linear</default>
```
ammen99, line 139: I think this is way too slow, I would personally set it to 150ms :)

ammen99 (commented):
LGTM now, tested it and it works :)

One question which remains is what to do if the user minimizes without a minimize hint. The zoom animation will still animate the view, but as if it were closing. Wonder whether we can set a default minimize hint to be something like one of the corners of the screen?

metadata/animate.xml
soreau: Sure, I have no objection to tweaking the defaults.

soreau:
> LGTM now, tested it and it works :)
> 
> One question which remains is what to do if the user minimizes without a minimize hint. The zoom animation will still animate the view, but as if it were closing. Wonder whether we can set a default minimize hint to be something like one of the corners of the screen?

With waybar, apparently not setting the minimize hint results in the view just disappearing during the animation. I was thinking, if we get 0's for the hint or it's otherwise not set, we could try to target minimizing to mouse pointer. Compiz did it, but I'm not sure the wayfire code would work for this.

ammen99:
> > LGTM now, tested it and it works :)
> > One question which remains is what to do if the user minimizes without a minimize hint. The zoom animation will still animate the view, but as if it were closing. Wonder whether we can set a default minimize hint to be something like one of the corners of the screen?
> 
> With waybar, apparently not setting the minimize hint results in the view just disappearing during the animation. I was thinking, if we get 0's for the hint or it's otherwise not set, we could try to target minimizing to mouse pointer. Compiz did it, but I'm not sure the wayfire code would work for this.

Following the pointer could be nice, but if it doesn't work well enough, we can just pick a random target .. That will be better than the view just disappearing, at least the user has a visual indication of what happened.

marcusbritanicus:
Why not do what the default minimize animation did? If the minimize hint is not set, then the view would minimize to the center of the screen.

soreau:
> Why not do what the default minimize animation did? If the minimize hint is not set, then the view would minimize to the center of the screen.

Sounds like it might work.

soreau:
> LGTM now, tested it and it works :)
> 
> One question which remains is what to do if the user minimizes without a minimize hint. The zoom animation will still animate the view, but as if it were closing. Wonder whether we can set a default minimize hint to be something like one of the corners of the screen?

I tinkered with this and it seems the 'best' option was to minimize to the bottom center of the output.

ammen99 (approved):
LGTM now, thanks!

PR #2410, opened 2024-07-27
workarounds: Clean up metadata

soreau:
This makes the long options appear in tooltips and adds a short field for the option name, in wcm.


ammen99 (approved):
LGTM, thanks!

PR #2412, opened 2024-07-28
animate: Add spin animation

soreau:


plugins/animate/spin.hpp
```diff
@@ -0,0 +1,100 @@
...
+            tr, wf::TRANSFORMER_HIGHLEVEL, spin_transformer_name);
+    }
+
+    bool step() override
+    {
+        auto transform = view->get_transformed_node()
+            ->get_transformer<wf::scene::view_2d_transformer_t>(spin_transformer_name);
+        auto progress = this->progression.progress();
+        transform->alpha   = progress;
+        transform->angle   = progress * M_PI * 4.0;
```
ammen99, line 77: Idea: let the user configure the `4.0` constant, that way they can configure how often the window will rotate before closing/opening completely.

ammen99 (approved):
Just as the other animation, looks simple but works well! We can merge after you rebase. I have added one suggestion, feel free to ignore it if you don't like it.

plugins/animate/spin.hpp
soreau: Already way ahead of you, the [wayfire-animations branch](https://github.com/soreau/wayfire/blob/wayfire-animations/plugins/animate/spin.hpp#L82-L83) has a version which provides the feature you request.

ammen99: great :)

soreau:
> Just as the other animation, looks simple but works well! We can merge after you rebase. I have added one suggestion, feel free to ignore it if you don't like it.

It's a good suggestion, I've implemented it. 👍

ammen99 (approved):
thanks :)

PR #2417, opened 2024-07-30
animate: Add helix animation

soreau:
Add helix animation.

### DIFF
--- a/plugins/animate/helix.hpp
+++ b/plugins/animate/helix.hpp
@@ -39,7 +39,7 @@
 #include "animate.hpp"
 
 
-static std::string transformer_name = "animation-helix";
+static std::string helix_transformer_name = "animation-helix";
 
 wf::option_wrapper_t<wf::animation_description_t> helix_duration{"animate/helix_duration"};
 
@@ -310,15 +310,15 @@ class helix_animation : public animation_base
         auto bbox = view->get_transformed_node()->get_bounding_box();
         auto tmgr = view->get_transformed_node();
         auto node = std::make_shared<wf::helix::helix_transformer>(view, bbox);
-        tmgr->add_transformer(node, wf::TRANSFORMER_HIGHLEVEL + 1, transformer_name);
+        tmgr->add_transformer(node, wf::TRANSFORMER_HIGHLEVEL + 1, helix_transformer_name);
         node->init_animation(type & HIDING_ANIMATION);
     }
 
     void pop_transformer(wayfire_view view)
     {
-        if (view->get_transformed_node()->get_transformer(transformer_name))
+        if (view->get_transformed_node()->get_transformer(helix_transformer_name))
         {
-            view->get_transformed_node()->rem_transformer(transformer_name);
+            view->get_transformed_node()->rem_transformer(helix_transformer_name);
         }
     }
 
@@ -335,7 +335,7 @@ class helix_animation : public animation_base
             return false;
         }
 
-        if (auto tr = tmgr->get_transformer<wf::helix::helix_transformer>(transformer_name))
+        if (auto tr = tmgr->get_transformer<wf::helix::helix_transformer>(helix_transformer_name))
         {
             auto running = tr->progression.running();
             if (!running)
@@ -354,7 +354,7 @@ class helix_animation : public animation_base
     {
         if (auto tr =
                 view->get_transformed_node()->get_transformer<wf::helix::helix_transformer>(
-                    transformer_name))
+                    helix_transformer_name))
         {
             tr->progression.reverse();
         }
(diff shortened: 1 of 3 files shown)


plugins/animate/helix.hpp
```diff
@@ -0,0 +1,329 @@
...
+
+    class simple_node_render_instance_t : public wf::scene::transformer_render_instance_t<helix_transformer>
+    {
+        wf::signal::connection_t<node_damage_signal> on_node_damaged =
+            [=] (node_damage_signal *ev)
+        {
+            push_to_parent(ev->region);
+        };
+
+        wayfire_view view;
```
ammen99, line 106: as far as I see this variable is not used anywhere.

plugins/animate/helix.hpp
```diff
@@ -0,0 +1,329 @@
...
+{
+static std::string helix_transformer_name = "animation-helix";
+using namespace wf::scene;
+using namespace wf::animation;
+class helix_animation_t : public duration_t
+{
+  public:
+    using duration_t::duration_t;
+};
+class helix_transformer : public wf::scene::view_2d_transformer_t
```
ammen99, line 91: Is there a reason for deriving from `view_2d_transformer_t` instead of `transformer_base_node_t` directly?

plugins/animate/helix.hpp
```diff
@@ -0,0 +1,329 @@
...
+                        .damage   = damage & self->animation_geometry,
+                    });
+        }
+
+        void transform_damage_region(wf::region_t& damage) override
+        {
+            damage |= wf::region_t{self->animation_geometry};
+        }
+
+        void render(const wf::render_target_t& target,
```
ammen99, line 138: Looking at this, I have an idea on how to make this less CPU-intensive, because I am getting 50%+ cpu usage even on 1080p (the effect will be much more pronounced on 4k displays since the algorithm is linear in height):

Compute a vertex buffer with strips spanning a simple [0,1]x[0,1] rectangle (the rectangle can be easily scaled to match the view size with a matrix multiplication in the vertex shader).

Based on the actual view height at runtime, we select how many strips to render.

Rotation is also doable in the vertex shader - for example by using something like `floor(gl_VertexID / 6.0` you can get the strip index, and from the strip index you can calculate the rotation for the strip. 

Does that sound about right? Maybe I misunderstand the implementation, but feels like the computation-intensive part of this effect can thus be offloaded completely to the GPU.

ammen99 (commented):
The effect works quite well but I think it uses lots of CPU and I also think we can make it work almost entirely on the GPU .. let me know what you think about that.

soreau:
> The effect works quite well but I think it uses lots of CPU and I also think we can make it work almost entirely on the GPU .. let me know what you think about that.

My initial thought is that it would be a lot of work for relatively little payoff.

> The effect works quite well but I think it uses lots of CPU and I also think we can make it work almost entirely on the GPU .. let me know what you think about that.

> Looking at this, I have an idea on how to make this less CPU-intensive, because I am getting 50%+ cpu usage even on 1080p (the effect will be much more pronounced on 4k displays since the algorithm is linear in height):
> 
> Compute a vertex buffer with strips spanning a simple [0,1]x[0,1] rectangle (the rectangle can be easily scaled to match the view size with a matrix multiplication in the vertex shader).
> 
> Based on the actual view height at runtime, we select how many strips to render.
> 
> Rotation is also doable in the vertex shader - for example by using something like `floor(gl_VertexID / 6.0` you can get the strip index, and from the strip index you can calculate the rotation for the strip.
> 
> Does that sound about right? Maybe I misunderstand the implementation, but feels like the computation-intensive part of this effect can thus be offloaded completely to the GPU.

Maybe, but I'll leave you to experiment with this idea.

soreau:
> The effect works quite well but I think it uses lots of CPU and I also think we can make it work almost entirely on the GPU .. let me know what you think about that.

You previously mentioned that helix takes about 50% cpu in your tests. I did my own tests, of each animation, running wayfire with nothing more than what is required to do the tests. Here are the results:

```
Duration: 5000ms
No animation (idle): 2% cpu
Fade:        14% - 2% = 12%
Zoom:        14% - 2% = 12%
Fire:        42% - 2% = 40%
Zap:         14% - 2% = 12%
Spin:        14% - 2% = 12%
Helix:       17% - 2% = 15%
Blinds:      24% - 2% = 22%
Shatter:     40% - 2% = 38%
Vortex:      15% - 2% = 13%
Squeezimize: 18% - 2% = 16%
```

It seems fire is the worst, at 40%, followed by shatter at 38% and blinds at 22%. Helix is nearly on par with the built-in transformer animations at 15%. Please let me know what you think in light of this.

soreau:
Closing this because this animation is now part of wayfire-plugins-extra 'extra-animations' plugin.

PR #2418, opened 2024-07-31
animate: Add blinds animation

soreau:
Add blinds animation.


soreau:
Closing this because this animation is now part of wayfire-plugins-extra 'extra-animations' plugin.

PR #2429, opened 2024-08-11
transactions: Use correct geometry when adjusting for gravity

soreau:
It is apparent from the comment following this line, that this was the intention. This fixes a decoration sizing bug with ignore_views matching 'maximized is true'. This can be seen when toggling maximized for apps such as wcm.

### DIFF
--- a/src/view/xdg-shell/xdg-toplevel.cpp
+++ b/src/view/xdg-shell/xdg-toplevel.cpp
@@ -201,7 +201,7 @@ void wf::xdg_toplevel_t::handle_surface_commit()
         return;
     }
 
-    adjust_geometry_for_gravity(_pending, toplevel_size);
+    adjust_geometry_for_gravity(_pending, wf::dimensions(pending().geometry));
     LOGC(VIEWS, "Client-initiated resize to geometry ", pending().geometry);
     auto tx = wf::txn::transaction_t::create();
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
LGTM, thanks!

ammen99 (changes requested):
Actually, just before I merge this, I figured out that this might not be entirely correct. adjust_geometry_for_gravity changes _pending.geometry to match toplevel_size. Why do you think that this change is needed?

soreau:
> Actually, just before I merge this, I figured out that this might not be entirely correct. adjust_geometry_for_gravity changes _pending.geometry to match toplevel_size. Why do you think that this change is needed?

I have explained in the initial comment. Is it not clear how to reproduce the problem? Do you need further explanation or a video?

soreau:
> > Actually, just before I merge this, I figured out that this might not be entirely correct. adjust_geometry_for_gravity changes _pending.geometry to match toplevel_size. Why do you think that this change is needed?
> 
> I have explained in the initial comment. Is it not clear how to reproduce the problem? Do you need further explanation or a video?

On second thought, I tested this against latest pixdecor and it does not fix the bug, where the size becomes larger, after toggling maximize with `ignore_views = maximized is true`. I have to look more into it again.

PR #2432, opened 2024-08-15
Add wlroots to DesktopNames

Vladimir-csp:
For convenience of making units or autostart entries common for all wlroots compositors


ammen99 (approved):
Alright, thanks!

PR #2434, opened 2024-08-17
resize: Respect client minimum and maximum sizes

soreau:
This is needed so decorators render correctly in these cases.

### DIFF
--- a/plugins/single_plugins/resize.cpp
+++ b/plugins/single_plugins/resize.cpp
@@ -59,6 +59,7 @@ class wayfire_resize : public wf::per_output_plugin_instance_t, public wf::point
     bool preserve_aspect = false;
     wf::point_t grab_start;
     wf::geometry_t grabbed_geometry;
+    wf::dimensions_t min_size, max_size;
 
     uint32_t edges;
     wf::option_wrapper_t<wf::buttonbinding_t> button{"resize/activate"};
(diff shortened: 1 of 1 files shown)


plugins/single_plugins/resize.cpp
```diff
@@ -349,9 +352,51 @@ class wayfire_resize : public wf::per_output_plugin_instance_t, public wf::point
...
+            desired.width =
+                std::max(min_size.width +
+                    (view->toplevel()->pending().margins.left + view->toplevel()->pending().margins.right),
+                    desired.width);
+        }
+
+        if (max_size.width > 0)
+        {
+            desired.width =
+                std::min(max_size.width -
```
ammen99, line 368: Why the subtraction here? max_size is what the client reports, but our full geometry can grow to client_max_size + margins from the decoration.

soreau: Can you make a suggestion? I don't know of a client that is resizable and sets a max size to test with. Are you saying it should be addition instead of subtraction, like the minimum cases?

ammen99: I mean that the reported max size comes from the max size provided by the client and applies to the main surface. However, when we set the geometry, we set the geometry together with the decorations. So if a client for example says 'max width is 100' and we have a decoration margins 5px on each size, the actual 'max' size is 110.

i'll push a fix shortly.

PR #2455, opened 2024-08-27
compilation times improvement

ammen99:
This PR aims to improve compilation times of Wayfire's core and plugins.

A benchmark on my own laptop (which uses AMD Ryzen 5600U) with 6C12T:

- Before: 1m1s, slowest file: scale.cpp which takes ~9s.
- After: 24s, slowest file: tile-plugin.cpp which takes ~6s.

This is achieved through a combination of techniques:

- Headers are shuffled around so that we include less header files overall.
- We pre-declare a few very frequently used templates (option templates).
- We split some of the header-only helpers (workspace-wall, move-drag-helper) to a separate cpp file and link the resulting library with the plugins that need it.
- We use a custom PCH solution for plugins (manually generating it because meson is not smart enough yet, see https://github.com/mesonbuild/meson/issues/4350)
- We switch away from nlohmann/json and into jsoncpp (an unfortunate breaking change).

TODOs:

- [ ] Benchmark the impact of jsoncpp vs nlohmann/json
- [ ] Look into whether the plugin PCH cannot be easily solved on the meson side, upstream seems to not be against solving it properly, but I have no knowledge of meson's internals.
- [ ] Clean up the [DONOTMERGE] commits.
- [ ] Fix build order (wobbly vs plugins/common)


ammen99:
Closing this in favor of #2581 #2591 #2592 #2604

PR #2456, opened 2024-08-28
render-manager: Don't call paint when the session is not active

soreau:
When running wayfire on one tty and having clients that damage constantly, switching to another tty causes wayfire to consume 100% cpu. This patch disconnects the paint handler when the session is not active and reconnects it when the session becomes active again.

### DIFF
--- a/src/view/wlr-surface-node.cpp
+++ b/src/view/wlr-surface-node.cpp
@@ -205,7 +205,7 @@ wf::touch_interaction_t& wf::scene::wlr_surface_node_t::touch_interaction()
 
 void wf::scene::wlr_surface_node_t::send_frame_done(bool delay_until_vblank)
 {
-    if (!surface)
+    if (!wf::get_core().session->active || !surface)
     {
         return;
     }
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -396,6 +396,11 @@ struct swapchain_damage_manager_t
      */
     void schedule_repaint()
     {
+        if (!wf::get_core().session->active)
+        {
+            return;
+        }
+
         wlr_output_schedule_frame(output);
         force_next_frame = true;
     }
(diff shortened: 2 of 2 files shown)


src/view/wlr-surface-node.cpp
```diff
@@ -205,7 +205,7 @@ wf::touch_interaction_t& wf::scene::wlr_surface_node_t::touch_interaction()
 
 void wf::scene::wlr_surface_node_t::send_frame_done(bool delay_until_vblank)
 {
-    if (!surface)
+    if (!wf::get_core().session->active || !surface)
```
ammen99, line 208: I don't think this is the correct place for such a check. Frame done events are sent in some specific edge-cases (pending transactions, surface not visible on any output but trying to get mapped, etc.) and when the output emits a `frame_done_signal`. In my estimation, the rest of your changes should be enough to fix the issue, if not, then probably we need a guard around the emission of `frame_done_signal` in render-manager.cpp

soreau:
For some reason, I've found that when starting from a tty 'manually', wayfire does not have this problem, even without this patch. However, if I start wayfire with lightdm, the problem happens only after the second switch to tty. Apparently, the session->active variable is not toggled, so it's unreliable anyway.

ammen99:
> For some reason, I've found that when starting from a tty 'manually', wayfire does not have this problem, even without this patch. However, if I start wayfire with lightdm, the problem happens only after the second switch to tty. Apparently, the session->active variable is not toggled, so it's unreliable anyway.

Well if this fixes some cases I'd say it is still a win :)

soreau:
This was completely fixed by an update to seatd, as far as I can tell. I can not reproduce the issue as of now. Previously, the version was seatd 0.8.0 3e9ef69f1. I do not think this patch is necessary at all now, closing.

PR #2459, opened 2024-08-29
ipc: Use AF_INET to accept remote connections

soreau:
Locally, this should change nothing, with the corresponding pywayfire patch. Remotely, you can set WAYFIRE_IPC_IP nd WAYFIRE_IPC_PORT to connect with a pywayfire script.


soreau:
This requires [pywayfire #5](https://github.com/WayfireWM/pywayfire/pull/5).

killown:
Here is how we could manage remote connections without #2459 

https://github.com/killown/pywayfire_examples/tree/main/examples/remote-ipc

ammen99:
NACK, no plans to support networking here.

PR #2473, opened 2024-09-26
squeezimize: Set output pointer and avoid shadowing

kode54:
This fixes a crash related to minimizing with squeezimize using multiple outputs.


ammen99 (changes requested):
When the view is moved to a different output (or when its output is destroyed), we are supposed to remove the whole transformer, see animate.cpp.

Therefore this is just a workaround and not a proper fix for the issue, sorry.

PR #2479, opened 2024-10-02
Input method: update input method popup positions on text input commit

lilydjwg:
A text input `commit` may come with a `set_cursor_rectangle` request, and we need to update the input method popup to follow the text cursor on the client side.

Previously we run the update on popup surface commit, but events may come out of order, i.e. the input method draws a new surface buffer as soon as it commits the input string, and then the client window will take some time to process the committed string and figure out the new position of cursor.

### DIFF
--- a/src/api/wayfire/unstable/wlr-text-input-v3-popup.hpp
+++ b/src/api/wayfire/unstable/wlr-text-input-v3-popup.hpp
@@ -7,7 +7,7 @@
 
 namespace wf
 {
-class text_input_v3_im_relay_interface_t
+class text_input_v3_im_relay_interface_t : public wf::signal::provider_t
 {
   public:
     virtual wlr_text_input_v3 *find_focused_text_input_v3() = 0;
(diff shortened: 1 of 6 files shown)


src/api/wayfire/unstable/wlr-text-input-v3-popup.hpp
```diff
@@ -7,11 +7,20 @@
 
 namespace wf
 {
+class text_input_v3_popup;
+
 class text_input_v3_im_relay_interface_t
 {
+  protected:
```
ammen99, line 14: This seems like the wrong place to store information about active popups.

Note that popups already store a pointer to the input method relay. So I think here we can use a signal (make the relay derive from signal::provider_t) and then emit a signal on commit. Any popups currently active can connect to this signal and update their internal state, and we don't have to maintain a list of popups.

lilydjwg: Signal is added. That list of popups needs to stay there for input method v2 because otherwise the popup would be destroyed as nothing references it.

ammen99 (approved):
LGTM now, thanks!

PR #2484, opened 2024-10-03
render-manager: Don't paint when session is not active

soreau:
This can happen when switching to another tty. A test is to run glxgears in a drm session with xwayland, then switch to another tty and run top. The wayfire cpu usage is 100+%. However with this patch, when the session is inactive, the cpu usage is negligible.

One note is that patches were pushed to seatd recently which make the session->active state reliable. At the time of this writing, seatd does not have a release containing these patches, so for this to work reliably, seatd git master is recommended.


ammen99 (approved):
lgtm, thanks!

PR #2485, opened 2024-10-04
core: Add legacy wayland drm option

soreau:
This can be used to work around broken drivers that accelerate the desktop but not the wayland clients.


ammen99 (approved):
Thanks!

PR #2490, opened 2024-10-07
Make socket path unique

marcusbritanicus:
The solution proposed by @killown to make the socket path unique. I have tested this and confirm that it works.

### DIFF
--- a/plugins/ipc/ipc.cpp
+++ b/plugins/ipc/ipc.cpp
@@ -327,6 +327,7 @@ class ipc_plugin_t : public wf::plugin_interface_t
         char *pre_socket  = getenv("_WAYFIRE_SOCKET");
         const auto& dname = wf::get_core().wayland_display;
         pid_t pid = getpid();
+        const char *runtime_dir = getenv("XDG_RUNTIME_DIR");
 
         std::string socket;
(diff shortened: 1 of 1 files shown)


ammen99:
If we're going to do this, move the socket to XDG_RUNTIME_DIR :)

killown:
> If we're going to do this, move the socket to XDG_RUNTIME_DIR :)

```
void init() override
{
    char *pre_socket = getenv("_WAYFIRE_SOCKET");
    const char *runtime_dir = getenv("XDG_RUNTIME_DIR");
    const auto& dname = wf::get_core().wayland_display;
    pid_t pid = getpid();

    std::string socket;
    if (pre_socket)
    {
        socket = pre_socket;
    }
    else if (runtime_dir)
    {
        socket = std::string(runtime_dir) + "/wayfire-" + dname + "-" + std::to_string(pid) + ".socket";
    }
    else
    {
        socket = "/tmp/wayfire-" + dname + "-" + std::to_string(pid) + ".socket";
    }

    setenv("WAYFIRE_SOCKET", socket.c_str(), 1);
    server->init(socket);
}

ammen99 (approved):
Thanks!

Issue #2360, opened 2024-05-16
hyprlock doesn' work - fails to bind to ext-session-lock-v1


d3-X-t3r:
**Describe the bug**
[hyprlock](https://github.com/hyprwm/hyprlock) fails to launch under Wayfire (latest git build), giving an error `[CRITICAL] Couldn't bind to ext-session-lock-v1, does your compositor support it?` 

I believe `ext-session-lock-v1` _was_ implemented as part of #2237 ?

**To Reproduce**
1. Launch hyprlock
2. Check stdout/stderr

**Expected behavior**
Hyprlock should work.

**Screenshots or stacktrace**
```
$ hyprlock
[LOG]   | got iface: wl_shm v1
[LOG]   | got iface: wl_drm v2
[LOG]   | got iface: zwp_linux_dmabuf_v1 v4
[LOG]    > Bound to zwp_linux_dmabuf_v1 v4
[LOG]   | got iface: wl_compositor v6
[LOG]    > Bound to wl_compositor v6
[LOG]   | got iface: wl_subcompositor v1
[LOG]   | got iface: wl_data_device_manager v3
[LOG]   | got iface: zwp_primary_selection_device_manager_v1 v1
[LOG]   | got iface: zwlr_data_control_manager_v1 v2
[LOG]   | got iface: zwlr_output_manager_v1 v4
[LOG]   | got iface: zwlr_output_power_manager_v1 v1
[LOG]   | got iface: xdg_wm_base v3
[LOG]   | got iface: zwlr_layer_shell_v1 v4
[LOG]   | got iface: xwayland_shell_v1 v1
[LOG]   | got iface: zwp_tablet_manager_v2 v1
[LOG]   | got iface: wl_seat v8
[LOG]    > Bound to wl_seat v8
[LOG]   | got iface: zwlr_screencopy_manager_v1 v3
[LOG]    > Bound to zwlr_screencopy_manager_v1 v3
[LOG]   | got iface: zwlr_gamma_control_manager_v1 v1
[LOG]   | got iface: zwlr_export_dmabuf_manager_v1 v1
[LOG]   | got iface: zxdg_output_manager_v1 v3
[LOG]   | got iface: wp_drm_lease_device_v1 v1
[LOG]   | got iface: zwlr_input_inhibit_manager_v1 v1
[LOG]   | got iface: ext_idle_notifier_v1 v1
[LOG]   | got iface: zwp_idle_inhibit_manager_v1 v1
[...]

PR #2500, opened 2024-10-25
Enable session-lock plugin by default

SvitlanochkaIsTheBest:
### What is this change:

Adding the `session-lock` plugin as a default one to the [wayfire.ini](https://github.com/WayfireWM/wayfire/blob/master/wayfire.ini) and the [core.xml](https://github.com/WayfireWM/wayfire/blob/master/metadata/core.xml) config.

Done as part of the #2360 discussion.

### How it was validated:

Built Wayfire with changes from this PR on a VM. Got successful locking results using both [swaylock](https://github.com/swaywm/swaylock) and [waylock](https://github.com/ifreund/waylock) tools.

Please, let me know if additional validation results are needed.

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -7,7 +7,7 @@
 		<option name="plugins" type="string">
 			<_short>Plugins</_short>
 			<_long>Loads the specified plugins, space-separated list.</_long>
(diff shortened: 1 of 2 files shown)


kode54:
LGTM, though it looks like maintainers need to approve the build tests for them to run

soreau:
Thanks, LGTM too. 👍

ammen99 (approved):
Thanks :)

PR #2508, opened 2024-11-06
move: Use post input signal for touch move grabs

### LINKED ISSUE #2505: Moving window using touchscreen window title drag causes jump of window; using mouse doesn't do it
**Description**

I have one touchscreen monitor connected (the monitor provides an HDMI connection and a USB connection).
First time using this on wayfire.

Now, when I try to drag a window (eg. firefox) from its titlebar using the touchscreen:
- when the motion is initiated, the window jumps in the screen (it should stay in place)
- when the motion continues, the relative motion follows the finger motion, but the absolute position is wrong due to the initial offset

** Repro steps **

0. Desktop computer, connect one screen (HDMI), connect one regular mouse, no keyboard ; no input or output configuration in wayfire config
1. Start wayfire
2. Start firefox remotely, and move it from title bar with mouse -> OK
3. Kill wayfire, swap mouse with touchscreen USB
4. Start wayfire
5. Start firefox remotely, and move it with touchscreen -> NG

**Wayfire version**
- Wayfire master branch or track-wlroots dev branch
- X11 disabled

**Note**: Will clarify description

soreau:
Without this, touch fingers are not set yet which causes a visual jump when attempting to move drag a window with touchscreen devices. Use post signal so that the fingers are properly detected and the correct coordinates are used in the first place.

Fixes #2505.

### DIFF
--- a/plugins/single_plugins/move.cpp
+++ b/plugins/single_plugins/move.cpp
@@ -144,8 +144,8 @@ class wayfire_move : public wf::per_output_plugin_instance_t,
         }
     };
 
-    wf::signal::connection_t<wf::input_event_signal<wlr_touch_down_event>> on_raw_touch_down =
-        [=] (wf::input_event_signal<wlr_touch_down_event> *ev)
+    wf::signal::connection_t<wf::post_input_event_signal<wlr_touch_down_event>> on_raw_touch_down =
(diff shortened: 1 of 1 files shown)


zougloub:
Tested here, OK.

ammen99 (approved):
Makes sense, thanks!

PR #2509, opened 2024-11-06
window-manager: Ignore fullscreen requests for current state

### LINKED ISSUE #2507: Window move of gnome nautilus using mouse drag on title causes jump of window
Maybe same root cause as #2505 ?

https://github.com/user-attachments/assets/74c8dff7-59ca-47ba-9ab2-f89b7154eec4

As part of troubleshooting an issue, we found out that when I try to drag a gnome window (eg. nautilus, gnome-calculator) from its titlebar using the mouse:

- when the motion is initiated, the window jumps in the screen (it should stay in place)
- when the motion continues, the relative motion follows the mouse cursor motion, but the absolute position is wrong due to the initial offset

** Repro steps **

0. Desktop computer, connect one screen (HDMI), connect one regular mouse, no keyboard
1. Configuration, reduced down to:
```
[autostart]
autostart_wf_shell = false
firefox = firefox
nautilus = nautilus

[core]
plugins = autostart move gtk-shell foreign-toplevel wayfire-shell
```
2. Start wayfire ; firefox and nautilus start
3. Move firefox from title bar with mouse -> :ok: 
4. Move nautilus in the same way -> :ng: 

Wayfire version: wayfire master branch ; X11 disabled

soreau:
This fixes dragging gtk4 apps with grid and/or wobbly disabled. The gtk4 toolkit sends a request for unfullscreen even though it is windowed, and this caused wayfire to set the geometry when dragging by titlebar, which made the window jump to an offset when grabbing the titlebar to move it.

Fixes #2507.

### DIFF
--- a/src/core/window-manager.cpp
+++ b/src/core/window-manager.cpp
@@ -244,6 +244,11 @@ void window_manager_t::tile_request(wayfire_toplevel_view view,
 void window_manager_t::fullscreen_request(wayfire_toplevel_view view,
     wf::output_t *output, bool state, std::optional<wf::point_t> ws)
 {
+    if (view->toplevel()->pending().fullscreen == state)
+    {
+        return;
+    }
+
     wf::output_t *wo = output ?: (view->get_output() ?: wf::get_core().seat->get_active_output());
     const wf::point_t workspace = ws.value_or(wo->wset()->get_current_workspace());
     wf::dassert(wo != nullptr, "Fullscreening should not happen with null output!");


zougloub:
Tested here, OK.

ammen99 (approved):
LGTM, thanks :)

mrsteve0924:
i was seeing this issue with pavucontrol and this commit fixed the window jumpiness. 

but now i get some weird behavior from the volume widget on wf-panel. the pop up volume slider appears then disappears with each mousescroll with each mouse scroll.  then somehow fixes itself but then doesn't disappear unless i click off somewhere.

ammen99:
> but now i get some weird behavior from the volume widget on wf-panel. the pop up volume slider appears then disappears with each mousescroll with each mouse scroll. then somehow fixes itself but then doesn't disappear unless i click off somewhere.

This is a separate issue, please report it separately, not on unrelated PRs ;)

mrsteve0924:
well it looks like the culprit is #2479 

ill submit it

Issue #2481, opened 2024-10-03
Add per-device keyboard layout support


fmleo:
Hi! My current setup has 2 keyboards, an external one which I use at home and the laptop one which is used mostly everywhere else.
My suggestion comes from my previous experience with other wlroots compositors:

sway
```
input type:keyboard {
    xkb_layout br
}

input "6700:39936:SEMICO___USB_Gaming_Keyboard" {
   xkb_layout us
   xkb_options 'compose:ralt'
}
```

hyprland
```
device:at-translated-set-2-keyboard {
    kb_layout = br
}

device:6700:39936:SEMICO___USB_Gaming_Keyboard {
   kb_layout = us
   kb_options = 'compose:ralt'
}
```

As seen in the examples above, in both compositors you can set keyboard options based on the device names returned by `swaymsg -t get_inputs` and the hyprland equivalent.

#2419 could be relevant to this issue.

zougloub:
The input config is documented to be in `[input]` (`input/...`)  however `get_input_device_section`  which can get per-device config, is not invoked in the `wf::keyboard_t::keyboard_t` ctor. Is it possible that calling it  then making the option wrappers use `section->get_option(...)` instead of `config.get_option("input/...")`would solve the problem?

soreau:
Preliminary [patch](https://termbin.com/honk) for per-keyboard layout. @zougloub has confirmed that it works.

PR #2510, opened 2024-11-06
Input device updates

soreau:
- Adds the ability to configure input devices that might have matching names
- Refactor so hotplugging a device only refreshes the configuration for that device, not all devices
- Add calibration option for touch devices
- Add a way to configure multiple keyboards and mice with different settings

### DIFF
--- a/src/api/wayfire/config-backend.hpp
+++ b/src/api/wayfire/config-backend.hpp
@@ -45,8 +45,8 @@ class config_backend_t
     /**
      * Find the output section for a given input device.
      *
-     * The returned section must be a valid output object as
-     * described in input-device.xml
+     * The returned section must be a valid config object as described
+     * in <@prefix>.xml (typically input.xml or input-device.xml).
      */
     virtual std::shared_ptr<config::section_t> get_input_device_section(
         std::string const & prefix, wlr_input_device *device);
--- a/src/api/wayfire/config-backend.hpp
+++ b/src/api/wayfire/config-backend.hpp
@@ -49,7 +49,7 @@ class config_backend_t
      * described in input-device.xml
      */
     virtual std::shared_ptr<config::section_t> get_input_device_section(
-        wlr_input_device *device);
+        std::string const & prefix, wlr_input_device *device);
 
     virtual ~config_backend_t() = default;
 
--- a/src/core/plugin.cpp
+++ b/src/core/plugin.cpp
@@ -44,7 +44,7 @@ static struct udev_property_and_desc
 };
 
 std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_section(
-    wlr_input_device *device)
+    std::string const & prefix, wlr_input_device *device)
 {
     auto& config = wf::get_core().config;
     std::shared_ptr<wf::config::section_t> section;
@@ -65,7 +65,7 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_sectio
                         continue;
                     }
 
-                    std::string name = std::string("input-device:") + nonull(value);
+                    std::string name = prefix + ":" + nonull(value);
                     LOGC(INPUT_DEVICES, "Checking for config section [", name, "] ",
                         pd.property_name, " (", pd.description, ")");
                     section = config.get_section(name);
@@ -80,7 +80,7 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_sectio
     }
 
     std::string name = nonull(device->name);
-    name = "input-device:" + name;
+    name = prefix + ":" + name;
     LOGC(INPUT_DEVICES, "Checking for config section [", name, "]");
     section = config.get_section(name);
     if (section)
@@ -90,7 +90,7 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_sectio
     }
 
     config.merge_section(
-        config.get_section("input-device")->clone_with_name(name));
+        config.get_section(prefix)->clone_with_name(name));
 
     LOGC(INPUT_DEVICES, "Using config section [", name, "]");
     return config.get_section(name);
--- a/src/core/plugin.cpp
+++ b/src/core/plugin.cpp
@@ -5,6 +5,7 @@
 #include "wayfire/signal-definitions.hpp"
 #include <wayfire/util/log.hpp>
 #include <wayfire/config-backend.hpp>
+#include <libudev.h>
 
 void wf::plugin_interface_t::fini()
 {}
@@ -27,18 +28,71 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_output_section(
     return config.get_section(name);
 }
 
+static struct udev_property_and_desc
+{
+    char const *property_name;
+    char const *description;
+} properties_and_descs[] =
+{
+    {"ID_PATH", "stable physical connection path"},
+    {"ID_SERIAL", "stable vendor+pn+sn info"},
+    {"LIBINPUT_DEVICE_GROUP", "stable libinput info"},
+    // sometimes it contains info "by path", sometimes "by id"
+    {"DEVPATH", "unstable devpath"},
+    // used for debugging, to find DEVPATH and match the right
+    // device in `udevadm info --tree`
+};
+
 std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_section(
     wlr_input_device *device)
 {
-    std::string name = nonull(device->name);
-    name = "input-device:" + name;
     auto& config = wf::get_core().config;
-    if (!config.get_section(name))
+    std::shared_ptr<wf::config::section_t> section;
+
+    if (wlr_input_device_is_libinput(device))
     {
-        config.merge_section(
-            config.get_section("input-device")->clone_with_name(name));
+        auto libinput_dev = wlr_libinput_get_device_handle(device);
+        if (libinput_dev)
+        {
+            udev_device *udev_dev = libinput_device_get_udev_device(libinput_dev);
+            if (udev_dev)
+            {
+                for (struct udev_property_and_desc const & pd : properties_and_descs)
+                {
+                    const char *value = udev_device_get_property_value(udev_dev, pd.property_name);
+                    if (value == nullptr)
+                    {
+                        continue;
+                    }
+
+                    std::string name = std::string("input-device:") + nonull(value);
+                    LOGC(INPUT_DEVICES, "Checking for config section [", name, "] ",
+                        pd.property_name, " (", pd.description, ")");
+                    section = config.get_section(name);
+                    if (section)
+                    {
+                        LOGC(INPUT_DEVICES, "Using config section [", name, "] for ", nonull(device->name));
+                        return section;
+                    }
+                }
+            }
+        }
     }
 
+    std::string name = nonull(device->name);
+    name = "input-device:" + name;
+    LOGC(INPUT_DEVICES, "Checking for config section [", name, "]");
+    section = config.get_section(name);
+    if (section)
+    {
+        LOGC(INPUT_DEVICES, "Using config section [", name, "]");
+        return section;
+    }
+
+    config.merge_section(
+        config.get_section("input-device")->clone_with_name(name));
+
+    LOGC(INPUT_DEVICES, "Using config section [", name, "]");
     return config.get_section(name);
 }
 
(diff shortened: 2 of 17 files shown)


src/core/plugin.cpp
```diff
@@ -27,18 +28,76 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_output_section(
...
+    // used for debugging, to find DEVPATH and match the right
+    // device in `udevadm info --tree`
+};
+
 std::shared_ptr<config::section_t> wf::config_backend_t::get_input_device_section(
     wlr_input_device *device)
 {
+    auto& config = wf::get_core().config;
+    std::shared_ptr<wf::config::section_t> section;
+    auto print_devpath = getenv("WF_PRINT_UDEV_DEVPATH");
```
ammen99, line 51: I think what you are looking for here are the debug categories, they are rather simple to add, maybe an `input-device` category?

Check these here: https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/debug.hpp#L42

src/core/plugin.cpp
```diff
@@ -27,18 +28,76 @@ std::shared_ptr<config::section_t> wf::config_backend_t::get_output_section(
...
+    {
+        auto libinput_dev = wlr_libinput_get_device_handle(device);
+        if (libinput_dev)
+        {
+            udev_device *udev_dev = libinput_device_get_udev_device(libinput_dev);
+            if (udev_dev)
+            {
+                for (struct udev_property_and_desc const & pd : properties_and_descs)
+                {
+                    if (!print_devpath && !strncmp(pd.property_name, "DEVPATH", strlen("DEVPATH")))
```
ammen99, line 63: Actually, wouldn't it be more appropriate to send this info over via IPC and print it in the scripts we already have for listing input devices? I can imagine that this would be much easier to use.

Otherwise, I think we can/should just print every property we are querying here (in the corresponding debug category).

src/core/seat/input-manager.hpp
```diff
@@ -49,6 +49,12 @@ class input_manager_t
      */
     void configure_input_devices();
 
+    /**
+     * Calibrate a touch device with a matrix. This function does nothing
+     * if called with a device that is not a touch device.
+     */
+    void calibrate_touch_device(wlr_input_device *dev, std::string const & cal);
```
ammen99, line 56: This doesn't seem like it should be in input-manager ... Because this is device-type-specific (mapping inputs to outputs works for tablets, touch and pointer). Ideally speaking, we'd want to add an input-device class for touch devices and load their specific options there, just like we do for pointers: https://github.com/WayfireWM/wayfire/blob/master/src/core/seat/pointing-device.hpp

And then we can create this device-specific object here: https://github.com/WayfireWM/wayfire/blob/56532427bea34ed9d777708b52356220c366de73/src/core/seat/input-manager.cpp#L39

At the very least, it seems like this should be somewhere in the input device class.

src/api/wayfire/config-backend.hpp
```diff
@@ -49,7 +49,7 @@ class config_backend_t
      * described in input-device.xml
      */
     virtual std::shared_ptr<config::section_t> get_input_device_section(
-        wlr_input_device *device);
+        std::string const & prefix, wlr_input_device *device);
```
ammen99, line 52: needs API/ABI version bump :)

src/core/seat/keyboard.cpp
```diff
@@ -121,14 +121,18 @@ void wf::keyboard_t::setup_listeners()
...
 {
-    model.load_option("input/xkb_model");
-    variant.load_option("input/xkb_variant");
-    layout.load_option("input/xkb_layout");
-    options.load_option("input/xkb_options");
-    rules.load_option("input/xkb_rules");
-
-    repeat_rate.load_option("input/kb_repeat_rate");
-    repeat_delay.load_option("input/kb_repeat_delay");
+    auto section =
```
ammen99, line 124: Seems like we'd want the same for pointer? https://github.com/WayfireWM/wayfire/blob/56532427bea34ed9d777708b52356220c366de73/src/core/seat/pointing-device.cpp#L12-L31

ammen99 (commented):
+10 for the idea. I was thinking of whether we can make this work fully dynamically (currently if you want to add a section, you have to restart Wayfire) but I don't think this can be done easily.

soreau:
> +10 for the idea. I was thinking of whether we can make this work fully dynamically (currently if you want to add a section, you have to restart Wayfire) but I don't think this can be done easily.

Thanks for the review! However, I don't think it is true that you need to restart wayfire, but only hot(re)plug the input device in question. Of course this might not be completely optimal, but a full restart is not needed as far as I understand.

src/core/plugin.cpp
soreau: I disagree that we should require ipc for this task. It should be as easy as `wayfire -d input-device` (or plural, input-devices).

src/core/seat/input-manager.hpp
soreau: Hm, ok..

src/api/wayfire/config-backend.hpp
soreau: Ah yes, I forgot to bump! :)

src/core/seat/keyboard.cpp
soreau: I was thinking what other bits we might want this for.. adding for pointer makes sense to me, will update.

src/core/seat/input-manager.cpp
```diff
@@ -52,10 +52,52 @@ void wf::input_manager_t::handle_new_input(wlr_input_device *dev)
...
+    auto cal = section->get_option("calibration")->get_value_str();
+    if (!cal.empty())
+    {
+        device->calibrate_touch_device(cal);
+    }
+
+    auto wo = wf::get_core().output_layout->find_output(mapped_output);
+    if (wo)
+    {
+        LOGD("Mapping input ", dev->name, " to output ", wo->to_string(), ".");
```
ammen99, line 91: should this be LOGC(INPUT_DEVICE)? we can leave it like this if you think it is better

ammen99 (approved):
LGTM, I'll test to see whether this doesn't break my setup and merge if it doesn't (shouldn't by the looks of it). Feel free to update the wiki :)

Btw. has someone on matrix/irc tested this with multiple devices etc? I am fairly certain we can close some issues with this PR.

src/core/seat/input-manager.cpp
soreau: I think leaving it this way makes it easier to see without the additional noise of `-d input-devices` and it is the same as it was before, so users can still use `-d` alone to see this particular output logged.

soreau:
> Btw. has someone on matrix/irc tested this with multiple devices etc? I am fairly certain we can close some issues with this PR.

I am not sure but @zougloub might be interested in testing.

ammen99:
I think this fixes at least #2481 right?

soreau:
> I think this fixes at least #2481 right?

Yes, it should.

src/core/seat/pointer.cpp
```diff
@@ -355,9 +355,26 @@ void wf::pointer_t::handle_pointer_axis(wlr_pointer_axis_event *ev,
     }
 
     /* Calculate speed settings */
-    double mult = ev->source == WLR_AXIS_SOURCE_FINGER ?
-        wf::pointing_device_t::config.touchpad_scroll_speed :
-        wf::pointing_device_t::config.mouse_scroll_speed;
+    bool touchpad = ev->source == WLR_AXIS_SOURCE_FINGER;
+    double mult   = 1.0;
+    for (const auto& device : wf::get_core_impl().input->input_devices)
```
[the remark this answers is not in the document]
soreau: I have force-pushed the top commit to implement this. Can you please review and test?

soreau:
@zougloub If you want to test this before it's merged, now would be a good time. It should be 'ready'. 👍

ammen99 (changes requested):
Looking at this again, some more open questions:

- We have the input-device.xml section. As far as I see, this will not be used anymore, as we are allowing per-device `[input:]` section. I think that therefore the option for mapping should be moved from `input-device` to `input` and input-device.xml should be removed as we dont use it.

- Previous point raises the question of backward compatibility, maybe we can check whether [input-device] sections still exist and if yes, then merge them into the main device section and print a warning to the user?

- The input section is no longer a `plugin` in the XML, but an `object` (there is no practical difference, but a convention that objects are instantiated with different suffixes).

src/api/wayfire/input-device.hpp
```diff
@@ -21,6 +21,12 @@ class input_device_t
      */
     bool set_enabled(bool enabled = true);
 
+    /**
+     * Calibrate a touch device with a matrix. This function does nothing
+     * if called with a device that is not a touch device.
+     */
+    void calibrate_touch_device(std::string const & cal);
+
```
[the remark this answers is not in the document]
ammen99: Am I seeing it wrongly or the conversation was resolved without fixing the actual issue?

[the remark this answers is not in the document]
soreau: I did make a change, but either I didn't commit the correct files or I read `input_device_impl_t` as `input_device_t`. But it seems like it's not resolved, so apparently I clicked the button without double checking.

src/api/wayfire/config-backend.hpp
ammen99: I think you must have forgotten to push a commit upstream, because this also hasn't been resolved ...

soreau:
>     * The input section should no longer be a `plugin` in the XML, but an `object` (there is no practical difference, but a convention that objects are instantiated with different suffixes).

I will just say that wcm doesn't support 'objects' so switching to this from 'plugin' will remove 'input' from wcm altogether (and I'm guessing there's some that might not appreciate this).

ammen99 (approved):
LGTM, thanks!

lilydjwg:
My device is called `TPPS/2 IBM TrackPoint` and I have the following config:

```
[input:TPPS/2 IBM TrackPoint]
mouse_cursor_speed = 0.2
```

It crashes:

> [build/wayfire/src/debug.cpp:442] Wayfire encountered error loading option "input:TPPS/2 IBM TrackPoint/left_handed_mode": No such option: input:TPPS/2 IBM TrackPoint/left_handed_mode. Usual reasons for this include missing or outdated plugin XML files, a bug in the plugin itself, or mismatch between the versions of Wayfire and wf-config. Make sure that you have the correct versions of all relevant packages and make sure that there are no conflicting installations of Wayfire using the same prefix. If this plugin was recently installed or updated, you need to restart Wayfire before using it.

Do I need to escape that `/` somehow?

soreau:
> My device is called `TPPS/2 IBM TrackPoint` and I have the following config:
> 
> ```
> [input:TPPS/2 IBM TrackPoint]
> mouse_cursor_speed = 0.2
> ```
> 
> It crashes:
> ...
> Do I need to escape that `/` somehow?

I would try `wayfire -d input-devices` and review the log to see what other identifying matches it's checking for that you can use.

ammen99:
> My device is called `TPPS/2 IBM TrackPoint` and I have the following config:
> 
> ```
> [input:TPPS/2 IBM TrackPoint]
> mouse_cursor_speed = 0.2
> ```
> 
> It crashes:
> 
> > [build/wayfire/src/debug.cpp:442] Wayfire encountered error loading option "input:TPPS/2 IBM TrackPoint/left_handed_mode": No such option: input:TPPS/2 IBM TrackPoint/left_handed_mode. Usual reasons for this include missing or outdated plugin XML files, a bug in the plugin itself, or mismatch between the versions of Wayfire and wf-config. Make sure that you have the correct versions of all relevant packages and make sure that there are no conflicting installations of Wayfire using the same prefix. If this plugin was recently installed or updated, you need to restart Wayfire before using it.
> 
> Do I need to escape that `/` somehow?

Oh ... / is not supported in an option or a section name ... This is a wf-config bug. Maybe we can work around for cases like these by splitting at the last / or by allowing escape sequences in section names.

lilydjwg:
Thanks. I'll use another name for now. Of the five strings, only two of them don't contain a `/`; one of them doesn't seem distinguishable, and I hope that the other won't change after a reboot.

```
platform-i8042-serio-2
noserial
11/2/a:synaptics-pt/serio0
/devices/platform/i8042/serio1/serio2/input/input31/event21
TPPS/2 IBM TrackPoint
```

Escaping is a better way to go I guess, because it's more versatile.

lilydjwg:
> I hope that the other won't change after a reboot

It has changed after I reloaded the kernel module, so no reliable workaround is possible for my case.

PR #2517, opened 2024-11-14
touch: make EDGE_SWIPE_THRESHOLD configurable

arpio23:
On mobile device with High-DPI screens and different scaling it would make sense to make that value configurable.

### DIFF
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -227,6 +227,13 @@
 				<min>0.1</min>
 				<max>5.0</max>
 			</option>
+			<option name="edge_swipe_threshold" type="int">
+				<_short>Touchscreen edge swipe threshold</_short>
+				<_long>Change the threshold of the built-in edge swipe detection. Higher values mean that a bigger area on the edge is used to trigger the edge swipe.</_long>
+				<default>10</default>
+				<min>5</min>
+				<max>50</max>
+			</option>
 			<option name="touchpad_scroll_speed" type="double">
 				<_short>Touchpad scroll speed</_short>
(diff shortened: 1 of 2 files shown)


metadata/input.xml
```diff
@@ -227,6 +227,13 @@
 				<min>0.1</min>
 				<max>5.0</max>
 			</option>
+			<option name="edge_swipe_threshold" type="int">
+				<_short>Touchscreen edge swipe threshold</_short>
+				<_long>Change the threshold of the built-in edge swipe detection. Higher values mean that a bigger area on the edge is used to trigger the edge swipe.</_long>
+				<default>10</default>
+				<min>5</min>
```
ammen99, line 234: I would make this 1 or so .. let the users have more freedom :)

src/core/seat/touch.cpp
```diff
@@ -427,24 +426,25 @@ static uint32_t find_swipe_edges(wf::touch::point_t point)
 {
     auto output   = wf::get_core().seat->get_active_output();
     auto geometry = output->get_layout_geometry();
+    wf::option_wrapper_t<int> edge_swipe_threshold{"input/edge_swipe_threshold"};
```
ammen99, line 429: This will cause a lookup for the option every time the function is called. I would make it a `static wf::option_wrapper_t<...>` so that it loads only the first time the function is called :)

ammen99 (changes requested):
The idea LGTM, two small things to fix and I'll merge :)

src/core/seat/touch.cpp
arpio23: Thanks, will fix up after new year.

PR #2519, opened 2024-11-20
Wayfire.ini

mrsteve0924:
added dbus-update-activation-environment --systemd WAYLAND_DISPLAY DISPLAY XAUTHORITY to [autostart]

I think most people are using GTK apps and this command is almost mandatory.
Commenting it out is fine.  As long as it is included in the file so people are aware of it.  Right now it's buried in the wiki tips & tricks. 

Including it here should help people get their install on track faster.


ammen99 (approved):
I guess this is a good point, thanks!

PR #2522, opened 2024-11-22
Edge swipe extra

arpio23:
The extra edge-swipe types are configurable and it should not break any existing configs. If edge_swipe_section_length is set to 0.0 (default) everything works as before.
Related wf-config pr https://github.com/WayfireWM/wf-config/pull/73

### DIFF
--- a/metadata/input.xml
+++ b/metadata/input.xml
@@ -227,6 +227,13 @@
 				<min>0.1</min>
 				<max>5.0</max>
 			</option>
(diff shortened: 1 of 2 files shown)


ammen99 (approved):
LGTM. Please update the wf-config submodule version as well, so that the CI passes (this will only be possible after I merge the wf-config PR which I hope to do in the next few minutes).

PR #2527, opened 2024-11-29
xdg-activation: add options to make it more strict

dkondor:
This adds three options to configure xdg-activation and allow more strict behavior. These include:
 - reject requests without a view supplied (disabled by default)
 - only allow activating with the last issued token (disabled by default)
 - set the timeout used by wlroots (default: 30s, which is also used by wlroots)

Currently, wlroots already rejects token creation if the client supplied a wl_surface which is not focused. However, if there is no surface, creating a token always succeeds, basically allowing any app to focus itself (e.g. using `gtk_window_present()` which falls back to self-generated tokens with no surface if none is available in the environment). This can be a good thing e.g. when interacting with already running apps from the terminal, but can lead to unexpected focus changes as well. With this PR, this behavior can be changed according to individual preferences.

A small test program is here: https://github.com/dkondor/xdg_activation_test -- it is a case of a parent passing tokens to a child process. Without this PR, all three cases in it always work. With the new, restrictive options enabled, only the first case works.

Note that this still does not do full focus stealing prevention, for that, we would need to keep track if the surface that requested the token loses focus. I would consider looking into that next if there is interest.

If this gets merged, I'm happy to add some notes to the Wiki on the options (and also on xdg-activation more generally).

### DIFF
--- a/plugins/protocols/xdg-activation.cpp
+++ b/plugins/protocols/xdg-activation.cpp
@@ -12,16 +12,29 @@
 class wayfire_xdg_activation_protocol_impl : public wf::plugin_interface_t
 {
   public:
+    wayfire_xdg_activation_protocol_impl()
+    {
+        set_callbacks();
+    }
+
     void init() override
     {
         xdg_activation = wlr_xdg_activation_v1_create(wf::get_core().display);
-        xdg_activation_request_activate.notify = xdg_activation_handle_request_activate;
+        if (timeout >= 0)
+        {
+            xdg_activation->token_timeout_msec = 1000 * timeout;
+        }
 
-        wl_signal_add(&xdg_activation->events.request_activate, &xdg_activation_request_activate);
+        xdg_activation_request_activate.connect(&xdg_activation->events.request_activate);
+        xdg_activation_new_token.connect(&xdg_activation->events.new_token);
     }
 
     void fini() override
-    {}
+    {
+        xdg_activation_request_activate.disconnect();
+        xdg_activation_new_token.disconnect();
+        last_token = nullptr;
+    }
 
     bool is_unloadable() override
     {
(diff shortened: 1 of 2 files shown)


plugins/protocols/xdg-activation.cpp
```diff
@@ -12,53 +12,135 @@
...
+            {
+                // note: for a valid surface, wlroots already checks that this is the active surface
+                LOGI("Not registering activation token, surface was not supplied");
+                token->seat  = nullptr; // this will ensure that this token will be rejected later
+                reject_token = true;
+            }
+
+            if (reject_token)
+            {
+                if (token == last_token)
```
ammen99, line 104: While this technically works, why not simply use the token's `destroy` event? This here (+leaving the pointer dangling) seems like quite the hack.

dkondor: I think I had a reason for it, but cannot remember what it was :)
Although I originally did something more complicated and might be a leftover from that. I'll try again and see.

dkondor: OK, I tried and it seems to work well with the destroy signal as well

ammen99 (commented):
Overall lgtm, I have two small questions, we can merge after they are resolved.

plugins/protocols/xdg-activation.cpp
```diff
@@ -12,53 +12,131 @@
...
+            xdg_activation_token_destroy.connect(&token->events.destroy);
+        });
+
+        xdg_activation_token_destroy.set_callback([this] (void *data)
         {
-            LOGI("Denying focus request, seat wasn't supplied");
-            return;
-        }
+            auto token = static_cast<struct wlr_xdg_activation_token_v1*>(data);
+            if (token == last_token)
```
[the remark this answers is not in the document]
dkondor: Actually, it should not matter, probably I added it as extra paranoia or as a remnant of a previous attempt where  I kept this signal connected for multiple tokens (technically, it can be false if we already set `last_token = nullptr` on line 64, but that does not matter). I'll remove the check to make things more clear.

plugins/protocols/xdg-activation.cpp
```diff
@@ -12,53 +12,131 @@
...
-        auto toplevel = wf::toplevel_cast(view);
-        if (!toplevel)
+            auto toplevel = wf::toplevel_cast(view);
+            if (!toplevel)
+            {
+                LOGE("Could not get toplevel view");
+                return;
+            }
+
+            LOGI("Activating view");
```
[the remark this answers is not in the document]
dkondor: Sure, I think LOGD is fine

dkondor:
I've added a commit to address the comments -- do you need me to squash commits and / or rebase on main given the recent changes?

dkondor:
OK, so the build failure looks weird and it seems related to the changes to master -- I'll rebase and try to fix them

ammen99 (approved):
LGTM now, thanks!

Issue #2422, opened 2024-08-04
layer-shell view steals the focus from scale plugin


dkondor:
**Describe the bug**
If scale is started when a layer-shell surface has the keyboard focus, it cannot grab the focus for itself.

**To Reproduce**
1. Enable the scale plugin
2. Ensure that a layer-shell surface has the keyboard focus
3. Start the scale plugin (keybinding or IPC works)

**Expected behavior**
Scale has keyboard interaction (e.g. arrows, ESC).

**Actual behavior**
Scale starts, but the layer-shell surface continues to have the keyboard focus. Using the mouse to interact with scale does work. Interestingly, the scale-title-filter plugin also works.

**Wayfire version**
git, 44e1fa9c62e1f8c9f35cedbf3e86c6a0247e0b79

This can be easily reproduced using the gtk-layer-demo program (setting keyboard interactivity to "On demand"). I'm also attaching a simple test case (in Python, but Github does not allow .py) that calls scale via IPC when a button is clicked.

Real-world case: [Cairo-Dock](https://github.com/Cairo-Dock/cairo-dock-core) with [this](https://github.com/dkondor/wayfire-scale-ipc) plugin

[ipc_window.txt](https://github.com/user-attachments/files/16489086/ipc_window.txt)

ammen99:
Believe it or not, this is intentional. Scale's input grab is below the layer where panels and docks usually are. This is done so that you can indeed interact with wf-panel and other panels while in scale, someone requested this feature awhile ago. The solution/expectation is to have your dock not be keyboard interactive unless it really needs to interact with the keyboard.

dkondor:
Thanks for the explanation! BTW is there a recommended way for "losing" the keyboard focus from a layer-shell view while keeping the availability to be focused later? Switching keyboard-interactivity to NONE and back to ON_DEMAND grabs again the keyboard focus.

ammen99:
It shouldn't grab on the switch back to on-demand, this is probably a bug. On-demand should be auto-focused only on map.

PR #2530, opened 2024-12-06
layer-shell: do not grab focus if keyboard interactivity is set to ON_DEMAND

### LINKED ISSUE #2450: layer-shell: setting keyboard-interactivity to ON_DEMAND always grabs the focus
**Describe the bug**
Per [here](https://github.com/WayfireWM/wayfire/issues/2422#issuecomment-2308661898), this is not intended.

**To Reproduce**
See Python code below. It creates a layer-shell view with buttons to toggle keyboard-interactivity + a separate button below the text entry ("Release keyboard") which tries to "lose" keyboard interactivity by (1) switching to NONE; (2) waiting 500 ms (to make what's happening more visible); and (3) switching back to ON_DEMAND

In any case when keyboard-interactivity is switched to ON_DEMAND, the layer-shell surface grabs keyboard focus.

**Expected behavior**
When clicking on "Release keyboard", the layer-shell surface should stay unfocused.

(Note: I'm unsure what should happen when clicking the "On demand" button; on the one hand, a mouse click should lead to keyboard focus; on the other hand, even in this case, switching to ON_DEMAND happens after the mouse click finished. In any case, this is not relevant to my use case and I'm fine with any behavior)

**Use case**
Primarily, I use ON_DEMAND to be able to register `modifier + click` on a layer-shell surface (while using standard GTK API). This requires ON_DEMAND to be the default state. I do want to "lose" the keyboard focus in specific cases: (1) when a menu is closed (e.g. by ESC or wf-shell::toggle_menu); (2) when triggering scale via IPC (e.g. see [here](https://github.com/WayfireWM/wayfire/issues/2422)).

**Wayfire version**
git 44e1fa9c62e1f8c9f35cedbf3e86c6a0247e0b79

**Test code**

```
#!/usr/bin/env python3

import os
import sys
import gi
import wayfire_socket as ws

gi.require_version('Gtk', '3.0')
gi.require_version('GtkLayerShell', '0.1')

from gi.repository import Gtk, GtkLayerShell, GLib


class KBWindow(Gtk.Window):
	def __init__(self):
		self.timeout_id = 0
		Gtk.Window.__init__(self)
		GtkLayerShell.init_for_window(self)
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.ON_DEMAND)
		lbl = Gtk.Label.new('Set keyboard mode')
		btn1 = Gtk.Button.new_with_label('None')
		btn1.connect("clicked", self.kb_none)
		btn2 = Gtk.Button.new_with_label('On demand')
		btn2.connect("clicked", self.kb_ondemand)
		btn3 = Gtk.Button.new_with_label('Exclusive')
		btn3.connect("clicked", self.kb_excl)
		btn4 = Gtk.Button.new_with_label('Quit')
		btn4.connect("clicked", Gtk.main_quit)
		btn5 = Gtk.Button.new_with_label('Release keyboard')
		btn5.connect("clicked", self.release_kb)
		entry = Gtk.Entry()
		box = Gtk.Box(orientation = Gtk.Orientation.VERTICAL)
		box.add(lbl)
		box.add(btn1)
		box.add(btn2)
		box.add(btn3)
		box.add(entry)
		box.add(btn5)
		box.add(btn4)
		self.add(box)
		self.connect('destroy', Gtk.main_quit)
		self.show_all()
		# GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.ON_DEMAND)

	def kb_none(self, btn):
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.NONE)
	
	def kb_ondemand(self, btn):
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.ON_DEMAND)
	
	def kb_excl(self, btn):
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.EXCLUSIVE)
	
	def reset_keyboard(self):
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.ON_DEMAND)
		self.timeout_id = 0
		return False

	def release_kb(self, btn):
		# we set keyboard mode to NONE first to lose keyboard focus
		GtkLayerShell.set_keyboard_mode(self, GtkLayerShell.KeyboardMode.NONE)
		if self.timeout_id == 0:
			# we wait 500 ms before setting it back to ON_DEMAND to see if focus really changes
			self.timeout_id = GLib.timeout_add(500, self.reset_keyboard)
	
def main(args):
	win1 = KBWindow()
	Gtk.main()
	return 0

if __name__ == '__main__':
	sys.exit(main(sys.argv[1:]))

```

dkondor:
Fixes #2450 

This change avoids automatically focusing layer-shell surfaces when their keyboard interactivity is changed to ON_DEMAND. As discussed in https://github.com/WayfireWM/wayfire/issues/2422#issuecomment-2308661898, this is not intended. However, this change can lead to unexpected changes if an app expects that this will grab focus. So it might not be the best solution after all.

Main change: Layer-shell views can now intentionally lose focus by setting keyboard interactivity to NONE and then back to ON_DEMAND (thus allowing them to be later refocused by normal user interactions; this was not possible before).

Possible drawback: if an app sets keyboard interactivity to ON_DEMAND as a response to user action (e.g. mouse click), it will not get keyboard focus until further user action. This is demonstrated e.g. by the test code in issue #2450 where after applying this patch, clicking on the "On demand" button will not grab focus. However, grabbing focus is still possible by first setting it to EXCLUSIVE and then back to ON_DEMAND (this is e.g. used by [wf-shell](https://github.com/WayfireWM/wf-shell/blob/f49f719effed6de47a85a64689672581cc18ec58/src/util/wf-autohide-window.cpp#L412) which continues to work fine with this patch applied).

### DIFF
--- a/src/view/layer-shell/layer-shell.cpp
+++ b/src/view/layer-shell/layer-shell.cpp
@@ -566,10 +566,10 @@ void wayfire_layer_shell_view::commit()
 
         if (prev_state.keyboard_interactive != state->keyboard_interactive)
         {
-            if ((state->keyboard_interactive >= 1) && (state->layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP))
+            if ((state->keyboard_interactive == 1) && (state->layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP))
             {
                 wf::get_core().seat->focus_view(self());
-            } else
+            } else if (state->keyboard_interactive == 0)
             {
                 wf::get_core().seat->refocus();
             }


ammen99 (approved):
LGTM, thanks!

PR #2541, opened 2024-12-20
foreign-toplevel: Fix potential crash when app-id mode is 'full'

soreau:
This fixes a potential crash with compositor views when workarounds option app-id-mode is set to 'full'. The crash can be caused when attempting to use magnifier, with this option set.


ammen99 (approved):
thanks :)

PR #2544, opened 2024-12-21
input method: don't forward key-release without correspinding key-press

tokyo4j:
This PR follows the fix in https://github.com/labwc/labwc/pull/2437.

After commit [e2189903](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e21899037a97dcade40149f05ab54229c0018644) in wlroots, when ctrl-f is pressed in firefox with a IME client running, the following key-release event for "f" is not sent, thus "f" is repeated like "ffffffffff..." in the input box of firefox.

The video below is the demonstration of the regression in labwc, but it also happens in wayfire:

https://github.com/user-attachments/assets/b87e7e7f-c288-4f50-8902-361a6f0be678

This happens in the following flow:
- The compositor receives key-press events for ctrl+f and sends them to Firefox
- Firefox enables text_input for the input box
- The compositor activates the IME client
- The compositor receives a key-release event for "f", and forward it to the IME client
- The IME client sends the key-release event for "f" to the compositor via a virtual keyboard, but wlroots ignores it because the key "f" has not been pressed in the virtual keyboard.
- As firefox doesn't see the key-release event, its repeat timer timeouts and it starts printing "fffffffff...".

So this PR fixes this problem by not forwarding the key-release event to the IME client unless the corresponding key-press event has also been forwarded.

### DIFF
--- a/src/core/seat/input-method-relay.cpp
+++ b/src/core/seat/input-method-relay.cpp
@@ -284,6 +284,19 @@ bool wf::input_method_relay::handle_key(struct wlr_keyboard *kbd, uint32_t time,
         return false;
     }
 
+    if (state == WL_KEYBOARD_KEY_STATE_PRESSED) {
+        pressed_keys.insert(key);
+    } else
+    {
+        // Don't forward the release event if the press event for the same key
+        // has also been forwarded.
+        if (!pressed_keys.count(key))
+        {
+            return false;
+        }
+        pressed_keys.erase(pressed_keys.find(key));
+    }
+
     wlr_input_method_keyboard_grab_v2_set_keyboard(keyboard_grab, kbd);
     wlr_input_method_keyboard_grab_v2_send_key(keyboard_grab, time, key, state);
     return true;
--- a/src/core/seat/input-method-relay.hpp
+++ b/src/core/seat/input-method-relay.hpp
@@ -7,6 +7,7 @@
 
 #include <vector>
 #include <memory>
+#include <set>
 
 namespace wf
 {
@@ -24,6 +25,8 @@ class input_method_relay : public text_input_v3_im_relay_interface_t
     std::optional<uint32_t> last_done_serial;
     uint32_t next_done_serial = 0;
     void send_im_done();
+
+    std::multiset<uint32_t> pressed_keys;
 
     text_input *find_focusable_text_input();
     void set_focus(wlr_surface*);


src/core/seat/input-method-relay.cpp
```diff
@@ -284,6 +284,19 @@ bool wf::input_method_relay::handle_key(struct wlr_keyboard *kbd, uint32_t time,
...
+        pressed_keys.insert(key);
+    } else
+    {
+        // Don't forward the release event if the press event for the same key
+        // has also been forwarded.
+        if (!pressed_keys.count(key))
+        {
+            return false;
+        }
+        pressed_keys.erase(key);
```
ammen99, line 297: you're using a multiset but here you would erase all the keys with matching keycodes .. I assume you want something like `pressed_keys.erase(pressed_keys.find(key))`

tokyo4j: Thanks. Fixed.

tokyo4j:
Now I'm not confident that this PR should be merged because I noticed this PR causes another regression of stuck modifier in Firefox when pressing Ctrl+F (even though repeated fffffff... is fixed). Labwc had the same issue and I made a [workaround](https://github.com/labwc/labwc/pull/2498) for that, but it's a dirty workaround specifically for Fcitx5 and I honestly don't want to be responsible for applying that to Wayfire too.

Fortunately, wlroots devs agreed that the commit which caused the regression (repeated fffffff...) should be reverted in 0.18.3, and the issue that the commit tried to solve should be addressed in 0.19, with some API changes (https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4912#note_2715401).

So for now, I suggest keeping this PR on hold and waiting for wlroots 0.18.3 release.

ammen99:
> Now I'm not confident that this PR should be merged because I noticed this PR causes another regression of stuck modifier in Firefox when pressing Ctrl+F (even though repeated fffffff... is fixed). Labwc had the same issue and I made a [workaround](https://github.com/labwc/labwc/pull/2498) for that, but it's a dirty workaround specifically for Fcitx5 and I honestly don't want to be responsible for applying that to Wayfire too.

If you want to use fcitx5, I very strongly recommend using the input-method-v1 plugin instead of the support for input-method-v2 available in core. We had a lot of bugs and finally had to implement the older protocol, which turns out to be much more meaningful. Maybe try that and see whether the bug you're experiencing won't be fixed with that protocol.

I personally don't want to add any fcitx5-specific workarounds in core, as I think that the underlying problems are on the protocol level and using the older protocol is just a better option.

tokyo4j:
Closing as [e2189903](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e21899037a97dcade40149f05ab54229c0018644) was reverted in wlroots 0.19.

PR #2546, opened 2024-12-24
core: Add option to run as root

### LINKED ISSUE #2540: give an option to run wayfire as root
Give an option to run wayfire on root account by simply adding a command line parameter just like in Hyprland's "--i-am-not-stupid" parameter for running Hyprland on root.

soreau:
To run wayfire as root, pass the option --with-great-power-comes-great-responsibility.

Fixes #2540.


ammen99 (approved):
Thanks:)

PR #2547, opened 2024-12-26
input-method-popup: Fix damage issues

### LINKED ISSUE #2539: When input method panel becomes shorter, the shortened region is still visible
**Describe the bug**

**To Reproduce**
1. Enable plugin `input-method-v1`
2. Find a text field and start typing using input method
3. When input method panel becomes shorter (but not moved), the shortened region is still visible

**Expected behavior**
The shortened region should disappear and whatever was below is now showing.

**Screenshots or stacktrace**

https://github.com/user-attachments/assets/19d8f7c3-4dc6-4101-ac85-7d7cd60b718d
(It's better to download the screencast and view frame by frame.)

**Wayfire version**
git

soreau:
Fixes #2539.

### DIFF
--- a/src/api/wayfire/unstable/wlr-text-input-v3-popup.hpp
+++ b/src/api/wayfire/unstable/wlr-text-input-v3-popup.hpp
@@ -39,6 +39,7 @@ class text_input_v3_popup : public wf::view_interface_t
   private:
     wf::geometry_t geometry{0, 0, 0, 0};
     wlr_box old_cursor_rect{0, 0, 0, 0};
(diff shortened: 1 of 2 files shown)


ammen99:
This fixes the issue only for input-method-popups, but not for other subsurfaces. And as we discussed in the issue, the underlying problem is that we don't accumulate damage from surface size changes. I will submit a new PR with a 'proper' fix which should hopefully also resolve the problem.

ammen99:
Ok, contrary to what I thought, this is still needed even with the other fixes. The issue is that we change both position and size at the same time.

PR #2550, opened 2024-12-27
session-lock: lock the screen even if the client provides no lock surface

dkondor:
Minimal changes to hide the screen contents in this case as well.

Simple test client: https://github.com/dkondor/simple_lock

### DIFF
--- a/plugins/protocols/session-lock.cpp
+++ b/plugins/protocols/session-lock.cpp
@@ -144,10 +144,9 @@ class lock_crashed_node : public lock_base_node<simple_text_node_t>
         // TODO: make the text smaller and display a useful message instead of a big explosion.
         set_text(text);
         auto layer_node = output->node_for_layer(wf::scene::layer::LOCK);
-        if (!is_displayed)
+        if (parent() == nullptr)
         {
             wf::scene::add_back(layer_node, shared_from_this());
-            is_displayed = true;
         }
 
         wf::get_core().seat->set_active_node(shared_from_this());
@@ -166,9 +165,6 @@ class lock_crashed_node : public lock_base_node<simple_text_node_t>
         result.local_coords = at;
         return result;
     }
-
-  private:
-    bool is_displayed = false;
 };
 
 class wf_session_lock_plugin : public wf::plugin_interface_t
--- a/plugins/protocols/session-lock.cpp
+++ b/plugins/protocols/session-lock.cpp
@@ -144,7 +144,12 @@ class lock_crashed_node : public lock_base_node<simple_text_node_t>
         // TODO: make the text smaller and display a useful message instead of a big explosion.
         set_text(text);
         auto layer_node = output->node_for_layer(wf::scene::layer::LOCK);
-        wf::scene::add_back(layer_node, shared_from_this());
+        if (!is_displayed)
+        {
+            wf::scene::add_back(layer_node, shared_from_this());
+            is_displayed = true;
+        }
+
         wf::get_core().seat->set_active_node(shared_from_this());
     }
 
@@ -161,6 +166,9 @@ class lock_crashed_node : public lock_base_node<simple_text_node_t>
         result.local_coords = at;
         return result;
     }
+
+  private:
+    bool is_displayed = false;
 };
 
 class wf_session_lock_plugin : public wf::plugin_interface_t
@@ -290,6 +298,12 @@ class wf_session_lock_plugin : public wf::plugin_interface_t
             {
                 disconnect_signals();
                 set_state(state == UNLOCKED ? DESTROYED : ZOMBIE);
+                if (state == ZOMBIE)
+                {
+                    // ensure that the crashed node is displayed in this case as well
+                    lock_all();
+                }
+
                 LOGC(LSHELL, "session lock destroyed");
             });
             destroy.connect(&lock->events.destroy);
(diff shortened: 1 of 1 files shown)


plugins/protocols/session-lock.cpp
```diff
@@ -318,7 +318,7 @@ class wf_session_lock_plugin : public wf::plugin_interface_t
             if (state == ZOMBIE)
             {
                 output->set_inhibited(true);
-                output_states[output]->crashed_node->display();
+                output_states[output]->crashed_node->display("💥");
```
ammen99, line 321: Maybe we can define this emoji somewhere as a constant in order to not repeat it? :)

dkondor: Yes, makes sense -- I've added a separate function for it

dkondor:
I've also realized after some re-testing that this does not work if a screenlocker crashes without ever displaying a surface. The last commit takes care of this (I

plugins/protocols/session-lock.cpp
```diff
@@ -134,20 +134,30 @@ class lock_crashed_node : public lock_base_node<simple_text_node_t>
...
             1280 /* font_size */,
             wf::color_t{0, 0, 0, 1} /* bg_color */,
             wf::color_t{0.9, 0.9, 0.9, 1} /* fg_color */);
         set_text_params(params);
         // TODO: make the text smaller and display a useful message instead of a big explosion.
-        set_text("💥");
+        set_text(text);
         auto layer_node = output->node_for_layer(wf::scene::layer::LOCK);
-        wf::scene::add_back(layer_node, shared_from_this());
+        if (!is_displayed)
```
ammen99, line 147: Do we need an additional bool variable for this? AFAICT all you want to know is whether the node has alraedy been added to the scenegraph. You could do that by checking whether its parent is NULL.

plugins/protocols/session-lock.cpp
```diff
@@ -346,12 +359,17 @@ class wf_session_lock_plugin : public wf::plugin_interface_t
...
-            if (output_state->surface_node)
+            if (state == ZOMBIE)
+            {
+                output_state->crashed_node->display_crashed();
+            } else if (output_state->surface_node)
             {
                 output_state->surface_node->display();
+            } else
+            {
+                // if the surface node has not yet been displayed we show an empty surface
```
ammen99, line 370: Should we hide the crashed node once the actual lock node is there? I think for example if one wanted to implement a blurred lockscreen, or animate the lockscreen surface, this would prevent it from working.

plugins/protocols/session-lock.cpp
dkondor: No need, I didn't realize `parent()` can be used for this :)

plugins/protocols/session-lock.cpp
dkondor: Potentially, however, after doing some tests with the current implementation, I'm a bit confused how this should happen:
 - with [Hyprlock](https://github.com/hyprwm/hyprlock/): things already seem to "work": the crashed node disappears, and fade-in starts from the unlocked desktop
 - with [shaderlock](https://github.com/RobinMcCorkell/shaderlock): the unlocked screen flashes briefly (which is weird, since I assumed that crashed node should be there), but then the crashed lock screen comes back and is the basis of further animations

This is especially strange given that both projects seem to use the wlr-screencopy protocol to get a screenshot of the current desktop contents. Do you happen to know of another lockscreen app that does not make a screenshot, but just simply displays a (partially) transparent surface? That would allow me to test this more cleanly.

On a second thought, actually it could even make sense to not show the desktop after the previous lockscreen has already crashed. Contrary to when originally locking the screen, it could be a security / privacy issue if the screen contents are revealed when restarting a crashed lockscreen (however briefly). But then it is very weird why the screen shows up with shaderlock, I'll investigate that separately a bit more.

ammen99: As a guess, maybe they take the screenshot at different times?

dkondor: Yes, that can be.

Now I realized what we have here does not actually matter, since a new instance is created when a new lock client connects: https://github.com/WayfireWM/wayfire/blob/4dc41b8d32236925a85ef6385b7f016114289958/plugins/protocols/session-lock.cpp#L460

So that would also explain why the desktop becomes visible even with shaderlock (I assume it takes the screenshot before creating the new lock then). Maybe it would make sense to refactor this so that the crashed node is kept until the new client creates a lock surface and thus the screen cannot be revealed accidentally. However, that would be a bigger change and IMO unrelated to the current PR -- I can create a separate issue and try writing a patch for it if you think this can be an issue. Note that I've only seen this with shaderlock, probably because it is a bit slow to display its initial lock surface (e.g. swaylock appears fast enough). In any case, this is likely not a common issue, since most lockscreens do not crash, and even in that case, the problem manifests only if there is some automated way to restart them which could trigger when an "unauthorized" user is present.

dkondor: > Note that I've only seen this with shaderlock, probably because it is a bit slow to display its initial lock surface (e.g. swaylock appears fast enough).

Actually, I've just tried gtklock and it is also slow enough that the unlocked desktop flashes, but in this case really briefly (barely possible to notice it).

ammen99: I will trust your judgement on this matter :)

ammen99 (approved):
LGTM, thanks!

PR #2553, opened 2024-12-30
nonstd: fix source file encoding

dkondor:
Found this while looking into the packaging status in Debian:
https://udd.debian.org/lintian/?packages=wayfire

There was one non-UTF-8 character in comments ('§'); instead of using the UTF-8 variant, I suggest to rephrase to use only ASCII characters for simplicity.


ammen99 (approved):
Thanks:)

PR #2562, opened 2025-01-12
plugin-loader: refuse loading plugins with no metadata

### LINKED ISSUE #2561: Crash when loading a newly installed plugin
**Describe the bug**
Wayfire crashes when trying to load a plugin that was installed after it had originally started. Might only happen for plugins that use keybinding, but haven't investigated in detail.

**To Reproduce**
1. Start Wayfire
2. Install a new plugin (tested with [this](https://github.com/dkondor/wayland-screenlock-proxy/tree/main/wayfire_plugin) plugin, but I expect the same behavior with others)
3. Enable it (WCM or config file)
4. Wayfire crashes

**Expected behavior**
No crash. Ideally, the new plugin should be able to load, but it is acceptable to required restarting Wayfire as long as it does not crash.

**Screenshots or stacktrace**
```
EE 12-01-25 15:24:52.738 - [src/debug.cpp:401] Wayfire encountered error loading option "screenlock_integration/lock": No such option: screenlock_integration/lock. Usual reasons for this include missing or outdated plugin XML files, a bug in the plugin itself, or mismatch between the versions of Wayfire and wf-config. Make sure that you have the correct versions of all relevant packages and make sure that there are no conflicting installations of Wayfire using the same prefix.
EE 12-01-25 15:24:52.867 - #1  wf::detail::option_wrapper_debug_message(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::runtime_error const&) ../src/debug.cpp:407
EE 12-01-25 15:24:53.557 - #2  wf::option_wrapper_t<wf::keybinding_t>::load_option(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../src/api/wayfire/option-wrapper.hpp:41 (discriminator 1)
EE 12-01-25 15:24:54.082 - #3  wf::option_wrapper_t<wf::keybinding_t>::option_wrapper_t(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ../src/api/wayfire/option-wrapper.hpp:32
EE 12-01-25 15:24:54.107 - #4  screenlock_integration::init() ../wayfire_plugin/screenlock_integration.cpp:47 (discriminator 4)
EE 12-01-25 15:24:54.324 - #5  wf::plugin_manager_t::reload_dynamic_plugins() ../src/core/plugin-loader.cpp:228
EE 12-01-25 15:24:54.559 - #6  wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}::operator()() const ../src/core/plugin-loader.cpp:23
EE 12-01-25 15:24:54.774 - #7  void std::__invoke_impl<void, wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}&>(std::__invoke_other, wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}&) /usr/include/c++/13/bits/invoke.h:61
EE 12-01-25 15:24:55.092 - #8  std::enable_if<is_invocable_r_v<void, wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}&>, void>::type std::__invoke_r<void, wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}&>(wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}&) /usr/include/c++/13/bits/invoke.h:117
EE 12-01-25 15:24:55.473 - #9  std::_Function_handler<void (), wf::plugin_manager_t::plugin_manager_t()::{lambda()#1}::operator()() const::{lambda()#1}>::_M_invoke(std::_Any_data const&) /usr/include/c++/13/bits/std_function.h:291
EE 12-01-25 15:24:55.625 - #10 std::function<void ()>::operator()() const ../src/util.cpp:591
EE 12-01-25 15:24:55.748 - #11 wf::wl_idle_call::execute() ../src/util.cpp:91
EE 12-01-25 15:24:55.845 - #12 handle_idle_listener(void*) ../src/util.cpp:27
EE 12-01-25 15:24:55.858 - #13 wl_event_loop_dispatch_idle ??:?
EE 12-01-25 15:24:55.870 - #14 wl_event_loop_dispatch ??:?
EE 12-01-25 15:24:55.884 - #15 wl_display_run ??:?
EE 12-01-25 15:24:55.968 - #16 main ../src/main.cpp:456
EE 12-01-25 15:24:56.154 - #17 __libc_start_call_main ../sysdeps/x86/libc-start.c:74
EE 12-01-25 15:24:56.375 - #18 call_init ../csu/libc-start.c:128
EE 12-01-25 15:24:58.836 - #19 _start ??:?
```

**Wayfire version**
git (actually [this](https://github.com/WayfireWM/wayfire/pull/2550) branch, but I don't expect any difference to master)
0.8.0+git20240110-2.1build1 (the version in Ubuntu 24.04)

**Additional info**

The problem seems to come from wf-config, which does not update its internal state when the new plugin is installed. Specifically, loading the plugin tries to read its config (`screenlock_integration/lock` in the above case), ending up [here](https://github.com/WayfireWM/wf-config/blob/b2e683fef467d25bd337db4d1fe946111229a375/src/config-manager.cpp#L77). The `get_section()` function then returns `nullptr`, since I assume there is no mechanism to look for xml files that were added since Wayfire was started.

This is turn results in throwing an exception from [here](https://github.com/WayfireWM/wf-config/blob/b2e683fef467d25bd337db4d1fe946111229a375/include/wayfire/config/option-wrapper.hpp#L89)

This is caught [here](https://github.com/WayfireWM/wayfire/blob/cf9afeda562ff4bb46bf2923723a744fbf832efd/src/api/wayfire/option-wrapper.hpp#L40) and the error message is printed followed by the stacktrace.

**Possible solutions**

1. Improve wf-config to dynamically rescan the metadata directory when it changes (or at least when a section is not found). This would be a complete solution, allowing to load freshly installed plugins, but seems complicated and can lead to weird edge cases.
2. Refuse to load plugins whose metadata is not loaded. This required restarting Wayfire after installing a plugin, which is still reasonable, and probably less error-prone. However, it needs some extra communication between the config backend and the plugin loader.

dkondor:
Fixes #2561

This is taking the second approach mentioned there, by not loading plugins that have no metadata (I assume that all plugins have an XML and it is loaded by wf-config, even if they have no options there).

### DIFF
--- a/src/core/plugin-loader.cpp
+++ b/src/core/plugin-loader.cpp
@@ -168,6 +168,19 @@ void wf::plugin_manager_t::reload_dynamic_plugins()
                     continue;
                 }
 
+                auto config_section = wf::get_core().config.get_section(plugin_name);
+                if (config_section == nullptr)
+                {
+                    LOGE(plugin_name,
+                        " has no configuration, skipping. "
+                        "Usual reasons for this include missing or outdated plugin XML files, "
+                        "or mismatch between the versions of Wayfire and wf-config. "
+                        "Make sure that you have the correct versions of all relevant packages and make sure that there ",
+                        "are no conflicting installations of Wayfire using the same prefix. "
+                        "Note that if this plugin was newly installed, you need to restart Wayfire before using it.");
+                    continue;
+                }
+
                 next_plugins.push_back(plugin_path.value());
             } else
             {


dkondor:
I did a bit more testing and realized that this only works if the plugin is completely "unknown".

However, if there is leftover configuration in `wafire.ini`, the crash from #2561 can still happen as apparently wf-config will create a section. Still, this fixes at least the case of completely newly installed plugins.

soreau:
This is a NAK because plugins without metadata are valid to load. If metadata is installed or changes, wayfire must be restarted. If dynamic reloading of metadata is implemented, it would start in wf-config, as far as I understand. But then you would have to plumb it through the relevant wf-config consumers and it might be easier to just restart wayfire.

dkondor:
Oh, I see, that does complicate things. I also feel that reloading metadata will have its own complications.
Another way to fix my original issue would be to `dlopen()` everything on startup in the plugin lib folder and later refuse to load plugins that are not already opened. Although this feels as a hack as well and could bring it's own problems. I might experiment with it if I have a bit more time.

PR #2563, opened 2025-01-16
workarounds: Add force frame sync option

soreau:
The force frame sync option causes glFinish() to be called after rendering but before commiting the new buffer. This can be used to workaround driver bugs, especially where dma-fences are not available for frame synchronization. This is the case with multi- threaded llvmpipe rendering, using drm or wayland backends.

### DIFF
--- a/metadata/workarounds.xml
+++ b/metadata/workarounds.xml
@@ -68,5 +68,10 @@
 			<_long>Enable certain damage optimizations which are based on a surfaces' opaque regions. In some cases, this optimization might give unexpected results (i.e background app stops updating) even though this is fine according to Wayland's protocol.</_long>
 			<default>false</default>
 		</option>
+		<option name="force_frame_sync" type="bool">
+			<_short>Force frame synchronization.</_short>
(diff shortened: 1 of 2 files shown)


metadata/workarounds.xml
```diff
@@ -68,5 +68,10 @@
 			<_long>Enable certain damage optimizations which are based on a surfaces' opaque regions. In some cases, this optimization might give unexpected results (i.e background app stops updating) even though this is fine according to Wayland's protocol.</_long>
 			<default>false</default>
 		</option>
+		<option name="force_frame_sync" type="bool">
+			<_short>Force frame synchronization.</_short>
+			<_long>This option can be used to workaround driver bugs that cause rendering artifacts, though can cause more resource usage. Leave disabled if unsure.</_long>
```
ammen99, line 73: I would add a comment saying that this helps with nvidia cards (because I expect that this will be the most common use-case for this option).

ammen99: LGTM otherwise.

ammen99 (approved):
LGTM, thanks!

Issue #2367, opened 2024-05-26
GTK Shell: implement global menu get-set properties


rilian-la-te:
```
/**
 *  In gnome-shell/mutter/meta windows/views keep track of the properties
 *  specified as arguments here.
 *  Currently only the app_id is implemented / required.
 */
static void handle_gtk_surface_set_dbus_properties(wl_client *client,
    wl_resource *resource,
    const char *application_id, const char *app_menu_path, const char *menubar_path,
    const char *window_object_path, const char *application_object_path,
    const char *unique_bus_name)
{
    auto surface = static_cast<wf_gtk_surface*>(wl_resource_get_user_data(resource));
    if (application_id)
    {
        wf::get_core().get_data_safe<wf_gtk_shell>()->surface_app_id[surface->wl_surface] = application_id;
    }
}
```

I think it will be good if it will be possible to query and set other properties, because it will open Global Menu to work. Maybe it is worth to make it separate plugin along with KDE global menu protocol.

PR #2569, opened 2025-01-31
DRAFT: proof of concept for global menus

dkondor:
Partially addresses #2367 

This is a draft so far, since the functionality is very basic and only works for GTK3 apps. I would eventually prefer exposing this via the `wlr-foreign-toplevel-management` protocol.

So far, this allows reporting the DBus properties set by [gtk-shell](https://wayland.app/protocols/gtk-shell#gtk_surface1:request:set_dbus_properties) to IPC clients (new event: `view-gtk-dbus-properties-changed`).

So far I've included an example client that displays the menus of the active app: `wf_gtk_global_menu.py` (once this PR is closer to ready, I'd move this to a separate repo, or open a separate PR in pywayfire).
Currently, this only works with GTK3 apps. To use this:
1. Install `appmenu-gtk3-module`
2. Run `python3 wf_gtk_global_menu.py` (it needs to start before any app whose menu to show)
3. Start a GTK3 app with setting the environment variable `GTK_MODULES=appmenu-gtk-module`
4. Focus the app (you might need to change focus for things to work); its app-id should show up in the "Active app" field
5. Select "Show menu" 

Next steps:
 - add also the [KDE appmenu](https://wayland.app/protocols/kde-appmenu) protocol (it would allow global menus for Qt apps)
 - experiment with adding a new event for the `wlr-foreign-toplevel-management` protocol (similar to [this](https://wayland.app/protocols/kde-plasma-window-management#org_kde_plasma_window:event:application_menu)) to make using this easier for panels without relying on Wayfire's IPC
 - likely this would require storing the properties (`app_menu_path`, `menubar_path`, etc.) so they can be reported for already running apps as well

### DIFF
--- a/plugins/ipc-rules/ipc-events.hpp
+++ b/plugins/ipc-rules/ipc-events.hpp
@@ -5,6 +5,7 @@
 #include "plugins/ipc/ipc-method-repository.hpp"
 #include <wayfire/per-output-plugin.hpp>
 #include <nlohmann/json.hpp>
+#include "plugins/protocols/gtk-shell.hpp"
 
 namespace wf
 {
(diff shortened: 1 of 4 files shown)


soreau:
If I understand correctly, here is some prior art, pre-ipc era: https://github.com/soreau/wf-extra-protos

dkondor:
> If I understand correctly, here is some prior art, pre-ipc era: https://github.com/soreau/wf-extra-protos

Thanks for looking into this! I think these are orthogonal, since here the menu content is supplied by the app; see also e.g. https://github.com/Alexays/Waybar/issues/288

soreau:
> Thanks for looking into this! I think these are orthogonal, since here the menu content is supplied by the app; see also e.g. [Alexays/Waybar#288](https://github.com/Alexays/Waybar/issues/288)

That seems like a nice concept! I at least know that users want right-click menu functionality whether it be on the desktop, window titlebar or even the taskbar/window-list with more than just minimize/maximize/close (like send window to workspace or desktop menu change workspace).

@dkondor I would like to take this opportunity to thank you for all of your contributions. They help make wayfire continuously better. Keep up the good work!

dkondor:
Closing in favor of #2576

PR #2571, opened 2025-02-03
Add to log message which plugin failed to load.

CarloWood:
This changes a message like:

[wayfire.git/src/core/plugin-loader.cpp:87] error loading plugin: /usr/lib/libcwd_r.so.9: undefined symbol: elf_nextscn

into:

[wayfire.git/src/core/plugin-loader.cpp:87] error loading plugin [/usr/lib/wayfire/libpixdecor.so]: /usr/lib/libcwd_r.so.9: undefined symbol: elf_nextscn

In other words, not just print the dlerror, but print which plugin failed to load.

### DIFF
--- a/src/core/plugin-loader.cpp
+++ b/src/core/plugin-loader.cpp
@@ -84,7 +84,7 @@ std::pair<void*, void*> wf::get_new_instance_handle(const std::string& path)
     void *handle = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
     if (handle == NULL)
     {
-        LOGE("error loading plugin: ", dlerror());
+        LOGE("error loading plugin [", path, "]: ", dlerror());
         return {nullptr, nullptr};
     }
 
--- a/plugins/ipc/ipc-method-repository.hpp
+++ b/plugins/ipc/ipc-method-repository.hpp
@@ -16,7 +16,7 @@ namespace ipc
 class client_interface_t
 {
   public:
-    virtual void send_json(nlohmann::json json) = 0;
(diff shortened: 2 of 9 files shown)


CarloWood:
I tried to create a new PR, but because the previous one isn't merged yet - it seems I can only append to this one :/.

src/core/plugin-loader.cpp
```diff
@@ -84,7 +84,7 @@ std::pair<void*, void*> wf::get_new_instance_handle(const std::string& path)
     void *handle = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
     if (handle == NULL)
     {
-        LOGE("error loading plugin: ", dlerror());
+        LOGE("error loading plugin [", path, "]: ", dlerror());
```
soreau, line 87: This appears to be the only change that the PR advertises. I'm not sure what the other changes are about, but they don't seem that they should be a part of this pull request.

CarloWood: That is correct. That was the first commit and PR of four days ago.
After that I made two more PR requests, but there is no way to make separate PR's. It merges everything into this one. I thought that soreau told me to just keep adding stuff, maybe I misunderstood?

soreau: The PR should be about an idea. If a commit is not part of the idea, it should go into a separate PR. If you have one PR that depends on another, you can submit the first one and then in the subsequent PRs, put a message that it depends on the first PR (number). Of course the dependent PRs won't build in CI until the first one is merged, but after it is, we can simply rerun CI and most times, you will not need to take further action.

CarloWood:
I'm closing this PR because I'll split it up into three different PR's.

PR #2573, opened 2025-02-07
Add to log message which plugin failed to load.

CarloWood:
This changes a message like:

[wayfire.git/src/core/plugin-loader.cpp:87] error loading plugin: /usr/lib/libcwd_r.so.9: undefined symbol: elf_nextscn

into:

[wayfire.git/src/core/plugin-loader.cpp:87] error loading plugin [/usr/lib/wayfire/libpixdecor.so]: /usr/lib/libcwd_r.so.9: undefined symbol: elf_nextscn

In other words, not just print the dlerror, but print which plugin failed to load.


ammen99 (approved):
thanks!

PR #2574, opened 2025-02-07
Fix compile warning.

CarloWood:
Fix fixes the compile warning,

../../../../github/wayfire/wayfire.git/src/debug.cpp: In function ‘std::string read_output(std::string)’: ../../../../github/wayfire/wayfire.git/src/debug.cpp:95:10: warning: ignoring return value of ‘char* fgets(char*, int, FILE*)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   95 |     fgets(buffer, MAX_FUNCTION_NAME, file);
      |     ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In this case there is a potential problem: if gets returns NULL, then nothing was read and the buffer is uninitialized. Not only would this function be returning uninitialized data, an arbitrary amount of memory could be allocated into `line` (and be returned) when `std::string line = buffer;` is searching for a terminating zero in the uninitialized buffer.

I kinda rewrote this function (while I was at it) to not write first to the stack and then copy that into `line`. Instead the `std::string` is created with the same space as the buffer was, and then we write directly into the allocated memory. Afterwards a `resize()` sets `line` to the correct size - removing any trailing newline at the same time.

In other words, this code is much faster too (as if that was needed, but alas)!

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -14,6 +14,7 @@
 #endif
 
 #include <cstdio>
+#include <cstring>
 #include <dlfcn.h>
 #include <sys/stat.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


soreau:
> I kinda rewrote this function (while I was at it) to not write first to the stack and then copy that into `line`. Instead the `std::string` is created with the same space as the buffer was, and then we write directly into the allocated memory. Afterwards a `resize()` sets `line` to the correct size - removing any trailing newline at the same time.

In my opinion,. this part should go into another commit, since it makes it unclear on how the warning was fixed.

CarloWood:
The warning was fixed by assigning the return of the call to fgets to a variable.

ammen99:
thanks :)

PR #2576, opened 2025-02-13
Add support for global menus

dkondor:
See https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/131

Requires also https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4986

Test program: https://github.com/dkondor/gtk_global_menu_test

This implements the proposed changes to `wlr-foreign-toplevel-managemen` + also implements the [KDE AppMenu](https://wayland.app/protocols/kde-appmenu) protocol.


killown:
this is a must have feature, awesome!

PR #2578, opened 2025-02-17
PR 05 Move safe-list.hpp to (updated) wf-config.

CarloWood:
Use the new wf-config with updated_handlers fix.


test/misc/meson.build
```diff
@@ -8,7 +8,7 @@ test('Tracking factory test', tracking_allocator)
 safe_list = executable(
     'safe_list',
     'safe-list-test.cpp',
-    include_directories: wayfire_api_inc,
+    include_directories: wfconfig_inc,
     dependencies: doctest,
```
soreau, line 12: The only change you need in this patch is:
```
-    dependencies: doctest,
+    dependencies: [doctest, wfconfig],
```

The rest is unnecessary.

ammen99 (approved):

PR #2581, opened 2025-02-18
Compilation time improvements (part 1)

ammen99:
Part 1 of my changes aimed at improving compilation times in Wayfire.
The changes here are pretty straightforward, just shuffling header files and definitions around so that we don't include stuff we don't need.

~~There are no breaking changes for plugins, but some plugins may be broken if they rely on transitive includes.~~

Breaking changes for plugins:

- `wf::get_core().config` is now a pointer.
- Output signals are defined in output-layout.hpp instead of signal-definitions.hpp
- `keyboard-focus-changed` and `seat-activity` signals are defined in seat.hpp

If plugins use these signals, they ought to now include the correct headers.

Benchmark on my 6C12T laptop shows 52s build time before, 46s build time after, so about 12% faster.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,4 +62,6 @@
 
 build/**
 
+.cache/**
+
 # End of https://www.toptal.com/developers/gitignore/api/c++,meson,ninja,linux
--- a/plugins/blur/blur.cpp
+++ b/plugins/blur/blur.cpp
@@ -21,6 +21,7 @@
 #include "wayfire/scene-render.hpp"
 #include "wayfire/scene.hpp"
 #include "wayfire/signal-provider.hpp"
+#include <list>
 
 using blur_algorithm_provider =
     std::function<nonstd::observer_ptr<wf_blur_base>()>;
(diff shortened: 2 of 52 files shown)


soreau:
Here on `Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz`, ctimes-v2 branch is slightly slower than [current master](https://github.com/WayfireWM/wayfire/commit/c171f623f4c9db91758e2c44a5a60ee50863be80). These are the commands I ran and the results:
```
git checkout master && git pull
ccache -C
rm -rf build
meson setup -Duse_system_wlroots=enabled -Duse_system_wfconfig=enabled build && time ninja -C build
...
[482/482] Linking target plugins/ipc-rules/libipc-rules.so

real	2m53.061s
user	6m51.206s
sys	0m48.135s

git checkout ctime-v2
ccache -C
rm -rf build
meson setup -Duse_system_wlroots=enabled -Duse_system_wfconfig=enabled build && time ninja -C build
...
[482/482] Linking target plugins/ipc-rules/libipc-rules.so

real	2m58.031s
user	6m2.048s
sys	0m44.082s
```

soreau:
> Huh weird, maybe I need to test gcc too instead of just clang. In any case, try adding pch, maybe that will tilt the timing in favor of the changes? `-Db_pch=true` to meson

According to the [meson docs](https://mesonbuild.com/Builtin-options.html), b_pch defaults to `true`.

PR #2588, opened 2025-02-25
Allow Vendored Configuration File

janvhs:
This PR adds support for vendoring the Wayfire configuration file.  In particular this means `${XDG_CONFIG_HOME}/wayfire/wayfire.ini` will take precedence over `${XDG_CONFIG_HOME}/wayfire.ini` while looking for a configuration file.

I opted to implement this change, since Wayfire already uses vendor-directories for plugin data in `${XDG_DATA_HOME}/wayfire/plugin`. Furthermore, I spent way too long trying to find out why my default configuration is not working.  I assume this change can reduce the maintenance burden, since people doing the same mistake will not open an issue or ask in IRC.
I copied it from `/usr/share/examples/wayfire/wayfire.ini` as packaged by my distro and therefore assumed wayfire uses the common vendor prefix pattern.  While researching my problem, I found several threads where people likely did the same mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ xbps-install -S wayfire
 
 ## Configuration
 
-Copy [`wayfire.ini`] to `~/.config/wayfire.ini`.
+Copy [`wayfire.ini`] to `~/.config/wayfire.ini` or `~/.config/wayfire/wayfire.ini`.
 Before running Wayfire, you may want to change the command to start a terminal.
 See the [Configuration] document for information on the options.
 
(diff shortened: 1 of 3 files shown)


ammen99:
> The CI is failing because some files I did not touch are not formatted properly. Should I format and commit those as well?

We use patched uncrustify to workaround some bugs. Look more closely at the CI output  the command before the failed check tells you what you need to fix (spoiler: it is a file you did change :)).

ammen99 (approved):
LGTM, thanks!

PR #2593, opened 2025-03-01
Add an option to disable tap-and-drag gesture

jihem-el:
I don't like the tap-and-drag feature with touchpads because I have a lot of false positives when I use it. So, I suggest to add a setting to disable it.
By default, the option is still enabled.


ammen99 (approved):
LGTM, thanks!

PR #2597, opened 2025-03-02
Pr 06 unidirectional maximization support

CarloWood:
This works, together with changes already made to pixdecor. Aka, it is tested. But on its own it should do precisely nothing: have no effect at all (aka, still compile everything as before and still behave in the same way), short from UB as a result of plugins not following the correct meaning of TILED_EDGES_* - because now each bit is actually treated as how to treat that edge.


CarloWood:
At this point the mouse button click on the maximize button are passed correctly to tile_request. It is almost surprising that after this the decoration plugin already seems to work; as nothing was changed to make it deal with unidirectional maximization internally (I had expected it to be completely confused and -say- draw the window fully maximized even if it is only partially maximized!).

PR #2604, opened 2025-03-17
Compilation times improvements (part 4)

ammen99:
This is the last of my attempts to improve compile time performance.
One of the big issues with plugins is the use of nlohmann, which has a very nice interface but very slow compilation times and is not very performant.

To solve this, this PR migrates Wayfire to a custom json wrapper which has very simple syntax for the basic operations.
In addition, the wrapper is currently implemented using yyjson, which is one of the fastest json parsers available.
The wrapper can be very easily changed to use a different json library if it turns out yyjson has issues.
The use of a wrapper ensures that subsequent changes won't be huge breaking changes for plugins.

Needs more testing as yyjson has a very basic interface so a lot of the operations had to be implemented manually like copying.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -36,7 +36,7 @@ pixman         = dependency('pixman-1')
 xkbcommon      = dependency('xkbcommon')
 libdl          = cpp.find_library('dl')
 udev           = dependency('libudev')
-json           = dependency('RapidJSON')
+json           = dependency('yyjson')
 
 # We're not to use system wlroots: So we'll use the subproject
 if get_option('use_system_wlroots').disabled()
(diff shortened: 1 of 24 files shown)


killown:
stipc.run_cmd is not starting any app besides the result is ok

In [1]: stipc.run_cmd("kitty")
Out[1]: {'result': 'ok', 'pid': 26113}

In [2]: stipc.run_cmd("kitty")
Out[2]: {'result': 'ok', 'pid': 26122}

In [3]: stipc.run_cmd("test")
Out[3]: {'result': 'ok', 'pid': 26126}

In [4]: stipc.run_cmd("librewolf")
Out[4]: {'result': 'ok', 'pid': 26130}

Issue #2598, opened 2025-03-05
Add an override option for client side decorations


allomanta:
I would like to wayfire to (forcefully) draw decorations (borders specifically) around all my windows, however it seems like all the current decoration plugins call `view->should_be_decorated()` which seems to check if the window wants server-side decorations or not.  My current method has been to clone the deco plugin and add a config option and change the should_decorate_view function in decoration.cpp to:
```
bool should_decorate_view(wayfire_toplevel_view view)
{
	return !ignore_decoration_of_view(view) &&
	       (override_client_side || view->should_be_decorated());
}
```
This has worked well enough, but I have to rebuild the plugin for each update, track changes and fix build issues (wlroots usually) and I was wondering if this override could be added to the default deco plugin. 

Since the last I updated I'm also having a weird bug with this fix where it's only drawing borders for windows  that are already open if I restart the plugin, but it doesn't work for new windows opened afterwards. Maybe this is just a build/environment issue as I was having quite a few of those, but I don't really know how to explain this behavior.

PR #2606, opened 2025-03-20
Decoration plugin optionally force decorations

allomanta:
This PR adds an option to the decoration plugin that ignores the clients request for client side decorations, see issue [2598](https://github.com/WayfireWM/wayfire/issues/2598).

### DIFF
--- a/metadata/decoration.xml
+++ b/metadata/decoration.xml
@@ -40,5 +40,10 @@
 			<_long>Disables window decoration for windows matching the specified criteria.</_long>
 			<default>none</default>
 		</option>
+		<option name="override_client_side" type="bool">
+			<_short>Override client side decorations</_short>
+			<_long>Specifies whether to add decorations around windows even if server side decorations aren't requested.</_long>
+			<default>false</default>
+		</option>
 	</plugin>
 </wayfire>
--- a/plugins/decor/decoration.cpp
+++ b/plugins/decor/decoration.cpp
@@ -17,6 +17,9 @@
 class wayfire_decoration : public wf::plugin_interface_t
 {
     wf::view_matcher_t ignore_views{"decoration/ignore_views"};
+	wf::option_wrapper_t<bool> override_client_side{
+		"decoration/override_client_side"
+	};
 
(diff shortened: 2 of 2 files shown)


ammen99:
If we do this, maybe it should be a criteria/view_matcher option so that users can force decorations for particular views only? I think this will make the option useful for more people.

allomanta:
Good point, I hadn't considered that. I'm personally just using this to have consistent borders around all my windows ala hyprland or cosmic. I changed it to a view matcher and updated the description and variables to better match the ignore config option. The commit is a little messier as uncrustify seems to have formatted a bunch as well.

Also I wanted to say thanks for developing wayfire!

ammen99:
I think the code style got messed up, please use uncrustify (we even currently use a fork due to upstream bugs)

allomanta:
Yeah it seems like some disagreement with my setup and running uncrustify messed up the code style. I have now undone that commit and added my changes back and I think the style should be correct now.

ammen99 (approved):
LGTM now, thanks for the PR!

allomanta:
Is there a way to see the error in the pipeline? Because running the command `git ls-files | grep "hpp$\|cpp$" | xargs uncrustify -c uncrustify.ini --no-backup` is not producing any changes in the 2 files here. Maybe I need to use that fork you mentioned?

ammen99:
> Is there a way to see the error in the pipeline? Because running the command `git ls-files | grep "hpp$\|cpp$" | xargs uncrustify -c uncrustify.ini --no-backup` is not producing any changes in the 2 files here. Maybe I need to use that fork you mentioned?

Yes, you can, just take a look at the second-to-last command in the CI output, it prints the full diff: https://github.com/WayfireWM/wayfire/actions/runs/14062744211/job/39378080329?pr=2606#step:8:1

It seems that the error is tabs vs spaces.

allomanta:
Ah I was looking at the next CI step, but I think it was tabs vs spaces, should hopefully be fixed now.

ammen99:
Looks good now :)

PR #2609, opened 2025-03-25
add get_cursor_position ipc method to fetch cursor coordinates

killown:


plugins/ipc/stipc.cpp
```diff
@@ -293,6 +293,7 @@ class stipc_plugin_t : public wf::plugin_interface_t
         method_repository->register_method("stipc/feed_key", feed_key);
         method_repository->register_method("stipc/feed_button", feed_button);
         method_repository->register_method("stipc/move_cursor", move_cursor);
+        method_repository->register_method("stipc/get_cursor_position", get_cursor_position);
```
ammen99, line 296: You need to unregister the method in fini(), otherwise lgtm

PR #2611, opened 2025-03-27
build: Add wrap file for when yyjson is not available

soreau:
This attempts to find a yyjson installation and falls back to building as a subproject via wrap file.


mrsteve0924:
wf-install's update_build.sh script fails to build if yyjson not installed on system.

should i assume this pr will get around that issue?

soreau:
> wf-install's update_build.sh script fails to build if yyjson not installed on system.
> 
> should i assume this pr will get around that issue?

Yes.

ammen99 (approved):
Thanks!

PR #2617, opened 2025-04-06
wlr_layer-shell: Fix init timing and surface validation

killown:
fix: EE 05-04-25 21:47:56.528 - [types/wlr_layer_shell_v1.c:296] A configure is sent to an uninitialized wlr_layer_surface_v1 0x5b50eeb11b00

### DIFF
--- a/src/view/layer-shell/layer-shell.cpp
+++ b/src/view/layer-shell/layer-shell.cpp
@@ -496,7 +496,6 @@ void wayfire_layer_shell_view::map()
     {
         this->app_id = nonull(lsurface->namespace_t);
         wf::view_implementation::emit_app_id_changed_signal(self());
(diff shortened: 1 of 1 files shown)


src/view/layer-shell/layer-shell.cpp
```diff
@@ -494,6 +496,7 @@ void wayfire_layer_shell_view::map()
     {
         this->app_id = nonull(lsurface->namespace_t);
         wf::view_implementation::emit_app_id_changed_signal(self());
+        wf_layer_shell_manager::get_instance().arrange_unmapped_view(this);
```
ammen99, line 499: Seems very weird to arrange_unmapped_view in the map handler, I think you can remove this call.

ammen99 (changes requested):

PR #2618, opened 2025-04-06
Try to avoid crashing when loading a plugin with invalid options

### LINKED ISSUE #2561: Crash when loading a newly installed plugin

dkondor:
Fixes #2561 -- second attempt.

This adds a flag to avoid crashing if an option is not found while loading a plugin; the idea is that most plugins will try to load their options on creation or in their `init()` function, so that most of these cases will be caught.

Potential issues:
 - This might lead to UB if the plugin tries to use the (now uninitialized) options? (I did not check if they are initialized to some defaults).
 - Also, this could lead to crashes / UB if plugins' `fini()` function or destructor relies on the `init()` function and / or the constructor having run without issues. In most cases, this would just mean still crashing (so no worse behavior), unless a plugin manages to mess up the internal state of Wayfire in an interesting way.
 - Avoiding a crash when using an invalid option might make things more complicated for plugin developers (although it should be clear that the plugin was not loaded).


ammen99 (approved):
Feels a bit hacky but I can't figure out a better way to do it (one which allows us to print the stacktrace properly in case we load an option later without having a flag like that). Thanks!

ammen99:
> * This might lead to UB if the plugin tries to use the (now uninitialized) options? (I did not check if they are initialized to some defaults).
> * Also, this could lead to crashes / UB if plugins' `fini()` function or destructor relies on the `init()` function and / or the constructor having run without issues. In most cases, this would just mean still crashing (so no worse behavior), unless a plugin manages to mess up the internal state of Wayfire in an interesting way.

I am not sure I understand what you mean by this?

dkondor:
> > * This might lead to UB if the plugin tries to use the (now uninitialized) options? (I did not check if they are initialized to some defaults).
> > * Also, this could lead to crashes / UB if plugins' `fini()` function or destructor relies on the `init()` function and / or the constructor having run without issues. In most cases, this would just mean still crashing (so no worse behavior), unless a plugin manages to mess up the internal state of Wayfire in an interesting way.
> 
> I am not sure I understand what you mean by this?

I was mainly thinking if the plugin class has a member variable that is now left uninitialized due to the constructor / init throwing, but then is accessed by the destructor. Also if init sets up something that is not properly cleaned up by the destructor, but we don't call fini (although this could be a bug in the plugin). I'm not sure how likely it is to hit these in practice (or they might not be relevant actually).

dkondor:
> FWIW I think we could try to dlclose() in case the loading failed, what do you think? Is it worth it?

I have no idea :) But I can try adding

ammen99:
> I was mainly thinking if the plugin class has a member variable that is now left uninitialized due to the constructor / init throwing, but then is accessed by the destructor. Also if init sets up something that is not properly cleaned up by the destructor, but we don't call fini (although this could be a bug in the plugin). I'm not sure how likely it is to hit these in practice (or they might not be relevant actually).

Ah, so you mean like if the init method loads the options, and it fails yeah .. I guess that could be a problem, sure, if we register random callbacks. That's the problem with carrying on legacy interfaces which don't use RAII :(

dkondor:
> Good PR, but what about validating the metadata before loading the plugin? If the metadata is invalid, simply don't attempt to load it?

The problem is if the two get out of sync, e.g. the user installs (or possibly updates) a plugin while Wayfire is running: in this case, the metadata will not be reloaded. Users should not do this, but it is better to try to avoid the crash.

dkondor:
> > I was mainly thinking if the plugin class has a member variable that is now left uninitialized due to the constructor / init throwing, but then is accessed by the destructor. Also if init sets up something that is not properly cleaned up by the destructor, but we don't call fini (although this could be a bug in the plugin). I'm not sure how likely it is to hit these in practice (or they might not be relevant actually).
> 
> Ah, so you mean like if the init method loads the options, and it fails yeah .. I guess that could be a problem, sure, if we register random callbacks. That's the problem with carrying on legacy interfaces which don't use RAII :(

Yes, currently if `init()` fails, I don't try to call `fini()` -- actually I could try doing it, hoping that it does not do something weird, but a crash can happen either way. Nevertheless, I think this is still an improvement, since now only limited cases will trigger a crash as opposed to always.

dkondor:
> > FWIW I think we could try to dlclose() in case the loading failed, what do you think? Is it worth it?
> 
> I have no idea :) But I can try adding

I've added this with the last commit, but it is actually not working for me: even if I update the plugin, Wayfire keeps using the old .so; however, since you have the `enable_so_unloading` option, I assume this is working for some users :)

killown:
> > Good PR, but what about validating the metadata before loading the plugin? If the metadata is invalid, simply don't attempt to load it?
> 
> The problem is if the two get out of sync, e.g. the user installs (or possibly updates) a plugin while Wayfire is running: in this case, the metadata will not be reloaded. Users should not do this, but it is better to try to avoid the crash.

Your solution is good and achieves the goal, but I was referring to something like this:

```
/* then skip plugin load if outdated */
bool is_metadata_up_to_date(const std::string& plugin_name) {
    std::string metadata_path = get_metadata_path(plugin_name);
    std::string so_path = get_plugin_so_path(plugin_name);

    if (!std::filesystem::exists(metadata_path)) {
        LOGW("Metadata file for plugin '", plugin_name, "' does not exist.");
        return false;
    }

    auto metadata_mtime = std::filesystem::last_write_time(metadata_path);
    auto so_mtime = std::filesystem::last_write_time(so_path);

    if (so_mtime > metadata_mtime) {
        LOGW("Metadata for plugin '", plugin_name, "' is outdated.");
        return false;
    }

    LOGI("Metadata for plugin '", plugin_name, "' is up-to-date.");
    return true;
}
```

But don't worry, I was just wondering about this solution xD

dkondor:
> Your solution is good and achieves the goal, but I was referring to something like this:
> 

I see, in some way, this seems "cleaner" by not loading code that will not work anyway. However, it is possible to have an up-to-date metadata installed, but not loaded by `compositor_core_t::config` (as far as I can tell, it only reads metadata once, when starting Wayfire).

dkondor:
> > > I was mainly thinking if the plugin class has a member variable that is now left uninitialized due to the constructor / init throwing, but then is accessed by the destructor. Also if init sets up something that is not properly cleaned up by the destructor, but we don't call fini (although this could be a bug in the plugin). I'm not sure how likely it is to hit these in practice (or they might not be relevant actually).
> > 
> > 
> > Ah, so you mean like if the init method loads the options, and it fails yeah .. I guess that could be a problem, sure, if we register random callbacks. That's the problem with carrying on legacy interfaces which don't use RAII :(
> 
> Yes, currently if `init()` fails, I don't try to call `fini()` -- actually I could try doing it, hoping that it does not do something weird, but a crash can happen either way. Nevertheless, I think this is still an improvement, since now only limited cases will trigger a crash as opposed to always.

OK, after rethinking, now I'm thinking that it is probably better to call `fini()` even if `init()` failed. A typical thing plugins might do in `init()` is to add a bunch of bindings in a row, e.g.
```
output->add_button(wf::option_wrapper_t<wf::buttonbinding_t>("plugin/some_button"), &btn_func);
output->add_key(wf::option_wrapper_t<wf::keybinding_t>{"plugin/some_key"), &key_func);
...
```

So it is important to disconnect these (which is safe to try even for unconnected bindings). I'm adding a commit to do this, but let me know if you prefer the original instead.

dkondor:
> > > FWIW I think we could try to dlclose() in case the loading failed, what do you think? Is it worth it?
> > 
> > 
> > I have no idea :) But I can try adding
> 
> I've added this with the last commit, but it is actually not working for me: even if I update the plugin, Wayfire keeps using the old .so; however, since you have the `enable_so_unloading` option, I assume this is working for some users :)

Actually, it is working, I forgot about the `workarounds/enable_so_unloading` option :)

PR #2620, opened 2025-04-09
ipc-rules: fix wset-index garbage values

### LINKED ISSUE #2368: [IPC] Various issues with IPC
**Clear description of the problem**
One of the common problems I am facing with IPC is garbage values for various properties. this generally happens because the said property is not set by wayfire. A good example is `view-app-id-changed` event. The first time this event is emitted, Wayfire has not yet "placed" the view. Therefore, it does not have a wset-index. And the json data contains garbage values like: `"wset-index": 18446744073709551615`.

This is not an isolated incident. Such behaviour can be observed in cases like moving a window with click-and-drag. The move plugin does not (probably, cannot) set the "from" workspace. Therefore when a window is moved, `view-workspace-changed` is emitted with garbage values for "from" workspace, much like the example above.

**What can be done**
For events like `view-app-id-changed`, we can completely ignore all the fields other than `"app-id"`, but that is not the most ideal solution. 

In general, it would be preferable to explicitly tell the user that the value of a field is invalid. For example, "wset-index" is, as a rule, never zero. The workspace indices can never be negative. So instead of letting garbage values being filled in, we can simply set them to these invalid values. This way, the events can be safely parsed and invalid values can be ignored.

killown:
fix #2368

### DIFF
--- a/plugins/ipc-rules/ipc-rules-common.hpp
+++ b/plugins/ipc-rules/ipc-rules-common.hpp
@@ -195,17 +195,22 @@ static inline wf::json_t view_to_json(wayfire_view view)
     description["minimized"]   = toplevel ? toplevel->minimized : false;
     description["activated"]   = toplevel ? toplevel->activated : false;
     description["sticky"]     = toplevel ? toplevel->sticky : false;
-    description["wset-index"] = [&]() -> int64_t {
-        if (!toplevel || !toplevel->get_wset()) {
+    description["wset-index"] = [&] () -> int64_t
+    {
+        if (!toplevel || !toplevel->get_wset())
+        {
             return -1;
         }
+
         auto index = toplevel->get_wset()->get_index();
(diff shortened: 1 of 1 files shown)


plugins/ipc-rules/ipc-rules-common.hpp
```diff
@@ -195,8 +195,22 @@ static inline wf::json_t view_to_json(wayfire_view view)
     description["minimized"]   = toplevel ? toplevel->minimized : false;
     description["activated"]   = toplevel ? toplevel->activated : false;
     description["sticky"]     = toplevel ? toplevel->sticky : false;
-    description["wset-index"] = toplevel && toplevel->get_wset() ? toplevel->get_wset()->get_index() : -1;
```
ammen99, line 198: It would be enough to cast the result of get_index to `int64_t` instead of writing a long lambda xD

ammen99 (approved):
thanks :)

PR #2621, opened 2025-04-09
Fix crash when applying window rules to constrained dialogs

### LINKED ISSUE #2619: wayfire crashing when accessing firefox.desktop file
**Describe the bug**
attempting to get properties of file with thunar crashes wayfire

**To Reproduce**
navigate to ~/.local/share/applications/firefox.dekstop
right click file and select properties
wayfire crashes

**Wayfire version**
wayfire 0.10.0 branch master wlroots-0.18.2
commit cee51fd0c5cac970d76ab61379fdee16bde39c6a

-Seems to work fine if i use pcmanfm though

[log2.txt](https://github.com/user-attachments/files/19624981/log2.txt)

killown:
Added logic to skip rule application for transient dialogs that are constrained to their parent window. 

The issue occurred because Wayfire tried to apply \`start_on_output\` rules to dialogs that could not leave their parent window. The fix ensures that such dialogs are ignored when applying output-changing rules.

fix #2619

### DIFF
--- a/plugins/window-rules/view-action-interface.cpp
+++ b/plugins/window-rules/view-action-interface.cpp
@@ -442,6 +442,11 @@ void view_action_interface_t::_start_on_output(std::string name)
         return;
     }
 
+    if (_view->parent)
+    {
+        return;
+    }
+
     if (_view->get_output() == output)
     {
         return;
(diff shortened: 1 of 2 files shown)


ammen99:
This doesn't seem to be the correct fix for the issue. We simply should not execute rules for setting the output of dialogs to any output other than their parent's output.

PR #2622, opened 2025-04-09
Add support for tilde (~) expansion in file paths

### LINKED ISSUE #2612: Cube plugin is unable to use the ~ in the file specification
**Describe the bug**
Whenever I use the cube plugin and I try to specify an image file that is located in one of my user directories for the skydome_texture config option, I get a green skydome, which I assume means the file was not found. I'm unsure if this problem is specific to just the cube plugin or to the entire Wayfire project, but I assume that the ~ shorthand, that's used for the user's home directory just doesn't work. Or is there just some alternative to the ~ that is used in the config?

tilda version
skydome_texture = ~/Documents/test.png <- this results in the green skydome (the file path is correct)

absolute path version:
skydome_texture = /home/harrisonfld/Documents/test.png <- this results in the correct skydome

**Steps to reproduce the behavior:**
1. Add the cube plugin to the ~/.config/wayfire.ini
2. Under the cube config heading specify:
```
[cube]
background_mode = skydome
skydome_texture = ~/someskydome.png
# replace someskydome.png with your actual test skydome

```
3. Use Wayfire, and try to use the cube view, should get a green skydome

**Expected behavior**
Skydome should just load, and the tilda should auto-expand into the user's home directory path.

**Screenshots or stacktrace**
~ version:
Image
absolute path version:
Image

**Wayfire version**
Installed from the chaotic AUR repo on Arch Linux
wayfire -v:
0.9.0-unknown (Jan 19 2025) branch unknown

killown:
- Adds `get_expanded_path` to expand ~ to the user's home directory.
- Modifies the cube plugin's skydome texture loading to support ~ in paths.

fix #2612

### DIFF
--- a/plugins/cube/skydome.cpp
+++ b/plugins/cube/skydome.cpp
@@ -4,7 +4,6 @@
 
 #include <wayfire/output.hpp>
 #include <wayfire/workspace-set.hpp>
-#include <wayfire/utils.hpp>
 
 #include <glm/gtc/matrix_transform.hpp>
 #include "shaders.tpp"
--- a/plugins/cube/skydome.cpp
+++ b/plugins/cube/skydome.cpp
@@ -4,7 +4,7 @@
 
 #include <wayfire/output.hpp>
 #include <wayfire/workspace-set.hpp>
-
+#include <wayfire/utils.hpp>
 
 #include <glm/gtc/matrix_transform.hpp>
 #include "shaders.tpp"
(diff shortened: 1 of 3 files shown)


soreau:
I haven't reviewed this but my first thought is that this might make more sense for wf-config to translate. This would mean that all wf-config users would benefit, i.e. wf-shell and wcm.

soreau:
> > I haven't reviewed this but my first thought is that this might make more sense for wf-config to translate. This would mean that all wf-config users would benefit, i.e. wf-shell and wcm.
> 
> That would require much more changes though as we'd have to add a whole new file path option type. Or do you mean to just move the helper function?

I'm not sure why you would think there would need to be a new option type.. the option type is string and a `hint` tag could be added to know if it's a file path. The `hint` tag is currently used by wcm to put a file chooser dialog button. [Here](https://github.com/WayfireWM/wayfire/blob/master/metadata/cube.xml#L115) is an example. I expect wf-config can simply parse hints and if it's a file, do the `wordexp()` dance.

ammen99:
> I'm not sure why you would think there would need to be a new option type.. the option type is string and a `hint` tag could be added to know if it's a file path. The `hint` tag is currently used by wcm to put a file chooser dialog button. [Here](https://github.com/WayfireWM/wayfire/blob/master/metadata/cube.xml#L115) is an example. I expect wf-config can simply parse hints and if it's a file, do the `wordexp()` dance.

This is essentially a new type of option, maybe a subclass of `option_t<std::string>`. I guess that when parsing the XML we could detect filetype hints, then instantiate that subclass instead of the regular `option_t<std::string>`. As it is a subclass, plugins can continue using the string option type in their code.

I agree this approach is better as it will automatically work for all users of wf-config. So @killown thanks for your PR but the approach suggested by @soreau is better, feel free to open another PR for wf-config if you want :)

PR #2626, opened 2025-04-12
xdg-activation: add focus stealing prevention when using tokens

dkondor:
Continuation of #2527

This implements focus stealing in a straightforward way by disallowing xdg-activation tokens if the view that generated them is not focused anymore. So this is about the case when the user switches to another view before a token is used.

### DIFF
--- a/metadata/xdg-activation.xml
+++ b/metadata/xdg-activation.xml
@@ -14,6 +14,11 @@
 			<_long>Whether to reject activation requests if a newer request has arrived since their creation.</_long>
 			<default>false</default>
 		</option>
+		<option name="focus_stealing_prevention" type="bool">
+			<_short>Prevent stealing focus by an activation request</_short>
+			<_long>Whether to reject an activation request if the user interacted with a different view since its creation.</_long>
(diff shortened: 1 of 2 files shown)


dkondor:
This works based on xdg-activation and only tracks which view has focus. Mainly, it is to prevent the following scenario:
1. User interacts with app A, which generates a token and passes it to app B
2. App B is slow to use its token, the user switches to app C in the meantime
3. Now app B uses the token to activate itself, which might be unexpected

In contrast, the plugin in wpe focuses on keyboard (if I understand correctly), to prevent taking focus while typing, regardless of whether the requesting app has a valid token or not. So I think these can complement each other, with this PR more focusing on users' attention and also reducing the cases that even trigger the fsp plugin in wpe.

dkondor:
Marking it a draft, since I this prevents focusing also in some valid cases, especially when used in combination with #2627 and desktop components, specifically when launching apps / actions from a menu opened from a layer-shell panel (such as with https://github.com/WayfireWM/wf-shell/pull/293)

PR #2627, opened 2025-04-12
Make focusing newly mapped views optional (strict focus stealing prevention)

dkondor:
Also continuation of #2527 and #2626, creating a (mostly) complete solution for focus stealing prevention.

This adds an option (`core/focus_on_map`) that determines whether newly mapped (opened) views are focused. The default is `true` which keeps the current behavior (as I imagine the newly introduced behavior would be too restrictive for most users). When set to false, views must explicitly use the xdg-activation protocol to gain focus, even when first mapped.

This PR is organized into 4 parts:
 - commit d3c872bea44e3271c7a8a6f56dd5c43725784c2a is just the same as #2626 (as I don't think there is much sense in merging this only)
 - the next two commits (b29982fea5fa72ccb81c2af6a257b257f63c6354 and 95ebff8bec296f531de9b0e815b776605631099d) implement this option
 - commit 65a1338efa754b9fd35db6cf2eb52142bbf693a5 adapts xdg-activation to be able to focus newly mapped views with a valid token
 - the last two commits (e7af9084a3a295950cf84da33f1484ac9e3a17d3 and 228478f15db7d97aa199b457f678ae80d6454607) generate and pass valid tokens to commands that are run by Wayfire, so that when starting GUI apps, they can still receive focus

This is currently a draft, since some functionality is still missing:
 - The main issue is that currently, newly mapped views are still stacked on top, but do not receive keyboard focus; ideally, they would be stacked below the currently active view, but I haven't yet figured out how to do this.
 - From a UI perspective, I actually don't like putting this option in `core`. it would be nice to collect all focus stealing / activation related options in one place (i.e. with xdg-activation); however, since this option needs to be accessed from core, I don't like the idea of putting it under a plugin. Maybe an alternative would be introducing one more signal, but that also sounds overkill.
 - Also, I'm unsure if this handles XWayland views correctly; I'm wondering if I should prevent focus [here](https://github.com/WayfireWM/wayfire/blob/099fea49800a5c2ed1a39f8c8513543e10362c18/src/view/xwayland/xwayland-unmanaged-view.hpp#L174) as well; in any case, I haven't tested with XWayland yet, but will do so next.
 - I'll need to do more testing, but potentially I want to separately track the active view for normal and self-generated tokens (my main concern is that some commands will not result in a new view opening, so they should not invalidate normal tokens, but this is of course and edge case).

### DIFF
--- a/metadata/core.xml
+++ b/metadata/core.xml
@@ -88,6 +88,11 @@
 			<_long>Whether to pass buttons through to the client when switching the focus via clicking</_long>
 			<default>true</default>
 		</option>
+		<option name="focus_on_map" type="bool">
+			<_short>Focus newly opened views</_short>
+			<_long>If unset, newly opened views will only be focused by using the xdg-activation protocol.</_long>
+			<default>true</default>
+		</option>
 		<option name="exit" type="key">
 			<_short>Wayfire Shutdown</_short>
 			<_long>Calls the shutdown routines for wayfire.</_long>
--- a/metadata/xdg-activation.xml
+++ b/metadata/xdg-activation.xml
@@ -14,6 +14,11 @@
 			<_long>Whether to reject activation requests if a newer request has arrived since their creation.</_long>
 			<default>false</default>
 		</option>
+		<option name="focus_stealing_prevention" type="bool">
+			<_short>Prevent stealing focus by an activation request</_short>
+			<_long>Whether to reject an activation request if the user interacted with a different view since its creation.</_long>
+			<default>true</default>
+		</option>
 		<option name="timeout" type="int">
 			<_short>Timeout for activation (in seconds)</_short>
 			<_long>Focus requests will be ignored if at least this amount of time has elapsed between creating and using it.</_long>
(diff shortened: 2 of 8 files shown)


killown:
I think it's easier with python and a few lines:
```

from wayfire import WayfireSocket
sock = WayfireSocket()

last_focused_view_id = sock.get_focused_view()["id"]

sock.watch(["view-mapped", "view-focused"])

while True:
    msg = sock.read_next_event()
    if msg["event"] == "view-mapped":
        sock.set_focus(last_focused_view_id)
    if msg["event"] == "view-focused":
        last_focused_view_id = sock.get_focused_view()["id"]
```

ammen99:
> I think it's easier with python and a few lines:
> 
> ```
> 
> from wayfire import WayfireSocket
> sock = WayfireSocket()
> 
> last_focused_view_id = sock.get_focused_view()["id"]
> 
> sock.watch(["view-mapped", "view-focused"])
> 
> while True:
>     msg = sock.read_next_event()
>     if msg["event"] == "view-mapped":
>         sock.set_focus(last_focused_view_id)
>     if msg["event"] == "view-focused":
>         last_focused_view_id = sock.get_focused_view()["id"]
> ```

While this is a good workaround in practice, we still switch the focus and then back. Depending on the app, this might not be what we want.

PR #2630, opened 2025-04-15
layer-shell: notify scenegraph of geometry changes

ammen99:
Updating the geometry triggers a recomputation of surface visibility. Without it, once the layer-shell view goes offscreen (for ex. during autohide), it is computed as not visible on an output and does not receive `wl_surface.frame` events, leading to choppy animations.

Fixes https://github.com/WayfireWM/wf-shell/issues/189
Fixes https://github.com/WayfireWM/wf-shell/issues/276
Fixes https://github.com/WayfireWM/wf-shell/issues/287


NamorNiradnug:
Thanks a lot!

PR #2631, opened 2025-04-16
fix incorrect ipc json keys

killown:


plugins/ipc-rules/ipc-rules.cpp
```diff
@@ -76,7 +76,12 @@ class ipc_rules_t : public wf::plugin_interface_t,
 
     wf::ipc::method_callback get_view_info = [=] (wf::json_t data)
     {
-        auto id = wf::ipc::json_get_uint64(data, "id");
+        auto id = wf::ipc::json_get_optional_uint64(data, "view-id");
+        if (!id.has_value())
+        {
+            // Deprecated API usage: 'id' field for view is deprecated.
+            id = wf::ipc::json_get_optional_uint64(data, "id");
```
ammen99, line 83: This one shouldn't be optional, if the first one fails, then we require `id`. Same in the other places.

plugins/ipc/ipc-helpers.hpp
```diff
@@ -55,7 +55,7 @@ WFJSON_GETTER_FUNCTION(bool, bool);
 #undef WFJSON_GETTER_FUNCTION
 
 
-inline wayfire_view find_view_by_id(uint32_t id)
+inline wayfire_view find_view_by_id(const std::optional<uint64_t>& id)
```
ammen99, line 58: this makes no sense at all as a change, the validation of id should be done before looking for the view, so that the user can differentiate. I.e if the user doesn't specify neither `id` nor `view-id`, then the error should be `missing id field` not `view not found` bc the ipc client did not even specify which client

killown:
The `client.configure_view` function is not working in this PR and requires additional work.

killown:
> The `client.configure_view` function is not working in this PR and requires additional work.

fixed

killown:
you can test the changes with WayfireWM/pywayfire#23

PR #2633, opened 2025-04-23
Fix include path

marcusbritanicus:
This adds `/usr/include/wayfire` to the -I include path.

### DIFF
--- a/plugins/ipc-rules/ipc-rules-common.hpp
+++ b/plugins/ipc-rules/ipc-rules-common.hpp
@@ -1,5 +1,5 @@
 #pragma once
-#include "plugins/ipc/ipc-helpers.hpp"
+#include "wayfire/plugins/ipc/ipc-helpers.hpp"
 #include <wayfire/output.hpp>
 #include <wayfire/workarea.hpp>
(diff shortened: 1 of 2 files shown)


marcusbritanicus:
@ammen99 Do have a look at [this](https://github.com/WayfireWM/wayfire/blob/master/plugins/ipc-rules/ipc-rules-common.hpp#L2). I use `#include <wayfire/plugins/ipc-rules/ipc-rules-common.hpp>`. The problem is that the compiler cannot find the header `plugins/ipc/ipc-helpers.hpp`. Hence this "fix".

marcusbritanicus:
> But the fix is to make this file include the header as `wayfire/plugins/ipc/ipc-helpers.hpp`

How will this help? If I include `wayfire/plugins/ipc/ipc-rules-common.hpp`, the compiler will search for `plugins/ipc/ipc-helpers.hpp` and will not find it.

PS: I initially thought including the header `wayfire/plugins/ipc/ipc-helpers.hpp` might work, but it does not.

marcusbritanicus:
Closing, as this is superceded by @kode54's #2640

PR #2638, opened 2025-05-06
resize: handle overflow due to margins

kode54:
Some apps, notably a lot of Electron apps, set maximum size to INT_MAX, which will overflow the int32_t type when the margins of the server side decorations are added on top.

Use a larger integer type to calculate margins, then clamp as necessary, based on the actual type.

Fixes 099fea49800a5c2ed1a39f8c8513543e10362c18

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,23 +21,15 @@ jobs:
     container:
         image: archlinux:latest
     steps:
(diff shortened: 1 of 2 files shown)


kode54:
Damn, I should have compile tested this first. I'll fix before I force push.

kode54:
I rebased this on top of my CI fix, so it should pass CI now, and should merge okay if that is merged first.

ammen99 (approved):
Alright, did not think that apps may report INT_MAX themselves, but it makes sense. Thanks :)

PR #2639, opened 2025-05-06
archlinux ci: stop using clang17 repository

kode54:
This should no longer be necessary now that clang 19 is in the repos.

Undoes 1ceafa69edb00b8bb373793bb1090446cc5f5550


ammen99 (approved):
Thanks :)

PR #2640, opened 2025-05-06
Include path fix

kode54:
This fixes an include path for ipc-rules plugin, for when the headers are installed system-wide and included from external projects. It should not break building Wayfire, and should fix any external plugins including the ipc-rules-common.hpp header from the system install. It prefers the local path first, so that internal building will prefer its own copy over the system copy.


kode54:
#2639 is included in this PR so that CI will pass, but that PR should be applied before either of my other current PRs.

kode54:
Crap. I forgot to rebase this on the CI fix.

ammen99:
I don't like putting plugins headers in the `src/` folder, otherwise this is the fix I envisioned ;) Please do it like `plugins/common` or `plugins/grid` do it.

marcusbritanicus:
@kode54 I think I wans't very clear in irc. What I meant is the following. Firstly, create the tree, `wayfire/plugins/ipc/` inside `plugins/ipc`. Then, move `ipc-activator.hpp`, `ipc-helpers.hpp`, and `ipc-method-repository.hpp` into `wayfire/plugins/ipc/` (which is inside `plugins/ipc`).

This is what @ammen99 also wants.

PR #2645, opened 2025-05-11
foreign-toplevel: Fix missing windows on hotplug

kode54:
This fixes a case where window references may no longer be tracked when an output is freshly attached to the compositor after logging in, causing apps such as wf-panel to fail to track buttons for any windows which are moved off of the existing output(s).

Thanks @soreau for producing this patch so quickly, which I am pretty sure I successfully uncrustify'd before committing with his author tag and pushing.

### DIFF
--- a/plugins/protocols/foreign-toplevel.cpp
+++ b/plugins/protocols/foreign-toplevel.cpp
@@ -63,6 +63,25 @@ class wayfire_foreign_toplevel
         wlr_foreign_toplevel_handle_v1_destroy(handle);
     }
 
+    void toplevel_send_state()
+    {
+        wlr_foreign_toplevel_handle_v1_set_maximized(handle,
+            view->pending_tiled_edges() == wf::TILED_EDGES_ALL);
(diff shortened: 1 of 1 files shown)


ammen99:
I am not sure I understand why this is needed exactly, can you elaborate which event(s) are missing otherwise? Why is this a bug in Wayfire and not in the client?

kode54:
I don't even remember why I filed this now. Someone else can deal with it.

ammen99:
I don't think this makes any sense. When we create a new output, wf-shell is supposed to pick this up via a new wl_output global. Then it creates a new panel instance, which in turn creates a new window-list widget. The new window-list widgets ought to bind to the foreign-toplevel global again, which WILL trigger a re-send of all toplevel details: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_foreign_toplevel_management_v1.c#L603-632

kode54:
And there are no toplevels to send for because when all the outputs were destroyed, so were the windows.

PR #2646, opened 2025-05-12
Add ipc method to send view to workspace

marcusbritanicus:


plugins/vswitch/vswitch.cpp
```diff
@@ -609,6 +611,67 @@ class wf_vswitch_global_plugin_t : public wf::per_output_plugin_t<vswitch>
 
         return wf::ipc::json_ok();
     };
+
+    wf::ipc::method_callback send_view = [=] (const wf::json_t& data)
+    {
+        uint64_t x = wf::ipc::json_get_uint64(data, "x");
+        uint64_t y = wf::ipc::json_get_uint64(data, "y");
+        uint64_t output_id = wf::ipc::json_get_uint64(data, "output-id");
+        std::optional<uint64_t> view_id = wf::ipc::json_get_optional_uint64(data, "view-id");
```
ammen99, line 620: I am a bit confused, if the method is `send-view` then why is the view optional? IPC clients can call `set-workspace` if they want to switch without a view ...

plugins/vswitch/vswitch.cpp
```diff
@@ -609,6 +611,67 @@ class wf_vswitch_global_plugin_t : public wf::per_output_plugin_t<vswitch>
...
+        {
+            return wf::ipc::json_error("Cannot grab unmapped view!");
+        }
+
+        if (view->get_output() != wo)
+        {
+            return wf::ipc::json_error("Cannot grab view on a different output!");
+        }
+
+        wf::view_change_workspace_signal signal;
```
ammen99, line 656: I believe that pretty much the same code has been implemented already in the same file, maybe we can have a shared function instead of duplicating the code?

ammen99 (changes requested):

plugins/vswitch/vswitch.cpp
```diff
@@ -24,6 +24,31 @@ class workspace_animation_t : public duration_t
     timed_transition_t dy{*this};
 };
 
+/**
+ * A mall helper function to move a view and it's children to workspace @to_ws.
```
ammen99, line 28:
```suggestion
 * A small helper function to move a view and it's children to workspace @to_ws.
```
```suggestion
 * A mall helper function to move a view and it's children to workspace @to_ws.
```

plugins/vswitch/vswitch.cpp
```diff
@@ -609,6 +637,37 @@ class wf_vswitch_global_plugin_t : public wf::per_output_plugin_t<vswitch>
...
+        {
+            return wf::ipc::json_error("Invalid view or view not toplevel!");
+        }
+
+        if (!view->is_mapped())
+        {
+            return wf::ipc::json_error("Cannot grab unmapped view!");
+        }
+
+        auto grid_size = view->get_wset()->get_workspace_grid_size();
```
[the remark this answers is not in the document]
marcusbritanicus: Done.

marcusbritanicus:
> lgtm, needs a code style fix and then we're good to go :)

Just uncrustified. Sorry, I had forgotten about this, been busy with work. :(

PR #2651, opened 2025-05-19
animate: Improve squeezimize shader

soreau:
This fixes a few different problems with squeezimize animation:
- Matches target width, even if surface width is smaller than target
- Fixes animation going out of damage bounds
- Fix disappearing surfaces when panel is on the bottom


ammen99 (approved):
i don't know the shader in detail, but I'll trust your judgement :)

PR #2655, opened 2025-05-24
ipc-rules: Fix view-fullscreen event

### LINKED ISSUE #2653: Feature request: Condition for Keycolor plugin / Global Alpha with exceptions
I want to implement a transparency and blur effect across desktop. Wayfire features the blur perfectly! 
But the transparency or alpha, as far as I understand isn't able to become global with exceptions on some windows (I wish I could do that so that's a feature request in itself) Alpha can only be adjusted temporarily for a window and it's not global.

So in my attempt I figured out that I can achieve a similar effect for a global transparency through adjusting the Keycolor plugin. Except unfortunately it doesn't have exceptions and neither does it have a toggle bind. I tried putting "toggle = " under it in the config but it didn't work. 

The reason I need to toggle, or in a perfect world make exceptions either for windows or for fullscreens is that I don't want transparency on games and when watching videos.
So what would be perfect is for either Keycolor plugin or the Alpha plugin to have Fullscreen exception or have specific window exception or at least have global toggle.
It would make for such eye candy to be able to do it and I hope we can have it!
Thank you!

soreau:
Closes #2653.

### DIFF
--- a/plugins/ipc-rules/ipc-events.hpp
+++ b/plugins/ipc-rules/ipc-events.hpp
@@ -118,7 +118,7 @@ class ipc_rules_events_methods_t : public wf::per_output_tracker_mixin_t<>
 
         {"view-tiled", get_generic_output_registration_cb(&_tiled)},
         {"view-minimized", get_generic_output_registration_cb(&_minimized)},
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
thanks, nice catch!

PR #2657, opened 2025-05-25
idle: Fix disable-initially option functionality

### LINKED ISSUE #2647: `idle/disable_initially` doesn't work
**Describe the bug**
When `idle` plugin is configured with `disable_initially=true`, it doesn't work, i.e. the `cube` effect is still activated after the timeout.

**Debug information**
When first entering the wayfire environment, it logs
```
creating idle inhibitor 0x55b2b8349f70 previous count: 0
```
But it seems that this idle inhibitor doesn't work. I then fullscreened a window and restored, and it logs
```
creating idle inhibitor 0x55b2b83a9360 previous count: 1
destroying idle inhibitor 0x55b2b83a9360 previous count: 2
```
Now the initial idle inhibitor `0x55b2b8349f70` works and the `cube` effect is never activated.

**Wayfire version**
`0.9.0`, Gentoo ebuild repository: [`gentoo/gui-wm/wayfire/wayfire-0.9.0.ebuild`](https://gitweb.gentoo.org/repo/gentoo.git/tree/gui-wm/wayfire/wayfire-0.9.0.ebuild)

soreau:
Fixes #2647.


plugins/single_plugins/idle.cpp
```diff
@@ -218,20 +220,28 @@ class wayfire_idle_plugin : public wf::per_output_plugin_instance_t
...
+        {
+            create_screensaver_timeout();
+        };
+
+        if (disable_initially)
+        {
+            return;
+        }
+
         update_fullscreen();
```
ammen99, line 235: I think we still need to call update_fullscreen even if it is initially disabled, so that we can properly keep track of the state (if i remember enough about the plugin, it is one of the plugins I have least idea about how it works).

ammen99: actually it doesn't seem to matter, reading the code. Thanks :)

ammen99 (approved):
LGTM, thanks :)

PR #2666, opened 2025-06-11
meson: Improve devenv support

kennylevinsen:
`meson devenv -C build` allows entering a build folder with all environment variables prepared for easy execution without system-wide installation, automatically fixes things like the LD_LIBRARY_PATH.

Wayfire's plugin and backend loader systems rely on their own path discovery and configuration that devenv is unaware of, and so wayfire cannot launch out of the box in a devenv.

Adjust meson build files to augment WAYFIRE_PLUGIN_PATH and WAYFIRE_PLUGIN_XML_PATH so that plugins and XML files work by default. Add the WAYFIRE_DEFAULT_CONFIG_BACKEND environment variable to allow the devenv to provide a default for that as well, as it does not live together with the plugins.

With this, running wayfire straight from the repository is as simple as:

    meson setup build
    ninja -C build
    meson devenv -C build
    ./src/wayfire

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -290,6 +290,10 @@ int main(int argc, char *argv[])
     std::vector<std::string> extended_debug_categories;
     bool allow_root = false;
 
+    if (char *default_config_backend = getenv("WAYFIRE_DEFAULT_CONFIG_BACKEND")) {
+        config_backend = default_config_backend;
+    }
+
     int c, i;
     while ((c = getopt_long(argc, argv, "c:B:d::DhRlrv", opts, &i)) != -1)
     {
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -290,6 +290,10 @@ int main(int argc, char *argv[])
     std::vector<std::string> extended_debug_categories;
     bool allow_root = false;
 
+    if (char *default_config_backend = getenv("WAYFIRE_DEFAULT_CONFIG_BACKEND")) {
```
ammen99, line 293:
```suggestion
    if (char *default_config_backend = getenv("WAYFIRE_DEFAULT_CONFIG_BACKEND"))
    {
```

ammen99 (approved):
Didn't know that meson has support for this, thanks!

PR #2675, opened 2025-06-16
squeezimize: create gles2 texture before gles2 subpass

### LINKED ISSUE #2673: Squeezemize animation does not work anymore
After new vulkan-capable rendering merged - this animation does not paint anything on screen. But seems to be activated - if I put a long duration for it selectively it takes time to un-minizmize window.

ammen99:
Avoids unwanted GL state changes.
Fixes #2673


mark-herbert42:
tested - works OK

PR #2676, opened 2025-06-16
Optimize setting cursors

### LINKED ISSUE #1015: Optimize setting cursors
We should not try to set the cursor image too often with `core::set_cursor`. We could cache the last cursor name (and clear that cache when a client sets the cursor, or when we hide the cursor), so that plugins don't care about how often they set the cursor.

Reason: setting the cursor very often results in lots of uploads of the cursor texture to the GPU.

killown:
trying to close #1015


ammen99 (approved):
thanks!

PR #2682, opened 2025-06-21
Add min-width and min-height options to the resize plugin.

### LINKED ISSUE #488: [resize plugin] make minimal size when resizing configurable
If wayfire SSD is used (mostly xwayland), the minimal size of a window should be the available buttons + 3-4 letters space as width and height of the window deco buttons.

Currently it is possible if the application does not have itself minimum size to resize the window until it becomes almost un-targetable, basically a square

killown:
close #488 

wcm-626

### DIFF
--- a/metadata/resize.xml
+++ b/metadata/resize.xml
@@ -10,6 +10,24 @@
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
 
+        <option name="min_size" type="bool">
+            <_short>Enable minimum view size</_short>
+            <_long>If enabled, views cannot be resized smaller than a usable size.</_long>
+            <default>true</default>
+        </option>
+        
+        <option name="min_width" type="int">
+            <_short>Minimal view width</_short>
+            <_long>If min_size is true, this sets a fixed minimum width. If not set, calculated based on buttons and margins.</_long>
+            <default>0</default>
+        </option>
+        
+        <option name="min_height" type="int">
+            <_short>Minimal view height</_short>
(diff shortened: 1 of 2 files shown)


ammen99:
I don't think we need a boolean option. If users don't want to set min size, they can just set min_width/height to 0. This effectively disables the feature.

plugins/single_plugins/resize.cpp
```diff
@@ -308,6 +311,16 @@ class wayfire_resize : public wf::per_output_plugin_instance_t, public wf::point
         min_size = wf::expand_dimensions_by_margins(min_size,
             view->toplevel()->pending().margins);
 
+        if (user_min_width)
+        {
+            min_size.width = user_min_width;
```
ammen99, line 316:
```suggestion
            min_size.width = std::max(user_min_width, min_size.width);
```

Idea: maybe the user says min size 50pixels, but app says min size 200 pixels, we can't resize lower than 200 pixels.

metadata/resize.xml
```diff
@@ -9,7 +9,19 @@
 			<_long>When the specified button is held down, you can drag a window to resize it.</_long>
 			<default>&lt;super&gt; BTN_RIGHT</default>
 		</option>
-
+        
+        <option name="min_width" type="int">
```
ammen99, line 13: indentation seems wrong, at least in the github UI :)

killown: lets fix it

ammen99 (approved):
Thanks!

PR #2683, opened 2025-06-22
workaround to disable middle click

### LINKED ISSUE #2425: Disable middle click / primary selection
I couldn't find anything in the settings for this. But do see where support was added https://github.com/WayfireWM/wayfire/issues/447

Maybe for reference this issue was adding an option to disable it in sway: https://github.com/swaywm/sway/issues/4511

Also it's an option in KDE settings.

It can seemingly be a decisive issue for linux users. Some strongly for middle click paste, some strongly against. For me it messes with me having it on, frequently ending up with pieces of clipboard randomly in my text files (maybe an accidental middle click while scrolling, or trying to close a tab), and also there's the chance of me accidentally pasting sensitive info in a random webpage/form/app

killown:
Close #2425 

If apps internally handle middle-click paste, there's nothing Wayfire can do about that: https://github.com/swaywm/sway/issues/4511#issuecomment-2561951039

### DIFF
--- a/metadata/workarounds.xml
+++ b/metadata/workarounds.xml
@@ -78,5 +78,10 @@
 			<_long>Maximum size in pixels of graphics buffers used for rendering. Needs to be set lower on some systems to avoid crashes and other issues.</_long>
 			<default>16384</default>
 		</option>
+		<option name="disable_primary_selection" type="bool">
+			<_short>Disable primary selection</_short>
+			<_long>Disable primary selection (middle-click copy/paste).</_long>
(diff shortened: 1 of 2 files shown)


soreau:
Can't you just make a binding with command plugin for middle click to run a command like `true`?

ammen99:
I think that would disable middle-click entirely, whereas the current PR only disables the primary selection protocol, leaving middle button intact for use in the other cases where it is used (for example blender, etc.)

PR #2688, opened 2025-06-25
Add keybinding to toggle tiling in current workspace

### LINKED ISSUE #2317: Switch to tile by key
I wonder if it was possible with IPC to switch to tile mode or vice versa by pressing a key?

for example being able to switch between the two modes or even remove the tile window decoration

killown:
close #2317

**feature:**
Toggles tiling mode for all currently tileable views from the current workspace. When enabled, eligible views are added to the tiling layout. 		When disabled, tiled views are removed from the layout and maximized

### DIFF
--- a/plugins/tile/tile-plugin.cpp
+++ b/plugins/tile/tile-plugin.cpp
@@ -276,36 +276,38 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
     {
         auto output = this->output;
         auto wset   = output->wset();
+        auto views = wset->get_views(wf::WSET_MAPPED_ONLY | wf::WSET_CURRENT_WORKSPACE);
         bool any_tiled = false;
-
-        for (auto& view : wset->get_views())
-        {
+    
+        for (auto& view : views)
+        {  
             if (tile::view_node_t::get_node(view))
             {
                 any_tiled = true;
                 break;
             }
         }
-
-        for (auto& view : wset->get_views())
-        {
-            if (!view->is_mapped() || !can_tile_view(view))
+     
+        for (auto& view : views)
+        {   
+            if (!can_tile_view(view))
             {
                 continue;
             }
-
+    
             auto node = tile::view_node_t::get_node(view);
             if (any_tiled && node)
             {
                 detach_view(view);
                 wf::get_core().default_wm->tile_request(view, 0);
                 wf::get_core().default_wm->tile_request(view, wf::TILED_EDGES_ALL);
-            } else if (!any_tiled && !node)
+            }
+            else if (!any_tiled && !node)
             {
                 attach_view(view);
             }
         }
-
+    
         return true;
     };
 
--- a/plugins/tile/tile-plugin.cpp
+++ b/plugins/tile/tile-plugin.cpp
@@ -276,38 +276,37 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
     {
         auto output = this->output;
         auto wset   = output->wset();
-        auto views = wset->get_views(wf::WSET_MAPPED_ONLY | wf::WSET_CURRENT_WORKSPACE);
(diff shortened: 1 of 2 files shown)


plugins/tile/tile-plugin.cpp
```diff
@@ -271,6 +272,48 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
         });
     };
 
+    wf::key_callback on_toggle_tile_current_ws = [=] (auto)
+    {
+        auto last_focused_view = wf::get_core().seat->get_active_view();
```
ammen99, line 277: We need to check that we can use the simple-tile plugin (to avoid for example triggering this while another plugin  is active, like expo or move-drag, changing the window during those operations could result in invalid compositor state. See how this is done in `conditioned_view_execute`, `especially the can_activate_plugin` check.

plugins/tile/tile-plugin.cpp
```diff
@@ -271,6 +272,48 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
...
+            if (!can_tile_view(view))
+            {
+                continue;
+            }
+
+            auto node = tile::view_node_t::get_node(view);
+            if (any_tiled && node)
+            {
+                detach_view(view);
+                wf::get_core().default_wm->tile_request(view, 0);
```
ammen99, line 304: I don't think the default handler should automatically maximize all views. If the user wants this, they could add a script, but I think in general it makes sense to switch between a tiling and a 'floating' layout (which doesn't mean all maximized).

Also btw setting edges to 0 and then to ALL seems rather pointless, one could put them to ALL directly (but I don't think we want this code here anyway).

plugins/tile/tile-plugin.cpp
```diff
@@ -348,6 +391,7 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
         output->add_button(button_move, &on_move_view);
         output->add_button(button_resize, &on_resize_view);
         output->add_key(key_toggle_tile, &on_toggle_tiled_state);
+        output->add_key(key_toggle_tile_current_ws, &on_toggle_tile_current_ws);
```
ammen99, line 394: This code uses keybindings which is possible for sure but I think the original issue wanted to have this as an IPC option. Same for key_toggle_tile. It would be easier to use ipc activators (example: https://github.com/WayfireWM/wayfire/pull/1926). The ipc activators need to be put in the global tile plugin (as there is one IPC socket, but multiple outputs).

killown: What I gathered from his messages was that he wanted "the easiest way to switch", likely choosing the IPC path because it seemed simpler than requesting plugin code changes https://github.com/WayfireWM/wayfire/issues/2317#issuecomment-2041451328 but ok, we could go for the IPC option

plugins/tile/tile-plugin.cpp
killown: Switching to floating mode works for developers but may annoys users, views could appear in random positions making an extra work to manually maximize it to fix. This maximize change reflects real usage, and the behavior you describe is exactly what Sway does, leading to the same script-heavy fixes. I could remove it, but as an end user, I’d disable the plugin after the PR is finished. Sway’s behavior isn’t viable in Wayfire, and here’s why. 

First I will show how it works in hyprland:

https://github.com/user-attachments/assets/555d419c-af22-49cf-9d55-68bb4a052df9

The view goes to the top with "bind = ALT, Tab, bringactivetotop" It doesn’t make it float, but it covers the entire output area, a desired behavior for many users. 

Sure, you could fix it with a script, but in the meantime, we’re stuck with this old view glitch bug in simple-tile, which is configuring a view to cover all  available screen, the view content will be stretched first than the whole view is set for the new geometry and the content will look like wide until the new geometry is set, which is very annoying.

PR #2689, opened 2025-06-25
Add keybind for maximizing tiled views

killown:


ammen99 (changes requested):
- We also have to be careful, when we have a maximized view, we cannot do any interactive move or resize (as the maximized view covers all the others). So we need to disallow interactive move/resize in these cases. This is same as the fullscreen case, so see the `has_fullscreen_view()` function, it just needs to have the show_maximized check as well.

- We need to consider what we want to happen when we switch focus. For example say we show_maximized view A, if we switch focus with say alt-tab or use the scale plugin and we focus another view B, what should happen? Do we unmaximize A (this is what happens with fullscreen)? Note that we already have a function to track focus changes, so we could do it like this, when we focus a new view, we check on the current workspace (same principle as has_fullscreen_view) if we can find a view which is shown maximized but is not the new focus. If yes, we unmaximize the view.

- What about the arrow keys, if we navigate to view left/right, do we want an option like for fullscreen, that when we switch to adjacent view, the current one stops being maximized and the next one is maximized? We could leave this as a feature for another PR if you feel it is too much.

plugins/tile/tree.cpp
```diff
@@ -458,10 +467,18 @@ void view_node_t::set_geometry(wf::geometry_t geometry, wf::txn::transaction_upt
     }
 
     wf::get_core().default_wm->update_last_windowed_geometry(view);
-    view->toplevel()->pending().tiled_edges = TILED_EDGES_ALL;
```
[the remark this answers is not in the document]
killown: solved thanks

killown:
> * We also have to be careful, when we have a maximized view, we cannot do any interactive move or resize (as the maximized view covers all the others). So we need to disallow interactive move/resize in these cases. This is same as the fullscreen case, so see the `has_fullscreen_view()` function, it just needs to have the show_maximized check as well.
> 
>     * We need to consider what we want to happen when we switch focus. For example say we show_maximized view A, if we switch focus with say alt-tab or use the scale plugin and we focus another view B, what should happen? Do we unmaximize A (this is what happens with fullscreen)? Note that we already have a function to track focus changes, so we could do it like this, when we focus a new view, we check on the current workspace (same principle as has_fullscreen_view) if we can find a view which is shown maximized but is not the new focus. If yes, we unmaximize the view.
> 
>     * What about the arrow keys, if we navigate to view left/right, do we want an option like for fullscreen, that when we switch to adjacent view, the current one stops being maximized and the next one is maximized? We could leave this as a feature for another PR if you feel it is too much.



> * We also have to be careful, when we have a maximized view, we cannot do any interactive move or resize (as the maximized view covers all the others). So we need to disallow interactive move/resize in these cases. This is same as the fullscreen case, so see the `has_fullscreen_view()` function, it just needs to have the show_maximized check as well.
> 
>     * We need to consider what we want to happen when we switch focus. For example say we show_maximized view A, if we switch focus with say alt-tab or use the scale plugin and we focus another view B, what should happen? Do we unmaximize A (this is what happens with fullscreen)? Note that we already have a function to track focus changes, so we could do it like this, when we focus a new view, we check on the current workspace (same principle as has_fullscreen_view) if we can find a view which is shown maximized but is not the new focus. If yes, we unmaximize the view.
> 
>     * What about the arrow keys, if we navigate to view left/right, do we want an option like for fullscreen, that when we switch to adjacent view, the current one stops being maximized and the next one is maximized? We could leave this as a feature for another PR if you feel it is too much.

>  switch to adjacent view, the current one stops being maximized and the next one is maximized? We could leave this as a feature for another PR if you feel it is too much.

Thanks a lot for all recommendations, you might agree that this PR **should not** be merged before the next wayfire release. As you described possible issues, let it sit for a while then we could implement such features: **"switch to adjacent view, the current one stops being maximized and the next one is maximized"** .

plugins/tile/tile-plugin.cpp
```diff
@@ -271,28 +289,58 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
         });
     };
 
+    wf::key_callback on_toggle_maximized = [=] (auto)
+    {
+        auto view = wf::get_core().seat->get_active_view();
```
ammen99, line 294: You should use the `conditioned_view_execute` helper function here (see how the other functions use it). This ensures the binding is not toggled while another plugin is active.

plugins/tile/tile-plugin.cpp
```diff
@@ -271,28 +289,58 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
...
+        auto toplevel = toplevel_cast(view);
+        // skip if the workspace has no views
+        if (!toplevel)
+        {
+            return false;
+        }
+        auto node = tile::view_node_t::get_node(toplevel);
```
ammen99, line 301: It is possible that `node` is null here, because we allow both floating and tiled views. So if the focused toplevel happens to not be tiled, node will be null and this will crash.

plugins/tile/tile-plugin.cpp
```diff
@@ -271,28 +289,58 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
...
+            auto current_node = tile::view_node_t::get_node(view);
+            bool was_maximized = current_node->show_maximized;
+    
+            if (was_maximized)
             {
-                /* This will lower the fullscreen status of the view */
-                view_bring_to_front(adjacent->view);
-                wf::get_core().seat->focus_view(adjacent->view);
-
-                if (was_fullscreen && keep_fullscreen_on_adjacent)
```
ammen99, line 288: either i am missing something or the fullscreen case was deleted entirely :)

plugins/tile/tile-plugin.cpp
```diff
@@ -348,6 +396,7 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
         output->add_button(button_move, &on_move_view);
         output->add_button(button_resize, &on_resize_view);
         output->add_key(key_toggle_tile, &on_toggle_tiled_state);
+        output->add_key(key_toggle_maximized, &on_toggle_maximized);
```
ammen99, line 399: you need to also remove the binding

plugins/tile/tree.cpp
```diff
@@ -372,6 +373,27 @@ void view_node_t::set_gaps(const gap_size_t& size)
...
 
+wf::geometry_t adjust_geometry_for_workspace(
+    std::shared_ptr<wf::workspace_set_t> wset,
+    const wf::geometry_t& base_geometry,
+    const wf::geometry_t& local_geometry)
+{
+    auto vp = wset->get_current_workspace();
+    auto size = wset->get_last_output_geometry().value_or(wf::tile::default_output_resolution);
+
+    int source_x = std::floor(1.0 * (local_geometry.x + vp.x * size.width) / size.width);
```
ammen99, line 384: I am wondering, why do you add vp.x * size.width, then two lines below, you subtract vp.x, this simply cancels out, doesn't it?

plugins/tile/tree.cpp
```diff
@@ -388,16 +410,17 @@ wf::geometry_t view_node_t::calculate_target_geometry()
...
-        int view_vp_x = std::floor(1.0 * geometry.x / size.width);
-        int view_vp_y = std::floor(1.0 * geometry.y / size.height);
-
-        local_geometry = {
-            (view_vp_x - vp.x) * size.width,
-            (view_vp_y - vp.y) * size.height,
-            size.width,
-            size.height,
-        };
+        auto base_geometry = wset->get_attached_output()->get_relative_geometry();
```
ammen99, line 413: the base_geometry should be the size variable from above :)

issue is wset->get_attached_output() could be null in very very specific cases (i think you can only get that with some weird sequences of ipc calls)

plugins/tile/tree.cpp
```diff
@@ -388,16 +410,17 @@ wf::geometry_t view_node_t::calculate_target_geometry()
...
+        local_geometry = adjust_geometry_for_workspace(wset, base_geometry, local_geometry);
+    } else if (this->show_maximized)
+    {
+        auto base_workarea = wset->get_attached_output()->workarea->get_workarea();
```
ammen99, line 417: same as above, in very bad cases, get_attached_output() is null, we need a check for this (in that case we can probably just skip showing maximized, if the wset is not attached to an output, the user doesn't see it anyway)

plugins/tile/tree.cpp
```diff
@@ -388,16 +410,17 @@ wf::geometry_t view_node_t::calculate_target_geometry()
...
+        auto source_geometry = view->get_geometry();
+        local_geometry = adjust_geometry_for_workspace(wset, base_workarea, source_geometry);
+        local_geometry.x     += gaps.left;
```
ammen99, line 420: One thing here, we have 3 types of gaps, inner gaps (between tiled views), outer horizontal and vertical gaps for the edges of the output. The `gaps` that you use here, they are calculated based on the normal tiling layout, so some of them can be equal to the inner gaps, others to the outer gaps, depends on how the layout is.

So I think it would make most sense to directly load the outer gaps from wf-config:

```cpp
static wf::option_wrapper_t<int> outer_horiz_gaps{simple-tile/outer_horiz_gap_size};
```

Same for vertical ofc. and use that.

If you are unsure why this is a problem, I would encourage you to try setting the outer gaps to be bigger than the inner gaps, then tile a few views and maximize one of them, you will see the issue :)

plugins/tile/tile-plugin.cpp
```diff
@@ -157,10 +158,27 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
...
+            }
+        }
+    }
+
     wf::signal::connection_t<view_mapped_signal> on_view_mapped = [=] (view_mapped_signal *ev)
     {
         if (auto toplevel = toplevel_cast(ev->view))
         {
+            auto wset = toplevel->get_wset();
+            unmaximize_all_views_on_workspace(wset);
```
ammen99, line 180: should we also do this on focus change? i.e if we alt-tab out of maximized view. in fact there is the function `consider_exit_fullscreen()` which is called exactly when attaching a new view, and when focus is changed, you could look into it :)

ammen99 (changes requested):
looking better, a few more things that I noticed. After these are fixed I think I will start testing the feature as well, to make sure we don't miss anything.

plugins/tile/tile-plugin.cpp
killown: What should we use instead? I'm confused

ammen99: you can see for the others, we do add_key (or even better, add_activator), then there is rem_binding() when unloading the plugin

plugins/tile/tree.cpp
killown: your old calc was always was returning 0,0 no matter what workspace the view logic belongs to, local_geometry.x is already relative to the current workspace if you're on workspace 1, 0 and the view was on workspace 1, 0, local_geometry.x will be 0, meaning that only the first workspace was working

plugins/tile/tile-plugin.cpp
killown: Handling this in on_focus_change would break the existing logic for maximized views during keyboard navigation. If a view is maximized, the next navigated one should also be maximized.

But consider_exit_fullscreen() instead of unmaximize_all_views_on_workspace makes sense, thanks for the recomendation

killown:
https://github.com/user-attachments/assets/d4caac6c-37af-4e13-90e1-b769d197db84

plugins/tile/tile-ipc.hpp
```diff
@@ -215,6 +215,40 @@ inline std::unique_ptr<tile::tree_node_t> build_tree_from_json(const wf::json_t&
...
+    bool show_maximized = wf::ipc::json_get_bool(params, "show-maximized");
+
+    auto node = tile::view_node_t::get_node(view);
+    if (!node)
+    {
+        return wf::ipc::json_error("View is not tiled");
+    }
+
+    autocommit_transaction_t tx;
+    node->show_maximized = show_maximized;
```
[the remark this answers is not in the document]
ammen99: do you then want this show maximized to be like a stacked / tabbed layout when you set it for all views? I thought it should work for just a single view. But ok, if that's what you like :)

killown:
time to review?

plugins/tile/tile-ipc.hpp
```diff
@@ -215,6 +215,38 @@ inline std::unique_ptr<tile::tree_node_t> build_tree_from_json(const wf::json_t&
     return root;
 }
 
+inline wf::json_t handle_ipc_set_show_maximized(const wf::json_t& params)
+{
+    if (!params.has_member("view-id"))
```
ammen99, line 220: No need for these checks. `json_get_uint64` will throw an exception if view-id is not found and it will be handled in the ipc method repository. Same for the other `.has_member('show-maximized')` check below.

plugins/tile/tile-plugin.cpp
```diff
@@ -271,25 +280,63 @@ class tile_output_plugin_t : public wf::pointer_interaction_t, public wf::custom
...
+            bool was_maximized = current_node->show_maximized;
+
+            if (was_maximized)
+            {
+                autocommit_transaction_t tx;
+                current_node->show_maximized = false;
+                current_node->set_geometry(current_node->geometry, tx.tx);
+            }
+
+            auto adjacent_node = tile::view_node_t::get_node(adjacent->view);
```
ammen99, line 321: This line is a bit confusing. `adjacent` is already a `view_node_t`. So you start with the node pointer (`adjacent`), get the view of it, then look for the node of this view. In other words, `adjacent` will always be the same as `adjacent_node`, no need for an extra lookup.

ammen99: When you realize this, you can actually merge the two `if`s together and use the same transaction for both `set_geometry` calls. The benefit is that by resizing both the old and the adjacent view with the same transaction, Wayfire will only start the animation when **both** views are ready for the resize.

ammen99 (changes requested):
A few small things, otherwise looks quite good to me.

ammen99 (approved):
Thanks!

PR #2691, opened 2025-06-27
don't try uninitialized surfaces in wlr_layer_surface_v1_configure

killown:


killown:
> I think the issue is actually on line `445` where we try to send the initial configure. We needed this before because wlroots did the waiting for us. Now instead we shouldn't trigger initial configure manually, but wait for the commit. Can you try removing the extra `.emit(NULL)` on line 445 and verify that no more messages are coming through?

indeed, this was the root issue

PR #2698, opened 2025-06-30
Fix double slash in XDG metadata path construction

killown:
Even though linux supports redundant slashes: 

`WW 30-06-25 10:59:25.157 - [subprojects/wf-config/src/file.cpp:575] Failed to open XML directory /home/myuser/.local/share//wayfire/metadata`


ammen99 (approved):
lgtm now, thanks!

PR #2699, opened 2025-07-01
Decoration plugin - add font color option

mark-herbert42:
Simple change - one option for titlebar font color instead of hardcoded white,  to improve looks for light themes.


ammen99 (approved):
lgtm, thanks :)

PR #2700, opened 2025-07-01
oswitch add directions to output switching

### LINKED ISSUE #1623: Ability to switch output based on relative position
Currently the output switcher (oswitch) plugin only enables switching to the next output in the list. While this functionality is perfect for working with two monitors, when working with three or more it may be desirable to choose the output based on the current position. 

For example, instead of having just `oswitch.next_output` we could have something like `oswitch.left_output,right_output,...`

killown:
close #1623

### DIFF
--- a/metadata/oswitch.xml
+++ b/metadata/oswitch.xml
@@ -24,5 +24,25 @@
 			<_long>Moves focus to the previous output with the focused window with the specified activator.</_long>
 			<default></default>
 		</option>
+		<option name="left_output" type="activator">
+			<_short>Left output</_short>
+			<_long>Moves focus to the output on the left.</_long>
+			<default>&lt;super&lt; Left</default>
+		</option>
+		<option name="right_output" type="activator">
+			<_short>Right output</_short>
+			<_long>Moves focus to the output on the right.</_long>
+			<default>&lt;super&lt; Right</default>
+		</option>
+		<option name="up_output" type="activator">
+			<_short>Up output</_short>
+			<_long>Moves focus to the output above.</_long>
+			<default>&lt;super&lt; Up</default>
+		</option>
+		<option name="down_output" type="activator">
+			<_short>Down output</_short>
+			<_long>Moves focus to the output below.</_long>
+			<default>&lt;super&lt; Down</default>
+		</option>
 	</plugin>
 </wayfire>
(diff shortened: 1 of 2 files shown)


metadata/oswitch.xml
```diff
@@ -24,5 +24,25 @@
 			<_long>Moves focus to the previous output with the focused window with the specified activator.</_long>
 			<default></default>
 		</option>
+		<option name="left_output" type="activator">
+			<_short>Left output</_short>
+			<_long>Moves focus to the output on the left.</_long>
+			<default>&lt;super&lt; Left</default>
```
ammen99, line 30: I think we should disable this by default (leave empty or `disabled` works too), when we add new bindings, they might conflict with existing bindings from users.

plugins/single_plugins/oswitch.cpp
```diff
@@ -33,8 +53,50 @@ class wayfire_oswitch : public wf::plugin_interface_t
...
+        double best_score = -INFINITY;
+
+        for (auto& output : wf::get_core().output_layout->get_outputs())
+        {
+            if (output == current_output)
+            {
+                continue;
+            }
+
+            auto geo  = output->get_layout_geometry();
```
ammen99, line 76: I am wondering about the logic here. You select the one with the biggest distance from the center, right? So say we have three monitors A, B, C in a row, left to right. Now if I am on A and press the binding to go to the right, I expect B to be focused. But your current logic would select C (if I understand correctly? As C would give the highest score with the longest distance). I am not sure what the best way to solve this is, maybe try to look for closest outputs to the right edge of the output (when going to the right?)

ammen99 (changes requested):
This is an old issue, I would have nowadays suggested that people can write an IPC script with whatever logic they want (moving to the sides, or to a particular output, or wrapping around, or whatever). Then again, if you're up for writing the code in C++, I wouldn't mind having it in the upstream plugin.

In any case, I don't think the output selection algorithm does what a user might expect in all scenarios.

metadata/oswitch.xml
```diff
@@ -24,5 +24,25 @@
...
+			<_short>Left output</_short>
+			<_long>Moves focus to the output on the left.</_long>
+			<default></default>
+		</option>
+		<option name="right_output" type="activator">
+			<_short>Right output</_short>
+			<_long>Moves focus to the output on the right.</_long>
+			<default></default>
+		</option>
+		<option name="up_output" type="activator">
```
ammen99, line 37: I was thinking whether the option names `output_left/right/above/below` don't sound better, what do you think?

PR #2703, opened 2025-07-02
Rendering fixes

### LINKED ISSUE #2539: When input method panel becomes shorter, the shortened region is still visible

### LINKED ISSUE #2667: Damage artefacts when maximize-restore gtk window
**Describe the bug**
When opening gtk3 app like wcm or caja on empty screen , maximizing it an then restoring back the artifacts remain in the places where  the background should be, and those artfacts remain flickering unntil  I move the window.

This does not happen with windows like firefox or thunderbird - those ones maximize-restore without artefacts.

**To Reproduce**

Open just one wcm (or caja) window. maximize and restore to original size.

**Wayfire version**
tried with git, and with vulkan patches applied  - no difference.

### LINKED ISSUE #2677: The blur plugin is creating screen artifacts.
**Describe the bug**
to reproduce the issue is necessary cube plugin and blur/alpha enabled on the view

~ ❯❯❯ cat /tmp/wayfire-log.txt | grep git
II 16-06-25 20:50:06.525 - [render/gles2/renderer.c:539] Using OpenGL ES 3.2 Mesa 25.2.0-devel (git-00dd0d0dd1)

https://github.com/user-attachments/assets/d33a7a59-25e9-4ab7-ae41-5e8ddd9b2422

### LINKED ISSUE #2678: wdisplays window does not have cursor painted over it. Same with screen recoding software when recording active
Tried different forkedversions besides the official release (1.1.1) - and always same result. Opening wdisplays - and cursor starts behave strange way. It is never painted over wdisplays window and blinking when moving it on other screen space. Wdisplays is functional  - "blind"  klicking with invisible cursor do press buttons and activate functions.  

Runnng same wdisplays in sway - no such issue with wdisplays.  Also not related to the latest vulkan patches - tried with older git version (just before vulkan patch) - same story.

Just did not use wdisplays for a long time but recently needed to connectl second monitor - and bad finding.

Also - tried ssr-wlroots (fork of simplescreenrecording, long abandoned by still the best screen capture soft). Same effect when recordng - cursor is not shown when moving, but appears back when mouse stops and remains static.

Same - with wf-recorder when active. Guess same trouble might happen with any screen sharing apps.

Tried with release - wlroots 17.4, wayfire 0.9 - no problem, cursor is behaving normal in all mentioned cases. As already mentioned - return to before-vulkan massive patchset did not help, so that happened before.

ammen99:
This PR contains several fixes related to rendering, including:

- **blur: reset opengl scissor before restoring saved pixels** (fixes #2677)
- **render-manager: render sw cursors in a separate render pass** (fixes #2678)
- **render-manager: rewrite depth buffer management logic**
- **wlr-surface-node: accumulate damage from surface resize** (fixes #2539 and fixes #2667)

@killown @mark-herbert42 @lilydjwg please try this PR and let me know if any of the bugs I mentioned are still present.


mark-herbert42:
wdisplays now works perfectly for me, and one more effect - thre was blinking of the eDP-1 indicating little window of wdisplays - not so signifficant to report the bug. But now it is gone - so wdisplays issue is gone even with vk-color-management enabled.

ammen99:
> Actually how it happens - I start wayfire , and I have mate-panel in start list. It crashes. Then I do start wayfire again- it starts but no mate-panel. Startinh mate-panel from terminal - get the responce can not start, mate panel is already running
> 
> killall mate-panel. Start mate panel. Now wayfire runs with mate-panel and works good.
> 
> Removed mate-panel from autostart - no start crashes. So seems the issue is between mate-panel and wayfire. But it used to work like a charm for a long time....

And when you enable address sanitizer it never crashes?

Can you try disabling most plugins to make sure this is not a random plugin crashing?

mark-herbert42:
I am not an expert developer so no idea what address sanitizer is....    The trace from the log seems logical - it crashes in gtk-layer-shell  which is exacgtly in sync with mate-panel being the reason for the crash.  I will rebuild mate-panel , and try again. And what is strange - why it starts manually Ok but does not do it in autostart.

ammen99:
> I am not an expert developer so no idea what address sanitizer is.... The trace from the log seems logical - it crashes in gtk-layer-shell which is exacgtly in sync with mate-panel being the reason for the crash. I will rebuild mate-panel , and try again. And what is strange - why it starts manually Ok but does not do it in autostart.

Address sanitizer is when you build wayfire with `meson setup build -Db_sanitize=address,undefined ...`. It allows a lot of additional runtime checks and gives very useful information in case of a crash. But I think I fixed the problem you were having, please try again with the new commits I pushed.

mark-herbert42:
tested - no crashes on autostart of mate-panel anymore.

lilydjwg:
I can still reproduce #2539:

https://github.com/user-attachments/assets/36334041-bb93-41c5-8ebb-4cb5139f6dc7

(The right edge is screen edge.)

ammen99:
> I can still reproduce #2539:
> 
>  a.mp4 
> (The right edge is screen edge.)

The video does not play, maybe did not upload properly?

lilydjwg:
It plays for me. Maybe your browser doesn't support the HEVC codec? Anyway, a H264 version here:

https://github.com/user-attachments/assets/f2a2f00a-caf9-4d11-b106-1e15582770eb

lilydjwg:
Unfortunately no. It's much better than without those commits, fixing most cases, but I can still reliably reproduce the slight issue at the start of screencast: when "en" turns into "五" and becomes narrower, there is a slim edge left for some time.

I also reproduced other minor artifacts in virtual machine (for whatever reasons the input method popup is a bit larger there), but very shortly and I can't reproduce them later. If I didn't make mistake, a horizontal strip at the top of the disappearing region was left visible in this case.

ammen99:
> Unfortunately no. It's much better than without those commits, fixing most cases, but I can still reliably reproduce the slight issue at the start of screencast: when "en" turns into "五" and becomes narrower, there is a slim edge left for some time.

Hmm, so soreau's original commits fix the issue, but my updates break it again? Very weird...

ammen99:
@lilydjwg I had the wrong implementation of the resize fix. Now, I don't need the other PR anymore, I can see no visual artifacts with fcitx5 with just this PR. Please test and let me know whether this is the case for you as well.

lilydjwg:
Yes, they are gone!

ammen99:
Ok, so after all, we don't need the other PR. Which is good, the less state we manage, the better :)

Issue #2707, opened 2025-07-04
Provide a better way to let the user understand that setting options via IPC will make subsequent config file changes ineffective


killown:
**Describe the bug**
```
    this->plugins_opt.set_callback([=] ()
    {
        /* reload when config reload has finished */
        idle_reload_plugins.run_once([&] () { reload_dynamic_plugins(); });
    });
```
Plugin reload/enable/disable will stop working after using set_option_values.

**To Reproduce**
```
def verify_required_wayfire_plugins():
    required_plugins = {
        "stipc",
        "ipc",
        "ipc-rules",
        "resize",
        "window-rules",
        "wsets",
        "session-lock",
        "wm-actions",
        "move",
        "vswitch",
        "grid",
        "place",
        "scale",
    }

    enabled_plugins = sock.get_option_value("core/plugins")["value"].split()

    for plugin_name in required_plugins:
        if plugin_name not in enabled_plugins:
            enabled_plugins.append(plugin_name)

    sock.set_option_values({"core/plugins": " ".join(enabled_plugins)})

verify_required_wayfire_plugins()
```
disable or enable a plugin in wayfire.ini

**Expected behavior**
plugin-loader should at least LOGE the reason why the plugins can't reload/enable/disable.

**Suggestion**
reload_config_signal somewhere in set_config_options will set a callback LOGI with the options that can't be modified by wayfire.ini

**Why consider the Suggestion**
It took me a while to find the issue, others might face the same problem without knowing the cause, some others could set random plugins options and complain later why these options is not changing through wayfire.ini.

ammen99:
Looking at your PR #2708, I am thinking, maybe we can add an std::set of options for which we have emitted a warning? The idea is that once we emit the warning, we don't want to spam it again. So you can get a warning that an option has been overridden by IPC just once for each option at most.

Maybe we should even limit it to just a few warnings and say after 5-10 warnings warn that further warnings of this type will be suppressed?

PR #2714, opened 2025-07-07
Log config options set by IPC

killown:
fixes https://github.com/WayfireWM/wayfire/issues/2707

**Message format:**
[plugins/ipc-rules/ipc-utility-methods.hpp:287] Config options set via IPC will not update from /home/myuser/.config/wayfire.ini until Wayfire restart

### DIFF
--- a/plugins/ipc-rules/ipc-utility-methods.hpp
+++ b/plugins/ipc-rules/ipc-utility-methods.hpp
@@ -281,6 +281,9 @@ class ipc_rules_utility_methods_t
             return wf::ipc::json_error("Options must be an object!");
         }
 
+        const char *config_file = std::getenv("WAYFIRE_CONFIG_FILE");
+        std::string config_file_str = config_file ? config_file : "wayfire.ini";
+
         for (auto& option : data.get_member_names())
         {
             auto opt = wf::get_core().config->get_option(option);
@@ -304,6 +307,9 @@ class ipc_rules_utility_methods_t
                         std::string(json_to_string(data[option])) + "!");
                 }
             }
+            LOGW("Config option '", option, "' modified via IPC. ",
+                "Changes in ", config_file_str, " will not affect this option ",
+                "until Wayfire is restarted.");
 
             opt->set_locked(true);
         }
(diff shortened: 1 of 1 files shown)


plugins/ipc-rules/ipc-utility-methods.hpp
```diff
@@ -281,6 +283,16 @@ class ipc_rules_utility_methods_t
...
         }
 
+        const char *config_file     = std::getenv("WAYFIRE_CONFIG_FILE");
+        std::string config_file_str = config_file ? config_file : "wayfire.ini";
+
+        if (warning_count <= MAX_WARNINGS)
+        {
+            LOGW("Config options set via IPC will not update from ", config_file_str,
+                " until Wayfire restart.");
+            warning_count++;
```
ammen99, line 293: I would say we should ++warning_count every time, and when we hit MAX_WARNINGS, print a warning that we will suppress future warnings. Just so that users don't wonder why there are no subsequent warnings xd

plugins/ipc-rules/ipc-utility-methods.hpp
```diff
@@ -23,6 +23,8 @@ class ipc_rules_utility_methods_t
   private:
     wlr_backend *headless_backend = NULL;
     std::set<uint64_t> our_outputs;
+    int warning_count = 0;
```
ammen99, line 26: We should make these static variables in `set_config_options`, this way it is clear that they are for that IPC method only.

ammen99 (changes requested):

plugins/ipc-rules/ipc-utility-methods.hpp
```diff
@@ -281,6 +283,16 @@ class ipc_rules_utility_methods_t
             return wf::ipc::json_error("Options must be an object!");
         }
 
+        const char *config_file     = std::getenv("WAYFIRE_CONFIG_FILE");
```
ammen99, line 286: I think strictly speaking we don't really need to print the config file every time, we say this at the beginning of the log. So we could just say that we won't laod from the config file. Consider that for example config backends could also use a different config file altogether not the ini file.

killown:
Thanks for the suggestions. We should find a better way to explain set_config_options' implications without spamming logs like add comments.

PR #2715, opened 2025-07-07
feat: Add config reload delay option and refactor reload logic

### LINKED ISSUE #2493: IPC: Saving modifications in wayfire.ini causes watch_events() to stop working
Modifying && Saving wayfire.ini is causing watch_events() to stop receiving IPC events. It’s possible that wayfire is unnecessarily reloading IPC plugins when changes are made to the configuration, which could be disrupting the event flow.

```
def watch_events():
    sock.watch()
    while True:
        msg = sock.read_message()
        print(msg)
```

Expected behavior: watch_events() should continue reading and printing messages after changes are made to wayfire.ini.

Current behavior: After modifying and saving wayfire.ini, the function stops reading messages.

ammen99:
Fixes #2493 and potentially other long standing issues related to config reload.

An open question is whether we want to set the default to something like 50 or leave it to 0 to keep the old behavior. Thoughts?

### DIFF
--- a/metadata/workarounds.xml
+++ b/metadata/workarounds.xml
@@ -93,5 +93,12 @@
 			<_long>Automatically reload the config file when it's modified.</_long>
 			<default>true</default>
 		</option>
+		<option name="config_reload_delay" type="int">
+			<_short>Config reload delay</_short>
+			<_long>Delay in milliseconds after a config file change before reloading.</_long>
+			<default>20</default>
+			<min>0</min>
+			<max>500</max>
+		</option>
 	</plugin>
 </wayfire>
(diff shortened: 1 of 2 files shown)


soreau:
> > The thing I would like to see fixed in all this is, starting wayfire with something like `wayfire -c /tmp/wayfire.ini` where /tmp/wayfire.ini does not exist, then while wayfire is running, creating that config file and changing it would work without having to restart wayfire.
> 
> Did we not fix this a long time ago? There is code for it in the default config backend at least.

Hm, I'm not sure if it works or not, would have to test it.

killown:
WCM is working with 50ms.

This PR doesn't just fix the IPC issue, it also fix the outputs being disabled due to the config reload issue. This no longer happens.

I believe 50ms should be the default option, as it will prevent many issues, this pr is bigger than you imagine, besides the simple code.

PR #2721, opened 2025-07-13
animate: Improve squeezimize animation

### LINKED ISSUE #2720: Squeezemize looks bad if minimize area is too narrow
If setting a minimize area (by [set_rectangle](https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1#zwlr_foreign_toplevel_handle_v1:request:set_rectangle)) with a very small width, the animation looks "bad": the view being minimized is squeezed too much too early, becoming a thin line.

Image

Simple test case: https://github.com/dkondor/minimize_test

Real-world use case: under some conditions, Cairo-Dock will try to set a rectangle with width = height = 1 (see [here](https://github.com/Cairo-Dock/cairo-dock-core/blob/master/src/gldit/cairo-dock-application-facility.c#L445) and [here](https://github.com/Cairo-Dock/cairo-dock-core/blob/master/src/implementations/cairo-dock-foreign-toplevel.c#L107)) This is easy to fix, but I thought to raise this here as a general issue.

I'll look into the squeezimize plugin and suggest an improvement if I can.

soreau:
This adjusts the progress and sigmoid curve to improve squeezimize. Namely, this should make it so the window is not squeezed too much when first minimized.

Fixes #2720.

### DIFF
--- a/plugins/animate/squeezimize.hpp
+++ b/plugins/animate/squeezimize.hpp
@@ -72,12 +72,12 @@ uniform int upward;
 
 void main()
 {
-    float y;
+    float y, sigmoid;
     vec2 uv_squeeze;
     float inv_w = 1.0 / (src_box.z - src_box.x);
     float inv_h = 1.0 / (src_box.w - src_box.y);
-    float progress_pt_one = clamp(progress, 0.0, 0.5) * 2.0;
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
thanks :)

PR #2723, opened 2025-07-13
cube: Invert y input coordinate

### LINKED ISSUE #2722: Cube: top/bottom rotation with the mouse is inverted
**Describe the bug**
Previously, when I rotated the cube with the mouse (using the shortcut <alt><ctrl>BTN_LEFT), moving the mouse to the top showed the bottom of the cube and moving the mouse to the bottom showed the top of the cube.

Since commit 641ed7f8b4a5fd33c785877c1732d1896a2183b5, the logic is inverted: moving the mouse to the top shows the top of the cube and moving the mouse to the bottom shows the bottom of the cube. It's a less intuitive behavior and I guess it's a regression, not the intended behavior.

**To Reproduce**
1. Start wayfire with the default configuration, or a configuration with the cube plugin enabled
2. Use the shortcut \<alt\>\<ctrl\>BTN_LEFT to rotate the cube

soreau:
This is a side effect of the fact that the y invertedness for buffers switched.

Fixes #2722.

### DIFF
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -699,7 +699,7 @@ class wayfire_cube : public wf::per_output_plugin_instance_t, public wf::pointer
         }
 
         double xdiff = ev->delta_x;
-        double ydiff = ev->delta_y;
+        double ydiff = ev->delta_y * -1.0;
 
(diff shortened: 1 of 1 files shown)


ammen99:
Thanks :)

PR #2724, opened 2025-07-13
core: plugins: s/mediump/highp/

soreau:
The intent here is to improve rendering accuracy overall. This
is especially noticeable when normally rectangular surfaces are
stretched as is the case with squeezimize and wobbly effects.


ammen99 (approved):
thanks :)

PR #2732, opened 2025-07-16
Squeezimize: horizontal version

dkondor:
Since I've started looking into the squeezimize shader, I though to try to create a horizontal version. This would be used for panels on the left / right edge of the screen. I belive the shared is working well now, but the heuristic for deciding whether to use the horizontal or vertical version could be improved: now, I just test which distance to the target box is larger, but this could unexpectedly choose the "wrong" animation, e.g. if the minimize location is close to a corner and the minimized view is on the other side of the screen. Potential alternatives:
 - Add an option where the user can select
 - Try to improve the heuristic, e.g. consider which screen edge the minimize position is closer to
 - Actually look up the wl_surface set in the [set_rectangle](https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1#zwlr_foreign_toplevel_handle_v1:request:set_rectangle) request and check which edge it is anchored on (at least if it is a layer shell surface).

I'm open to experimenting with any of these variants.

Notes:
 - you can easily test with this example program: https://github.com/dkondor/minimize_test
 - the second commit is just to simplify the calculations IMO; if it is not making things simpler for you, I can revert it :)
 - I'm not sure if it is efficient to set a potentially different shader every time the animation starts (I assume these are cached somehow?); I could just create two separate programs if that would help

### DIFF
--- a/plugins/animate/squeezimize.hpp
+++ b/plugins/animate/squeezimize.hpp
@@ -328,26 +328,43 @@ class squeezimize_transformer : public wf::scene::view_2d_transformer_t
                 (this->minimize_target.y + this->minimize_target.height) - bbox.y),
                 (bbox.y + bbox.height) - this->minimize_target.y);
 
-        bool horiz   = false;
-        auto src_box = view->get_bounding_box();
-        auto output  = view->get_output();
-        auto geom    = output->get_relative_geometry();
-        auto d_horiz = std::max(std::max(0, src_box.x - (minimize_target.x + minimize_target.width)),
-            std::max(0, minimize_target.x - (src_box.x + src_box.width)));
-        auto d_vert = std::max(std::max(0, src_box.y - (minimize_target.y + minimize_target.height)),
-            std::max(0, minimize_target.y - (src_box.y + src_box.height)));
+        bool horiz;
+        // auto src_box = view->get_bounding_box();
+        auto output = view->get_output();
+        auto geom   = output->get_relative_geometry();
 
-        if (((d_vert == 0) && (d_horiz > 0)) ||
-            ((d_vert > 0) && (d_horiz > ((double)d_vert * geom.width) / geom.height)))
+        // check which edge the target is closest to
+        double x  = minimize_target.x + minimize_target.width / 2.0;
+        double y  = minimize_target.y + minimize_target.height / 2.0;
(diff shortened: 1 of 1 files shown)


soreau:
Or, maybe check the center point of the minimize target to see which output edge it's closest to.

dkondor:
> Or, maybe check the center point of the minimize target to see which output edge it's closest to.

Good idea! I'll try this a bit later and update the PR.

soreau:
> > Or, maybe check the center point of the minimize target to see which output edge it's closest to.
> 
> Good idea! I'll try this a bit later and update the PR.

Bonus points for dividing the output into four triangles with borders of diagonal lines crossing the output, for the check. 😆

plugins/animate/squeezimize.hpp
```diff
@@ -110,6 +108,60 @@ void main()
...
+    vec2 uv_squeeze;
+    float inv_w = 1.0 / (src_box.z - src_box.x);
+    float inv_h = 1.0 / (src_box.w - src_box.y);
+    float progress_pt_one = pow(clamp(progress, 0.0, 0.25) * 4.0, 2.0);
+    float progress_pt_two = pow(progress, 2.0);
+
+    uv_squeeze.x = inv_w * (uv.x - src_box.x);
+    uv_squeeze.y = inv_h * (1.0 - uv.y - src_box.y);
+
+    if (upward == 1)
```
soreau, line 138: I would s/upward/direction/.

plugins/animate/squeezimize.hpp
```diff
@@ -276,15 +328,49 @@ class squeezimize_transformer : public wf::scene::view_2d_transformer_t
...
+                // top edge
+                horiz = false;
+                this->upward = true;
+            }
+        }
+
         wf::gles::run_in_context_if_gles([&]
         {
-            program.compile(squeeze_vert_source, squeeze_frag_source);
+            program.compile(squeeze_vert_source, horiz ? squeeze_frag_source_horiz : squeeze_frag_source);
```
soreau, line 372: We could have two programs compiled upfront and ready to use, but this is probably fine too, since the shaders are cached.

soreau (approved):
I don't have much to say here, other than 'Great job!'.

ammen99 (approved):
thanks :)

Issue #2734, opened 2025-07-17
xdg popup won't appear until you move the mouse cursor


killown:
**Describe the bug**
The xdg popup won't appear until you move the mouse cursor.

**To Reproduce**

https://github.com/user-attachments/assets/87089d51-5022-46c0-8a5d-e8e2d16f2a39

**Wayfire version**
0.10.0-79f220ee (Jul 17 2025) branch master wlroots-0.19.0

```
[workarounds]
app_id_mode = "stock"
all_dialogs_modal = true
disable_primary_selection = false
discard_command_output = true
dynamic_repaint_delay = false
enable_input_method_v2 = false
enable_opaque_region_damage_optimizations = true
enable_so_unloading = false
focus_main_surface_instead_of_popup = false
force_frame_sync = false
force_preferred_decoration_mode = false
max_buffer_size = 8192
remove_output_limits = false
use_external_output_configuration = true
```

PR #2735, opened 2025-07-18
Fix popup rendering

killown:
https://github.com/WayfireWM/wayfire/issues/2734


killown:
> It is interesting this fixes the issue. I suppose that the commit() handler of the xdg-popup needs to be called after the commit handler of the main surface. However I am not sure why this is important. I'll take another deeper look soon to figure out what the root issue is, so that we can avoid such issues in the future too.

on_surface_commit is happening before the popup_parent is fully initialized, like position calculation not complete while the callback happens meanwhile, then causing these issues. To avoid this issue, a simple comment // telling why not move the line to the top ever again

ammen99:
can't reproduce this issue actually. using current wayfire-git

killown:
> can't reproduce this issue actually. using current wayfire-git

You can't reproduce it if anything triggers a repaint, like playing a random youtube video or your panel has some kind of animation, the popup will always work. Try using Nautilus in fullscreen, but don’t move the cursor when you click.

ammen99:
Thanks for investigating this, however, a 'real' solution which should be a bit more robust and follows core's design principles is available in #2752, so closing this in favor of that PR.

PR #2742, opened 2025-07-19
WF shell protocol: keyboard lang manager implementation

WShad:
Added keyboard layout manager to wayfire-shell protocol extension.
Wayfire shell's panel now able to display current keyboard layout and request to set it.
Any suggestions?)

Wayfire shell's corresponding PR: https://github.com/WayfireWM/wf-shell/pull/301

### DIFF
--- a/plugins/protocols/wayfire-shell.cpp
+++ b/plugins/protocols/wayfire-shell.cpp
@@ -411,6 +411,109 @@ static void handle_surface_destroy(wl_resource *resource)
     wl_resource_set_user_data(resource, nullptr);
 }
 
+static void handle_keyboard_lang_manager_set_layout(wl_client *client, wl_resource *resource, uint32_t layout_idx)
+{
+    auto seat     = wf::get_core().get_current_seat();
+    auto keyboard = wlr_seat_get_keyboard(seat);
+
+    if (!keyboard)
+    {
+        return;
+    }
+
+    if (layout_idx >= xkb_keymap_num_layouts(keyboard->keymap))
+    {
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


ammen99:
Hey, have you considered using the already existing IPC methods for this in wf-shell? This way we don't have to provide duplicated functionality.

WShad:
Thought about it, but IPC's methods tightly coupled with IPC-rules plugin and JSON.
Should i extract this methods as utility?

ammen99:
> Or did you mean to use IPC instead of wf-shell protocol? If so: using IPC and wf-shell protocol at same time - is a little bit weird.

That was what I was thinking :) I get your point, yet we use the wayfire-shell protocol mostly for historical reasons - it is from before the time wayfire had IPC. However at this point Wayfire's IPC is way easier to use (at least IMO) and it has many more features that might turn out useful for future wf-shell widgets and functionalities.

For example if wf-shell had support for Wayfire's IPC, we could also add many more actions to right-click of window-list. Or show the current workspace. Things like that. It doesn't have to happen in the same PR of course :) But these are just examples where the IPC already has everything we need. So I expect it would be a good move to start adding IPC to wf-shell instead of extending the protocol with functionality we have already implemented.

WShad:
> Wayfire's IPC is way easier to use (at least IMO) and it has many more features that might turn out useful for future wf-shell widgets and functionalities.

Agree.

> So I expect it would be a good move to start adding IPC to wf-shell instead of extending the protocol with functionality we have already implemented.

Let's go then!))

ammen99:
If we're clear on that, I suppose this PR should be closed, and only the wf-shell PR needs to be updated to use IPC.

PR #2747, opened 2025-07-22
window-manager: add function to directly store windowed geometry

### LINKED ISSUE #2585: Prevent Waybar from hiding.
**Describe the bug**  
I've been going crazy for days and can't figure out how this issue started. I updated both Wayfire to the latest version and Waybar, and from that moment on, this problem appeared: when I open a fullscreen app, especially with Wine, Waybar leaves an empty space in the bar area and doesn't go fully fullscreen.  

I'm not sure if it's an IPC request timing issue or something else. However, if I toggle fullscreen on and off from within the game, it works correctly.  

Additionally, in some games like *Crysis 2*, all mouse position events become misaligned, making the cursor behave erratically.  

**To Reproduce**  
1. Open a game in fullscreen.  

**Expected behavior**  
I expect that any app running in fullscreen should cover the entire screen and that mouse input should remain accurate.  

**Screenshots or stacktrace**  

Image  

Image

**Wayfire version**  
git

ammen99:
Fixes #2585

@killown @AgostinoA please test

Ref https://github.com/ammen99/wayfire-tests/commit/af28073d2143db90d371dcf89807a86629422488

### DIFF
--- a/src/api/wayfire/plugin.hpp
+++ b/src/api/wayfire/plugin.hpp
@@ -107,7 +107,7 @@ using wayfire_plugin_load_func = wf::plugin_interface_t * (*)();
 /**
  * The version is defined as macro as well, to allow conditional compilation.
  */
-#define WAYFIRE_API_ABI_VERSION_MACRO 2025'07'12
+#define WAYFIRE_API_ABI_VERSION_MACRO 2025'07'21
 
 /**
  * The version of Wayfire's API/ABI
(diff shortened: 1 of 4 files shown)


killown:
nice, worked!

AgostinoA:
Guys, I've been trying since this morning, at first I thought I hadn't done the checkout properly, but I actually did. I've recompiled several times, but the change has no effect. I tried both with the test using GLFW and with Xterm (fullscreen).

@killown have you tried both with GLFW and Xterm ?

I don’t know what to say, I even tried with the base config: I only added `waybar`.
I’ll also add that I tested with both `wf-panel`.

[test_fullscreen.zip](https://github.com/user-attachments/files/21382484/test_fullscreen.zip)

PR #2764, opened 2025-07-27
window-manager: Ignore unfullscreen requests if already windowed

soreau:
This fixes dragging gtk4 apps with grid and/or wobbly disabled. The gtk4 toolkit sends a request for unfullscreen even though it is windowed, and this caused wayfire to set the geometry when dragging by titlebar, which made the window jump to an offset when grabbing the titlebar to move it. This was fixed before but broken by 60231d5d.


soreau:
Here is a [wayfire-test](https://github.com/ammen99/wayfire-tests/pull/8) for this case.

ammen99:
Thanks!

PR #2779, opened 2025-08-15
subsurfaces: Only damage where the client requests

### LINKED ISSUE #2774: Firefox triggers a full window repaint upon webpage content changes.
**Describe the bug**
I used the `repaint` plugin to monitor window redraws and noticed that in firefox, when certain elements change such as a blinking text cursor or a small gif, wayfire redraws the entire window instead of just the affected region. This causes a noticeable increase in GPU usage on my Intel integrated graphics.

I compiled a relatively old version of wayfire (0.7.5), and this issue does not occur there and chromium doesn't have this issue either. I'm not sure whether the problem lies with wayfire or firefox.

**To Reproduce**
1. Using firefox under wayfire to access such a gif gif. 
2. After enable the repaint plugin, I can observe the entire window is being redrawn.

**Expected behavior**
Only redraw the changed regions.

**Wayfire version**
- `firefox`: 141.0.3-1
- `wayfire`: 0.9.0.r221.gd627275b-1 built from commit d627275b

**Related issues**
https://github.com/WayfireWM/wayfire/issues/2567

soreau:
Fixes #2774.

### DIFF
--- a/src/view/wlr-subsurface-controller.cpp
+++ b/src/view/wlr-subsurface-controller.cpp
@@ -90,9 +90,6 @@ bool wf::wlr_subsurface_root_node_t::update_offset(bool apply_damage)
     } else if (changed)
     {
         set_offset(offset);
-    } else if (apply_damage)
-    {
-        scene::damage_node(this, get_bounding_box());
     }
 
     return changed;


ammen99 (approved):
Thanks!

PR #2780, opened 2025-08-16
xwayland: offer focus in global active cases

ssfdust:
Not all diglogs want focus though they are focusable. In some strange cases, (e.g. WeChat, a famous chat application in China), the application's popup has both NORMAL and UTILITY type, we need to call wlr_xwayland_surface_offer_focus() for these views, in case they actually don't want focus.

### DIFF
--- a/src/view/xwayland/xwayland-toplevel-view.hpp
+++ b/src/view/xwayland/xwayland-toplevel-view.hpp
@@ -376,8 +376,72 @@ class wayfire_xwayland_view : public wf::toplevel_view_interface_t, public wayfi
         wf::adjust_view_output_on_map(this);
         do_map(surface, false);
         on_surface_commit.connect(&surface->events.commit);
+        bool wants_focus = false;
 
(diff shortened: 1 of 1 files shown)


src/view/xwayland/xwayland-toplevel-view.hpp
```diff
@@ -376,8 +376,67 @@ class wayfire_xwayland_view : public wf::toplevel_view_interface_t, public wayfi
...
+           * Globally Active client windows may receive a WM_TAKE_FOCUS
+           * message from the window manager. If they want the focus, they
+           * should respond with a SetInputFocus request.
+           */
+          case WLR_ICCCM_INPUT_MODEL_GLOBAL:
+            /*
+             * Assume that NORMAL and DIALOG windows are likely to
+             * want focus. These window types should show up in the
+             * Alt-Tab switcher and be automatically focused when
+             * they become topmost.
```
ammen99, line 417: I would explain here that when we offer focus, we don't really focus the dialog/whatever. Instead, by offering focus like that, we let the client switch focus between its surfaces without involving the compositor.

Otherwise lgtm.

ammen99 (approved):
Thanks!

PR #2781, opened 2025-08-16
output-layout: Move output request state handler from render-manager to output-layout

soreau:
This fixes i.e. wlr-randr always reporting the mode dimensions in config wayland backend WL-1. This also fixes direct scanout criteria for client geometry must match output geometry.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -336,7 +336,7 @@ struct output_layout_output_t
     int current_bit_depth = RENDER_BIT_DEPTH_DEFAULT;
 
     std::unique_ptr<wf::output_impl_t> output;
-    wl_listener_wrapper on_destroy, on_commit;
+    wl_listener_wrapper on_destroy, on_commit, on_request_state;
 
     std::shared_ptr<wf::config::section_t> config_section;
     wf::option_wrapper_t<wf::output_config::mode_t> mode_opt;
@@ -365,6 +365,7 @@ struct output_layout_output_t
     {
         this->handle = handle;
         on_destroy.connect(&handle->events.destroy);
+        on_request_state.connect(&handle->events.request_state);
         initialize_config_options();
 
         is_nested_compositor = wlr_output_is_wl(handle);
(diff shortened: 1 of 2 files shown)


src/core/output-layout.cpp
```diff
@@ -1292,6 +1294,16 @@ class output_layout_t::impl
             remove_output(output);
         });
 
+        on_request_state.set_callback([=] (void *data)
+        {
+            auto ev = static_cast<wlr_output_event_request_state*>(data);
+            wlr_output_commit_state(ev->output, ev->state);
+            send_wlr_configuration();
+            outputs[output]->output->render->damage_whole();
```
ammen99, line 1302: `outputs[output]->output` could be NULL for disabled outputs. I doubt wlroots will send a signal for such outputs but it is not forbidden so I would add a check for that.

Also fwiw if we send a wlr configuration, then we also need to emit the output-configuration-changed signal too.

soreau: I have used `lo` instead of `outputs[output]->output` and installed the output-configuration-changed signal emission, which actually fixed a bug where resizing a second WL output and then the first would leave the second partially rendered until resizing it again.

src/core/output-layout.cpp
```diff
@@ -1292,6 +1293,15 @@ class output_layout_t::impl
             remove_output(output);
         });
 
+        lo->on_request_state.set_callback([=] (void *data)
+        {
+            auto ev = static_cast<wlr_output_event_request_state*>(data);
+            wlr_output_commit_state(ev->output, ev->state);
+            send_wlr_configuration();
+            lo->output->render->damage_whole();
```
ammen99, line 1301: `lo->output` is not guaranteed to be non-NULL either ...

src/core/output-layout.cpp
```diff
@@ -1744,6 +1754,11 @@ class output_layout_t::impl
 
         wlr_output_manager_v1_set_configuration(output_manager,
             wlr_configuration);
+
```
ammen99, line 1757: This is the wrong place for this code, the function sends the wlr_output_manager configuration, not the wayfire signals.

In addition, wee need to send the configuration changed signal for each output (as you do it) but also the global output configuration changed signal for the whole output layout.

soreau: I have moved it now, please check again.

src/core/output-layout.cpp
```diff
@@ -1292,6 +1293,28 @@ class output_layout_t::impl
...
+            send_wlr_configuration();
+            for (auto& output : outputs)
+            {
+                output.second->emit_configuration_changed(
+                    wf::OUTPUT_POSITION_CHANGE | wf::OUTPUT_MODE_CHANGE);
+            }
+
+            output_layout_configuration_changed_signal signal;
+            wf::get_core().output_layout->emit(&signal);
+            if (!lo->output)
```
ammen99, line 1309: We should check this maybe after the output_commit_state, because the emit_configuration_changed functions also don't have a check for whether the output is NULL, we can either add a check there too, or move this check upwards in this function so that we don't emit_configuration_changed on an output with NULL wf::output_t object.

Issue #2494, opened 2024-10-13
Wallpaper blinks when moving mouse cursor


sladkkr:
**Description**
After startup of a Wayfire session the wallpaper blinks when moving the mouse cursor, this behavior occurs before launching any application, after that no blinking. Blinking is synchronized with the cursor movement, looks like every pixel traveled by the cursor corresponds to change either into a default Wayfire desktop color or back into the set wallpaper.

**To Reproduce**
1. Add any wallpaper utility to autostart section of a configuration. (tried swaybg, wbg, swww).
2. Start a Wayfire session.
3. Move the mouse cursor before launching any application.

**Expected behavior**
No blinking obviously.

**Wayfire version**
0.9.0 for Nixpkgs (using the NixOS Unstable)

ammen99:
Does this happen with latest wayfire?

killown:
Same issue when dpms is off, related, you can reproduce this with swww

https://github.com/user-attachments/assets/fd258617-76f0-4025-a8ba-8386049adb02

ammen99:
Most probably fixed with #2798, please reopen if it still happens.

PR #2798, opened 2025-08-29
Fix black screen on DPMS-on: Emit config change before damage

### LINKED ISSUE #2793: Output issue after DPMS off/on
**Describe the bug**
output only shows mouse cursor after dpms off/on (no reliable way to reproduce) it's random.

**To Reproduce**
```
~ ❯❯❯ wlopm --off OUTPUT_NAME
~ ❯❯❯ wlopm --on OUTPUT_NAME
```

Workaround: disable/enable the output

**Screenshots or stacktrace**

https://github.com/user-attachments/assets/4a50993f-4375-4f9c-9a3a-22f239a435c8

In the video, I am grabbing and moving views.

**Wayfire version**
0.10.0-2aeeda42 (Aug 26 2025) branch update-subproject wlroots-0.19.0

killown:
I need to test it for a while.

I assume you're already aware of why this may fix, so no explanation is necessary.

fix #2793 #2494

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -990,8 +990,8 @@ struct output_layout_output_t
             pending_state.commit(handle);
 
             ensure_wayfire_output(get_effective_size());
-            output->render->damage_whole();
             emit_configuration_changed(changed_fields);
+            output->render->damage_whole();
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
Nice catch, thanks!

PR #2803, opened 2025-08-31
update uncrustify.ini to fix deprecation warnings

killown:


ammen99:
If you want to do that, you have to update the CI to newer uncrustify versions, as we used a fixed old version which I patched to fix a few bugs. I was looking to testing newer uncrustify, but it simply crashes for me .. so I am not sure updating to a newer uncrustify is a good idea either.

killown:
> If you want to do that, you have to update the CI to newer uncrustify versions, as we used a fixed old version which I patched to fix a few bugs. I was looking to testing newer uncrustify, but it simply crashes for me .. so I am not sure updating to a newer uncrustify is a good idea either.

sounds like this issue should be solved by you someday, thanks for the information

PR #2818, opened 2025-09-08
Prevent use-after-free in auxilliary_buffer_t

killown:
I'm not sure you want this, but it does look cleaner and safer than the old way.
   
```
this->texture = other.texture; 
this->buffer  = other.buffer;    
other.buffer.buffer = NULL;     
```

Old code seems unsafe and confusing? It copies the buffer first, then sets the original to NULL

### DIFF
--- a/src/render.cpp
+++ b/src/render.cpp
@@ -24,11 +24,8 @@ wf::auxilliary_buffer_t& wf::auxilliary_buffer_t::operator =(auxilliary_buffer_t
         return *this;
     }
 
-    this->texture = other.texture;
-    this->buffer  = other.buffer;
-    other.buffer.buffer = NULL;
-    other.texture     = NULL;
-    other.buffer.size = {0, 0};
+    this->texture = std::exchange(other.texture, nullptr);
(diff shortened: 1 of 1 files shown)


killown:
> Also btw I don't see any use-after-free as you indicated - can you elaborate on why you think there is any use-after-free? To me this change is purely about making the code a bit more readable.

(confusing == not intuitive)

this->texture = other.texture;      -> 1
this->buffer  = other.buffer;       -> 2
other.buffer.buffer = NULL;         -> 3
other.texture = NULL;

-> 2 performs a copy of the buffer struct

If the other object is destroyed immediately after this copy somehow, other.buffer.buffer was only nulled in -> 3 

This is just defensive coding and more readable indeed, in your mind it's 100% correct and we never got any use-after-free before, but if somehow it's not 100% like in future code changes, there will be no issues also.

ammen99:
> If the other object is destroyed immediately after this copy somehow, other.buffer.buffer was only nulled in -> 3

This can't quite happen because Wayfire is single-threaded :)

PR #2822, opened 2025-09-09
core: load input/output options with section+option name

### LINKED ISSUE #2804: Input devices sections containing '/' are not properly handled by option loading
If we have an input device which contains a '/' in its name (for example 'A/B'), we may try to load input options for it by looking for the option 'A/B/name', which is ambiguous as it can mean option with name `name` in section `A/B` or name `B/option` in section `A`.

To solve this issue, we should ideally add APIs to wf-config and Wayfire to load options from a given section with a given name, which would align very closely with the existing code:

https://github.com/WayfireWM/wayfire/blob/8bbea76d3b47a7e78e368739ef2d5e6e53331b1f/src/core/seat/keyboard.cpp#L125-L133

The suggested API would be something like `option.load(section, "name")`

ammen99:
Requires https://github.com/WayfireWM/wf-config/pull/84, therefore the wf-config submodule was bumped.

Fixes #2804

@lilydjwg please test :)

### DIFF
--- a/src/api/wayfire/option-wrapper.hpp
+++ b/src/api/wayfire/option-wrapper.hpp
@@ -46,6 +46,19 @@ class option_wrapper_t : public base_option_wrapper_t<Type>
         }
     }
 
+    void load_option(std::shared_ptr<wf::config::section_t> section, const std::string& option_name)
+    {
+        try {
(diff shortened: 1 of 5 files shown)


ammen99:
> It works nicely!

Thanks for the confirmation :)

PR #2835, opened 2025-09-12
Add view properties

### LINKED ISSUE #1832: Per-window animations & effects / Window rules for animations & effects
I've been looking through the docs for quite a bit now trying to find something about animating specific windows with specific animations.

For instance animating a notification window with a fade when closed while normal windows still have the zoom effect.
Another use for this would be applying specific blur-settings on a per-window basis, for example having more blur on a file manager and less blur on your taskbar.

This could possibly be integrated into the window rules, as an `animate` or `blur` etc. command.

Hyprland has a very refined system for applying settings per window rules which I personally really like.
https://wiki.hyprland.org/Configuring/Window-Rules/#rules

ammen99:
This PR adds a more streamlined way to attach custom data (integers, strings, etc.) to views under a given key - properties. They are meant to be used for setting metadata which is unrelated to the 'main' view state (geometry, minimized/tiled/fullscreen state, etc.), as well as custom properties that various plugins can use.

Planned properties for this first PR:

- [x] `startup-x` and `startup-y` to replace `is_positioned` in the view-mapped signal. This way Xwayland can indicate it has a custom position for the views and the place plugin should not move them. Similar could be set via IPC or any custom plugin.
- [x] A way to create IPC scripts which delay the map transaction, allowing IPC clients to set custom properties (and also set initial geometry, etc.)
- [x] ~~A way to set properties via window-rules or similar~~ (will be IPC only for now, PRs welcome for window-rules properties)
- [x] `open-animation-type`, `open-animation-duration`, `close-animation`, `close-animation-duration` for the `animate` plugin (fixes #1832)

### DIFF
--- a/plugins/animate/animate.cpp
+++ b/plugins/animate/animate.cpp
@@ -295,46 +295,80 @@ class wayfire_animation : public wf::plugin_interface_t, private wf::per_output_
         wf::animation_description_t duration;
     };
 
-    view_animation_t get_animation_for_view(
-        wf::option_wrapper_t<std::string>& anim_type, wayfire_view view)
+
+    view_animation_t get_animation_for_view(const std::string& anim_type,
+        wf::option_wrapper_t<std::string>& anim_type_cfg, wayfire_view view)
     {
-        /* Determine the animation for the given view.
-         * Note that the matcher plugin might not have been loaded, so
-         * we need to have a fallback algorithm */
+        std::string desired_effect_name;
+        std::optional<wf::animation_description_t> desired_duration;
+
+        // Default animation options (open/close/minimize_animation + enabled_for)
+        if (animation_enabled_for.matches(view))
+        {
+            desired_effect_name = (std::string)anim_type_cfg;
+        }
+
+        // Specialized animation options (fade/zoom/fire enabled for specific view classes)
         if (fade_enabled_for.matches(view))
         {
-            return {"fade", fade_duration};
+            desired_effect_name = "fade";
+            desired_duration    = fade_duration.value();
         }
 
         if (zoom_enabled_for.matches(view))
         {
-            return {"zoom", zoom_duration};
+            desired_effect_name = "zoom";
+            desired_duration    = zoom_duration.value();
         }
 
         if (fire_enabled_for.matches(view))
         {
-            return {"fire", fire_duration};
+            desired_effect_name = "fire";
+            desired_duration    = fire_duration.value();
         }
 
-        if (animation_enabled_for.matches(view))
-        {
-            if (!effects_registry->effects.count(anim_type))
-            {
-                if ((std::string)anim_type != "none")
-                {
-                    LOGE("Unknown animation type: \"", (std::string)anim_type, "\"");
-                }
+        // Lastly, check view properties. They take the highest priority.
+        const std::string anim_type_prop     = anim_type + "-animation-type";
+        std::optional<std::string> type_prop = view->get_property<std::string>(anim_type_prop);
+        desired_effect_name = type_prop.value_or(desired_effect_name);
 
-                return {"none", wf::animation_description_t{0, {}, ""}};
+        const std::string anim_duration_prop = anim_type + "-animation-duration";
+        std::optional<std::string> duration_prop_str =
+            view->get_property<std::string>(anim_duration_prop);
+        if (auto duration_prop_str = view->get_property<std::string>(anim_duration_prop))
+        {
+            auto duration_prop =
+                wf::option_type::from_string<wf::animation_description_t>(*duration_prop_str);
+
+            if (duration_prop)
+            {
+                desired_duration = duration_prop;
+            } else
+            {
+                LOGE("Invalid animation duration property \"", *duration_prop_str, "\"");
+            }
+        }
+
+        if (!effects_registry->effects.count(desired_effect_name))
+        {
+            if ((std::string)desired_effect_name != "none")
+            {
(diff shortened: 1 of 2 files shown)


killown:
> In case anyone is wondering what this is all about:
> 
>     * This PR adds 'properties' which is nothing else but an extension of the existing custom data mechanism. With properties, we can add metadata of various types to a view (named integer/string/bool/etc properties). The most interesting feature of this PR are the animation properties, which allow us a very fine grained control over which views get what animations (you can even use it to have a random animation every time you open a new window, etc.).
> 
>     * Currently, IPC is completely async. For example, when a new view is opened, it is placed on the screen somewhere. Then we send view-mapped in the IPC client. The IPC client can then configure the view in any way it wants. However, this is not ideal as the IPC script cannot set the 'initial' geometry/etc. of the view. With this PR, IPC clients can request that Wayfire delays the map operation, until the IPC client is done setting it up.
> 
> 
> Example IPC script showing how this works in practice:
> 
> ```python
> import wayfire
> 
> s = wayfire.WayfireSocket()
> # With pre-map events, Wayfire will delay mapping the view until we unblock it (or until transaction timeout happens)
> s.watch(['view-pre-map'])
> cnt = 0
> 
> while True:
>     ev = s.read_next_event()
>     # We get only view-pre-map events as those are the only events we subscribe to
>     view = ev['view']
>     if view['title'] == "simple-egl":
>         # set spin animation for weston-simple-egl only
>         s.set_view_property(view['id'], 'open-animation-type', 'spin')
>     else:
>         # rotate through different animations for each other view
>         cnt = (cnt + 1) % 3
>         if cnt == 0:
>             # Use default animation settings
>             pass
>         if cnt == 1:
>             # zoom animation for every third view, starting from the second view
>             s.set_view_property(view['id'], 'open-animation-type', 'zoom')
>         if cnt == 2:
>             # fire animation with a custom duration for every third view, starting from the third view
>             s.set_view_property(view['id'], 'open-animation-type', 'fire')
>             s.set_view_property(view['id'], 'open-animation-duration', '1s circle')
> 
>     # Tell Wayfire we have finished setting up the view and it can be mapped now.
>     s.unblock_view_map(view['id'])
> ```
> 
> Note: view-pre-map event is not subscribed by default when we do a `socket.watch()` call in pywayfire, because otherwise it would introduce a lot of delays on the Wayfire side, unless the client **explicitly** subscribes to view-pre-map, then it is responsible for calling `unblock_view_map` to avoid these delays.

how the custom property works?

```
In [13]: sock.set_view_property(12, "sticky", True)
Out[13]: {'result': 'ok'}

In [14]: sock.get_view(id)
Out[14]: 
{'id': 12,
 'pid': 103832,
 'title': 'IPython: home/neo',
 'app-id': 'kitty',
 'base-geometry': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'parent': -1,
 'geometry': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'bbox': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'output-id': 1,
 'output-name': 'DP-1',
 'last-focus-timestamp': 11698151762696,
 'role': 'toplevel',
 'mapped': True,
 'layer': 'workspace',
 'tiled-edges': 15,
 'fullscreen': False,
 'minimized': False,
 'activated': True,
 'sticky': False,
 'wset-index': 1,
 'min-size': {'width': 0, 'height': 0},
 'max-size': {'width': 0, 'height': 0},
 'focusable': True,
 'type': 'toplevel'}

In [15]: sock.set_view_property(12, "hello", True)
Out[15]: {'result': 'ok'}

In [16]: sock.get_view(id)
Out[16]: 
{'id': 12,
 'pid': 103832,
 'title': 'IPython: home/neo',
 'app-id': 'kitty',
 'base-geometry': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'parent': -1,
 'geometry': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'bbox': {'x': 0, 'y': 36, 'width': 1920, 'height': 1044},
 'output-id': 1,
 'output-name': 'DP-1',
 'last-focus-timestamp': 11698151762696,
 'role': 'toplevel',
 'mapped': True,
 'layer': 'workspace',
 'tiled-edges': 15,
 'fullscreen': False,
 'minimized': False,
 'activated': True,
 'sticky': False,
 'wset-index': 1,
 'min-size': {'width': 0, 'height': 0},
 'max-size': {'width': 0, 'height': 0},
 'focusable': True,
 'type': 'toplevel'}
```

ammen99:
The properties are not visible over IPC, do you need that? I could add it but the issue is that properties can store any type of value, so you can't easily serialize them to json.

ammen99:
Also it would be easy to add a query for a particular property (for example it is easy to query for a property with a name 'open-animation-type' and value type 'string'), if that is enough for your use-case.

killown:
> Also it would be easy to add a query for a particular property (for example it is easy to query for a property with a name 'open-animation-type' and value type 'string'), if that is enough for your use-case.

My use case would be something like `view['icon'] = "some_icon"`, so the IPC broadcast sends this view data to subscribers, eliminating the need to query for an icon. Not only that, I am sure I would have much more use cases, but this one I picked randomly


> Ah, also @killown, properties is not the same as what you see in the json list. What you see in the json currently are fixed parts of the definition of a view, whereas properties are like random attached metadata, not part of the view as seen in the json description.

if I understood well, your PR is not specifically about custom data yet xd

PR #2851, opened 2025-09-24
wayfire.ini: desc on ime and ipc

liumingedwin:
instrumental words on ime and ipc appended. 
these configurations ARE NOT DOCUMENTED.

### DIFF
--- a/wayfire.ini
+++ b/wayfire.ini
@@ -79,7 +79,12 @@ plugins = \
 # Feel free to add it to the list if you want it.
 # You can find its documentation here:
 # https://github.com/WayfireWM/wayfire/wiki/Configuration#blur
-
+# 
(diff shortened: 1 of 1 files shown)


ammen99:
Hi! I think that unfortunately the example ini file is not the correct place to put these comments, as helpful as they are. But you can put them in the wiki, so that other people can benefit from your knowledge as well. For example I think it would be good to put IME information in the tips and tricks here: https://github.com/WayfireWM/wayfire/wiki/Tips-&-Tricks

About plugin documentation, feel free to add the missing plugins here: https://github.com/WayfireWM/wayfire/wiki/Configuration

The wiki is editable by anyone, so we mostly rely on the community to keep it up-to-date :)

That being said, thank you for your PR, this is definitely important information and we should put it in the wiki.

Issue #2558, opened 2025-01-06
Wayfire failed to build from source in AUR


leandrocunha526:
In the AUR it is not possible to rebuild and I am still checking the possible causes for this since I am just getting to know this software. See the log below.

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wayfire

```
FAILED: test/geometry/geometry_test 
c++  -o test/geometry/geometry_test test/geometry/geometry_test.p/geometry_test.cpp.o -Wl,--as-needed -Wl,--no-undefined -pie -Wl,--whole-archive -Wl,--start-group src/liblibwayfire.a -Wl,--no-whole-archive -rdynamic -Wl,-E -lstdc++fs -Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,-z,pack-relative-relocs -flto=auto -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wp,-D_GLIBCXX_ASSERTIONS -g -ffile-prefix-map=/home/arch/git/archlinux/aur/wayfire/src=/usr/src/debug/wayfire -flto=auto '-Wl,-rpath,$ORIGIN/../../subprojects/wf-utils:/usr/lib/wlroots0.17' -Wl,-rpath-link,/home/arch/git/archlinux/aur/wayfire/src/build/subprojects/wf-utils -Wl,-rpath-link,/usr/lib/wlroots0.17 proto/libwl_protos.a subprojects/wf-utils/libwf-utils.so.0.4.0 subprojects/wf-touch/libwftouch.a /usr/lib/libwayland-server.so /usr/lib/wlroots0.17/libwlroots.so /usr/lib/libxkbcommon.so /usr/lib/libinput.so /usr/lib/libpixman-1.so /usr/lib/libdrm.so /usr/lib/libEGL.so /usr/lib/libGLESv2.so /usr/lib/libglm.a -ldl /usr/lib/libwf-config.so /usr/lib/libxcb.so /usr/lib/libjpeg.so /usr/lib/libpng16.so /usr/lib/libwayland-client.so -Wl,--end-group
{standard input}: Assembler messages:
[...]

PR #2864, opened 2025-10-20
seat: ensure activated state does not remain on minimized toplevel

### LINKED ISSUE #2858: Minimized windows don't lose their zwlr_foreign_toplevel_handle_v1::state activated flag
**Describe the bug**
When an active window is minimized, it does not automatically lose its active status. It only loses it if another window takes that status. So if the minimized window is the only one present, it retains its active status even though it is no longer active.

This can break the functionality of panels or other components that use wlr-foreign-toplevel-management and expect that a minimized window is never active.

Downstream issue: https://gitlab.xfce.org/xfce/xfce4-panel/-/issues/954

**To Reproduce**
1. Open any app window
2. Minimize it
3. Trace its active status and see that it is not cleared

**Expected behavior**
Minimized windows should never be active.

**Wayfire version**
0.10.0 from the AUR on arch linux

ammen99:
Fixes #2858

@Tamaranch please test

### DIFF
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -344,7 +344,7 @@ class wayfire_wm_actions_output_t : public wf::per_output_plugin_instance_t
 
     wf::activator_callback on_bring_to_front = [=] (auto ev) -> bool
     {
-        return execute_for_selected_view(ev.source, [this] (wayfire_view view)
+        return execute_for_selected_view(ev.source, [] (wayfire_view view)
         {
             view_bring_to_front(view);
             return true;


ammen99:
> It's slightly better, but not enough. From the user's point of view, the key point is really that the layer-shell surface loses focus. As long as it hasn't lost focus, the minimized window retains the `active` flag. After this fix, all I have to do is bring up a context menu on the panel for it to lose focus (at least momentarily), and then the minimized window loses the `active` flag. Before, this wasn't possible.
> 
> But with the test program proposed in #2858, it doesn't work, because there is no way to artificially cause the layer-shell surface to lose focus.

Sorry I am having trouble understanding what you mean here. Can you describe it step by step so that I can see which case doesn't work with this PR applied?

Tamaranch:
> fwiw as a workaround you can set `workarounds/keep_last_toplevel_activated` to false in the config file and that should fix the issue.

No, unfortunately that doesn't fix the issue. Let's say it changes it to its opposite: the window loses its `active` flag as soon as you click on the layer-shell surface, so if the minimize action is based on that flag, as in my test program in #2858 (`xfw_screen_get_active_window`), and as in xfce4-panel, the window can never be minimized.

Tamaranch:
Yes, I can confirm that this fixes the problem, both with my test program in #2558 and with xfce4-panel. Thank you!

PR #2875, opened 2025-11-03
fix musl compilation

clemperorpenguin:
changed uint to standard type unsigned int, which fixes a compilation error on musl systems.


ammen99 (approved):
Thanks!

PR #2876, opened 2025-11-04
ipc-rules: add custom events and get-view-property

### LINKED ISSUE #2438: Add custom signals to Wayfire's IPC events
**Hello,**

could you enable a way to add custom signals to Wayfire's IPC events? I am working on a Wayfire plugin and need to handle additional IPC events.

_The goal is to inject custom signals within the IPC events_

Thank you for considering this request!

ammen99:
This PR adds support for get-view-property. This can be useful if IPC clients want to attach custom metadata to a view (string, double, int, bool) such that other IPC clients can then read them. Note that there is no way to list all available properties, they need to be looked up by their name.

Another feature added in this PR is the ability for plugins to send custom events to IPC clients. Custom events should end in an `#` and can be used like this:

```diff
diff --git a/plugins/wm-actions/wm-actions.cpp b/plugins/wm-actions/wm-actions.cpp
index ccacd61a..53c8d751 100644
--- a/plugins/wm-actions/wm-actions.cpp
+++ b/plugins/wm-actions/wm-actions.cpp
@@ -7,6 +7,7 @@
 #include "wayfire/plugins/ipc/ipc-method-repository.hpp"
 #include "wayfire/plugins/ipc/ipc-helpers.hpp"
 #include "wayfire/plugins/ipc/ipc-activator.hpp"
+#include "../ipc-rules/ipc-rules-common.hpp"
 #include "wayfire/core.hpp"
 #include "wayfire/plugin.hpp"
 #include "wayfire/scene-operations.hpp"
@@ -83,6 +84,11 @@ class wayfire_wm_actions_output_t : public wf::per_output_plugin_instance_t
         wf::wm_actions_above_changed_signal data;
         data.view = view;
         output->emit(&data);
+
+        wf::json_t ev;
+        ev["view"] = wf::ipc_rules::view_to_json(view);
+        ev["above"] = above;
+        wf::ipc_rules::send_event_to_subscribes(ev, "view-above-changed#");
         return true;
     }
```

If this patch were applied, IPC clients either watching for the event (`socket.watch()` or `socket.watch(["view-above-changed#"])`) will get an event every time the always-on-top state changes. This mechanism is aimed at external plugins so that they can integrate with IPC.

Fixes #2438

cc @killown

### DIFF
--- a/plugins/ipc-rules/ipc-rules.cpp
+++ b/plugins/ipc-rules/ipc-rules.cpp
@@ -151,19 +151,39 @@ class ipc_rules_t : public wf::plugin_interface_t,
         auto value = data["value"];
         if (value.is_bool())
         {
-            view->set_property<bool>(property, value.as_bool());
+            if (!view->set_property<bool>(property, value.as_bool()))
+            {
+                return wf::ipc::json_error("failed to set property " + property +
+                    " as bool value, wrong property type?");
+            }
         } else if (value.is_int64())
         {
-            view->set_property<int64_t>(property, value.as_int64());
+            if (!view->set_property<int64_t>(property, value.as_int64()))
+            {
+                return wf::ipc::json_error("failed to set property " + property +
+                    " as int64 value, wrong property type?");
+            }
         } else if (value.is_uint64())
         {
-            view->set_property<uint64_t>(property, value.as_uint64());
+            if (!view->set_property<uint64_t>(property, value.as_uint64()))
+            {
+                return wf::ipc::json_error("failed to set property " + property +
+                    " as uint64 value, wrong property type?");
+            }
         } else if (value.is_string())
         {
-            view->set_property<std::string>(property, value.as_string());
+            if (!view->set_property<std::string>(property, value.as_string()))
+            {
+                return wf::ipc::json_error("failed to set property " + property +
+                    " as string value, wrong property type?");
+            }
         } else if (value.is_double())
         {
-            view->set_property<double>(property, value.as_double());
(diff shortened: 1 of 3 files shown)


ammen99:
> @ammen99 amazing work, thank you!
> 
> Not an issue but if we set a view property as a bool, we can't change its type later:
> 
> ```python
> In [15]: sock.set_view_property(id, "hello_world!", True)
> Out[15]: {'result': 'ok'}
> 
> In [16]: sock.get_view_property(id, "hello_world!")
> Out[16]: {'result': 'ok', 'value': True}
> 
> In [17]: sock.set_view_property(id, "hello_world!", "Hello!")
> Out[17]: {'result': 'ok'}
> 
> In [18]: sock.get_view_property(id, "hello_world!")
> Out[18]: {'result': 'ok', 'value': True}
> 
> In [19]: sock.set_view_property(id, "hello_world!", False)
> Out[19]: {'result': 'ok'}
> 
> In [20]: sock.get_view_property(id, "hello_world!")
> Out[20]: {'result': 'ok', 'value': False}
> 
> In [21]: sock.set_view_property(id, "hello_world!", True)
> Out[21]: {'result': 'ok'}
> 
> In [22]: sock.get_view_property(id, "hello_world!")
> Out[22]: {'result': 'ok', 'value': True}
> ```

Yes, properties are supposed to have a fixed type. If you want dynamic type, you can always use string and dynamically parse it as int/bool/whatever. However you raise a good point, we should add an error for setting a property with the wrong type, so that the user knows they are using the wrong type.

PR #2890, opened 2025-11-14
tile: keep fullscreen and maximized on adjacent focus and new floating views

ammen99:
@killown Any objections?

### DIFF
--- a/plugins/tile/tile-wset.hpp
+++ b/plugins/tile/tile-wset.hpp
@@ -67,6 +67,20 @@ class tile_workspace_set_data_t : public wf::custom_data_t
         outer_vert_gaps.set_callback(update_gaps);
     }
 
+    ~tile_workspace_set_data_t()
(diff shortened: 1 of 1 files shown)


killown:
> @killown Any objections?

Your PR is unmaximizing all views from every workspace when a new view is created. Wouldn’t it make more sense to unmaximize only the views from the current workspace?

killown:
> @killown Updated, please review / test :)

Amazing, working as intended!

Since this PR is about fixes for simple title, I'll report this here instead of creating a new issue. Not a big issue but in case you consider the following a bug:

When you disable simple-title, all tiled views won't resizing and maximizing. You can no longer change the view configuration state. The only fix is to close the app and start a new view

To reproduce, simply try resize a view after disabling the plugin

ammen99:
@killown It should work now with the latest commit.

PR #2913, opened 2025-11-28
window-rules: allow using X11-style geometry

js202510:
As an alternative to integer x/y/w/h geometry, allow an X11-style geometry "WxH+x-y" as string value for the "set geometry" action.

It has two advantages:
- it allows the window position to be specified from the right or bottom edge of the workarea (e.g. "-0-0" puts the view into the lower right corner)
- it allows to specify the positon without changing the size (use the applications default size)

Maybe use of the X11 geometry syntax is controversial, but this functionality is missing, I hope it could be added to wayfire in one way or other.

### DIFF
--- a/plugins/window-rules/view-action-interface.cpp
+++ b/plugins/window-rules/view-action-interface.cpp
@@ -17,6 +17,7 @@
 #include <cfloat>
 #include <string>
 #include <vector>
+#include <sstream>
 
 namespace wf
 {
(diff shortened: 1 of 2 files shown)


ammen99:
Hi and thanks for your PRs! I wonder whether you have looked at Wayfire's IPC though, it can be used to achieve this without resorting to a weird syntax.

plugins/window-rules/view-action-interface.cpp
```diff
@@ -338,6 +339,80 @@ std::tuple<bool, float> view_action_interface_t::_validate_alpha(
...
+            int x = bb.x, y = bb.y, w = bb.width, h = bb.height;
+            char c;
+
+            c = g.peek();
+            if ((c != 'x') && (c != '-') && (c != '+'))
+            {
+                g >> w;
+                if (g.fail())
+                {
+                    goto err;
```
ammen99: Not a fan of goto, I wouldn't add it in newer code. Also have you considered falling back to C-style sscanf for this? I think it would make parsing way simpler to read and follow.

ammen99: (this == the parsing of the geometry string)

PR #2916, opened 2025-12-01
Move detection and calculation of grid snapping to grid plugin from move plugin

BwackNinja:
Second attempt working towards https://github.com/WayfireWM/wayfire/pull/1619

The grid plugin now responds to moves and shows previews unless it isn't loaded. When it isn't loaded, the move plugin still handles maximizing windows and the associated preview.

### DIFF
--- a/plugins/grid/wayfire/plugins/grid.hpp
+++ b/plugins/grid/wayfire/plugins/grid.hpp
@@ -8,16 +8,28 @@ namespace wf
 {
 namespace grid
 {
+enum move_op_t
+{
+    MOVE_OP_CLEAR_PREVIEW  = 0,
+    MOVE_OP_UPDATE_PREVIEW = 1,
+    MOVE_OP_DROP           = 2,
+};
+
 /**
  * name: request
  * on: core
  * when: Emitted before move renders a grid indicator and sets the slot.
  * carried_out: true if a plugin can handle move request to grid.
  */
-struct grid_request_signal
+struct grid_handle_move_signal
 {
     /* True if a plugin handled this signal */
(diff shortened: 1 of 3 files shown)


plugins/grid/wayfire/plugins/grid.hpp
```diff
@@ -38,6 +26,24 @@ enum slot_t
...
+ * when: Emitted before move renders a grid indicator and sets the slot.
+ * carried_out: true if a plugin can handle move request to grid.
+ */
+struct grid_request_signal
+{
+    /* True if a plugin handled this signal */
+    bool carried_out = false;
+    bool force_off = false;
+    wf::output_t *output;
+    wf::point_t input;
```
ammen99: This one would need a docstring to clarify what it is actually (something like "input position relative to the indicated output")

ammen99:
@BwackNinja seems quite good to me now, doesn't compile however (and I expect you'd have to fix the style, look at the CI output, there is a diff showing the needed changes)

ammen99:
@BwackNinja For me, the move snap feature is completely broken now, does it work on your end? Have you tested it in a nested wayland-backend wayfire?

ammen99:
If/when you implement your alternative grid plugin, feel free to share your work here or as a link in the GitHub wiki (https://github.com/WayfireWM/wayfire/wiki/Related-projects) :)

BwackNinja:
Thanks! Initial implementation is here: https://github.com/BwackNinja/wayfire-plugin-grid-plus

I need to stop using the grid's configuration options and use its own instead, but it does seem to work.

PR #2935, opened 2025-12-26
switcher: render in stacking order 

### LINKED ISSUE #2907: [switcher] Incorrect background window switching
Sometimes, when switching between a maximized and non-maximized window, after the transition ends, the window in the background switches incorrectly

**To Reproduce**
1. Enable switcher plugin
2. Open 3 windows: 2 Fullscreen, 1 windowed
3. Alt tab from a fullscreen one to the windowed one

I can't get it to reproduce reliably.

**Expected behavior**
When using `switcher`, the background window should be the same one as from the transition

**Screenshots or stacktrace**
N/A

**Wayfire version**
`0.11.0-4cc90a06 (Nov 18 2025) branch master wlroots-0.19.2`

ammen99:
Fixes #2907

@MarcusBritanicus @CCobaltDev does this fix the issue as I think it should?

### DIFF
--- a/plugins/single_plugins/switcher.cpp
+++ b/plugins/single_plugins/switcher.cpp
@@ -297,7 +297,7 @@ class WayfireSwitcher : public wf::per_output_plugin_instance_t, public wf::keyb
 
         if (active)
         {
-            arrange(false);
+            arrange(0);
         } else
         {
             cleanup_views([=] (SwitcherView& sv)
@@ -327,7 +327,7 @@ class WayfireSwitcher : public wf::per_output_plugin_instance_t, public wf::keyb
         {
             active = true;
             input_grab->grab_input(wf::scene::layer::OVERLAY);
-            arrange(true);
+            arrange(dir);
             activating_modifiers = wf::get_core().seat->get_keyboard_modifiers();
         } else
         {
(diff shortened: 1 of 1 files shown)


ammen99:
> It fixes the rendering problem (i think?) but makes switching feel really weird
> 
> i'd expect it to switch by last-switched-to/by-focus

This PR only changes the rendering order but I think I see the issue is more general than I thought. The order of views during the animation is last-focused, but the end state is stacking, so a proper solution would require some special animations to transition between the two. Thanks for testing so far...

ammen99:
@CCobaltDev I had some time today and looked again, let me know what you think about the newest version of the code. It is not perfect if you set it to a very slow animation (at 3-4 seconds length you can see artifacts here and there) but I think it should be relatively smooth for normal animation durations (<1s).

CCobaltDev:
I like the switcher transparency
There are some window shaking/jitter artifacts with 2 windows

But doing the switch still has some odd behavior - not sure how to describe it so here's my exact case

I have 3 windows open: Firefox, Vesktop, Alacritty
When I focus Firefox and alt+tab, it shows Vesktop, Firefox, Alacritty - Firefox should be last, right? This also leads to where pressing alt+tab twice with 3 windows open has 0 net effect after the animation - same window focused as before the switcher activation
@ammen99

ammen99:
@CCobaltDev I pushed some further changes which I think should improve the cases you mentioned, let me know what you think :)

CCobaltDev:
Seems to be working great!

CCobaltDev:
Thank you so much!

ammen99:
> Thank you so much!

Thanks for testing :)

PR #2936, opened 2025-12-26
IPC: send info about the "above" state of views

dkondor:
This does not seem to be available for IPC clients, making it difficult to create an UI that allows changing it based on the current state.

Client implementation: https://github.com/dkondor/cairo-dock-core/tree/wf_actions_above

Note:
 - I chose "above" / "view-above" as the name of the property / event to be consistent with the wm-actions plugin; let me know if the more verbose "always-on-top" would be better.
 - I changed the plugin ABI version as this seems to be the standard way for IPC clients to detect the IPC API version. Let me know if I missed something and there is a better way for this.


ammen99 (approved):
LGTM, thanks!

Issue #2927, opened 2025-12-16
Add support for explicit sync (nvidia xwayland glitches)


xiaomx32:
**Describe the bug**
Open `visual studio code` in `wayfire`, the screen flashes, tears, and artifacts crazily.
When moving the mouse to the blinking and unable to load area, the situation improved slightly.
When using `wayfire` normally, there may also be slight flickering, especially after opening multiple windows and switching to other desktops(`Tab` + `Alt`).

**To Reproduce**
1. `wayfire`
2. `yay -S visual-studio-code-bin`
3.  `code test.cpp`

**Expected behavior**
the `code` window can display normally

**Software version**
```bash
pacman -Qi wayfire-lily-git 
Name            : wayfire-lily-git
Version         : 0.11.0.20251127.144214.g23a101a0-1
Description     : 3D wayland compositor (lilydjwg's version)
Architecture    : x86_64
URL             : https://github.com/lilydjwg/wayfire
Licenses        : MIT
Groups          : None
Provides        : wayfire  wayfire-git
Depends On      : wlroots-lily-git  cairo  pango  glm  libjpeg  yyjson  libyyjson.so=0-64
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : wayfire  wayfire-git
Replaces        : None
Installed Size  : 8.71 MiB
Packager        : lilydjwg <lilydjwg@gmail.com>
Build Date      : Thu 27 Nov 2025 11:13:02 PM CST
Install Date    : Fri 28 Nov 2025 11:31:39 AM CST
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : SHA-256 Sum  Signature
```
```bash
pacman -Qi wlroots-lily-git 
Name            : wlroots-lily-git
Version         : 0.19.2.20251104.031626-1
Description     : Modular Wayland compositor library (with patches)
Architecture    : x86_64
URL             : https://gitlab.freedesktop.org/lilydjwg/wlroots
Licenses        : MIT
Groups          : None
Provides        : libwlroots-0.19.so=libwlroots-0.19.so-64  wlroots0.19=0.19.2.20251104.031626
Depends On      : libdisplay-info.so=3-64  libglvnd  libinput  liblcms2.so=2-64  libliftoff.so=0-64  libpixman-1.so=0-64  libseat.so=1-64  libudev.so=1-64  libvulkan.so=1-64
[...]

PR #2938, opened 2025-12-28
xwayland: fix map-time positioning for self-positioned clients

killown:
This pr will fix frame pacing jitter in games.

https://github.com/user-attachments/assets/e05408fd-d3b4-4365-ab6a-6d142472fcc6

### DIFF
--- a/src/view/xwayland/xwayland-toplevel-view.hpp
+++ b/src/view/xwayland/xwayland-toplevel-view.hpp
@@ -53,7 +53,6 @@ class wayfire_xwayland_view : public wf::toplevel_view_interface_t, public wayfi
             wlr_xwayland_surface_configure(xw, ev->x, ev->y, ev->width, ev->height);
             if ((ev->mask & XCB_CONFIG_WINDOW_X) && (ev->mask & XCB_CONFIG_WINDOW_Y))
             {
-                this->self_positioned = true;
                 toplevel->pending().geometry.x = ev->x - output_origin.x;
                 toplevel->pending().geometry.y = ev->y - output_origin.y;
             }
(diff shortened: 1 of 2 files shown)


ammen99:
> I'll test further, but it's consistent: the game runs smoothly in Sway/Hyprland, yet lags in Wayfire even at 144 FPS. I will make sure first if this is really fixed, I tested like 20 times because I also could not find a reason why this worked. Draft for now!

Probably you simply had to restart wayfire. Alternatively, could be something with the place plugin, as that is the only plugin which cares about the startup-x/y properties that you remove here.

killown:
> > I'll test further, but it's consistent: the game runs smoothly in Sway/Hyprland, yet lags in Wayfire even at 144 FPS. I will make sure first if this is really fixed, I tested like 20 times because I also could not find a reason why this worked. Draft for now!
> 
> Probably you simply had to restart wayfire. Alternatively, could be something with the place plugin, as that is the only plugin which cares about the startup-x/y properties that you remove here.

I found the root issue, the lag stems from windowed mode, not true fullscreen. Most games default to a borderless window, which will cause the lag only on wayfire. In GNOME, both modes (windowed and exclusive fs) worked fine, confirming the issue isn't the game itself, also I tested with more games, The PR makes sense yet, I put as a draft because the code didn't fix really.

I tested with wayfire.ini from source code, the only change I did was the output setup and the lag was still happening :)

killown:
# SWAY

https://github.com/user-attachments/assets/220afdd9-99e8-4cf7-a07f-91d313a70263



# WAYFIRE
https://github.com/user-attachments/assets/4384ab29-f52b-4430-b2c4-ae5bde31626f

killown:
if it's not random, I tried for a few times and the Vulkan renderer fixed my issue. That explains why I never had problems with gaming, I've used Vulkan renderer for a long time and only recently GLES2. #2927 had the same solution. Since Wayfire's future is Vulkan, I'll drop GLES2.

ammen99:
> > > if it's not random, I tried for a few times and the Vulkan renderer fixed my issue. That explains why I never had problems with gaming, I've used Vulkan renderer for a long time and only recently GLES2. #2927 had the same solution. Since Wayfire's future is Vulkan, I'll drop GLES2.
> > 
> > 
> > Maybe the issue was in some of the GLES plugins like blur, or wobbly? Have you tried reproducing the issue with as little effects enabled as possible?
> 
> I tried the wayfire.ini from your source code, but even with that, if a plugin enabled by default can cause such gaming issues, I'd rather drop GLES2 anyway.

Well if we find that a plugin causes this, maybe we can fix the plugin for gles2 users too. Currently gles2 features many features people like to use, so it is not bad to fix gles2 plugins too.

PR #2943, opened 2026-01-04
core: include active renderer in --version output

killown:
Output:
```
wayfire --version  
0.11.0-0bdb2e06 (Jan  4 2026) branch master wlroots-0.19.2 [vulkan]
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -20,15 +20,22 @@
 static std::string get_version_string()
 {
     std::string version = std::string(WAYFIRE_VERSION) + "-" + wf::version::git_commit +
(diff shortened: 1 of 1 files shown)


ammen99 (changes requested):
If at all, I suggest printing the environmental variable for the wlroots renderer (`WLR_RENDERER`) if set. Printing the version should not require setting up the full wlroots backend+renderer.

soreau:
> If at all, I suggest printing the environmental variable for the wlroots renderer (`WLR_RENDERER`) if set. Printing the version should not require setting up the full wlroots backend+renderer.

It seems like this should only be printed in the log, not from --help or --version where the renderer environment variable is not likely to be set.

PR #2959, opened 2026-01-17
Global menu support (via IPC)

dkondor:
This is an alternative to #2576 that does not require any new Wayland protocol, so can be deployed as is. The downside is that the client displaying the global menus will need to use Wayfire's IPC as well, so will need a specialized implementation.

Note: currently, how this works is that the IPC plugins sends out the set of DBus properties whenever sending information about a view. This could be optimized by having a separate event for it, since these are not expected to change.

A simple test client is available here: https://github.com/dkondor/gtk_global_menu with basic instructions on how to make things work.

My medium-term plan is to add support in Cairo-Dock (although the Global-Menu plugin is still considered unstable and actually does not provide a nice user experience) which should be very easy to do. I might also look into creating a very simple layer-shell client that actually displays a menu bar (however, that might need something like what you did for wf-shell [here](https://github.com/WayfireWM/wf-shell/blob/master/src/panel/widgets/tray/dbusmenu.cpp) to translate the dbusmenu objects to a menu model).

### DIFF
--- a/src/api/wayfire/plugin.hpp
+++ b/src/api/wayfire/plugin.hpp
@@ -107,7 +107,7 @@ using wayfire_plugin_load_func = wf::plugin_interface_t * (*)();
 /**
  * The version is defined as macro as well, to allow conditional compilation.
  */
-#define WAYFIRE_API_ABI_VERSION_MACRO 2025'12'26
+#define WAYFIRE_API_ABI_VERSION_MACRO 2026'01'22
 
 /**
  * The version of Wayfire's API/ABI


dkondor:
> > My medium-term plan is to add support in Cairo-Dock
> 
> Hi @dkondor, this sounds great! Regarding cairo-dock specifically though, some folks on chat have wondered about cairo-dock and while I know you have it working, I do not know which repos or branches you use for it. So, can you please say which cairo-dock repos and branches you use to have the 'best' cairo-dock experience on wayfire?

My changes have been merged upstream, so it is best to use the official repositories:
https://github.com/Cairo-Dock/cairo-dock-core
https://github.com/Cairo-Dock/cairo-dock-plug-ins

dkondor:
Hi,

I've updated the PR using properties as suggested + updated the example code: https://github.com/dkondor/gtk_global_menu/ -- now it includes a very minimal client that actually displays a menu bar.

The only things that I could not figure out is whether it is possible to remove properties -- did I miss something? (Although in practice, it is not expected that clients remove these properties, so this is likely not an issue; note: the code style check is probably failing due to the TODO comment I left in there about this)

ammen99:
> Hi,
> 
> I've updated the PR using properties as suggested + updated the example code: https://github.com/dkondor/gtk_global_menu/ -- now it includes a very minimal client that actually displays a menu bar.
> 
> The only things that I could not figure out is whether it is possible to remove properties -- did I miss something? (Although in practice, it is not expected that clients remove these properties, so this is likely not an issue; note: the code style check is probably failing due to the TODO comment I left in there about this)

There is no way to remove properties currently, but it should be easy to add a new method for this purpose.

dkondor:
Looking at the code again, now I've realized that I could just use the [erase_data](https://github.com/WayfireWM/wayfire/blob/3c9a7712f88e520f025eca331dbe57c94904362a/src/api/wayfire/object.hpp#L99) function to delete a property :)

@ammen99 do you still prefer to add a separate, `erase_property()` function for clarity? (that just calls `erase_data()`)

PR #2974, opened 2026-02-12
output-layout: Add different automatic modes

hustlerone:
This adds two extra `AUTO` modes, whose behaviors differ from the main `AUTO` one.

`AUTO_HIGHRR` checks which mode has the highest refresh rate, then find the highest resolution which can still provide the refresh rate.

`AUTO_HIGHRES` just chooses the highest resolution mode available (which matches `AUTO` if no preferred video mode is specified by the display).

Depends on https://github.com/WayfireWM/wf-config/pull/87

### DIFF
--- a/include/wayfire/config/types.hpp
+++ b/include/wayfire/config/types.hpp
@@ -506,19 +506,12 @@ enum mode_type_t
 struct mode_t
 {
     /**
-     * Initialize an OFF or AUTO mode.
+     * Initialize a mode.
      *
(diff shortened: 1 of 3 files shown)


hustlerone:
Ok so this works but switching configurations might boot you back to `auto` which you don't want, `apply_configuration` in src/core/output-layout.cpp might be of interest.

hustlerone:
20 years of evolution and bash can't remember where my last command was so now i accidentally nuked this

hustlerone:
reopen

PR #2976, opened 2026-02-12
window-rules: fix view-action-interface output detection logic

mrsteve0924:
I added logic so the  make+model+serial number could be used with window-rules in addition to the usual connector names like DP-1, HDMI-A-1. This helps when connector names change especially when using a laptop dock connected to external monitors.

I had to update literal.cpp in wf-utils. see separate PR for that. thanks


mrsteve0924:
i found a regression. when wstroke plugin is enabled, activating the button modifier causes system to freeze.  wstroke probably not expecting output names other than the connector names.

ammen99 (approved):
One small nitpick, otherwise lgtm.

PR #2978, opened 2026-02-20
Inhibit output rendering from wayfire directly instead of relying on clients

soreau:
This also disables inhibiting outputs from client side, as this could lead to a black screen forever if the client does not uninhibit.


plugins/protocols/wayfire-shell.cpp
```diff
@@ -287,26 +279,21 @@ class wfs_output
 
     void inhibit_output()
     {
-        ++this->num_inhibits;
         if (this->output)
         {
-            this->output->render->add_inhibit(true);
+            LOGI(
```
ammen99: LOGW or LOGE

ammen99 (approved):
LGTM, thanks!

PR #2984, opened 2026-02-24
Fix "vswitch" plugin: undesired 'last' direction

WShad:
Fixes wrong direction on "last" caused by changing workspace by external methods.
Example: switch workspace by activating window via wf-panel, as a result _control_bindings_t_ knows nothing about this transition and remembers wrong "last direction".


ammen99:
I implemented the approach in #3088 , thanks for your PR nonetheless!

PR #2998, opened 2026-03-19
plugins: fix cairo text transparency

### LINKED ISSUE #2911: Crashed lock screen has small sections of desktop visible.
**Describe the bug**
When a lockscreen has crashed, the underlying desktop may be seen through the semi-transparent pixels of the crash image.

**To Reproduce**
Place an identifiable pattern on desktop, or a text heavy document. I used a large on screen keyboard.
Lock the screen, SSH in and kill it again. 

**Expected behavior**
None of the desktop should be visible through this screen

**Screenshots or stacktrace**

Image
Zooming in may be necessary, but text is visible from the UI below.

**Wayfire version**
Latest git

ziadkhouri:
fixes: https://github.com/WayfireWM/wayfire/issues/2911

Use [CAIRO_OPERATOR_OVER](https://cairographics.org/operators/) when rendering text with a background

### DIFF
--- a/plugins/common/wayfire/plugins/common/cairo-util.hpp
+++ b/plugins/common/wayfire/plugins/common/cairo-util.hpp
@@ -256,7 +256,7 @@ struct cairo_text_t
         x += xpad;
         y += ypad;
 
-        cairo_set_operator(cr, CAIRO_OPERATOR_SOURCE);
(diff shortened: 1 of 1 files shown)


ammen99 (approved):
LGTM, thanks!

PR #2999, opened 2026-03-19
subprojects: wlroots: Update to 0.19.3

soreau:


ammen99:
I think we should switch to wlroots 0.20 directly, but thanks for the PR nonetheless :)

soreau:
> I think we should switch to wlroots 0.20 directly, but thanks for the PR nonetheless :)

Indeed.. please see #3002.

PR #3009, opened 2026-04-14
xwayland: fix workarea calculation for workspaces other than 0,0

### LINKED ISSUE #2971: #2954  makes SAP GUI for Java unusable on all virtual screens except 1.
i have desktop configured as 3х2

1. On first desktop (upper left in wall) - all OK. But if I chhose other desktop - unusable.

Opening SAP Logon screen - sometimes it opens, sometimes not. Looks like it tries to open - and then almost immediately disappear. Same with all spawning windows - when I try to logon, open new session, some pop-up dialog. The java process related to disappeared window continues to run, it is not crached.  Looks like the window is moved to other desktop (desktop 1) - but usually it can not be found anywhere (but sometimes it is on desktop1 - if I open window on 1, drag it to any other and do new session from there)

Reverting #2954 solves the issue.

ammen99:
@mark-herbert42 Can you please test whether this PR fixes your issue?

Fixes #2971

### DIFF
--- a/src/view/xwayland/xwayland-toplevel-view.hpp
+++ b/src/view/xwayland/xwayland-toplevel-view.hpp
@@ -106,9 +106,10 @@ class wayfire_xwayland_view : public wf::toplevel_view_interface_t, public wayfi
         // View workspace relative to current workspace
         if (main_view->get_wset())
         {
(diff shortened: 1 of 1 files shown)


mark-herbert42:
Fixes the issue.

ammen99:
> Fixes the issue.

Thanks for testing!

PR #3012, opened 2026-04-16
xwayland: set cursor scale

### LINKED ISSUE #3011: When force_xwayland_scaling is true, cursor is big and blurry
## Wayfire version

1c7d7775

## GPU / Driver

AMD Radeon RX 6650 XT (RADV NAVI23) Mesa 26.0.4-arch1.1

## Describe the bug

When force_xwayland_scaling is true and the output scale is 2, cursor on Xwayland is big and blurry.

## To Reproduce

1. Set output scale to 2 and Xft.dpi to 192
2. Set force_xwayland_scaling to true
3. Run an Xwayland application, e.g. `env -u WAYLAND_DISPLAY firefox`

## Expected behavior
Xwayland cursor should take the scale into account.

Same Firefox window and cursor position, one is Xwayland and the other is native Wayland.

Xwayland screenshot
native wayland screenshot

ammen99:
Fixes #3011

@lilydjwg This was my first idea about how to fix the issue. I think it might not work very well with mixed DPI, however. Let me know whether this is the correct approach, or maybe we need to do something else?

### DIFF
--- a/src/core/seat/cursor.cpp
+++ b/src/core/seat/cursor.cpp
@@ -236,6 +236,11 @@ void wf::cursor_t::set_cursor(
         }
     }
 
+    if (ev->surface)
+    {
+        wf::xwayland_adjust_cursor(ev->surface);
+    }
+
     wlr_cursor_set_surface(cursor, ev->surface, ev->hotspot_x, ev->hotspot_y);
 
     last_cursor_name.clear();
--- a/src/view/view-impl.hpp
+++ b/src/view/view-impl.hpp
@@ -77,6 +77,7 @@ void xwayland_update_default_cursor();
 void xwayland_bring_to_front(wlr_surface *surface);
 int xwayland_get_pid();
 wl_client *xwayland_get_client();
+void xwayland_adjust_cursor(wlr_surface *cursor_surface);
 
 void init_desktop_apis();
 void fini_desktop_apis();
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -347,3 +347,64 @@ wl_client*wf::xwayland_get_client()
     return nullptr;
 #endif
 }
+
+#if WF_HAS_XWAYLAND
+static void update_xwayland_cursor_surface_scale(wlr_surface *surface, int32_t scale)
+{
+    if (!surface || (scale <= 0))
+    {
+        return;
+    }
+
+    surface->current.scale = scale;
+
+    int width  = surface->current.buffer_width;
+    int height = surface->current.buffer_height;
+    wlr_output_transform_coords(surface->current.transform, &width, &height);
+
+    width  /= scale;
+    height /= scale;
+
+    surface->current.width  = width;
+    surface->current.height = height;
+}
+
+#endif
+
+void wf::xwayland_adjust_cursor(wlr_surface *cursor_surface)
+{
+#if WF_HAS_XWAYLAND
+    if (wl_resource_get_client(cursor_surface->resource) == wf::xwayland_get_client())
+    {
+        static wf::option_wrapper_t<bool> force_xwayland_scaling{"workarounds/force_xwayland_scaling"};
+        static wf::wl_listener_wrapper on_cursor_destroy;
+        static wf::wl_listener_wrapper on_cursor_commit;
+        static wlr_surface *last_cursor_surface = nullptr;
+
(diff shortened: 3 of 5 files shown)


lilydjwg:
It doesn't work. The code is executed but nothing has changed.

ammen99:
@lilydjwg Try the latest commit. Does it at least change things for xterm? Here, I can see quite a big difference.

lilydjwg:
No, nothing has changed.

ammen99:
> No, nothing has changed.

What is your ini file? Also are you sure that firefox is running Xwayland mode to begin with? I just tested gedit and firefox with the x11 backend and the cursor has the same size as wayland native applications with my patch.

lilydjwg:
Here is my [wayfire.ini](https://github.com/lilydjwg/dotconfig/blob/10d5ae379223caa6ba9df8f848c5541d67f36620/wayfire.ini).

Besides Firefox, I also tested with xterm.

lilydjwg:
I gave codex gpt-5.4 a try and it ended up copying some logic from wlroots' `surface_finalize_pending`, and it worked as expected. Here's a slightly-revised patch from codex:

```diff
diff --git a/src/view/xwayland.cpp b/src/view/xwayland.cpp
index 3fa17742..0fc60185 100644
--- a/src/view/xwayland.cpp
+++ b/src/view/xwayland.cpp
@@ -215,6 +215,39 @@ class xwayland_view_controller_t
 static wlr_xwayland *xwayland_handle = nullptr;
 static wf::wl_listener_wrapper on_xwayland_surface_created;
 static wf::wl_listener_wrapper on_xwayland_ready;
+
+static void update_xwayland_cursor_surface_scale(wlr_surface *surface, int32_t scale)
+{
+    if (!surface || (scale <= 0))
+    {
+        return;
+    }
+
+    surface->current.scale = scale;
+
+    int width  = surface->current.buffer_width;
+    int height = surface->current.buffer_height;
+    wlr_output_transform_coords(surface->current.transform, &width, &height);
+
+    if (surface->current.viewport.has_src)
+    {
+        width  = surface->current.viewport.src.width;
+        height = surface->current.viewport.src.height;
+    } else
+    {
+        width  /= scale;
+        height /= scale;
+    }
+
+    if (surface->current.viewport.has_dst)
+    {
+        width  = surface->current.viewport.dst_width;
+        height = surface->current.viewport.dst_height;
+    }
+
+    surface->current.width  = width;
+    surface->current.height = height;
+}
 #endif
 
 void wf::init_xwayland(bool lazy)
@@ -372,7 +405,7 @@ void wf::xwayland_adjust_cursor(wlr_surface *cursor_surface)
             {
                 auto& ol = wf::get_core().output_layout;
                 auto wo  = ol->find_closest_output(wf::get_core().get_cursor_position());
-                last_cursor_surface->current.scale = wo->handle->scale;
+                update_xwayland_cursor_surface_scale(last_cursor_surface, wo->handle->scale);
             });
 
             on_cursor_destroy.connect(&cursor_surface->events.destroy);
```

ammen99:
I feel like the src/dst bits are wrong but we also need to scale the size, right. I am not sure why it seems to work here ...

lilydjwg:
Maybe some code path isn't taken in my case? (I looked at this place yesterday but didn't figure out a working solution.)

ammen99:
> Maybe some code path isn't taken in my case? (I looked at this place yesterday but didn't figure out a working solution.)

Yes, I don't expect many if any clients at all use viewport for cursors, definitely not xwayland. If you delete all of these branches, does it still work? If yes, we could merge.

lilydjwg:
Yes, I simplified the function to the following:

```cpp
static void update_xwayland_cursor_surface_scale(wlr_surface *surface, int32_t scale)
{
    if (!surface || (scale <= 0))
    {
        return;
    }

    surface->current.scale = scale;

    int width  = surface->current.buffer_width;
    int height = surface->current.buffer_height;
    wlr_output_transform_coords(surface->current.transform, &width, &height);

    width  /= scale;
    height /= scale;

    surface->current.width  = width;
    surface->current.height = height;
}
```

And it still works.

ammen99:
@lilydjwg Should I push the changes or will you send a PR to this branch?

lilydjwg:
You can just push the changes.

ammen99:
Hopefully I pushed everything correctly, let me know if I messed something up @lilydjwg

PR #3019, opened 2026-04-30
Extending the wayfire-shell-unstable-v2 protocol

marcusbritanicus:
- <del>Added create_positioned_hotspot request to zwf_output_v2 interface.</del>
- Added create_custom_hotspot request to zwf_output_v2 interface.
- Added proximity_changed event to zwf_hotspot_v2 interface.
- All hotspots (edge, corner, positioned) will now get proximity_changed signals


proto/wayfire-shell-unstable-v2.xml
```diff
@@ -115,12 +173,45 @@
...
       </description>
     </event>
+
+    <event name="proximity_changed" since="3">
+      <description summary="Proximity to trigger point changed">
+        Sent after an enter event and before the corresponding leave event,
+        whenever the input pointer moves within the hotspot area.
+
+        Proximity indicates the current distance in pixels from the input pointer
+        to the trigger point of the hotspot:
```
ammen99: I find this wording a bit confusing, is there a 'trigger point of the hotspot'? I thought hotspots can be rectangles.

plugins/protocols/wayfire-shell.cpp
```diff
@@ -296,59 +353,144 @@ class wfs_output
         }
     }
 
-    void create_hotspot(uint32_t hotspot, uint32_t threshold, uint32_t timeout,
-        uint32_t id)
+    wf::geometry_t calculate_trigger_geometry(uint32_t edge_mask)
+    {
+        wf::geometry_t output_geom = this->output->get_layout_geometry();
+        int edge_count = 0;
```
ammen99: This looks quite inefficient, can't we replace it by `__builtin_popcount`? We already use these types of functions in other places.

plugins/protocols/wayfire-shell.cpp
```diff
@@ -71,92 +73,144 @@ class wfs_hotspot
...
+            return false;
+        }
+
+        // Exact distance check
+        return calculate_proximity(cursor) <= threshold;
+    }
+
+    void send_proximity_if_needed(const wf::point_t& cursor)
+    {
+        if (wl_resource_get_version(hotspot_resource) < 2)
```
ammen99: there should be a macro defined as the version in which proximity was defined (in the protocol header created by wayland-scanner).

plugins/protocols/wayfire-shell.cpp
```diff
@@ -71,92 +73,144 @@ class wfs_hotspot
...
+
+        // Clamp to output boundaries
+        wf::geometry_t result = wf::geometry_intersection(expanded, output_geom);
+
+        return result;
+    }
+
+    uint32_t calculate_proximity(const wf::point_t& cursor) const
+    {
+        const auto& rect = trigger;
```
ammen99, line 124: There is a wlroots function to do this for you :) https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/util/box.h?ref_type=heads#L53

ammen99 (commented):
lgtm overall, a few things I noticed, I trust you have tested this :)

proto/wayfire-shell-unstable-v2.xml
marcusbritanicus: Yes, hotspots can be rectangles. I tried to write a general explanation, which was way too confusing. In case of the rectangle, there are simply 2(l+b) trigger points along the perimeter of the rectangle. So, in the rectangular case, the point on the perimeter closest to the mouse is the "trigger point".

I agree that it can be a bit misleading. If you have a better wording, I am happy to incorporate it.

ammen99: `Proximity is defined as the length of the shortest segment connecting the input pointer to the boundary of the hotspot rectangle` or something like it?

plugins/protocols/wayfire-shell.cpp
marcusbritanicus: @ammen99 Sure, I could use that. But to what end? I still have to calculate the distance to the box. Which means, I either have to do all the calculations. And I have to handle the edge cases separately: I allow the box to be empty, wlr dose not.

plugins/protocols/wayfire-shell.cpp
marcusbritanicus: I think what you mean is 

`if (wl_resource_get_version(hotspot_resource) < ZWF_SHELL_HOTSPOT_V2_PROXIMITY_SINCE_VERSION)`

Please correct me if I am wrong.

proto/wayfire-shell-unstable-v2.xml
marcusbritanicus: This is what I have come up with:

Proximity is defined as the Euclidean distance in pixels from the input pointer to the hotspot's trigger geometry, clamped to the threshold:

plugins/protocols/wayfire-shell.cpp
ammen99: @marcusbritanicus Sure but to me a code which looks like this:

```
auto point = closest_point(box, target)
return dist(point, target)
```

looks way simpler.

plugins/protocols/wayfire-shell.cpp
ammen99: yes indeed!

ammen99:
@marcusbritanicus The PR looks good from my side aside from the one simplification I'd like to see, otherwise we can merge.

plugins/protocols/wayfire-shell.cpp
marcusbritanicus: @ammen99 I think the current code is what it should be: Consider a corner hot-spot - it's a point. So `trigger` is a point (a rectangle with zero width & height), meaning `wlr_box_closest_point` will return `NaN`. So we will have to handle that case separately. Similarly, it's the case with the edges: `trigger` is a line (a rectangle with zero width/height).

You _can_ argue that I should use 1px width/height as appropriate for the `trigger` in cases like corner or edges. It will no longer be pixel perfect, but it will work. If you want pixel perfect, then using `wlr_box_closest_point` will make the code longer.

ammen99: Alllright, I don't feel that strongly .. Let's leave it as it is.

PR #3023, opened 2026-05-09
Hdr fixes

kode54:
This branch was:

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---

Some of the changes were hand edits or tweaks, such as the gamma 2.2 change to the add_rect color, which required multiple rewrites before I realized what was up with that:

- The target blend operation is alpha blended with pre-multiplied alpha, so the color that is coming in needs to have its R/G/B divided by the alpha before adjusting it, then scaled by the color multiplier in linear space, then converted back to sRGB, then pre-multiplied by the alpha again. I opted to use gamma 2.2 instead of piecewise sRGB for the space conversions.
- There was a place where the agent decided to try setting a NULL preferred image description, which I discovered is not the legal way to unset the preferred image description. So instead, I opted to write it to set a fallback of sRGB / gamma 2.2, in the case where a "best" format is not found for a surface.
- All the rest of the code did what it was supposed to right out of the generator, so it didn't need any major alterations.
- Except for Uncrustify. I had to fix up a few files with that, and the line breaking comment wrap found a bug in the Uncrustify implementation that needs to run a second pass to check for discontinuity with equals sign spacing. I'll report that on ammen99's uncrustify repository.

### DIFF
--- a/plugins/common/workspace-wall.cpp
+++ b/plugins/common/workspace-wall.cpp
@@ -259,9 +259,17 @@ class workspace_wall_t::workspace_wall_node_t : public scene::node_t
 
                 auto bbox = workspaces[i][j]->get_bounding_box();
 
+                // The aux buffer stores a linear-space scene composite (target_tf == EXT_LINEAR).
+                // When the output is HDR (PQ), HDR sources contribute SDR-relative linear values up to
+                // ~49.26 (PQ peak / SDR reference white), which would clip in an 8-bit linear buffer.
+                const auto *img_desc = wall->output->handle->image_description;
+                const bool is_hdr    = img_desc &&
+                    img_desc->transfer_function == WLR_COLOR_TRANSFER_FUNCTION_ST2084_PQ;
+
                 aux_buffers[i][j].allocate(wf::dimensions(bbox), wall->output->handle->scale,
                     wf::buffer_allocation_hints_t{
                         .needs_alpha = false,
+                        .hdr_linear  = is_hdr,
                     });
                 aux_buffer_damage[i][j] |= bbox;
                 aux_buffer_current_scale[i][j]  = 1.0;
--- a/plugins/common/workspace-wall.cpp
+++ b/plugins/common/workspace-wall.cpp
@@ -259,17 +259,10 @@ class workspace_wall_t::workspace_wall_node_t : public scene::node_t
 
                 auto bbox = workspaces[i][j]->get_bounding_box();
 
-                // The aux buffer stores a linear-space scene composite (target_tf == EXT_LINEAR).
-                // When the output is HDR (PQ), HDR sources contribute SDR-relative linear values up to
-                // ~49.26 (PQ peak / SDR reference white), which would clip in an 8-bit linear buffer.
-                const auto *img_desc = wall->output->handle->image_description;
-                const bool is_hdr    = img_desc &&
-                    img_desc->transfer_function == WLR_COLOR_TRANSFER_FUNCTION_ST2084_PQ;
-
                 aux_buffers[i][j].allocate(wf::dimensions(bbox), wall->output->handle->scale,
                     wf::buffer_allocation_hints_t{
                         .needs_alpha = false,
-                        .hdr_linear  = is_hdr,
+                        .hdr_linear  = wall->output && wall->output->is_hdr(),
                     });
                 aux_buffer_damage[i][j] |= bbox;
                 aux_buffer_current_scale[i][j]  = 1.0;
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -98,13 +98,7 @@ class wayfire_cube : public wf::per_output_plugin_instance_t, public wf::pointer
 
                 damage ^= bbox;
 
-                // Per-workspace aux buffers are linear scene composites (target_tf == EXT_LINEAR).
-                // On HDR (PQ) outputs, HDR sources land in the SDR-relative linear domain at values
-                // up to ~49.26 (PQ peak / SDR reference white) and need an FP16 backing to avoid
-                // clipping.
-                const auto *img_desc = self->cube->output->handle->image_description;
-                const bool is_hdr    = img_desc &&
-                    img_desc->transfer_function == WLR_COLOR_TRANSFER_FUNCTION_ST2084_PQ;
+                const bool is_hdr = self->cube->output && self->cube->output->is_hdr();
 
                 for (int i = 0; i < (int)ws_instances.size(); i++)
                 {
--- a/plugins/cube/cube.cpp
+++ b/plugins/cube/cube.cpp
@@ -98,11 +98,20 @@ class wayfire_cube : public wf::per_output_plugin_instance_t, public wf::pointer
 
                 damage ^= bbox;
 
+                // Per-workspace aux buffers are linear scene composites (target_tf == EXT_LINEAR).
+                // On HDR (PQ) outputs, HDR sources land in the SDR-relative linear domain at values
+                // up to ~49.26 (PQ peak / SDR reference white) and need an FP16 backing to avoid
+                // clipping.
+                const auto *img_desc = self->cube->output->handle->image_description;
+                const bool is_hdr    = img_desc &&
+                    img_desc->transfer_function == WLR_COLOR_TRANSFER_FUNCTION_ST2084_PQ;
+
                 for (int i = 0; i < (int)ws_instances.size(); i++)
                 {
                     const float scale = self->cube->output->handle->scale;
                     auto bbox = self->workspaces[i]->get_bounding_box();
-                    framebuffers[i].allocate(wf::dimensions(bbox), scale);
+                    framebuffers[i].allocate(wf::dimensions(bbox), scale,
+                        wf::buffer_allocation_hints_t{.hdr_linear = is_hdr});
 
                     wf::render_target_t target{framebuffers[i]};
                     target.geometry = self->workspaces[i]->get_bounding_box();
--- a/src/api/wayfire/render.hpp
+++ b/src/api/wayfire/render.hpp
@@ -437,11 +437,28 @@ struct render_target_t : public render_buffer_t
 
     /**
      * Set a new color transform for the render target.
+     *
+     * @param transform The wlr_color_transform to apply when rendering to this target. Wayfire
+     *  takes a reference; the caller retains ownership of its own reference.
+     * @param target_tf The transfer function that the @transform encodes to. Used to compute
+     *  per-texture luminance multipliers when SDR content is rendered to HDR targets and vice
+     *  versa.
      */
-    void set_color_transform(wlr_color_transform *transform);
+    void set_color_transform(wlr_color_transform *transform,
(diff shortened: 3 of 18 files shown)


src/output/render-manager.cpp
```diff
@@ -1041,12 +1186,91 @@ class wf::render_manager::impl
...
     void update_bound_output(wlr_buffer *buffer)
     {
         /* Make sure the default buffer has enough size */
         postprocessing->allocate(output->handle->width, output->handle->height);
         postprocessing->set_current_buffer(buffer);
 
+        // Allocate the linear scene intermediate when running on a renderer that supports
+        // FP16 render targets and the inverse-EOTF color transforms HDR needs. On the
+        // GLES2 renderer we keep the original single-pass pipeline (see start_output_pass).
+        if (!wf::get_core().is_gles2())
```
ammen99: This intermediate pass is done by the wlroots vulkan render pass automatically - depending on the color transforms we say we want to have. We don't really need this, provided we have set all the render pass options correctly.

kode54: Eliminated.

src/api/wayfire/render.hpp
```diff
@@ -237,6 +250,13 @@ struct render_buffer_t
 struct buffer_allocation_hints_t
 {
     bool needs_alpha = true;
+    /**
+     * If set, prefer a half-float (FP16) per-channel format so the buffer can store
+     * extended-range linear values without clipping or banding. Used for HDR linear
+     * compositing intermediates. Falls back to a 16-bit fixed format and then to 8-bit
+     * if no FP16 format is available.
+     */
+    bool hdr_linear = false;
```
ammen99, line 270: We don't use this anywhere in the code, so maybe it would be best to remove it.

Alternatively, I don't know what happens if you use a plugin like Expo. Does Expo work well with HDR? It may need to allocate 16-bit intermediary buffers. If this is the case, I think we need to add some helper function which will populate the buffer allocation hints depending on the target output, so that plugins can easily find the correct format.

kode54: I just noticed Expo doesn't render HDR. It clips HDR to SDR range.

kode54: Now it's used in some plugins. And it's open to be used in other plugins as necessary.

kode54:
Okay, this has an outstanding issue with direct scanout of size mismatched XWayland surfaces, possibly also size mismatched Wayland surfaces, I've committed a fix, will push it once I verify it works correctly.

plugins/wobbly/wobbly.cpp
```diff
@@ -1023,7 +1032,7 @@ class wobbly_render_instance_t :
             auto buffer = find_buffer(state.get_context(), total_size);
             buffer->write(unified_buffer.data(), total_size);
 
-            auto texture  = get_texture(data.target.scale);
```
ammen99, line 1026: If I understand the motivation for the change above correctly, then I think this is not needed. Custom vulkan shaders do take color transfer functions in consideration - if what we have right now is not enough, then we need to extend that support, not force more copies.

ammen99 (changes requested):
In general, I am highly doubtful about some of the latest changes. Why skip the zero-copy path if we don't need HDR / special color transfer functions? Maybe the best solution is something like this:

- Extend get_texture() to also include a parameter saying which color spaces the caller expects
- Make sure get_texture() does not go the zero-copy path if the texture does not match the expected color space
- In the GLES2 codepath, expect SRGB or whatever
- In the Vulkan codepath, say we support all transfer functions.

src/api/wayfire/render.hpp
```diff
@@ -24,6 +24,28 @@ namespace vk
 class command_buffer_t;
 }
 
+/**
```
ammen99: Do we need to define this in the header file?

src/view/view-3d.cpp
```diff
@@ -295,6 +295,12 @@ class view_2d_render_instance_t :
         auto source = self->get_updated_contents(self->get_children_bounding_box(),
             data.target.scale, this->children, this->_shown_on);
 
+        // inner_content is EXT_LINEAR (SDR-relative); the bound FBO is PQ-linear on HDR outputs.
+        // Bridge the domains in the default shader since we bypass wlr_render_pass_add_texture.
+        const float multiplier = wf::compute_luminance_multiplier(
+            WLR_COLOR_TRANSFER_FUNCTION_EXT_LINEAR,
```
ammen99: Shouldn't the source function come from the `source` texture instead of hardcoding it?

ammen99: (also the same in other places)

plugins/cube/cube.cpp
```diff
@@ -131,7 +139,47 @@ class wayfire_cube : public wf::per_output_plugin_instance_t, public wf::pointer
 
             void render(const wf::scene::render_instruction_t& data) override
             {
-                self->cube->render(data, framebuffers);
+                // Allocate the cube's intermediate buffer matching data.target's size (which has
+                // already been translated to origin by schedule_instructions). hdr_linear so HDR
+                // contents above SDR reference white aren't clipped by an 8-bit linear backing.
+                const auto *img_desc = self->cube->output->handle->image_description;
```
ammen99: It seems to me we shouldn't be looking at the output directly. Doesn't `data.target` include all the information we need to know about the target we are rendering to? If not, we need to extend the render target class. Then, we simply look: is the target colorspace something we support? If so, we can render directly; otherwise, we need aux buffers.

plugins/common/workspace-wall.cpp
```diff
@@ -259,9 +259,17 @@ class workspace_wall_t::workspace_wall_node_t : public scene::node_t
 
                 auto bbox = workspaces[i][j]->get_bounding_box();
 
+                // The aux buffer stores a linear-space scene composite (target_tf == EXT_LINEAR).
+                // When the output is HDR (PQ), HDR sources contribute SDR-relative linear values up to
+                // ~49.26 (PQ peak / SDR reference white), which would clip in an 8-bit linear buffer.
+                const auto *img_desc = wall->output->handle->image_description;
```
ammen99: We have these checks so often that it would make sense to have a helper function `output->render->is_hdr()`, same for `render_target->is_hdr()`.

src/output/render-manager.cpp
```diff
@@ -809,6 +809,115 @@ class wf::render_manager::impl
...
+     * The output color transform that matches the output's currently-committed image description.
+     * For non-sRGB output primaries, this is a pipeline of [sRGB→output-primaries matrix,
+     * inverse-EOTF]; otherwise it is just the inverse-EOTF. The wlroots Vulkan two-pass renderer
+     * composites every add_texture into an sRGB-primaries FP16 blend image, so without the
+     * primaries-conversion stage Rec.2020/PQ outputs would have sRGB-primaries values encoded
+     * via PQ — which the display interprets as Rec.2020 primaries, producing oversaturated
+     * colors. Cached so that it is not recreated each frame.
+     */
+    wlr_color_transform *output_inverse_eotf = nullptr;
+    wlr_color_transfer_function output_inverse_eotf_tf = (wlr_color_transfer_function)0;
```
ammen99: is 0 a valid color transfer function? if not, seems like a candidate for `std::optional<wlr_color_transfer_function>` instead of invalid values.

src/output/render-manager.cpp
```diff
@@ -1010,17 +1138,26 @@ class wf::render_manager::impl
         render_pass_params_t params;
         params.instances = &damage_manager->instance_manager->get_instances();
 
+        // Render directly to the output target. wlroots' Vulkan renderer already runs a
```
ammen99: I am not sure what the purpose of this comment is.

src/render.cpp
```diff
@@ -4,14 +4,88 @@
...
+        // SDR source → HDR target: scale [0,1] relative down so 1.0 maps to the SDR
+        // reference white in the PQ-relative range.
+        return wf::SDR_REFERENCE_WHITE_NITS / wf::PQ_MAX_NITS;
+    }
+
+    // HDR source → SDR target: scale up so the SDR reference luminance maps to 1.0.
+    return wf::PQ_MAX_NITS / wf::SDR_REFERENCE_WHITE_NITS;
+}
+
+static float float_max(float a, float b)
```
ammen99: This one looks really redundant in c++, where we have std::max.

ammen99 (changes requested):

kode54:
> In general, I am highly doubtful about some of the latest changes. Why skip the zero-copy path if we don't need HDR / special color transfer functions? Maybe the best solution is something like this:
> 
> * Extend get_texture() to also include a parameter saying which color spaces the caller expects
> * Make sure get_texture() does not go the zero-copy path if the texture does not match the expected color space
> * In the GLES2 codepath, expect SRGB or whatever

The altered wlroots code base expects linear in the GLES2 code path as well. And maybe it's worth just getting rid of this branch instead, since I doubt that will ever be merged into wlroots, considering it breaks on Nvidia.

> * In the Vulkan codepath, say we support all transfer functions.

kode54:
It's quite possible I should just discard everything after https://github.com/WayfireWM/wayfire/pull/3023/commits/2443c5f75c233e290dab870f07f18286464636f0 ... since that's mostly adapting everything to the GLES2 rewrite that will likely never be approved or employed anywhere.

kode54:
Could you please review again, I'm not sure how much of the above comments are regarding code I decided to move elsewhere for future dealings.

src/api/wayfire/render.hpp
```diff
@@ -19,6 +19,19 @@ namespace wf
 {
 class output_t;
 struct auxilliary_buffer_t;
+
```
ammen99, line 22: Since this is an output-related function, wouldn't it make sense to put it in output.hpp?

src/output/render-manager.cpp
ammen99: Still remains a point for improvement

src/view/wlr-surface-node.cpp
```diff
@@ -385,6 +406,17 @@ class wf::scene::wlr_surface_node_t::wlr_surface_render_instance_t : public rend
             return direct_scanout::OCCLUSION;
         }
 
+        // Direct scanout commits the surface buffer straight to the primary plane. With
```
ammen99: I don't think it is a problem to have a scale != 1, the issue is if we don't match the resolution of the output.

ammen99 (commented):
there are still a few small comments, but overall lgtm. If you fix them and are generally happy with how the PR works on your monitors, we can merge.

src/view/wlr-surface-node.cpp
kode54: I'll do a resolution check instead. I don't know why my analyzer decided it was entirely about scale, and I stupidly trusted it.

ammen99:
@kode54 I added some more commits to fix the remaining issues, please take a look at my latest changes, test and if it still works ok, I will merge.

ammen99 (approved):

kode54:
Thanks, so much, for looking at this, and for the changes you made to fix my mistake, and also pointing out the really smart use of smart pointers there.

I'll take a look in a few hours, when my evening winds down, and I'll let you know how it worked out.

PR #3032, opened 2026-05-23
API: move logical geometry to floating point

ammen99:
This is a huge breaking change, aiming to make it possible to get pixel-perfect layouts on outputs with fractional scaling. The entire codebase (geometries and damage coordinates) have now been moved to floating point numbers. This PR has been largely generated with the help of LLMs and still needs to be reviewed very carefully, thus the draft state.

It is a breaking change for plugins, which will most likely have to adapt, however, aside from the int -> double conversion, very little has changed semantically, so updating plugins should be easy.


mark-herbert42:
I remember there was a bug in firefox when different rounding in this floating poing geometry was causing a lot of crashes when size of the frame did not match the size of the video being decoded into that frame. The difference was really funny like 0.0050 something but that was enough to crash wayland.  So it might be a challenge to implement...

PR #3037, opened 2026-06-03
seat: select drag-and-drop action from keyboard modifiers

coffeejunk:
The `drag_handle_keyboard_modifiers` implementation in [wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/raw/master/types/data_device/wlr_drag.c) is empty, so `ctrl`/`shift` have no effect on a drag. 

```c
static void drag_handle_keyboard_modifiers(struct wlr_seat_keyboard_grab *grab,
		const struct wlr_keyboard_modifiers *modifiers) {
	//struct wlr_keyboard *keyboard = grab->seat->keyboard_state.keyboard;
	// TODO change the dnd action based on what modifier is pressed on the
	// keyboard
}
```

Separately, it was stated that this should be implemented by the compositor https://github.com/swaywm/sway/issues/8219#issuecomment-2395368597

Map the modifiers to `wlr_data_source::compositor_action` on each modifier's event while a drag is active: `ctrl` copies, `shift` moves, `ctrl+shift` asks.

As an example, Thunar:

Without the patch, dragging a file between two windows while pressing `CTRL` moves it:

https://github.com/user-attachments/assets/1b0cc409-5feb-45ff-bfe9-28a77e567710

With the patch applied, dragging a file between two windows while pressing `CTRL` copies it:

https://github.com/user-attachments/assets/4eb826b7-56c1-43e1-99e5-4ac26968499e

### DIFF
--- a/src/core/seat/seat-impl.hpp
+++ b/src/core/seat/seat-impl.hpp
@@ -91,8 +91,15 @@ struct seat_t::impl
 
     // Current drag icon
     std::unique_ptr<wf::drag_icon_t> drag_icon;
+    // The active drag, or NULL if no drag is in progress. Note we can have a
+    // drag without a drag icon, and active_drag->source may be NULL.
+    wlr_drag *active_drag = nullptr;
+
     // Is dragging active. Note we can have a drag without a drag icon.
-    bool drag_active = false;
+    bool is_drag_active() const
+    {
+        return active_drag != nullptr;
+    }
 
     /** Update the position of the drag icon, if it exists */
     void update_drag_icon();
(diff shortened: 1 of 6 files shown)


src/core/seat/seat-impl.hpp
```diff
@@ -93,6 +93,8 @@ struct seat_t::impl
     std::unique_ptr<wf::drag_icon_t> drag_icon;
     // Is dragging active. Note we can have a drag without a drag icon.
     bool drag_active = false;
+    // The active drag, valid while drag_active. Its source may be NULL.
```
ammen99: If we store the `active_drag` object, then we don't need `drag_active` at all. The latter should be removed, and all the places that use it can simply compare the active drag against a null pointer (or, if you prefer, you could add a helper function `is_drag_active()` which does the check). Otherwise lgtm.

ammen99 (approved):
LGTM, thanks!

PR #3040, opened 2026-06-08
Add support for vertical/horizontal maximization

rpaciorek:
vertical/horizontal maximization by middle/right click on maximize button (like in Openbox)

### DIFF
--- a/plugins/decor/deco-layout.cpp
+++ b/plugins/decor/deco-layout.cpp
@@ -260,7 +260,7 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
             }
         }
 
-        if (area && (area->get_type() & DECORATION_AREA_RESIZE_BIT) && button == BTN_LEFT)
+        if (area && (area->get_type() & DECORATION_AREA_RESIZE_BIT) && (button == BTN_LEFT))
         {
             return {DECORATION_ACTION_RESIZE, calculate_resize_edges()};
         }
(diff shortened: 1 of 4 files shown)


plugins/decor/deco-layout.cpp
```diff
@@ -260,24 +260,25 @@ decoration_layout_t::action_response_t decoration_layout_t::handle_press_event(
...
-        if (area && (area->get_type() & DECORATION_AREA_RESIZE_BIT))
+        if (area && (area->get_type() & DECORATION_AREA_RESIZE_BIT) && button == BTN_LEFT)
         {
             return {DECORATION_ACTION_RESIZE, calculate_resize_edges()};
         }
 
-        if (area && (area->get_type() == DECORATION_AREA_BUTTON))
+        if (area && (area->get_type() == DECORATION_AREA_BUTTON) && button == BTN_LEFT)
         {
             area->as_button().set_pressed(true);
```
ammen99, line 270: wouldn't we want to set the pressed state no matter which button is being pressed?

rpaciorek: Yes, we want. Fixed.

ammen99 (approved):
LGTM, I believe it will need a few style fixes (look at the CI output, it will include a diff to apply)

rpaciorek:
I applied style fixes, based on the output from CI - my local `uncrustify` (0.78.1_f) crashes semi-quiet on previous files, and running only on files covered by PR gave a different result.

ammen99:
> I applied style fixes, based on the output from CI - my local `uncrustify` (0.78.1_f) crashes semi-quiet on previous files, and running only on files covered by PR gave a different result.

Yeah, that's the reason we use a custom uncrustify in the CI ...

ammen99:
Thanks!

PR #3042, opened 2026-06-09
allow to customize button and title size

rpaciorek:
* repleced hardcoded scale for window buttons by scale set in options
* repleced hardcoded scale for window title text by scale set in options

### DIFF
--- a/metadata/decoration.xml
+++ b/metadata/decoration.xml
@@ -14,6 +14,13 @@
 			<_long>Sets the font color to use for the title bars.</_long>
 			<default>#ffffffff</default>
 		</option>
+		<option name="font_scale" type="double">
+			<_short>Font scale</_short>
+			<_long>Sets the font scale to use for the title bars.</_long>
+			<default>0.8</default>
+			<max>1.0</max>
+			<min>0.1</min>
+		</option>
 		<option name="title_height" type="int">
 			<_short>Title bar height</_short>
 			<_long>Sets the height of the title bars in pixels.</_long>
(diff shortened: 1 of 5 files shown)


marcusbritanicus:
@rpaciorek ammen is most likely to reject this PR. The default decoration plugin is minimal by design. If you want a more advanced one you should look at [pixdecor](https://github.com/soreau/pixdecor)

rpaciorek:
@marcusbritanicus Thanks for the link to pixdecor.

I wouldn't describe this PR as something that break minimalism (it's just extracting hard-coded values ​​into the config), but of course, I'll leave it to ammen to decide.

ammen99:
@rpaciorek If I remember correctly, the font option can also be used to set a font size, isn't that the case?

rpaciorek:
font size is overwrite on: https://github.com/WayfireWM/wayfire/blob/48f2ccdf2600818aa291ab15d3c37051b15efe1c/plugins/decor/deco-theme.cpp#L73 to scale the title text relative to the size of the header bar, via `font_size` set on: https://github.com/WayfireWM/wayfire/blob/48f2ccdf2600818aa291ab15d3c37051b15efe1c/plugins/decor/deco-theme.cpp#L65-L66 so the size cannot be set in the `font` option

ammen99 (approved):
Looks good, thanks! I indeed want to keep the plugin minimalistic, but as you said, these things are already present, only hardcoded, so adding an option doesn't seem like it will hurt.

ammen99:
Thanks!

PR #3043, opened 2026-06-11
ext-toplevel: Assign toplevel data to the corresponding view

soreau:


metadata/ext-toplevel.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" ?>
 <wayfire>
 	<plugin name="ext-toplevel">
-		<_short>Ext Foreign Toplevel List Protocol</_short>
+		<_short>Foreign Toplevel List Protocol</_short>
```
ammen99: I think we should mention this is the ext- version, not the wlr- version.

src/api/wayfire/core.hpp
```diff
@@ -167,10 +167,7 @@ class compositor_core_t : public wf::object_base_t, public signal::provider_t
         wlr_data_control_manager_v1 *data_control;
         wlr_gamma_control_manager_v1 *gamma_v1;
         wlr_screencopy_manager_v1 *screencopy;
-        wlr_ext_foreign_toplevel_list_v1 *foreign_toplevel_list;
-        wlr_ext_image_copy_capture_manager_v1 *image_copy_capture;
```
ammen99, line 171: We also advertise the output capture source global (the field below), which makes sense - so we can keep advertising image copy capture in core as well. Just the toplevel image capture should be left for a plugin.

ammen99 (changes requested):
a few small things

ammen99 (approved):
LGTM, thanks!

PR #3044, opened 2026-06-13
foreign-toplevel: support capturing windows.

lcolitti:
This commit adds support for capturing windows using foreign_toplevel_image_capture_source_manager.

Tested using Chrome webrtc screensharing via xdg-desktop-portal.

### DIFF
--- a/plugins/protocols/ext-toplevel.cpp
+++ b/plugins/protocols/ext-toplevel.cpp
@@ -36,6 +36,7 @@ class wayfire_ext_foreign_toplevel
 {
     wayfire_toplevel_view view;
     wlr_ext_foreign_toplevel_handle_v1 *handle;
+    wlr_scene *scene;
 
   public:
     wayfire_ext_foreign_toplevel(wayfire_toplevel_view view, wlr_ext_foreign_toplevel_handle_v1 *hndl) :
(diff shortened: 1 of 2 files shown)


soreau:
This is a NAK from me. Wayfire doesn't use wlr_scene and the wlr_scene implementation for foreign-toplevels is subpar compared to the wf-copy-capture plugin. For example when I tried it, it did not support resizing toplevels or even mouse cursor support. It was also very buggy for renderings in general. Currently to capture toplevels, enable the ext-toplevel and wf-copy-capture plugins and run a toplevel copy capture client.

soreau:
Other foreign-toplevel copy capture test clients you can try are:

- [wf-recorder](https://github.com/ammen99/wf-recorder/pull/347)
- [wf-stream-chooser](https://github.com/WayfireWM/wf-shell/tree/master/src/stream-chooser)
- [wf-panel](https://github.com/WayfireWM/wf-shell/pull/483)

ammen99:
@lcolitti Thanks for the PR. As @soreau already explained, he has gotten a version working with a custom implementation which supports many features. Your current implementation is close to what I'd have done myself (using wlr_scene) but the issue is that Wayfire supports custom subsurfaces like the decoration plugins, which wlroots does not know about and thus wouldn't be included with your current approach. If we wanted to use wlr_scene, I'd personally render to an aux buffer and then attach that buffer as a wlr scene node.

lcolitti:
Thanks for the quick response. I didn't know about wf-copy-capture. I'll just tried it. I was able to get it to share Chrome windows for a while but then wayfire segfaulted (without a stacktrace, otherwise I'd open an issue). I'll poke at it some more.

soreau:
@lcolitti If you do file an issue, please make it on the wf-copy-capture repo. I am curious to know what steps you took to crash it.

lcolitti:
Filed https://github.com/soreau/wf-copy-capture/issues/2 .

PR #3045, opened 2026-06-13
wraparound by desktop numeric id mode in vswitch

rpaciorek:


ammen99 (approved):
Thanks! Feel free to edit the wiki to add the new option values and an explanation of how it works.

PR #3046, opened 2026-06-16
seat: update last_cursor_name when client sets cursor

pakhromov:
Fixes the bug when cursor shape is not changed to default when opening expo. If a client sets a cursor shape (for example to "xterm" or "hand2"), last_cursor_name remains "left_ptr" from the previous compositor call. When expo then calls set_cursor("default"), the early-return fires ("left_ptr" == last_cursor_name) and the cursor is not updated.

https://github.com/user-attachments/assets/dd612c52-a3b1-40b2-9f74-34b4d182c788


ammen99 (approved):
Nice catch, thanks!

PR #3049, opened 2026-06-17
window-rules: make assign_workspace work with simple-tile

pakhromov:
simple-tile overrides the geometry of the view so offsetting it with _view->move has no effect. Instead use wset->move_to_workspace() for the view and its children, then emit view_change_workspace_signal so it works with the tile plugin.
Using the same logic as in move_view function of vswitch plugin.


ammen99 (approved):
LGTM, thanks!

PR #3054, opened 2026-06-20
wayfire-shell: Allow clients to use < version 3 without crashing

soreau:


ammen99 (commented):
Nice catch! Two things:

- I see no reason to pass and store the full resource, it would be enough to have the uint32_t client_version
- All checks should be in the `min(client_version, 3)` format, hardcoding between 1 and 3 simply means we'd have to change all of this again if we ever upgrade the protocol in the future. Even better: define the current version as a constant in the beginning of the file (`static constexpr uint32_t WAYFIRE_SHELL_VERSION = 3;` and then use that everywhere :)

soreau:
> Nice catch! Two things:
> 
> * I see no reason to pass and store the full resource, it would be enough to have the uint32_t client_version

Yes I considered this but decided to follow surrounding convention, which is less churn.

> * All checks should be in the `min(client_version, 3)` format, hardcoding between 1 and 3 simply means we'd have to change all of this again if we ever upgrade the protocol in the future. Even better: define the current version as a constant in the beginning of the file (`static constexpr uint32_t WAYFIRE_SHELL_VERSION = 3;` and then use that everywhere :)

I believe this information is already in the protocol headers.

ammen99 (approved):
LGTM, thanks!

PR #3060, opened 2026-07-06
render-manager: prevent infinite commit loop in direct scanout

### LINKED ISSUE #3033: Higher CPU usage when there are no layer-shell programs
## Wayfire version
0.11.0-4ffa601d (May 11 2026) branch master wlroots-0.20.0

## GPU / Driver
NVIDIA proprietary (595.71.05-2)

## Describe the bug
CPU usage of wayfire becomes higher when there are no layer-shell applications running. If you are not running any bar/panel and do not have any other widget type applications opened - the CPU usage of wayfire process gets higher.

## To Reproduce
Open a system monitoring tool and see the idle resource usage of wayfire process (not overall system cpu usage), then close all layer-shell applications and see the idle usage again - it becomes higher.

## Expected behavior
Wayfire should not have higher CPU usage when there are no layer-shell programs opened.

I uploaded video on YouTube (was higher than 10mb github limit) where I have btop opened and focused on wayfire process. Look at the CPU usage when I start and close different layer-shell programs. When there is such a program running - the idle CPU usage of wayfire itself is 0.0%, when I close such application - the usage becomes 0.5% and never drops below. 
https://youtu.be/yORlY8yecas

pakhromov:
Whenever a frame is committed, the next repaint is scheduled (the commit triggers the next frame event on the output, which calls paint() again).

The composited path prevents this cycle on idle, since start_frame() returns null when there is no damage and no frame request.

The direct scanout commits the client's buffer even when the client has not submitted anything new. This leads to an infinite commit loop when there is no need for it.

Fix this by commiting only when the content has changed since the last presented frame. should_repaint() mirrors the conditions start_frame() uses for its check.

Fixes #3033

### DIFF
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -242,8 +242,12 @@ struct swapchain_damage_manager_t
 
     bool force_next_frame = false;
 
-    // Set when new content was damaged, cleared when a frame (composited or direct scanout) is presented.
     // Tracks whether a new frame is needed at all.
+    // Set when new content was damaged, cleared when a frame (composited or direct scanout) is presented.
+    // We cannot use the empty damage ring for this check, because the ring is only cleared for 'regular'
+    // composited frames. Direct scanout does not clear the ring: the compositor's own render buffers are
+    // not updated during scanout, so the accumulated damage must be kept to prevent corrupted frames
+    // when transitioning from scanout to compositing.
     bool pending_frame_request = false;
 
     /**
@@ -1200,7 +1204,7 @@ class wf::render_manager::impl
         effects->run_effects(OUTPUT_EFFECT_PRE);
         effects->run_effects(OUTPUT_EFFECT_DAMAGE);
 
-        // Optimization: the output doesn't need a new frame (so isn't damaged), so we can
+        // Optimization: the output doesn't need a new frame, so we can
         // just skip the whole repaint
         if (!damage_manager->should_repaint())
         {
--- a/src/output/render-manager.cpp
+++ b/src/output/render-manager.cpp
@@ -256,9 +256,9 @@ struct swapchain_damage_manager_t
     }
 
     /**
-     * Notify the damage manager that a frame was presented via direct scanout.
+     * Notify the damage manager that a new frame is about to be presented.
      */
-    void handle_scanout_frame()
+    void frame_started()
     {
         force_next_frame = false;
         pending_frame_request = false;
@@ -266,23 +266,15 @@ struct swapchain_damage_manager_t
 
     /**
      * Start rendering a new frame.
-     * If the operation could not be started, or if a new frame is not needed, the function returns false.
-     * If the operation succeeds, true is returned, and the output (E)GL context is bound.
+     * The caller is responsible for checking that a new frame is needed (should_repaint()).
+     * If the operation could not be started, the function returns null.
+     * If the operation succeeds, a frame object is returned, and the output (E)GL context is bound.
      */
     std::unique_ptr<frame_object_t> start_frame()
     {
         auto buffer_extents = this->get_buffer_extents();
         pixman_region32_intersect_rect(&damage_ring.current, &damage_ring.current,
             buffer_extents.x, buffer_extents.y, buffer_extents.width, buffer_extents.height);
-        const bool needs_swap = force_next_frame | output->needs_frame |
-            pixman_region32_not_empty(&damage_ring.current) | (constant_redraw_counter > 0);
-        force_next_frame = false;
-        pending_frame_request = false;
-
-        if (!needs_swap)
-        {
-            return {};
-        }
 
         auto next_frame = std::make_unique<frame_object_t>();
         next_frame->state.committed |= WLR_OUTPUT_STATE_DAMAGE;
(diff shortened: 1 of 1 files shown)


pakhromov:
To add more context to #3033, I suspect that hardware cursor is not working at all on NVIDIA, at the very least it is not working on my machine. That means whenever the cursor is visible - direct scanout cannot be triggered. When the cursor is hidden and only one toplevel occupies the whole screen - the direct scanout starts, in my test case kitty was hiding the cursor and starting it. The application that I was running inside kitty (btop) was updating only once every second, however wayfire was still committing kitty's buffer on every refresh of my monitor since there was nothing preventing it from doing so. Btop then presented the average CPU usage for the last second and it was noticeably higher than when not in a direct scanout.

The worst-case scenario is static content displayed in fullscreen, like a paused video, or game menu, or just a shell - when I am looking at zsh in kitty and doing nothing, wayfire is doing 144 commits a second (144 Hz monitor), when it should not be doing anything. On systems with working hardware cursor it is even worse, since wlroots allows direct scanout even when the cursor is shown (looked into wlroots implementation), assuming nothing else is shown on the screen like a bar or notification.

The thing that puzzles me is how can VRR work with this. I do not have a VRR capable monitor so cannot really tell, but I suspect that it was not working correctly without this patch when in a direct scanout. VRR works by timing the monitor refresh to when the page flip arrives, and every committed buffer by wayfire leads to a page flip. If direct scanout always commits at the monitor's maximum refresh - from the VRR perspective, all the page flips are valid and it should be stuck on the max refresh forever assuming this is what is actually produced by the client, even though the client may produce frames at a much lower rate.

ammen99 (changes requested):
I see what the issue is, but wouldn't it be enough to move the scanout check after the existing `skip_frame` call? It feels like this PR duplicates the check instead of fixing the order.

pakhromov:
That would not trigger direct scanout at all, `skip_frame` is called if `start_frame` returned nothing, but it would run fine even if direct scanout can be used. What I am trying to say is that check for scanout needs to be before calling `start_frame`, otherwise `start_frame` will do the commit itself.

But we can prevent the duplication by just removing the check for needing a redraw inside start_frame. That way the flow would be: paint() called => check if the repaint is needed at all first => try direct scanout => if direct scanout failed, call start_frame but do not check for needing a repaint inside it since it was already checked.

ammen99:
> That would not trigger direct scanout at all, `skip_frame` is called if `start_frame` returned nothing, but it would run fine even if direct scanout can be used. What I am trying to say is that check for scanout needs to be before calling `start_frame`, otherwise `start_frame` will do the commit itself.
> 
> But we can prevent the duplication by just removing the check for needing a redraw inside start_frame. That way the flow would be: paint() called => check if the repaint is needed at all first => try direct scanout => if direct scanout failed, call start_frame but do not check for needing a repaint inside it since it was already checked.

That was my thought as well, even if maybe I didn't explain it properly. The check for "is repaint needed at first" is also already existing, but happens only after the direct scanout check. So maybe we don't need to add any new flags and functions, just swap the places of the existing if blocks and see whether it produces the desired result.

pakhromov:
> That was my thought as well, even if maybe I didn't explain it properly. The check for "is repaint needed at first" is also already existing, but happens only after the direct scanout check. So maybe we don't need to add any new flags and functions, just swap the places of the existing if blocks and see whether it produces the desired result.

That would be great, unfortunately it's not so easy, the main problem is with damage ring, direct scanout cannot clear it. If damage ring is cleared on direct scanout - when you exit from scanout -  whatever was in the swapchain before you entered it starts being displayed and only the newly damaged regions are displayed correctly (see recording).

The existing check in `start_frame` works when the damage ring is cleared which cannot be used for scanout, therefore scanout must consume something else.

https://github.com/user-attachments/assets/3475e47f-14a0-4a0a-8e64-c6b5605e28bd

src/output/render-manager.cpp
```diff
@@ -239,24 +241,40 @@ struct swapchain_damage_manager_t
     }
 
     bool force_next_frame = false;
+
+    // Set when new content was damaged, cleared when a frame (composited or direct scanout) is presented.
+    // Tracks whether a new frame is needed at all.
+    bool pending_frame_request = false;
```
ammen99, line 251: Alright, I think I understand the proposed PR better now, thanks for your explanations. I still wonder whether we need this `pending_frame_request` flag though, isn't it the same as checking whether the damage ring is currently empty? like this: `pixman_region32_not_empty(&damage_ring.current)` ?

pakhromov: Here is the picture as I see it:
Composited path: damage arrives -> damage ring grows -> frame presented -> ring is cleared.
Direct scanout path: damage arrives -> damage ring grows -> scanout commit happens -> nothing clears damage ring. If we check pixman_region32_not_empty(&damage_ring.current) instead of this flag flag, it would stay true forever after the very first damage event, which is exactly the infinite-loop this PR tries to fix.

The ring isn't cleared on scanout because the compositor's own render buffers are stale. If scanout ever stops (for example, a notification pops up) the first composited frame needs to redraw everything that was changed while scanout was working, and that is exactly the accumulated region in damage ring. Clearing it would corrupt the next composited frame after leaving scanout.

ammen99: I see, this is a tricky issue. With all of this in mind, I think your PR makes sense. My only request would be to document these cases as comment so that we don't forget: for example, explaining that the damage ring is only used for 'regular' frames, and the repaint flag is used for both normal render and scanout. Otherwise LGTM.

ammen99 (approved):
Thanks for the patient explanations, I understand the problem and your proposed solution now. With a bit of documentation we can merge!

ammen99 (approved):
Thanks!

PR #3064, opened 2026-07-16
workarounds: add use_native_buffer_size option for fractional scaling

### LINKED ISSUE #2697: Everything is blurrier with fractional scaling (Wayland-native, not Xwayland)
**Describe the bug**
When using fractional scaling, all fonts and images that have to be upscaled or downscaled are blurrier than what they would be when using 1x or 2x scaling (or using x11). This happens everywhere, from my panel to browser to code editor, etc.

I require 1.75x scaling because of the very high DPI of my screen but then the blurryness caused by it, while not incapacitating, makes it harder to read as if i had slight vision loss.

**Steps To Reproduce**
1. Set scaling to 1x or 2x and view the sharp fonts and images
2. Set scaling to a fractional value like 1.75 or 1.99 and notice the difference

**Expected behavior**
Fonts and images should remain as sharp as they would on integer scaling

**Screenshots or stacktrace**
#### 2x scaling:
Image
#### 1.99x scaling:
Image
#### 2x scaling:
Image
#### 1.99x scaling:
Image

While I used 1.99x to examplify, this problem happens on any scaling that isnt an integer value.

**Wayfire version**
0.9.0

ammen99:
This option can be used to force rendering of client textures to the actual buffer size they provide, leaving up to 1 pixel off the boundary of the surface in order to ensure crisp rendering.

Fixes #2697

**To use, enable the config option** like this:

```ini
[workarounds]
use_native_buffer_size = true
```

### DIFF
--- a/metadata/workarounds.xml
+++ b/metadata/workarounds.xml
@@ -110,5 +110,10 @@
 			<_long>Disables the default blurry scaling for Xwayland and forces client-side scaling for Xwayland windows dependent on the DPI settings.</_long>
 			<default>false</default>
 		</option>
+		<option name="use_native_buffer_size" type="bool">
+			<_short>Use the native buffer size for client surfaces</_short>
+			<_long>When this option is enabled, Wayfire will use the buffer size in pixels to determine a surface size, instead of relying on the client-provided logical size. This ensures that text remains crisp at fractional scales, but may result in cropping or empty pixels at surface boundaries.</_long>
+			<default>false</default>
+		</option>
 	</plugin>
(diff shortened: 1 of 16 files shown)


ammen99:
Note that this works only for fractional-scale clients, for regular clients using the standard wayland scaling mechanism (i.e rendering at 2x resolution and then downscaling) this won't work at all, because there isn't anything Wayfire can do in this case (i.e if we have a scale of 1.5 but the client renders at scale 2.0, we can't align the buffer to the pixels no matter what we do).

PR #3086, opened 2026-07-27
Unmapped view requests

### LINKED ISSUE #3078: on created ... set geometry_ppt resizes the view instead of setting the initial size
## Wayfire version
latest git

## GPU / Driver
Nvidia 610.43.03

## Describe the bug
Window rule `on created ... set geometry_ppt` does not create a new view with the size that I specify, instead it creates the view with its default size (as the application itself wants) and then quickly resizes it to the specified size. This creates a problem if the default size of the client is bigger than what I want it to be, especially noticeable when using the zoom animation - instead of gradually zooming from a smaller size, the window appears bigger initially and then shrinks instantly.

https://github.com/user-attachments/assets/50afb68d-cf05-45d3-a217-2d790bf7b2be

I looked at the existing signals and found `view_pre_map_signal`, but as I understand it is emitted already after the initial commit from the client which does not solve the problem. Will it make sense to add a new signal just for this? Something like `view_pre_configure_signal` which would be emitted before `configure_surface_with_state` in `handle_surface_commit()` (xdg-toplevel.cpp)

pakhromov:
Fixes #3078
Also makes improves snap rules in the same way, and `on created ... then minimize` now does not show anything at all, the new window just appears on the panel.

### DIFF
--- a/plugins/grid/wayfire/plugins/crossfade.hpp
+++ b/plugins/grid/wayfire/plugins/crossfade.hpp
@@ -266,7 +266,13 @@ class grid_animation_t : public wf::custom_data_t
         };
 
         // A not mapped view has no geometry on to animate from, just set the desired state in that case.
-        if ((type != CROSSFADE) || !view->is_mapped())
+        if (!view->is_mapped())
+        {
+            set_state();
+            return destroy();
+        }
+
+        if (type != CROSSFADE)
         {
             /* Order is important here: first we set the view geometry, and
              * after that we set the snap request. Otherwise the wobbly plugin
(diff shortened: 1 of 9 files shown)


plugins/grid/wayfire/plugins/crossfade.hpp
```diff
@@ -265,7 +265,8 @@ class grid_animation_t : public wf::custom_data_t
             tx->add_object(view->toplevel());
         };
 
-        if (type != CROSSFADE)
+        // A not mapped view has no geometry on to animate from, just set the desired state in that case.
+        if ((type != CROSSFADE) || !view->is_mapped())
```
ammen99: note this will trigger wobbly as well, maybe we need a separate if for !view->mapped

ammen99 (approved):
LGTM, thanks!

PR #3088, opened 2026-07-28
fix 1390

### LINKED ISSUE #1390: Switching to last workspace only works when last switch is done by vswitch
**Describe the bug**

#1371 doesn't fully implement #1296.

**To Reproduce**
1. Enable plugin vswitch and expo
2. Switch workspace with expo
3. Try switch back with vswitch
4. vswitch tries to reverse its last action instead, which fails or is wrong because the last switch is done by other plugins

**Expected behavior**
Workspace should be switched back.

**Screenshots or stacktrace**

**Wayfire version**
git

ammen99:
Fixes #1390

### DIFF
--- a/metadata/resize.xml
+++ b/metadata/resize.xml
@@ -16,14 +16,22 @@
 			<default>0</default>
 			<min>0</min>
 		</option>
-		
+
 		<option name="min_height" type="int">
 			<_short>Minimum view height</_short>
 			<_long>If set, views cannot be resized smaller than this height.</_long>
 			<default>0</default>
 			<min>0</min>
 		</option>
-        
+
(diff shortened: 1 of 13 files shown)


lilydjwg:
I find a case in which it still doesn't work:

Start expo but don't switch to another workspace and exit it, then `binding_last` will not switch at all.

ammen99:
> I find a case in which it still doesn't work:
> 
> Start expo but don't switch to another workspace and exit it, then `binding_last` will not switch at all.

I imagine expo emits workspace-changed events in this case (or calls set_workspace), which shouldn't emit a signal when there is no workspace change.

PR #3095, opened 2026-07-31
output-layout: add support for mirroring resolution

### LINKED ISSUE #2348: When external monitor is set to mirror laprop screen it get hard-swched to low 1920x1200 resolution
Wayfire git version,   Mesa git, 6.9-rc7.  

I use GPD win max 2 laptop with 2560x1600 screen, and connect  Zeuslap 16" monitor with exactly same 2560x1600 resolution.

It connects fine - and I can choose the resolution  I want up to native 2560 and 144Hz.  But - until I do 

 
[output:DP-1]
mode =  mirror eDP-1

When I do so - external monitor falls back to 1920x1200 and I can not do anything with it so mirror is totally unusable. But I do have both screens capable to have 2560 resolution natively when they run independently. And my  internal e-DP screen continues running native resolution - so only mirrored output is affected by this.   
If I try to set native resolution on external monitor - it stops mirroring and just works as separate output.

ammen99:
Resubmit of #1716

Original work by @TRPB

Fixes #2348

@mark-herbert42: Could you please test this PR? I know you wanted this feature in the past.

### DIFF
--- a/src/core/output-layout.cpp
+++ b/src/core/output-layout.cpp
@@ -429,7 +429,10 @@ bool output_state_t::operator ==(const output_state_t& other) const
     if (source == OUTPUT_IMAGE_SOURCE_MIRROR)
     {
         return other.source == OUTPUT_IMAGE_SOURCE_MIRROR &&
-               mirror_from == other.mirror_from;
+               mirror_from == other.mirror_from &&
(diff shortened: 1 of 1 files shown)


mark-herbert42:
so far  works good

Issue #1621, opened 2022-11-19
Flickering at 4k 144hz


TRPB:
**Describe the bug**

When running a monitor at 4k 144hz I get very regular flickering.

Video: https://youtu.be/mR3Ku8zlq1o 

This doesn't happen at 60hz.

I also tried Sway, it doesn't happen there so I don't believe it's a wlroots or wayland issue, but specific to Wayfire.

I've tried disabling/enabling adaptive sync (wayfire itself doesn't appear to have an option for this but enabling it the monitor config seems to work) 

**Wayfire version**

0.7.4-0c3621c

It looks like Sway had a similar issue in the past: https://github.com/swaywm/sway/issues/5076 I wonder if Wayfire needs to implement the same fix?

TRPB:
As noted in https://github.com/WayfireWM/wayfire/issues/1397 this is fixed with `WLR_DRM_NO_MODIFIERS=1` I'm not entirely sure what that does but it definitely makes the desktop environment feel less snappy. Less annoying than the flicker though.

soreau:
Does it help if you try fiddling with [this option](https://github.com/WayfireWM/wayfire/wiki/Configuration#coremax_render_time)?

TRPB:
interesting. Thanks! I assume that's in milliseconds? Setting it to 1 fixes it, setting it to 7 (roughly 1000/144), 1000 or -1 the flicker still exists. So it appears that (again assuming it's in milliseconds) a `max_render_time` lower than the time it takes the display to draw one frame fixes it.

So I'd expect 6 to fix it, it makes it a lot less frequent but still get the occasional flicker.

Does this imply it's dropping some frames then? It doesn't seem to cause a noticeable stutter, even on 0 or 1.

From the description I'd expected that lower values would cause the problem rather than resolve it. I'm curious why `0` doesn't break everything entirely or does Wayfire sensibly just use 1 if it's set to 0?

ammen99:
> interesting. Thanks! I assume that's in milliseconds? Setting it to 1 fixes it, setting it to 7 (roughly 1000/144), 1000 or -1 the flicker still exists. So it appears that (again assuming it's in milliseconds) a `max_render_time` lower than the time it takes the display to draw one frame fixes it.
> 
> So I'd expect 6 to fix it, it makes it a lot less frequent but still get the occasional flicker.
> 
> Does this imply it's dropping some frames then? It doesn't seem to cause a noticeable stutter, even on 0 or 1.
> 
[...]

PR #3096, opened 2026-08-01
adaptive repaint delay

### LINKED ISSUE #2377: RetroArch vsync rate is halved when using the Vulkan backend and Max Swapchain Images is set to 2
**Describe the bug**
In RetroArch (latest stable and also GIT version), using the `Vulkan` backend with `Settings->Video->Synchronization->Max Swapchain Images` set to `2`, the framerate is halved, so in a 60Hz display everything will run at 30FPS.
It can be observed without loading any cores, in `Settings->Video->Output->Estimated Screen Refresh Rate`.

Setting `Settings->Video->Synchronization->Max Swapchain Images` to `3` works around the problem, but introduces a frame of lag.

All sensible video/audio settings have been tried in RetroArch to try and isolate the cause.

Seems to be an specific compositor bug and not an WLRoots bug, because it doesn't happen in Labwc (https://github.com/labwc/labwc), but it also happens on Sway which is a bit puzzling.

**To Reproduce**
-Run RetroArch
-Go to `Settings->Video->Synchronization->Max Swapchain Images` and set it to `2`
-`Settings->Video->Output->Estimated Screen Refresh Rate` to numerically observe the halved vsync rate (can also be observed empirically by simply navigating the menus).

**Expected behavior**
Setting `Settings->Video->Synchronization->Max Swapchain Images`  to  `2` should not halve the vsync rate.

**Wayfire version**
Latest stable (0.8.1) and latest GIT code both show this bug.
Using latest stable WLRoots (0.17.3) and latest stable MESA (24.1.1) with Intel N100 GPU and I5-1235U with XE graphics.

ammen99:
This PR aims to solve some issues with frame scheduling that Wayfire has had for a long time. Here is a summary of the impotant changes:

- Per-output `min_render_budget` (formerly `core/max_render_time`): if set to a value >= 0, it indicates the time Wayfire should reserve for its own rendering and buffer submission. The remaining time (in one refresh interval) Wayfire gives to clients to update their buffers, resulting in less input-to-screen latency if the client manages to update before Wayfire submits its frame.
- A new dynamic repaint delay algorithm (`workarounds/dynamic_repaint_delay`): if enabled, the `min_render_budget` for each output is treated like a lower bound for the rendering time. In addition, core will estimate the real rendering time and submission time and use a custom algorithm to adjust the repaint delay as the workload changes. The algorithm seems to work fine, but needs a lot of testing before it can be enabled by default. The value of an algorithm is that during less-GPU-intense tasks, the repaint delay can be increased to lower latency, and it is decreased during heavy effects like for example Expo.
- The dynamic repaint delay algorithm also keeps track of composited vs scanned out frames separately, so hopefully can fix #2377
- There is a per-output option `vrr_idle_refresh_rate` which enforces a minimal refresh rate, potentially resolving issues like #1621

Overall this is a big change so if anyone is interested, please test, especially with dynamic repaint delay and max_render_time/min_render_budget set to -1.


mark-herbert42:
so far so good. Before this patch I always disabled dynamc delay for heavy GPU (and battery ) usage, Now seems pretty fair.

ammen99:
> so far so good. Before this patch I always disabled dynamc delay for heavy GPU (and battery ) usage, Now seems pretty fair.

Thanks for testing!

PR #3104, opened 2026-08-09
fix hermeticity

hustlerone:
can't build in nix without patching this in

### DIFF
--- a/plugins/common/wayfire/plugins/common/cairo-util.hpp
+++ b/plugins/common/wayfire/plugins/common/cairo-util.hpp
@@ -11,7 +11,7 @@
 #include <wayfire/render.hpp>
 
 // TODO: do we need some kind of dependency here?
-#include <drm_fourcc.h>
+#include <libdrm/drm_fourcc.h>
 
 namespace wf
 {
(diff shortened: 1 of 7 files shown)


plugins/common/wayfire/plugins/common/cairo-util.hpp
```diff
@@ -11,7 +11,7 @@
 #include <wayfire/render.hpp>
 
 // TODO: do we need some kind of dependency here?
-#include <drm_fourcc.h>
+#include <libdrm/drm_fourcc.h>
```
ammen99, line 14: This seems wrong, on my system, `pkg-config --cflags libdrm` gives `-I/usr/include/libdrm` where `drm_fourcc.h` is located. wlroots also includes the header with just `drm_fourcc.h` so I think that after adding the drm dep, you shouldn't need this.

teatwig: I think adding `drm` to the meson.build is the only change that should be necessary.
Even in nixpkgs pkg-config correctly returns `.../libdrm` as an include dir:
```sh
$ pkg-config --cflags libdrm 
-I/nix/store/jl06f30bsw7ya4qvlpawjqx23x52hm62-libdrm-2.4.133-dev/include -I/nix/store/jl06f30bsw7ya4qvlpawjqx23x52hm62-libdrm-2.4.133-dev/include/libdrm
```

PR #3106, opened 2026-08-10
squeezimize: fix default behavior

hustlerone:
The default behavior was broken since commit `3b6d5d5286265a58e784e18c1ece5ea867295c52`, where minimize_target (the function input) was being used for calculations instead of using this->minimize_target (the member of the class that stores the final target).

As of now (without this PR), if you run squeezimize you will see it will not squeeze into anything, it'll just float around then pop out of existence.

### DIFF
--- a/plugins/animate/squeezimize.hpp
+++ b/plugins/animate/squeezimize.hpp
@@ -338,8 +338,8 @@ class squeezimize_transformer : public wf::scene::view_2d_transformer_t
         auto geom   = output->get_relative_geometry();
 
         // check which edge the target is closest to
-        double x  = minimize_target.x + minimize_target.width / 2.0;
-        double y  = minimize_target.y + minimize_target.height / 2.0;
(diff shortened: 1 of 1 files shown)


soreau:
@dkondor ping on this, can you have a look?

dkondor:
Hi,

I don't have time to test this right now, but looking into the change and the [original commit](https://github.com/WayfireWM/wayfire/commit/3b6d5d5286265a58e784e18c1ece5ea867295c52), it was likely an oversight from me not to use `this->minimze_target` that this PR would correct.

Note: I think the problematic behavior only occurs if no client (taskbar / dock) sets a minimize location, since this is the only case when `this->minimize_target` and `minimize_target` will differ. This explains why we did not notice when testing.

soreau:
@dkondor Thanks. Just by looking, I arrived at the same conclusions. It seems this patch is rightfully doing what the commit message says.

@hustlerone Thanks for your pull request.

ammen99:
Thanks!

PR #3107, opened 2026-08-10
wrot: lower FOV

hustlerone:
un-exaggerates the perspective effect. The view is way flatter and the distortion toward the near plane that comes with the projection is pretty much unnoticeable with this change.

I've personally settled with PI/12, since for my 1080p display it's the highest fov with least amount of distortion.

https://www.mediafire.com/file/3o262gj2umzo70n/DISTORTION_FILES.zip/file

### DIFF
--- a/src/view/view-3d.cpp
+++ b/src/view/view-3d.cpp
@@ -351,7 +351,7 @@ void view_2d_transformer_t::gen_render_instances(
 }
 
 /* -------------------------------- 3d view --------------------------------- */
-const float view_3d_transformer_t::fov = PI / 4;
+const float view_3d_transformer_t::fov = PI / 12;
 glm::mat4 view_3d_transformer_t::default_view_matrix()
 {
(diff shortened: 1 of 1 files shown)


ammen99:
This changes not just wrot but also the switcher plugin. Have you tested with that? Also, maybe we should simply switch to a simple orthographic projection, or if that is worse, try to derive the best angle depending on the resolution (since you imply this relies on the resolution)?

hustlerone:
> This changes not just wrot but also the switcher plugin. Have you tested with that? Also, maybe we should simply switch to a simple orthographic projection, or if that is worse, try to derive the best angle depending on the resolution (since you imply this relies on the resolution)?

Not sure if it relies on resolution but I don't how if it's vertical or horizontal FOV. If it's vertical FOV then it should be agnostic.

As for the switcher it looked fine to me. Although I'd like your opinion on this change

PR #3108, opened 2026-08-11
added titlebar text alignment options for the decoration plugin

mrsteve0924:
added horizontal and vertical alignment options

title_halign = left, center, right
title_valign = top, center, bottom

### DIFF
--- a/metadata/decoration.xml
+++ b/metadata/decoration.xml
@@ -21,6 +21,40 @@
 			<max>1.0</max>
 			<min>0.1</min>
 		</option>
+		<option name="title_halign" type="string">
+			<_short>Title horizontal alignment</_short>
+			<_long>Sets the horizontal alignment of the title text. Valid values are left, center and right.</_long>
+			<default>left</default>
+			<desc>
+				<value>left</value>
+				<_name>left</_name>
+			</desc>
+			<desc>
+				<value>center</value>
+				<_name>center</_name>
+			</desc>
+			<desc>
+				<value>right</value>
+				<_name>right</_name>
+			</desc>
+		</option>
+		<option name="title_valign" type="string">
+			<_short>Title vertical alignment</_short>
+			<_long>Sets the vertical alignment of the title text. Valid values are top, center and bottom.</_long>
+			<default>top</default>
+			<desc>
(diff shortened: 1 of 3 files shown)


metadata/decoration.xml
```diff
@@ -14,13 +14,27 @@
 			<_long>Sets the font color to use for the title bars.</_long>
 			<default>#ffffffff</default>
 		</option>
+
```
soreau: There should be no newline in between options.

metadata/decoration.xml
```diff
@@ -14,13 +14,27 @@
...
 		<option name="font_scale" type="double">
 			<_short>Font scale</_short>
 			<_long>Sets the font scale to use for the title bars.</_long>
 			<default>0.8</default>
 			<max>1.0</max>
 			<min>0.1</min>
 		</option>
+
+		<option name="title_halign" type="string">
```
soreau, line 24: In my opinion, these two alignment options should be like the `deform` option for cube. This way, it will appear as a drop-down in wcm.

soreau (approved):
Looks fine to me aside from the nitpicks below.

metadata/decoration.xml
mrsteve0924: i added the drop downs

metadata/decoration.xml
mrsteve0924: fixed

plugins/decor/deco-theme.cpp
```diff
@@ -79,15 +79,52 @@ cairo_surface_t*decoration_theme_t::render_text(std::string text,
     PangoFontDescription *font_desc;
     PangoLayout *layout;
 
-    // render text
+// render text
```
ammen99: looks misindented?

ammen99: nvm fixed it myself

ammen99 (approved):
lgtm too, thanks!

plugins/decor/deco-theme.cpp
mrsteve0924: that's strange because it passed uncrustify. oh well.  thanks for fixing

Issue #3030, opened 2026-05-17
command: always + repeatable binding


trigg:
## Describe the feature
We currently have variants of command which can be used while in lockscreen (among other situations) and ones that repeat when held, but not one which does both.

## Use case / Example
With an always + repeatable variant we could have volume keys which repeat and are usable on lockscreen

## Benefits
Same as example

PR #3112, opened 2026-08-18
command : new configuration syntax and more flexible command execution command conditions

AKArien:
Remove the always_bindings category. Instead, make all other types of bindings configurable as `always`.

I don’t quite expect it to be merged as is, as it breaks configuration and might not be the cleanest/best way to do this, but i am looking for feedback/someone to have a better way of handling things.


soreau:
It is not clear what problem this might solve. Can you elaborate?

AKArien:
> It is not clear what problem this might solve. Can you elaborate?

Currently, you can only have a binding take place always (when another plugin is running, and when the screen is locked) that behaves like a regular binding. This allows having this « always » behaviour for repeatable and release bindings.

ammen99:
Fwiw I think you could add a new category of bindings, which can have repeatable, always and release as boolean options instead of being separate lists. Then we say everything in the old options is deprecated and remove it after 1-2 releases.

ammen99:
Also fwiw the ini syntax is ugly (not your fault, just how it is), if you have any better ideas of how to encode the information, that would be great. Maybe a command prefix of always$, repeat$<cmd>, for ex. `command = always$terminator`. I don't think this is necessarily better, simply saying that I'm open for other suggestions too.

AKArien:
@ammen99 atfer some thinking, i have a one-line declaration syntax that is `:selector1,selector2…:<command> = <activator>`, where selectors are any of normal/repeat/release/always.

This would make a binding work for any of these modes, so for example, to take back the example of press-to-speak in the wiki, you’d have one toggle binding for both normal and release, and it would of course allow getting an `always` and `repeat` for anything.

Am i correct in my understanding that it’s fine to drop the « name » (so, `terminal` in `binding_terminal = <super> KEY_ENTER`, `command_terminal = alacritty`) of the command, and that it’s not used for anything in the ipc or otherwise ?

And does this config syntax sound good to you ?

ammen99:
> @ammen99 atfer some thinking, i have a one-line declaration syntax that is `:selector1,selector2…:<command> = <activator>`, where selectors are any of normal/repeat/release/always.
> 
> This would make a binding work for any of these modes, so for example, to take back the example of press-to-speak in the wiki, you’d have one toggle binding for both normal and release, and it would of course allow getting an `always` and `repeat` for anything.
> 
> Am i correct in my understanding that it’s fine to drop the « name » (so, `terminal` in `binding_terminal = <super> KEY_ENTER`, `command_terminal = alacritty`) of the command, and that it’s not used for anything in the ipc or otherwise ?
> 
> And does this config syntax sound good to you ?

I am not sure I understand, could you share an example of how it will look in the config file?

AKArien:
> I am not sure I understand, could you share an example of how it will look in the config file?

Of course !

Say you have the following :
```
repeatable_binding_vup = KEY_VOLUMEUP
command_vup = wpctl set-volume -l 1.0 @DEFAULT_AUDIO_SINK@ 2.5%+

repeatable_binding_vdown = KEY_VOLUMEDOWN
command_vdown = wpctl set-volume -l 1.0 @DEFAULT_AUDIO_SINK@ 2.5%-

always_binding_mute = KEY_MUTE
command_mute = wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle

binding_terminal = <super> KEY_ENTER
command_terminal = alacritty
```

And for a press-to-toggle microphone, i’d think it would be something like :
```
binding_pmicmute = KEY_F20
command_pmicmute = wpctl set-mute @DEFAULT_AUDIO_SOURCE@ toggle

release_binding_rmicmute = KEY_F20
command_rmicmute = wpctl set-mute @DEFAULT_AUDIO_SOURCE@ toggle
```
But this doesn’t work on my system, i probably messed it up in my initial patch that i still have installed ? Either way :

Which, with these changes, become :
```
:repeat, always : wpctl set-volume -l 1.0 @DEFAULT_AUDIO_SINK@ 2.5%+ = KEY_VOLUMEUP
:repeat, always : wpctl set-volume -l 1.0 @DEFAULT_AUDIO_SINK@ 2.5%- = KEY_VOLUMEDOWN
:always: wpctl set-mute @DEFAULT_AUDIO_SINK@ toggle = KEY_MUTE
:normal, release, always : wpctl set-mute @DEFAULT_AUDIO_SOURCE@ toggle = KEY_F20
::alacritty = <super> KEY_ENTER
```

Note that this is defaulting empty/invalid values to `normal` bindings, hence why the alacritty one works, and that `repeat` also include the initial press like `normal` instead of being only while held down.

ammen99:
@AKArien what about commands which contain = in them? Also I just realized the reason why I was trying to use the compound lists in the first place, if you have hjson or similar, then it would be nice to have `{binding: <whatever>, repeatable: True, always: True}` or similar. But maybe we should simply support both formats, one big compound list, and one hacky list like you proposed. @marcusbritanicus what do you think, with hjson, would my proposal work?

AKArien:
> @AKArien what about commands which contain = in them?

Wouldn’t that simply be wf-config’s job to allow escaping them by preceeding them with a backslash ?

ammen99:
> > @AKArien what about commands which contain = in them?
> 
> Wouldn’t that simply be wf-config’s job to allow escaping them by preceeding them with a backslash ?

wf-config does not allow escaping `=`, because so far option names have been 'nice' .This would be (if implemented) the very first options to even be allowed to contain a `=` sign. Have to say I am not too huge a fan of the idea. Maybe we should stick to something like:

```
command_1 = <whatever>
binding_1 = <whatever>
is_repeatable_1 = True
is_always_enabled_1 = True
is_release_binding_1 = False
```

Alternatively:


```
command_1 = <whatever>
binding_1 = <whatever>
flags_1 = repeatable,always # or release
```

Does this make sense? It would be one compound option, easy to set from hjson or any other non-ini backend, both variants preserve backwards compatibility too. We can keep the existing repeatable etc bindings for a while as a legacy option.

Otherwise I feel like we'd have to invent syntax like what you're proposing:

```
b1 = repeat,always:<super>KEY_P:<whatever here>
```

It would require more manual parsing but maybe is preferred. What do people here prefer?

AKArien:
> Maybe we should stick to something like:
> 
> ```
> command_1 = <whatever>
> binding_1 = <whatever>
> is_repeatable_1 = True
> is_always_enabled_1 = True
> is_release_binding_1 = False
> ```

I have always found this multi-part syntax clunky and annoying to work with, probably in part because it leads to writing the same word over and over again, and also just isn’t very easy to read, hence why i was excited to replace it.

> both variants preserve backwards compatibility too. We can keep the existing repeatable etc bindings for a while as a legacy option.

is that any different than just keeping compatibility with the old options though ? this is still a different format than before, just more similar ?

> wf-config does not allow escaping `=`, because so far option names have been 'nice' .This would be (if implemented) the very first options to even be allowed to contain a `=` sign.

Sounds to me like it’s the occasion to, no ? Is there much of a reason to not want to allow that ?

> Otherwise I feel like we'd have to invent syntax like what you're proposing:
> 
> ```
> b1 = repeat,always:<super>KEY_P:<whatever here>
> ```

That would work as well, and shouldn’t be too hard, but at this point might as well add support to wf-config to allow escaping =, and get to use both « sides » of the option no ?
 
> It would require more manual parsing but maybe is preferred. What do people here prefer?

My preference here should be obvious :P

soreau:
> I have always found this multi-part syntax clunky and annoying to work with, probably in part because it leads to writing the same word over and over again, and also just isn’t very easy to read, hence why i was excited to replace it.

Wayfire is equipped to use different config backends, so you can make up any type of syntax you want. For example, @marcusbritanicus wrote a backend for `hjson` if I remember correctly. You basically have to replace [this](https://github.com/WayfireWM/wayfire/blob/master/src/default-config-backend.cpp) and load it with [this option](https://github.com/WayfireWM/wayfire/blob/master/src/main.cpp#L39).

AKArien:
> You basically have to replace [this](https://github.com/WayfireWM/wayfire/blob/master/src/default-config-backend.cpp) and load it with [this option](https://github.com/WayfireWM/wayfire/blob/master/src/main.cpp#L39).

I see, it seems i was misunderstanding how config works (at least more than now, oops), my apologies.

Then, i guess i would think the `flags` section Ammen mentionned :

> ```
> command_1 = <whatever>
> binding_1 = <whatever>
> flags_1 = repeatable,always # or release
> ```

Is probably the best way to go ? As a syntax like (again, as Ammen mentionned) :

> `b1 = repeat,always:<super>KEY_P:<whatever here>`

Would be the « correct » way to the one-line, but just be really clunky and annoying and not worth it at all ?

marcusbritanicus:
@ammen99 @soreau @AKArien Hi. I have been out of touch with github for a while. 

> @AKArien what about commands which contain = in them? Also I just realized the reason why I was trying to use the compound lists in the first place, if you have hjson or similar, then it would be nice to have {binding: <whatever>, repeatable: True, always: True} or similar. But maybe we should simply support both formats, one big compound list, and one hacky list like you proposed. @marcusbritanicus what do you think, with hjson, would my proposal work?

It would work great. With (h)json you get nearly limitless flexibility. Instead of hjson, we could use json - after all we have native support for json in wayfire. For example,

```json
"command": {
    "terminal": {
        "binding": "<super> KEY_ENTER",
        "repeatable": false,
        "always": false,
        "on_release": true,
    },
}
```

As for ini version, we could use a syntax like @ammen99 proposed (perhaps with a few changes?):
> Otherwise I feel like we'd have to invent syntax like what you're proposing:
> ```b1 = repeat,always:<super>KEY_P:<whatever here>```
> It would require more manual parsing but maybe is preferred. What do people here prefer?

Personally, I would prefer it to be a bit more verbose:
```ini
[command]
terminal=command:"<a very long command with various characters>",binding:<super> KEY_ENTER,repeat:false,release:false,always:false
```
If I am not mistaken, this will be natively parsed as an array by ini parsers. We take each element and get all the information we want. Also this version would make the command syntax highly extensible. We could add more "options" with full backward compatibility in the future.

> Wayfire is equipped to use different config backends, so you can make up any type of syntax you want. For example, @marcusbritanicus wrote a backend for hjson if I remember correctly. You basically have to replace [this](https://github.com/WayfireWM/wayfire/blob/master/src/default-config-backend.cpp) and load it with [this option](https://github.com/WayfireWM/wayfire/blob/master/src/main.cpp#L39).

@AKArien You can find it here: https://gitlab.com/wayfireplugins/wf-config-hjson
While it "works", it could do with a bit of fine tuning.

AKArien:
Just realised issue #3030 exists, which is about the same problem this tries to remedy

Issue #3027, opened 2026-05-13
[feature request] Locking the zoom


vdegenne:
## Describe the feature
The zoom feature follows the cursor by default. There is no option to avoid that.

## Benefits
The main advantage is to be able to work on a specific area of the screen and then being able to move the cursor around to do precise work (text editing, design, gaming(?), ...) without making other valuable elements going out of focus.

soreau:
It might not help your use-case, but I wanted to mention that there is also `mag` plugin in wayfire-plugins-extra.

vdegenne:
Also I found https://github.com/labwc/labwc/issues/1703 
Someone is mentioning a "slightly modified version of the standard zoom plugin" I don't know if that's the `mag` plugin you are referring to. The link seems to be pointing to the standard zoom plugin, I'm not sure.

soreau:
The mag plugin opens a window that is your zoom area. The zoom window can be moved or resized. The common use-case is to move the zoom window to another output while using it.

vdegenne:
I was getting too excited and thought I implemented it first try...

https://github.com/user-attachments/assets/a343801e-ecef-4498-a9bd-d21f765c961c

Visually it is working but the mouse events are still bound to the absolute values of the viewport (in the video you can clearly see it on the links as I move the mouse around them.)
Is there an easy fix?

soreau:
I looked into this and the answer is no, there is not an easy fix. :-)

Here's what I came up with:

```
diff --git a/plugins/single_plugins/zoom.cpp b/plugins/single_plugins/zoom.cpp
index b8054eb5..cf0b2d4f 100644
--- a/plugins/single_plugins/zoom.cpp
+++ b/plugins/single_plugins/zoom.cpp
@@ -1,10 +1,44 @@
 #include <wayfire/per-output-plugin.hpp>
+#include <wayfire/signal-definitions.hpp>
 #include <wayfire/output.hpp>
 #include <wayfire/render.hpp>
 #include <wayfire/render-manager.hpp>
 #include <wayfire/util/duration.hpp>
+#include <wayfire/view-transform.hpp>
+#include <linux/input-event-codes.h>
 
-class wayfire_zoom_screen : public wf::per_output_plugin_instance_t
+
+namespace wf
+{
+namespace zoom
+{
+using namespace wf::scene;
+const std::string zoom_transformer_name = "zoom_transformer";
+
+class locked_pointer_input_translation_node_t : public wf::scene::view_2d_transformer_t
[...]

PR #3121, opened 2026-09-02
[zoom plugin] no mouse follow

vdegenne:
I leave that here in case you want to help me collaborate.

It's just a plain implementation of no mouse follow for the native zoom plugin.

The main issue left is mouse events are not properly scaling with the zoom factor.

should implement https://github.com/WayfireWM/wayfire/issues/3027

note: Also `no follow` should be a toggle flag, it's on the todo-list.


soreau:
This code looks a lot more complicated than I'd expect for this feature. The way I see it, you should be able to install an activator binding which toggles a boolean and records the cursor position. Then, if the boolean is set, use the saved cursor position instead of the actual cursor position. Bonus points to add a transformer that smooth-moves from saved cursor position to actual cursor position when the keybinding is toggled 'off'.

vdegenne:
@ammen99 I took a little of both world. I tested with the changes you've made in `render-manager.cpp` and it works great without transformers now.
Also lock is now a toggleable option as inspired by @soreau's idea in proposed patch.
Your edge idea is also great so it was introduced as a check option as well we can opt in or out.
Here's a demo:


https://github.com/user-attachments/assets/4b273d3b-000f-442f-9397-9cbd7028b7aa

Note: the lock releases when we are completely zoomed out, the workflow goes basically like zoom/aim → lock → adjust(if push edges is on), zoom/aim → lock → adjust, ...

metadata/zoom.xml
```diff
@@ -4,22 +4,38 @@
 		<_short>Zoom</_short>
 		<_long>A plugin to zoom in the desktop with the mouse.</_long>
 		<category>Accessibility</category>
+
```
soreau: This extra newline should be omitted.

metadata/zoom.xml
```diff
@@ -4,22 +4,38 @@
...
 		<option name="modifier" type="key">
 			<_short>Zoom in the desktop by scrolling with modifier</_short>
 			<_long>Scrolls with the specified modifier to zoom in and out.</_long>
 			<default>&lt;super&gt;</default>
 		</option>
+
```
soreau: There shouldn't be newlines between the options.

metadata/zoom.xml
```diff
@@ -4,22 +4,38 @@
...
+		<option name="lock" type="activator">
+			<_short>Toggle locked zoom</_short>
+			<_long>Locks or unlocks the zoomed viewport.</_long>
+			<default>&lt;super&gt; BTN_MIDDLE</default>
+		</option>
+
```
soreau: Extra newline.

metadata/zoom.xml
```diff
@@ -4,22 +4,38 @@
...
 			<default>0.01</default>
 			<precision>0.001</precision>
 		</option>
+
 		<option name="smoothing_duration" type="animation">
 			<_short>Smoothing duration</_short>
 			<_long>Sets the smoothing duration in milliseconds.</_long>
 			<default>300ms linear</default>
 		</option>
+
```
soreau: No newlines around options.

metadata/zoom.xml
```diff
@@ -4,22 +4,38 @@
...
+		<option name="edge" type="bool">
+			<_short>Push edges</_short>
+			<_long>Allows the locked viewport to move when the cursor reaches its edge.</_long>
+			<default>false</default>
+		</option>
+
```
soreau: Extra newline.

plugins/single_plugins/zoom.cpp
```diff
@@ -85,21 +102,95 @@ class wayfire_zoom_screen : public wf::per_output_plugin_instance_t
...
         const float tw     = std::clamp(w / factor, 0.0f, w - x1);
         const float th     = std::clamp(h / factor, 0.0f, h - y1);
-        auto filter_mode   = (interpolation_method == (int)interpolation_method_t::NEAREST) ?
-            WLR_SCALE_FILTER_NEAREST : WLR_SCALE_FILTER_BILINEAR;
-        destination.blit(source, {x1, y1, tw, th}, {0.0, 0.0, (double)w, (double)h}, filter_mode);
+
+        wf::geometry_t source_box{x1, y1, tw, th};
+
+        if (locked)
+        {
```
soreau, line 122: Is there really a need for all this additional complex logic? As far as I can tell, all you need to do is supply the coordinates saved at the time of lock as I did in my patch.

plugins/single_plugins/zoom.cpp
```diff
@@ -13,11 +13,18 @@ class wayfire_zoom_screen : public wf::per_output_plugin_instance_t
...
     wf::option_wrapper_t<double> speed{"zoom/speed"};
     wf::option_wrapper_t<wf::animation_description_t> smoothing_duration{"zoom/smoothing_duration"};
+    wf::option_wrapper_t<bool> edge{"zoom/edge"};
     wf::option_wrapper_t<int> interpolation_method{"zoom/interpolation_method"};
+
     wf::animation::simple_animation_t progression{smoothing_duration};
+
     bool hook_set = false;
+    bool locked = false;
+    bool has_locked_region = false;
```
soreau, line 26: `has_locked_region` seems unnecessary.

soreau (changes requested):
Thanks for working on this @vdegenne. I think it mostly looks good with the exception of the newlines in the xml metadata and all the additional code in render.

plugins/single_plugins/zoom.cpp
vdegenne: This is for when push edges is on I guess, so it knows where to relocate the viewport?
Mainly the code was taken from ammen99's PR and since I'm not a C++ expert I'd prefer if you guys could take over from here, because it seems like optimisation but the code is already doing what's intended.

soreau: @vdegenne I have taken your branch and fixed it up and pushed it to WayfireWM as `zoom-no-follow`. So if you could checkout that branch and see that it works ok, that would help. I think it works ok except multiple outputs there are some rough edges.

vdegenne: I tried out your branch and it works great yes. Just curious, are your changes meant to make the code more readable or are they ultimately leading to a real performance gain? (either way it's good)
Also I tried with multi-display and haven't particularly seen any issues. However, [this](https://github.com/WayfireWM/wayfire/pull/3130#issuecomment-5641241705) is something that has previously caught my attention yes, but it's probably not related to the zoom right? I could open this as a new issue just for reference.

Waiting your instructions from now, since your branch has both our commits in line, should I close this PR?

soreau: > I tried out your branch and it works great yes. Just curious, are your changes meant to make the code more readable or are they ultimately leading to a real performance gain? (either way it's good)

The main point is to have clean, concise code for readability. But having less variables and code in general can cut down on resource usage like ram space and cpu cycles, though admittedly not by much compared to your code. The trip came for me, when I realized there was also the new 'edge' feature introduced.

> Also I tried with multi-display and haven't particularly seen any issues.

I just pushed a commit latest on zoom-no-follow that fixed the multi-output issue I was seeing.

> However, [this](https://github.com/WayfireWM/wayfire/pull/3130#issuecomment-5641241705) is something that has previously caught my attention yes, but it's probably not related to the zoom right? I could open this as a new issue just for reference.
> 
> Waiting your instructions from now, since your branch has both our commits in line, should I close this PR?

We have a bit of a mess going on here, there is the 'bleeding cursor' issue which I figure @ammen99 should address. Then there's #3130 that's still open, zoom-no-follow branch without a PR (yet) and your #3121. I'm guessing we should close the two open PR's, clean up the zoom-no-follow branch and make sure everyone is given credit for their work and make a PR out of it. First though, I think we should wait to see what @ammen99 has to say, especially about the visual cursor problem.

PR #3125, opened 2026-09-04
Refined the Squeezimize animation

126com200103646728200303:
# Refined the Squeezimize animation to make its visual effect smoother and more consistent with the target box position.

The main changes are:

- Reduced the amount of window distortion, resulting in a softer and less aggressive animation.
- Adjusted the curvature on both sides of the window based on the position of the target box.
  Previously, the window could bend toward a fixed direction first and then curve back in the opposite direction. This change makes the curvature adapt more naturally to the target box position.
- No noticeable performance impact was observed compared to the original implementation.

## Before and After
Notice: The GPU usage shown in the video may appear somewhat high; there is no need to worry, as this is due to the poor performance of my integrated graphics.

### Before

https://github.com/user-attachments/assets/0b760a84-541a-4779-813f-27bb85ca335a

https://github.com/user-attachments/assets/c4b3659b-141f-44dc-96f3-29f0ca96673b

### After

https://github.com/user-attachments/assets/d2fa3898-1b7e-4f97-9ebb-6025c62e35c3

https://github.com/user-attachments/assets/0cabe45a-42c8-4c60-8718-2a5fed41fd01

## Testing

The updated animation was tested with common window operations,
No noticeable performance regression was observed.


plugins/animate/squeezimize.hpp
```diff
@@ -86,17 +86,18 @@ void main()
     {
         y = uv.y;
         uv_squeeze.y += -progress_pt_two * (inv_h - target_box.w);
-        sigmoid = 1.0 / (1.0 + pow(2.718, -(y * (1.0 / (src_box.w - target_box.w)) * 15.0 - 10.0)));
+        sigmoid = 1.0 / (1.0 + pow(1.718, -(y * (1.0 / (src_box.w - target_box.w)) * 14.5 - 6.0)));
```
soreau, line 89: Can you explain why these lines were changed? The changes to the four lines like these seem unnecessary. In fact, I reverted these changes and the result is much smoother. Just for your information, these values have been hand tweaked and I believe the current code yields the best result.

plugins/animate/squeezimize.hpp
```diff
@@ -86,17 +86,18 @@ void main()
...
     } else
     {
         y = 1.0 - uv.y;
         uv_squeeze.y -= -progress_pt_two * (inv_h - target_box.y + target_box.w);
-        sigmoid = 1.0 / (1.0 + pow(2.718, -(y * (1.0 / (target_box.w - src_box.y)) * 15.0 - 10.0)));
+        sigmoid = 1.0 / (1.0 + pow(1.718, -(y * (1.0 / (target_box.w - src_box.y)) * 14.5 - 6.0)));
```
soreau, line 94: Same here.

plugins/animate/squeezimize.hpp
```diff
@@ -139,17 +140,18 @@ void main()
     {
         y = 1.0 - uv.x;
         uv_squeeze.x += progress_pt_two * (inv_w - target_box.z);
-        sigmoid = 1.0 / (1.0 + pow(2.718, -(y * (1.0 / (src_box.z - target_box.z)) * 15.0 - 10.0)));
+        sigmoid = 1.0 / (1.0 + pow(1.718, -(y * (1.0 / (src_box.z - target_box.z)) * 14.0 - 8.0)));
```
soreau, line 143: Same here.

plugins/animate/squeezimize.hpp
```diff
@@ -139,17 +140,18 @@ void main()
...
     } else
     {
         y = uv.x;
         uv_squeeze.x -= progress_pt_two * (inv_w - target_box.x + target_box.z);
-        sigmoid = 1.0 / (1.0 + pow(2.718, -(y * (1.0 / (target_box.z - src_box.x)) * 15.0 - 10.0)));
+        sigmoid = 1.0 / (1.0 + pow(1.718, -(y * (1.0 / (target_box.z - src_box.x)) * 14.0 - 8.0)));
```
soreau, line 148: Same here.

soreau (changes requested):
Thanks for the PR. I have tested this briefly, and it seems to work as advertised, but it introduces jerky motion, as if progress_pt_two is much faster than the first. Aside from the comments below, I approve these changes.

plugins/animate/squeezimize.hpp
126com200103646728200303: Thank you very much for your feedback. The reason for modifying these parts is that I felt the original animation curves were somewhat abrupt and not smooth enough, or to put it another way: it appears unnatural and too rigid. You can take a look at these two images:

Before: 
Screenshot_20260904_170818

After: 
Screenshot_20260904_170925


I think the second one is better, what do you think?

126com200103646728200303:
And you mentioned 'jerky motion.' Could you elaborate on it? Or send a video clip to show it? I can't tell what it looks like. Or is it that your duration is set too short, making it look too fast?

126com200103646728200303:
Maybe you thought the latter part of the video '  ScreenRecording_20260904_160620-bottom-mine.mp4' was lagging? That’s because I clicked the icon on the Dock again before it finished playing. I wanted to show what it looks like when it's interrupted, it wasn’t actually lagging.

126com200103646728200303:
Some people care a lot about animation interruptions, so I showed it, which might have caused you some misunderstanding.

mark-herbert42:
As I understand this animation was inspired by magic lamp animation - where the window was supposed to hide in waves like Genie into magic lamp. So I definetely like begore vatiant much more than after - just straightforward and boring hide.

mark-herbert42:
https://www.youtube.com/watch?v=ZEDIej7EbL8

Thats how original one looked like.

126com200103646728200303:
> As I understand this animation was inspired by magic lamp animation - where the window was supposed to hide in waves like Genie into magic lamp. So I definetely like begore vatiant much more than after - just straightforward and boring hide.

> https://www.youtube.com/watch?v=ZEDIej7EbL8
> 
> Thats how original one looked like.


Alright, but obviously ours won't wiggle back and forth like on YouTube to change direction. It will only wiggle once in the opposite direction and then fly straight to the icon on the Dock. For someone used to macOS (like me), it looks pretty confusing, and I even think it might be a bug 😂

126com200103646728200303:
> Genie

Anyway, I didn't feel like "it’s like a Genie." To me, it feels more like it got drunk, flew the wrong way, and then suddenly came back 😅

mark-herbert42:
If i remember correctly from 201* the number of wiggles was configurable - so there it was possible to do straight pull into dock  as in new version(0 wiggles), do one - like it does now, or select 3 and get lke on video.

soreau:
> Thank you very much for your feedback. The reason for modifying these parts is that I felt the original animation curves were somewhat abrupt and not smooth enough, or to put it another way: it appears unnatural and too rigid. You can take a look at these two images:

I think by far, the 'Before' image is much better. This is not about shape though, this is about the smoothness of the animation itself.

You have marked all conversations resolved, and the issue I have identified has not been resolved. To be clear, I will not approve changes in the 4 lines of code I highlighted. I will un-resolve the review comments for you.

> And you mentioned 'jerky motion.' Could you elaborate on it? Or send a video clip to show it? I can't tell what it looks like. Or is it that your duration is set too short, making it look too fast?

I do not have time to make videos demonstrating the jerky motion, but I can tell you how to reproduce it. If you have a window that is about half the screen height tall and a very thin width, put the window on the opposite side of where the minimize target is on the screen and try it. The problem is that the first part of the animation happens relatively 'normally' and the latter half happens very quickly, causing a jerkiness that is unpleasant. This has nothing to do with duration, as it is the same for any duration. I don't intend to explain much further, the changes I mentioned in my review are unacceptable. The other test I have worked with is somewhat opposite, where the height is minimal and the width is extended. In the square case it might be less noticeable but the problem is still present.

> Maybe you thought the latter part of the video ' ScreenRecording_20260904_160620-bottom-mine.mp4' was lagging? That’s because I clicked the icon on the Dock again before it finished playing. I wanted to show what it looks like when it's interrupted, it wasn’t actually lagging.

No, I tested myself, I am not judging anything from your video clips.

> Some people care a lot about animation interruptions, so I showed it, which might have caused you some misunderstanding.

It seems you are the one that is not understanding. When I wrote this I tested extensively, so I am quite sensitive to changes that are damaging to the aesthetics.

@dkondor Might you have anything to add here?

126com200103646728200303:
Since you believe this 'damaging aesthetic' and the bug you mentioned does not exist on my computer at all, or the method you referred to for triggering it is completely ineffective, I believe this PR is no longer necessary and it will be closed.

Furthermore, saying 'no time, no further explanation' is not an ideal way to collaborate in open source projects.

Goodbye.

dkondor:
Hi All,

I've just seen this; have not tested, only looking at the proposed changes:
- lines 97-100 (and 151-154) seem like a nice way to simplify the calculation and at a first glance are likely correct (equivalent or similar to the original), but I'd need to use a pen and paper to verify it to have a more definite opinion; happy to look into this if there is a chance to resurrect this MR
- looking at the two images in the comments and the videos I also have a preference for the "before" version, but this is highly subjective (my reference is Compiz, where it seems to be actually inbetween the two cases here); I could imagine having a config option for setting the "sharpness" of the sigmoid, but I know it's always a trade-off between having more configuration options vs too much complexity (I'm usually in favor of more options, but I know that good UI / UX is considered to minimize the complexity and make things "just work")
- in any case, I'd be strongly against changing the base of the pow function (line 89), but this is also subjective preference for sticking with nice numbers :) actually, changing from `pow(2.718, ...)` to `exp(...)` could make sense to be clearer about the intent; so if add an option, I'd suggest it to be an additional multiplicative factor in the exponent
- a nice thing would be to add some waves (as a configurable parameter of course); I guess this would be achieved by shifting or multiplying `uv_squeeze.y` by a sinusoid -- I've actually thought of this before, but never had the time to look into it

soreau:
> * in any case, I'd be strongly against changing the base of the pow function (line 89), but this is also subjective preference for sticking with nice numbers :) actually, changing from `pow(2.718, ...)` to `exp(...)` could make sense to be clearer about the intent; so if add an option, I'd suggest it to be an additional multiplicative factor in the exponent

@dkondor Thanks for your reply, it was helpful.

I probably didn't implement this correctly but here is a proof-of-concept of what an option might look like:

https://github.com/user-attachments/assets/632836b8-4aba-4cae-a63b-7d8e93359042

@126com200103646728200303 A few things I would like to say:

- Your work is valid and an improvement, aside from the disagreement about shape.
- I found that the jerky behavior I noticed with different shapes is mitigated when adjusting progress part one to not use pow().
- I would like to see this patch in wayfire upstream.
- I think @dkondor makes a good point about adding an option here.
- If you would like to implement an option, I can offer the patch which is based on your work.
- I am still open to collaboration on this, I'm not really sure why you closed your pull request here.

PR #3126, opened 2026-09-04
plugin-loader: adjust path for plugins installed with wayfire-plugin

### LINKED ISSUE #3124: wayfire-plugin meson libdir can be incorrect (Ubuntu 24.04)
## Wayfire version
0.12 (latest git)

## GPU / Driver
Not relevant

## Describe the bug
I'm trying out the new `wayfire-plugin` command with my plugins, but ran into some problems. Specifically, I'm trying to follow the instructions [here](https://github.com/WayfireWM/wayfire/blob/master/docs/wayfire-plugin-author.md) to adjust the build of a plugin.

## To Reproduce

1. Write a simple plugin that uses meson
2. Use the recommended command to install, e.g.
```
scale_ipc_filter_lib = shared_module('scale_ipc_filter', sources,
    dependencies: [wayfire],
    install: true,
    install_dir: get_option('prefix') / get_option('libdir') / 'wayfire',
    cpp_args: ['-Wno-unused-parameter', '-Wno-format-security','-DWAYFIRE_PLUGIN'],
    link_args: '-rdynamic')
```
3. Plugin gets installed under: `~/.local/share/wayfire/plugin-manager/install/lib/x86_64-linux-gnu/wayfire/`
4. Wayfire cannot load it as it is looking for plugins under `~/.local/share/wayfire/plugin-manager/install/lib/wayfire` -- see [here](https://github.com/WayfireWM/wayfire/blob/1287ff2d5ae48cea835793e2230e1526e8089980/src/core/plugin-loader.cpp#L329)

Note: the plugin shows up as correctly installed with `wayfire-plugin`:
```
wayfire-plugin list
wayfire-scale-ipc: 0.12.0 ABI 20260801
```

## Expected behavior
Wayfire can find and load the plugin.

Apparently, on Ubuntu (and I assume Debian as well), `get_option('libdir')` expands to `lib/x86_64-linux-gnu` by default. Everything works as expected if I change the meson install command to include:
```
install_dir: get_option('prefix') / 'lib' / 'wayfire',
```

I suggest to either:
 - update the [doc](https://github.com/WayfireWM/wayfire/blob/master/docs/wayfire-plugin-author.md) to suggest using 'lib' as that is where Wayfire will look for (and there does not seems to be a downside of a hardcoded path, since this is a directory only used by Wayfire)
 - change the [path discovery](https://github.com/WayfireWM/wayfire/blob/1287ff2d5ae48cea835793e2230e1526e8089980/src/core/plugin-loader.cpp#L329) to use the "libdir" that meson does via `config.h` (that now only contains a full path to system plugins)

Let me know which solution is preferred, and I can create a PR.

dkondor:
Fixes #3124 

Ensure that we use that path that will be generated by meson using `get_option('prefix') / get_option('libdir') / 'wayfire'`.

Note: the 'libdir' component can vary among systems; while most often it is just `lib`, e.g. on Ubuntu, it is `lib/x86_64-linux-gnu'.

Note: I'm not that happy with the `REL_LIBDIR` name for this variable, happy to hear alternate suggestions :)


soreau:
> Fixes #3124
> 
> Ensure that we use that path that will be generated by meson using `get_option('prefix') / get_option('libdir') / 'wayfire'`.

Nice work, this is basically exactly what I had in mind. (aside from the `REL_LIBDIR` name, heh)

> Note: I'm not that happy with the `REL_LIBDIR` name for this variable, happy to hear alternate suggestions :)

How about `WF_LIBDIR`?

src/core/plugin-loader.cpp
```diff
@@ -326,8 +326,9 @@ std::vector<std::string> wf::get_plugin_paths()
 
     if (xdg_data_dir != "")
     {
-        plugin_prefixes.push_back(xdg_data_dir + "/wayfire/plugin-manager/install/lib/wayfire");
-        plugin_prefixes.push_back(xdg_data_dir + "/wayfire/plugin-manager/install/lib64/wayfire");
+        // REL_LIBDIR from meson.build: get_option('libdir') to be compatible with wayfire-plugin
```
ammen99, line 329: can't we push the old paths just so that this is backwards compatible?

dkondor: Yes, we could -- my thinking was that cases where things worked with plugins using `get_option('libdir')` will continue to do so. The only breakage will come from when the plugin build was manually adjusted to the hardcoded path (which I expect not to be common, since I was the first one reporting the issue). But in any case, better to be safe, so I'll add the old paths explicitly.

dkondor: OK, just pushed an update to include the original paths here. BTW, I was wondering whether it would make sense to use `fs::path` here as well as it is done in `plugin.cpp` (e.g. [here](https://github.com/WayfireWM/wayfire/blob/4ab4428dedab438840078ac09e47957af097dd0a/src/core/plugin.cpp#L137))

ammen99: would be nice, if you feel like updating the code once again :)

dkondor: Yes, I'm happy to look into this, but might be only during the weekend / next week

ammen99 (approved):
lgtm, thanks!

PR #3130, opened 2026-09-10
zoom: implement locked pointer

### LINKED ISSUE #1380: Zoom in, mouse shrinks.
**Describe the bug**
Even after applying a bigger size to my mouse, it always becomes a teensy tiny dot when I zoom in on my desktop. Often times, I am left guessing where it went or zoom out entirely to find it again. Because sometimes, the mouse seems to actually shrink.

**To Reproduce**
1. Enable plugin zoom
2. Do zoom in
3. Cursor does not zoom - and sometimes it even shrinks.

**Expected behavior**
I expected the zoom to either keep the incrased pointer size, or even zoom it in, which I wouldn't mind either.

**Screenshots or stacktrace**
After waking my MacBook from sleep and zooming in and out whilst pointing my phone at it didn't make it shrink the cursor. I guess it felt like behaving... Anyway, this is what it looks like zoomed in. Its smol.
smol dot

**Wayfire version**
0.7.2, package via chaotic-aur on Garuda

### LINKED ISSUE #2095: Cursor always in the middle in zoom mode
I don't know how hard would this be to implement,
Currently when using "zoom" the screen is bound to the corners, which means we have to move the cursor really near 0,0 to be able to see an icon near 0,0.
It'd be really nice to be able to have an absolute view of the screen at the pointer location, that would mean the cursor would always show in the middle and if the cursor goes at 0,0 then the top and left half of the screen would be black (out of bounds) but the pointer should be blocked in order to not being able to go further.
If I remember correctly that is how Windows Magnifier works.

### LINKED ISSUE #3027: [feature request] Locking the zoom

ammen99:
- **zoom: add locked mode**
- **render-manager: pass sw cursors through post processing**

This PR changes the way core handles cursors and post-processing effects: cursors are now passed in through post-processing. This fixes #1380

Also, I asked GPT to implement alternative zoom modes:

```ini
[zoom]
mode = locked # follow (default, old behavior) or edge
```

This fixes #2095 and fixes #3027

Supersedes #3121


ammen99:
@vdegenne @soreau check this out

soreau:
> * **zoom: add locked mode**
> * **render-manager: pass sw cursors through post processing**
> 
> This PR changes the way core handles cursors and post-processing effects: cursors are now passed in through post-processing. This fixes #1380

Sounds good..

> 
> Also, I asked GPT to implement alternative zoom modes:

..but I disagree with this. The way I see the flow is that you zoom in 'normally', position the zoom area where you want it, then use a keybinding/activator to lock/unlock the zoom area. Then when unlocked, the zoomed area animates back to wherever the cursor is. This way, you don't have to modify the config to do what you want, you can lock/unlock on the fly.

ammen99:
> > * **zoom: add locked mode**
> > * **render-manager: pass sw cursors through post processing**
> > 
> > This PR changes the way core handles cursors and post-processing effects: cursors are now passed in through post-processing. This fixes #1380
> 
> Sounds good..
> 
> > Also, I asked GPT to implement alternative zoom modes:
> 
> ..but I disagree with this. The way I see the flow is that you zoom in 'normally', position the zoom area where you want it, then use a keybinding/activator to lock/unlock the zoom area. Then when unlocked, the zoomed area animates back to wherever the cursor is. This way, you don't have to modify the config to do what you want, you can lock/unlock on the fly.

feel free to implement whatever on top of this, the "hard" part are the core changes

soreau:
> feel free to implement whatever on top of this, the "hard" part are the core changes

Yes, I would rather the 'fine tuning' be left to either @vdegenne or me.

The only thing is, CI is not happy. Can you back out the zoom patch and fix CI?

vdegenne:
I agree, this PR should have been only the core change.
And we can keep the PR I opened to focus on the zoom plugin minimal changes, which with [the patch @soreau proposed](https://sharey.org/doJecC) would give the plugin a much better UX than what ChatGPT suggested.

It should behave like this:

- The zoom initially has the current behavior (follows the pointer, "free" mode?)
- We make so the "lock" mode is togglable/customizable from the config (currently hardcoded to `<modifier>+L` in the patch, but a better default IMO would be `<modifier>+middle mouse click`, it feels more natural to click to "place the anchor" after zooming with the same wheel, and that means we don't have to move right hand mouse->keyboard->mouse.)
- Ideally, when we zoom out and go back to normal the mode should toggle back to "free" mode so next time we zoom again we are not *locked* from the previous anchor point.

soreau:
I notice the cursor bleeds until it is moved, i.e. it keeps rendering each frame without damage, so the cursor gets surrounded with a progressively darkening blackness.

soreau:
I managed to fix the tests [here](https://github.com/WayfireWM/wayfire/commit/b1190788824253ca8a4860e1d99a9b0f43885aaa) but the software cursor still apparently renders without damage, causing the symptom described above.

soreau:
I have fixed the bleeding cursor issue [here](https://github.com/WayfireWM/wayfire/commit/484ace2ff8d23998bd394c433fb9c83813023719). Please let me know what you think.

vdegenne:
> I have fixed the bleeding cursor issue [here](https://github.com/WayfireWM/wayfire/commit/484ace2ff8d23998bd394c433fb9c83813023719). Please let me know what you think.

So it was the zoom! It works great, Thanks.

PR #3132, opened 2026-09-13
animate: don't log empty animation type as an error

mrsteve0924:
Handle an empty animation type the same way as "none" and return without logging an error. Unknown non-empty animation types continue to produce an error.

Assisted-by: ChatGPT:GPT-5.6 Sol
Signed-off-by: steve finkel <mrsteve0924@verizon.net>


mrsteve0924:
`desired_effect_name` is initialized as an empty string and remains
empty when a view does not match any of the animation matchers and
has no animation type property set.

This is a valid state indicating that no animation was selected, but
it was being treated as an unknown animation type, resulting in log
messages such as:

    Unknown animation type: ""

This change handles an empty animation type the same way as "none"
and returns without logging an error.

Unknown non-empty animation types continue to produce an error.

Tested by confirming that the empty animation type errors no longer
appear, while an intentionally invalid animation type such as "bogus"
still produces:

    Unknown animation type: "bogus"

ammen99:
wouldn't it be simpler to initialize the variable with "none" instead of handling multiple values the same way?

mrsteve0924:
> wouldn't it be simpler to initialize the variable with "none" instead of handling multiple values the same way?

took your suggestion

Issue #3099, opened 2026-08-02
title bar is rendered off to the side of the the app window


mrsteve0924:
## Wayfire version
0.12.0-dea6fabd (Aug  1 2026) branch master wlroots-0.20.2

## GPU / Driver
NVIDIA GeForce GTX 1660 Ti/PCIe/SSE2
NVIDIA 610.43.03

## Describe the bug
When running VIrtualBox (and this seems to be only happening in VirtualBox), the title bar is rendered off to the side of the the app window.  If I move the window the title bar snaps back into place. see video

## To Reproduce
run VirtualBox from terminal or desktop file

## Screenshots / Videos / Stacktrace

https://github.com/user-attachments/assets/69c69f83-6b8d-4932-9fd5-6f39acb9ffcb

## Additional context
I've tried running VirtualBox with different env settings but issue persists. 
QT_QPA_PLATFORMTHEME=qt6ct VirtualBox
QT_QPA_PLATFORMTHEME=qt5ct VirtualBox
QT_QPA_PLATFORMTHEME=xcb VirtualBox

I am using pixdecor.
i did a git bisect on my current wayfire and identified edd429e656651862f1c8131c2e53cbbc2ae7eff3 as the first bad commit. 
this bug doesn't appear if using the decorate plugin. so not sure if specifically a wayfire bug or pixdecor bug.

ammen99:
probably pixdecor bug, floating point geometry requires careful alignment, esp if you use fractional scale (but alignment issues exist even on integer scales)

mrsteve0924:
i was debugging an issue with transmission-gtk not rendering its icon on the window-list widget in wf-panel.  I ran `GDK_BACKEND=x11 transmission-gtk` and it fixed the icon issue, but triggered this window title bar bug again.  But I switched to using wayfire's decoration plugin for a few weeks and this is the first time since I was able to trigger the bug again.  So I think now the issue isnt with the previous decoration plugin I was using, pixdecor.

And as before the title bar fixes itself if i move the window with the keybindings.

PR #3134, opened 2026-09-15
Resize the server-side decoration when a view is mapped.

mrsteve0924:
Some XWayland views can reach their final geometry without emitting a geometry-changed signal after the decorator is created. In that case the decoration node can remain at its initial small size until another event, such as a workspace change or window move, triggers a resize.

Resize the decorator on view map so it uses the view's final geometry immediately.

Assisted-by: ChatGPT:GPT-5.6 Sol
Signed-off-by: steve finkel mrsteve0924@verizon.net


mrsteve0924:
this addresses issue #3099

ammen99 (approved):
I see, nice catch. Thanks!

PR #3136, opened 2026-09-17
decor: ellipsize long window titles

sanyarnd:
Long window titles can currently wrap or run into the titlebar buttons because the Pango layout has a width but no single-line constraint.

This keeps the existing horizontal and vertical alignment settings, while limiting the title to one line and ellipsizing it at the end when it does not fit. Short titles render as before.

Tested locally, built and run against wayland 1.24.0 and the bundled wlroots 0.20.2. This is one of a few fixes I made for my own Wayfire setup and figured was worth sending upstream.


ammen99:
> * Full local configuration is currently blocked because the host provides Wayland 1.22 while the bundled wlroots 0.20 requires Wayland 1.24.

The way I understand it, this means that either 1) you haven't tested this change or 2) you generated this with an LLM and haven't bothered to read and edit the LLM output to reflect the truth (that you did test the change)

Am I missing something? There is no point in submitting a PR for review in either of those cases.

sanyarnd:
Did test the change locally, the description was generated, I read it, but lack of sleep did it work 🥲 
Basically it's a part of fixes I did for myself and decided they're good enough to contribute back.

ammen99:
> Did test the change locally, the description was generated, I read it, but lack of sleep did it work 🥲 Basically it's a part of fixes I did for myself and decided they're good enough to contribute back.

Good! But please update the PR description then, it seemed quite weird.

PR #3139, opened 2026-09-19
animate: squeezimize: Fix shader so that opposing sides have the same curve

soreau:


ammen99 (approved):
You're the expert on this animation, if you say it is better this way, I trust your judgement :)

So lgtm and thanks!

PR #3140, opened 2026-09-20
decor: prevent unintended moves on titlebar double-click

mrsteve0924:
Require the pointer to move more than 5 pixels from the press position before starting a window move. This prevents small pointer movements during a double-click from initiating a move instead of toggling maximization.

Also clear the grabbed state when handling a double-click release. Otherwise, the geometry change caused by maximizing or restoring a tiled window can leave the decoration in a stale grabbed state and trigger an unintended move.


soreau:
LGTM, thanks for helping support pixdecor!

